PR #84, opened 2020-09-15
Remove gitrepo directory structure restrictions

ibuildthecloud:
Three breaking changes

   1. fleet.yaml structure is more nested now
   2. GitRepo bundleDirs is not paths
   3. Overlays section and concept is removed from bundleSpec


lindhe:
I find no documentation on Sops for Fleet. Is it meant to be used by end users, or just some internal component?

Issue #94, opened 2020-09-19
CIS v2 scan on a hardened cluster fails when fleet is deployed


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- On a cluster - CIS 1.5 hardened - Deploy CIS v2, OPA, Logging, monitoring and istio charts
- The CIS 1.5 hardened cluster scan when run, shows 2 tests - 5.1.5 and 5.3.2 as failures.

- following namespaces break test 5.1.5
```
kubectl get serviceaccounts --all-namespaces -o json | jq -r '.items[] | select(.metadata.name=="default") | select((.automountServiceAccountToken == null) or (.automountServiceAccountToken == true))' | jq .metadata.namespace
"cattle-gatekeeper-system"
"cattle-logging-system"
"cattle-monitoring-system"
"dashboard-shells"
"default"
"fleet-system"
"istio-system"
"kube-system"
```

**Expected Result:**
The charts should comply with hardening steps

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head - commit id: `8fd81a9b4`
- Installation option (single install/HA): Single node

StrongMonkey:
https://github.com/rancher/rancher/pull/29054 addresses `dashboard-shells`

sowmyav27:
**On master-head - commit id:`2229f5cd7`**

- Deploy a CIS 1.5 hardened cluster. (verified both 1.19.2 and 1.18.9 k8s clusters)
- Test 5.1.5 fails.
- `"fleet-system"` fails and is resulted in the below script when run

```
kubectl get serviceaccounts --all-namespaces -o json | jq -r '.items[] | select(.metadata.name=="default") | select((.automountServiceAccountToken == null) or (.automountServiceAccountToken == true))' | jq .metadata.namespace
```
Reopening the issue

prachidamle:
After the above PR is merged, we will need a fleet release and new fleet-agent chart

sowmyav27:
Reopening the issue as the fleet's tag deployed on latest master - commit id: `3950b8e` is `fleet:v0.3.0-beta300` and fix is not available.

prachidamle:
fleet-agent chart is not updated with the latest fleet release: 0.3.0-beta5

[...]

PR #97, opened 2020-09-23
Fix packaging script that does tag replacement to only replace dev tag

prachidamle:
https://github.com/rancher/fleet/issues/94

Following kubectl image's tag in the  fleet-agent chart should not be replaced when we package fleet-agent chart:

```
  kubectl:
    repository: rancher/kubectl
    tag: v1.18.6
```

Otherwise the fleet-agent chart gets the wrong kubectl image tag which cause the post install hook to not work. 

Currently this is what gets generated under rancher/charts/fleet-agent/Values.yaml
```
  kubectl:
    repository: rancher/kubectl
    tag: v0.3.0-beta3
```

Hence fixing the script to only replace tags `dev`

### DIFF
--- a/scripts/package-helm
+++ b/scripts/package-helm
@@ -33,7 +33,7 @@ sed -i \
     build/charts/fleet-agent/Chart.yaml
 
 sed -i \
-    -e 's/tag:.*/tag: '${HELM_TAG}'/' \
+    -e 's/tag: dev/tag: '${HELM_TAG}'/' \
     build/charts/fleet-agent/values.yaml
 
 go run ./pkg/codegen crds ./build/charts/fleet-crd/templates/crds.yaml


StrongMonkey (approved):

scripts/package-helm
```diff
@@ -33,7 +33,7 @@ sed -i \
     build/charts/fleet-agent/Chart.yaml
 
 sed -i \
-    -e 's/tag:.*/tag: '${HELM_TAG}'/' \
+    -e 's/tag: dev/tag: '${HELM_TAG}'/' \
```
ibuildthecloud, line 36:
```suggestion
    -e 's/tag: v0.*/tag: '${HELM_TAG}'/' \
```

PR #99, opened 2020-09-24
Allow mkdocs development on Mac

cjellick:
(because Linux desktop users only care about themselves)

Couldn't run the docs from a mac using `make serve-docs`. There were a
couple problems:
- --net=host wouldn't allow for port forwarding, needed to use a
specific port instead
- mkdocs serve only listens on localhost by default, needed to tell it
to listen on all interfaces
- listening on all interfaces caused mkdocs to emit a warning and
strict=true means that would cause it to immediately exit, needed to
turn off strict mode :grimace:

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -1,6 +1,6 @@
 site_name: Fleet - GitOps at Scale
 repo_url: https://github.com/rancher/fleet
-strict: true
+strict: false
 theme:
   name: material
   palette:
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -1,6 +1,7 @@
 site_name: Fleet - GitOps at Scale
 repo_url: https://github.com/rancher/fleet
-strict: false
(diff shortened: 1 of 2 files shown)


cjellick:
really feel like im being ignored

ibuildthecloud:
@cjellick Can you remove strict? For development you can make that local change, but for CI I need strict on otherwise docs fail.

mkdocs.yml
```diff
@@ -1,6 +1,6 @@
 site_name: Fleet - GitOps at Scale
 repo_url: https://github.com/rancher/fleet
-strict: true
+strict: false
```
cjellick, line 3:
```suggestion
# Set to false for local docs development
strict: true
```

cjellick:
fixed via github suggestion UI, so it added a commit.  you'll want to squash and merge 😸

PR #122, opened 2020-10-09
Fix overlays definition in example fleet.yaml in docs

bashofmann:
In fleet 0.3.0 the list of yaml overlays must be in a yaml object instead of directly in the customization object (https://github.com/rancher/fleet/blob/786b72d2c0093f21f3a565efc86d083e7555dd1a/pkg/apis/fleet.cattle.io/v1alpha1/bundle.go#L175)


ibuildthecloud:
Thanks!

PR #143, opened 2020-11-04
Add HTTP proxy settings to fleet gitjob subchart

somaritane:
This PR adds support for HTTP proxy settings to [`gitjob`](https://github.com/rancher/gitjob) controller.
The change complements https://github.com/rancher/gitjob/pull/20, as `gitjob` is deployed together with `fleet` as subchart.

### DIFF
--- a/charts/fleet/charts/gitjob/values.yaml
+++ b/charts/fleet/charts/gitjob/values.yaml
@@ -14,4 +14,4 @@ global:
 # proxy: http://<username>@<password>:<url>:<port>
 
 # comma separated list of domains or ip addresses that will not use the proxy
-noProxy: 127.0.0.0/8,10.0.0.0/8,172.16.0.0/12,192.168.0.0/16
+noProxy: 127.0.0.0/8,10.0.0.0/8,172.16.0.0/12,192.168.0.0/16,.svc,.cluster.local
--- a/charts/fleet/charts/gitjob/values.yaml
+++ b/charts/fleet/charts/gitjob/values.yaml
@@ -9,3 +9,9 @@ tekton:
 global:
   cattle:
     systemDefaultRegistry: ""
+
+# http[s] proxy server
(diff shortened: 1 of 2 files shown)


charts/fleet/charts/gitjob/values.yaml
```diff
@@ -9,3 +9,9 @@ tekton:
 global:
   cattle:
     systemDefaultRegistry: ""
+
+# http[s] proxy server
+# proxy: http://<username>@<password>:<url>:<port>
+
+# comma separated list of domains or ip addresses that will not use the proxy
+noProxy: 127.0.0.0/8,10.0.0.0/8,172.16.0.0/12,192.168.0.0/16
```
ytsarev, line 17: Probably not really required for gitjob, but usually, it makes sense to include `10.43.0.1` (local access to kubernetes api) to `NO_PROXY`

somaritane: > Probably not really required for gitjob, but usually, it makes sense to include `10.43.0.1` (local access to kubernetes api) to `NO_PROXY`

Thanks @ytsarev, nice catch. I guess I should also add `.svc` and `.cluster.local`, according to https://github.com/rancher/rancher/pull/29491, will amend.

somaritane: @ytsarev, It looks like [Service Cluster CIDR is configurable](https://github.com/rancher/docs/blob/master/content/k3s/latest/en/installation/install-options/server-config/_index.md#networking), so it's better to leave it for user to override.
However I've added `.svc` and `.cluster.local` domains.

ytsarev: Perfect :+1:

ytsarev (approved):

StrongMonkey:
Changes are included in here https://github.com/rancher/fleet/pull/154

PR #151, opened 2020-11-17
Run as non-root user 1000

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rancher/issues/29696

PR #165, opened 2020-12-01
Fallback to use custom cacert

StrongMonkey:


pkg/controllers/cluster/import.go
```diff
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"k8s.io/client-go/rest"
```
nickgerace, line 7: nit: included in the std modules

pkg/controllers/cluster/import.go
```diff
@@ -233,3 +235,27 @@ func (i *importHandler) importCluster(cluster *fleet.Cluster, status fleet.Clust
...
+
+	if raw.Contexts[raw.CurrentContext] != nil {
+		cluster := raw.Contexts[raw.CurrentContext].Cluster
+		if raw.Clusters[cluster] != nil {
+			_, err := http.Get(raw.Clusters[cluster].Server)
+			if err == nil {
+				raw.Clusters[cluster].CertificateAuthorityData = nil
+			}
+		}
+	}
```
nickgerace, line 258: We may want to return an error if `raw.Contexts[raw.CurrentContext]` is `nil`.

StrongMonkey:
https://github.com/rancher/fleet/issues/239

pkg/controllers/cluster/import.go
StrongMonkey: We might not want to handle validation at here. This piece of code is to check if the current cluster is behind a public CA and if so remove the cacert data in kubeconfig. So if there is no cluster it will fail to establish a contection and other places will return an error.

nickgerace: Ah, that makes sense given the context.

nickgerace (approved):

dramich (approved):

PR #169, opened 2020-12-08
Use deep copy in merge function

boxboatmatt:
Currently when `Calculate` calls `merge` it is not passing a deep copy of `spec.BundleDeploymentOptions`, which contains pointers. This results in the `merge` incorrectly modifying the `base` `BundleDeploymentOptions`. 

This PR will address an issue where targetCustomizations by one cluster override the base. i.e. helm values for a specific target apply to all targets that don't have a helm values override


boxboatjeremy (approved):
Looks good to me!

pariscc (approved):

PR #194, opened 2020-12-23
Add content controller to purge orphaned contents

boxboatmatt:

Currently fleet does not cleanup the `Content` objects that are created for a `Bundle`. Over time this will result in a very large number of `Content` objects being stored on the `fleet-controller` cluster. Added a `ContentController` to purge orphans when bundleDeployments are updated.

### DIFF
--- a/pkg/controllers/content/controller.go
+++ b/pkg/controllers/content/controller.go
@@ -1,0 +1,110 @@
+package content
+
+import (
+	"context"
+	"time"
+
+	fleet "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
+	fleetcontrollers "github.com/rancher/fleet/pkg/generated/controllers/fleet.cattle.io/v1alpha1"
+	corecontrollers "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
+	"github.com/rancher/wrangler/pkg/kv"
+	"github.com/rancher/wrangler/pkg/ticker"
+	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type handler struct {
+	content          fleetcontrollers.ContentController
+	bundleDeployment fleetcontrollers.BundleDeploymentController
+	namespaces       corecontrollers.NamespaceClient
+}
+
+type contentRef struct {
+	content      fleet.Content
+	safeToDelete bool
+	bundleCount  int
+}
+
+func Register(ctx context.Context,
+	content fleetcontrollers.ContentController,
+	bundleDeployment fleetcontrollers.BundleDeploymentController,
+	namespaces corecontrollers.NamespaceController) {
+
+	h := &handler{
+		content:          content,
+		bundleDeployment: bundleDeployment,
+		namespaces:       namespaces,
+	}
+
+	go func() {
+		h.purgeOrphaned(ctx)
+	}()
+
+}
+
+func (h *handler) purgeOrphaned(ctx context.Context) {
+
+	deleteRefs := make(map[string]*contentRef)
+
+	for range ticker.Context(ctx, time.Minute*5) {
+		logrus.Debugf("Checking for orphaned content objects")
+		namespaces, err := h.namespaces.List(metav1.ListOptions{})
+		if err != nil {
+			logrus.Warnf("Error reading namespaces %v", err)
+			continue
+		}
+		var bundleDeployments []fleet.BundleDeployment
+		for _, ns := range namespaces.Items {
+			nsBundleDeployments, err := h.bundleDeployment.List(ns.Name, metav1.ListOptions{})
+			if err == nil {
+				bundleDeployments = append(bundleDeployments, nsBundleDeployments.Items...)
+			}
+		}
+
+		contentRefs := make(map[string]*contentRef)
+
+		contents, err := h.content.List(metav1.ListOptions{})
+		if err != nil {
+			logrus.Warnf("Error reading contents %v", err)
+			continue
+		}
+
+		for _, content := range contents.Items {
+			contentRefs[content.Name] = &contentRef{
+				content:     content,
+				bundleCount: 0,
+			}
+		}
+
+		for _, bd := range bundleDeployments {
+			deployManifestID, _ := kv.Split(bd.Spec.DeploymentID, ":")
+			if val, ok := contentRefs[deployManifestID]; ok {
+				val.bundleCount++
+			}
+
+			stagedManifestID, _ := kv.Split(bd.Spec.StagedDeploymentID, ":")
+			if val, ok := contentRefs[stagedManifestID]; ok && stagedManifestID != deployManifestID {
+				val.bundleCount++
+			}
+		}
+
+		for _, cr := range contentRefs {
+			deleteRef, deleteCandidate := deleteRefs[cr.content.Name]
+			if cr.bundleCount == 0 && deleteCandidate && deleteRef.safeToDelete {
+				logrus.Infof("Deleting orphaned content[%s]", cr.content.Name)
+				_ = h.content.Delete(cr.content.Name, &metav1.DeleteOptions{})
+			} else if cr.bundleCount > 0 {
+				if deleteCandidate {
+					delete(deleteRefs, cr.content.Name)
+				}
+			} else {
+				logrus.Infof("Marking orphaned content[%s] for deletion", cr.content.Name)
+				deleteRefs[cr.content.Name] = &contentRef{
+					content:      cr.content,
+					bundleCount:  0,
+					safeToDelete: true,
+				}
+			}
+		}
+	}
+}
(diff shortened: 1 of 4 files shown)


StrongMonkey:
@boxboatmatt I would suggest to put the cleanup logic into a goroutine that runs every X minutes instead of triggering by bundleDeployment. Then X can be a global setting in fleet so we don't purge too aggresively.

ibuildthecloud:
This definitely needs to be address, but the approach needs to be a bit more like GC.  So run on an interval and then also give consideration to race conditions where a bundledeployment is being created for content X but then at the same time it is being deleted.

StrongMonkey:
@boxboatmatt To solve the race condition mentioned by @ibuildthecloud, you can re-enqueue all the bundles after cleanup to make sure content isn't being immediately deleted while waiting for bundledeployment to be created. Enqueue bundles will make sure all the needed contents will be created. Let me know when you can make the changes and I can help testing your PR also. Thanks for contributing!

boxboatmatt:
@StrongMonkey moved the logic into a goroutine. To solve the race condition I added a label to mark it for purging and then the 2nd time I see that content if it is still not referenced I remove it. Basically if it is an orphan for 10 minutes as currently written.

boxboatmatt:
@StrongMonkey what do you need me to do to get this into the next fleet release?

StrongMonkey:
@boxboatmatt I will try to review your PR again :+1:

pkg/controllers/content/controller.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	go func() {
+		h.purgeOrphaned(time.Minute * 5)
+	}()
+}
+
+func (h *handler) purgeOrphaned(duration time.Duration) {
+
+	for true {
```
StrongMonkey, line 44: There is a package in wrangler you can use to write a 5 minute ticker: https://github.com/rancher/wrangler/blob/master/pkg/ticker/ticker.go#L8

```
for range ticker.Context(ctx, 5 * time.Minute)
   // do GC
```

pkg/controllers/content/controller.go
```diff
@@ -0,0 +1,104 @@
...
+		logrus.Debugf("Checking for orphaned content objects")
+		namespaces, err := h.namespaces.List(metav1.ListOptions{})
+		if err == nil {
```
StrongMonkey, line 47: In general It would be nice to deal with err first. Something like 

```
namespaces, err := h.namespaces.List(metav1.ListOptions{})
if err != nil {
   // log error and return
}
// do things
```

pkg/controllers/content/controller.go
```diff
@@ -0,0 +1,104 @@
...
+					if val, ok := contentRefs[stagedManifestID]; ok && stagedManifestID != deployManifestID {
+						val.bundleCount++
+					}
+				}
+
+				for _, cr := range contentRefs {
+					if cr.bundleCount == 0 && cr.content.Labels["purge"] == "true" {
+						logrus.Infof("Deleting orphaned content[%s]", cr.content.Name)
+						_ = h.content.Delete(cr.content.Name, &metav1.DeleteOptions{})
+					} else if cr.bundleCount > 0 {
```
StrongMonkey, line 87: I think updating labels is fine but might be a little bit overhead. Idealy I think in-memory map would work. So you can put a map that contains the content that can be deleted on the next run.

pkg/controllers/controllers.go
```diff
@@ -5,6 +5,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/rancher/fleet/pkg/controllers/content"
+
```
StrongMonkey, line 9: Nit: remove space.

boxboatmatt:
@StrongMonkey rebased on `master` and addressed comments... not sure why the CI build is failing though.

pkg/controllers/content/controller.go
```diff
@@ -0,0 +1,110 @@
...
+	for range ticker.Context(ctx, time.Minute*5) {
+		logrus.Debugf("Checking for orphaned content objects")
+		namespaces, err := h.namespaces.List(metav1.ListOptions{})
+		if err != nil {
+			logrus.Warnf("Error reading namespaces %v", err)
+			continue
+		}
+		var bundleDeployments []fleet.BundleDeployment
+		for _, ns := range namespaces.Items {
+			nsBundleDeployments, err := h.bundleDeployment.List(ns.Name, metav1.ListOptions{})
```
StrongMonkey, line 58: Make sure this error is also printed in case of a failure

pkg/controllers/content/controller.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+type handler struct {
+	content          fleetcontrollers.ContentController
+	bundleDeployment fleetcontrollers.BundleDeploymentController
+	namespaces       corecontrollers.NamespaceClient
+}
+
+type contentRef struct {
+	content      fleet.Content
```
StrongMonkey, line 23: Do you need to put content in contentRef struct given that you already have the key in map?

boxboatmatt: Good catch... don't need it anymore

pkg/controllers/content/controller.go
```diff
@@ -0,0 +1,110 @@
...
+			if val, ok := contentRefs[stagedManifestID]; ok && stagedManifestID != deployManifestID {
+				val.bundleCount++
+			}
+		}
+
+		for _, cr := range contentRefs {
+			deleteRef, deleteCandidate := deleteRefs[cr.content.Name]
+			if cr.bundleCount == 0 && deleteCandidate && deleteRef.safeToDelete {
+				logrus.Infof("Deleting orphaned content[%s]", cr.content.Name)
+				_ = h.content.Delete(cr.content.Name, &metav1.DeleteOptions{})
```
StrongMonkey, line 95: Worth to log the error.
```suggestion
				if err := h.content.Delete(cr.content.Name, &metav1.DeleteOptions{}): err != nil && err != errors.IsNotFound(err) {
				// log error
				}
```

pkg/controllers/content/controller.go
```diff
@@ -0,0 +1,110 @@
...
+				val.bundleCount++
+			}
+
+			stagedManifestID, _ := kv.Split(bd.Spec.StagedDeploymentID, ":")
+			if val, ok := contentRefs[stagedManifestID]; ok && stagedManifestID != deployManifestID {
+				val.bundleCount++
+			}
+		}
+
+		for _, cr := range contentRefs {
```
StrongMonkey, line 91: @boxboatmatt Do you think you can split the logic into two parts for more readiblity? 
1. Check contentRefs and update deleteRefs.
2. Check deleteRefs, delete content and update deleteRefs.

StrongMonkey:
@boxboatmatt I think there are some transient errors on CI, make sure you repush to see if it fixes itself.

pkg/controllers/content/controller.go
boxboatmatt: Are you suggesting two loops... something like this?

```go
for contentName, cr := range contentRefs {
	_, deleteCandidate := deleteRefs[contentName]
	if cr.bundleCount > 0 {
		if deleteCandidate {
			delete(deleteRefs, contentName)
		}
	} else {
		if deleteCandidate {
			deleteRefs[contentName].safeToDelete = true
		} else {
			logrus.Infof("Marking orphaned content[%s] for deletion", contentName)
			deleteRefs[contentName] = &contentRef{
				bundleCount:  0,
				markForDeletion: true,
				safeToDelete: false,
			}
		}
	}
}

for contentName, dr := range deleteRefs {
	if dr.safeToDelete {
		logrus.Infof("Deleting orphaned content[%s]", contentName)
		if err := h.content.Delete(contentName, &metav1.DeleteOptions{}); err != nil && !errors.IsNotFound(err) {
			logrus.Warnf("Error deleting contentbundle %v", err)
		}
	}
}
```

StrongMonkey: yes, something like this

boxboatmatt:
> @boxboatmatt I think there are some transient errors on CI, make sure you repush to see if it fixes itself.

Still failing on latest push... I have no insight on what the error is since the CI is private.

StrongMonkey (approved):
LGTM, need to figure out what CI failed

boxboatmatt:
@StrongMonkey are you able to see CI logs? It looks like all of the recent PRs have failed CI... Let me know if there is something I need to do on my end

StrongMonkey:
@boxboatmatt Can you pin these two to `alpine:3.12.3`? https://github.com/rancher/fleet/blob/307463c0fb900996c7e8c1bd5743c3624107065d/package/Dockerfile.agent#L1
https://github.com/rancher/fleet/blob/70d3e5f679979c63fe61d2338b3db98a071637ba/package/Dockerfile#L1

I wonder if `alpine:3.13.0` is breaking the change. We didn't pin the version in Dockerfile so once cache is invalid it will fetch the latest version of alpine.

boxboatmatt:
@StrongMonkey looks like that was the issue.... CI passed

nickgerace (approved):

PR #195, opened 2020-12-23
Add Windows support and consolidate drone pipelines

nickgerace:
Relevant issue: https://github.com/rancher/rancher/issues/30516

This PR adds preliminary Windows support. The `fleet-agent` has a Windows manifest, and all other components use `tolerations` and `nodeSelector` settings to be scheduled on Linux nodes.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,22 +1,19 @@
 ---
 kind: pipeline
-name: amd64
-
+name: linux-amd64
 platform:
   os: linux
   arch: amd64
-
 steps:
 - name: build
-  image: rancher/dapper:v0.4.1
+  image: rancher/dapper:v0.5.5
   environment:
-    CROSS: "true"
+    CROSS: "windows"
   commands:
   - dapper ci
   volumes:
   - name: docker
     path: /var/run/docker.sock
-
 - name: github_binary_release
   image: plugins/github-release
   settings:
@@ -37,7 +34,34 @@ steps:
     - refs/tags/*
     event:
     - tag
-
+- name: upload-tag
+  pull: default
+  image: plugins/gcs
+  settings:
+    acl:
+      - allUsers:READER
+    cache_control: "no-cache,must-revalidate"
+    source: ./dist/artifacts
+    target: releases.rancher.com/fleet/${DRONE_TAG}
+    token:
+      from_secret: google_auth_key
+  when:
+    event:
+      - tag
+- name: upload-latest
+  pull: default
+  image: plugins/gcs
+  settings:
+    acl:
+      - allUsers:READER
+    cache_control: "no-cache,must-revalidate"
+    source: ./dist/artifacts
+    target: releases.rancher.com/fleet/latest
+    token:
+      from_secret: google_auth_key
+  when:
+    event:
+      - tag
 - name: docker-publish-agent
   image: plugins/docker
   settings:
@@ -45,9 +69,11 @@ steps:
     password:
       from_secret: docker_password
     repo: "rancher/fleet-agent"
-    tag: "${DRONE_TAG}-amd64"
+    tag: "${DRONE_TAG}-linux-amd64"
     username:
       from_secret: docker_username
+    build_args:
+      - "ARCH=amd64"
   when:
     instance:
     - drone-publish.rancher.io
@@ -56,7 +82,6 @@ steps:
     - refs/tags/*
     event:
     - tag
-
 - name: docker-publish
   image: plugins/docker
   settings:
@@ -64,9 +89,11 @@ steps:
     password:
       from_secret: docker_password
     repo: "rancher/fleet"
-    tag: "${DRONE_TAG}-amd64"
+    tag: "${DRONE_TAG}-linux-amd64"
     username:
       from_secret: docker_username
+    build_args:
+      - "ARCH=amd64"
   when:
     instance:
     - drone-publish.rancher.io
@@ -75,29 +102,24 @@ steps:
     - refs/tags/*
     event:
     - tag
-
 volumes:
 - name: docker
   host:
     path: /var/run/docker.sock
-
 ---
 kind: pipeline
-name: arm64
-
+name: linux-arm64
 platform:
   os: linux
   arch: arm64
-
 steps:
 - name: build
-  image: rancher/dapper:v0.4.1
+  image: rancher/dapper:v0.5.5
   commands:
-  - dapper ci
+  - dapper build
   volumes:
   - name: docker
     path: /var/run/docker.sock
-
 - name: github_binary_release
   image: plugins/github-release
   settings:
@@ -118,7 +140,6 @@ steps:
     - refs/tags/*
     event:
     - tag
-
 - name: docker-publish-agent
   image: plugins/docker
   settings:
@@ -126,9 +147,11 @@ steps:
     password:
       from_secret: docker_password
     repo: "rancher/fleet-agent"
-    tag: "${DRONE_TAG}-arm64"
+    tag: "${DRONE_TAG}-linux-arm64"
     username:
       from_secret: docker_username
+    build_args:
+      - "ARCH=arm64"
   when:
     instance:
     - drone-publish.rancher.io
@@ -137,7 +160,6 @@ steps:
     - refs/tags/*
     event:
     - tag
-
 - name: docker-publish
   image: plugins/docker
   settings:
@@ -145,9 +167,11 @@ steps:
     password:
       from_secret: docker_password
     repo: "rancher/fleet"
-    tag: "${DRONE_TAG}-arm64"
+    tag: "${DRONE_TAG}-linux-arm64"
     username:
       from_secret: docker_username
+    build_args:
+      - "ARCH=arm64"
   when:
     instance:
     - drone-publish.rancher.io
@@ -156,29 +180,25 @@ steps:
     - refs/tags/*
     event:
     - tag
-
 volumes:
 - name: docker
   host:
     path: /var/run/docker.sock
-
 ---
 kind: pipeline
-name: arm
-
+name: linux-arm
 platform:
   os: linux
   arch: arm
-
 steps:
+# Update this tag once dapper arm v7 works on drone.
 - name: build
   image: rancher/dapper:v0.4.1
   commands:
-  - dapper ci
+  - dapper build
   volumes:
   - name: docker
     path: /var/run/docker.sock
-
 - name: github_binary_release
   image: plugins/github-release
   settings:
@@ -199,7 +219,6 @@ steps:
     - refs/tags/*
     event:
     - tag
-
 - name: docker-publish-agent
   image: plugins/docker
   settings:
@@ -207,9 +226,11 @@ steps:
     password:
       from_secret: docker_password
     repo: "rancher/fleet-agent"
-    tag: "${DRONE_TAG}-arm"
+    tag: "${DRONE_TAG}-linux-arm"
     username:
       from_secret: docker_username
+    build_args:
+      - "ARCH=arm"
   when:
     instance:
     - drone-publish.rancher.io
@@ -218,7 +239,6 @@ steps:
     - refs/tags/*
     event:
     - tag
-
 - name: docker-publish
   image: plugins/docker
   settings:
@@ -226,9 +246,11 @@ steps:
     password:
       from_secret: docker_password
     repo: "rancher/fleet"
-    tag: "${DRONE_TAG}-arm"
+    tag: "${DRONE_TAG}-linux-arm"
     username:
       from_secret: docker_username
+    build_args:
+      - "ARCH=arm"
   when:
     instance:
     - drone-publish.rancher.io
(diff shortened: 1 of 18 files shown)


nickgerace:
Graduated this PR from a draft to a WIP (work in progress).

.drone.yml
```diff
@@ -48,6 +89,14 @@ steps:
     tag: "${DRONE_TAG}-amd64"
     username:
       from_secret: docker_username
+    build_args:
+      - "VERSION=${DRONE_TAG}"
+      - "RELEASES=releases.rancher.com"
+    context: package/
+    custom_dns: 1.1.1.1
+  volumes:
```
StrongMonkey, line 97: Do you need this docker volume? I think publish step doesn't really need it

.drone.yml
```diff
@@ -67,6 +115,14 @@ steps:
     tag: "${DRONE_TAG}-amd64"
     username:
       from_secret: docker_username
+    build_args:
```
StrongMonkey, line 118: same above

.drone.yml
```diff
@@ -75,50 +131,19 @@ steps:
     - refs/tags/*
     event:
     - tag
-
 volumes:
 - name: docker
   host:
     path: /var/run/docker.sock
-
+depends_on:
```
StrongMonkey, line 138: same question above...If you combined build in one step you probably don't need `depend_on`

.drone.yml
```diff
@@ -156,50 +196,19 @@ steps:
...
 ---
 kind: pipeline
 name: arm
-
 platform:
   os: linux
   arch: arm
-
 steps:
-- name: build
```
StrongMonkey, line 174: Any reason arm is removed?

charts/fleet-agent/values.yaml
```diff
@@ -31,6 +31,19 @@ internal:
   systemNamespace: fleet-system
   managedReleaseName: fleet-agent
 
+# The nodeSelector and tolerations for the agent deployment
+nodeSelector: {}
```
StrongMonkey, line 35: duplicated keys?

charts/fleet/charts/gitjob/values.yaml
```diff
@@ -15,3 +15,12 @@ global:
 
 # comma separated list of domains or ip addresses that will not use the proxy
 noProxy: 127.0.0.0/8,10.0.0.0/8,172.16.0.0/12,192.168.0.0/16,.svc,.cluster.local
+
+nodeSelector:
```
StrongMonkey, line 19: For gitjob you have to change the chart at https://github.com/rancher/gitjob.git and update it.

pkg/agent/manifest.go
```diff
@@ -32,7 +32,7 @@ func Manifest(namespace, image, pullPolicy, generation, checkInInterval string)
 		},
 	)
 
-	dep := basic.Deployment(namespace, DefaultName, image, pullPolicy, DefaultName)
+	dep := basic.Deployment(namespace, DefaultName, image, pullPolicy, DefaultName, false)
```
StrongMonkey, line 35: We shouldn't need this once agent image support windows, correct?

modules/cli/controllermanifest/template.go
```diff
@@ -63,7 +63,7 @@ func objects(namespace, controllerImage string, cfg string, crdsOnly bool) ([]ru
 	objs := []runtime.Object{
 		basic.Namespace(namespace),
 		serviceAccount,
-		basic.Deployment(namespace, genericName, controllerImage, "", serviceAccount.Name),
+		basic.Deployment(namespace, genericName, controllerImage, "", serviceAccount.Name, true),
```
StrongMonkey, line 66: I don't think your will need this though.. The method is not used anymore

.drone.yml
```diff
@@ -1,22 +1,25 @@
 ---
 kind: pipeline
-name: amd64
-
+type: docker
+name: build
```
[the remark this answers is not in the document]
nickgerace: Yes, I can change the name. The `docker` type is not needed as well. That comes from the rancher/dapper `.drone.yaml`, which this was based off of.

.drone.yml
```diff
@@ -1,22 +1,25 @@
...
+  image: rancher/dapper:v0.5.4
   environment:
     CROSS: "true"
   commands:
   - dapper ci
   volumes:
   - name: docker
     path: /var/run/docker.sock
-
+  when:
```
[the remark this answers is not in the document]
nickgerace: Agreed

.drone.yml
```diff
@@ -37,7 +40,45 @@ steps:
...
+      from_secret: google_auth_key
+  when:
+    event:
+      - tag
+volumes:
+- name: docker
+  host:
+    path: /var/run/docker.sock
+---
+kind: pipeline
```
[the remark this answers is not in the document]
nickgerace: Referencing my other comment (https://github.com/rancher/fleet/pull/195#discussion_r556902814), building has been separated from packaging, similar to [rancher/dapper](https://github.com/rancher/dapper/blob/master/.drone.yml). We shorten the PR time by packaging in parallel (linux-arm, windows, linux-amd64, etc.).

.drone.yml
```diff
@@ -48,6 +89,14 @@ steps:
     tag: "${DRONE_TAG}-amd64"
     username:
       from_secret: docker_username
+    build_args:
+      - "VERSION=${DRONE_TAG}"
```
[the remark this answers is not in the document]
nickgerace: Ah thank you.

.drone.yml
```diff
@@ -48,6 +89,14 @@ steps:
...
+      - "RELEASES=releases.rancher.com"
```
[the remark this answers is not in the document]
nickgerace: This is for the GCE bucket upload, although I may have to ensure that it's working.

.drone.yml
nickgerace: Ah, good point. I'll remove it

.drone.yml
nickgerace: Agreed

.drone.yml
nickgerace: I don't think it is.
https://github.com/nickgerace/rancher-fleet/blob/master-windows/.drone.yml#L205-L269

charts/fleet-agent/values.yaml
nickgerace: I went back and forth on how to do this. Essentially, we need separate `nodeSelector` and `tolerations` settings for the agent because the rancher/kubectl container does not have Windows manifests. We might want to have a different way to convey this though. Perhaps...
```yaml
fleetAgent:
  nodeSelector:
  tolerations:
kubectl:
  nodeSelector:
  tolerations:
```

charts/fleet/charts/gitjob/values.yaml
nickgerace: Ah, that makes sense.  I was unsure if this chart was manually pulled in and modified. Will do.

pkg/agent/manifest.go
nickgerace: Once the `rancher/fleet` (manager) image does, then correct, we will not need this. However, since the manger image does not support Windows currently, we need to be able to create...

1. a manager deployment: needs to be scheduled on Linux nodes
2. an agent deployment: has a Windows manifest

modules/cli/controllermanifest/template.go
nickgerace: The method is not used for the manager image anymore?

.drone.yml
```diff
@@ -37,7 +40,45 @@ steps:
     - refs/tags/*
     event:
     - tag
-
+- name: upload-tag
```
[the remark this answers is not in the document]
nickgerace: Now that I think about it, @luthermonson's motivation here could be to ensure that we not publish developer binaries on GitHub. I need to re-evaluate the `when:` field here.

.drone.yml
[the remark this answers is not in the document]
StrongMonkey: Got it, But looks like you should make build pipeline on on every platform. Looks like right now it will only run on linux/amd64

manifest-agent.tmpl
```diff
@@ -0,0 +1,23 @@
...
+    platform:
+      architecture: arm
+      os: linux
+  -
+    image: rancher/fleet-agent:{{#if build.tag}}{{build.tag}}{{else}}{{replace "release/" "" build.branch }}-head{{/if}}-arm64
+    platform:
+      architecture: arm64
+      os: linux
+  -
+    image: rancher/fleet-agent:{{#if build.tag}}{{build.tag}}{{else}}{{replace "release/" "" build.branch }}-head{{/if}}-windows-amd64-1909
```
luthermonson, line 19: you can add 2004 and 20H2 if youre motivated, you already have v0.5.4 dapper and those two are not available there. just copy pasta and be done with it now.

scripts/windows/package.ps1
```diff
@@ -0,0 +1,40 @@
...
+
+$DIR_PATH = Split-Path -Parent $MyInvocation.MyCommand.Definition
+$SRC_PATH = (Resolve-Path "$DIR_PATH\..\..").Path
+cd $SRC_PATH\package\windows
+
+$ARCH = $env:ARCH
+$TAG = $env:TAG
+$REPO = $env:REPO
+$IMAGE = "$REPO/fleet-agent:$TAG"
+$HOST_RELEASE_ID = $env:HOST_RELEASE_ID
```
luthermonson, line 12: you grabbed this host release id stuff from rancher-agent build and you dont really use it. it's an old if/else to do hyperv builds for windows versions not matching and trying to run it in hyperv. i dont think this stuff ever worked and you can get rid of it and do the docker build normally and ignore the stuff in the else down below.

charts/fleet-agent/values.yaml
nickgerace: Going with the above design.

.drone.yml
nickgerace: @luthermonson do we need this? I believe it's in rancher/dapper

charts/fleet/charts/gitjob/values.yaml
nickgerace: PR: https://github.com/rancher/gitjob/pull/29

manifest-agent.tmpl
nickgerace: Done

scripts/windows/package.ps1
nickgerace: Didn't even need this file

scripts/package
```diff
@@ -21,16 +21,7 @@ function build-image() {
...
-for i in bin/fleet-*; do
-    if [ -e "$i" ]; then
-        if [ "$i" = fleet-windows-amd64 ]; then
-            cp $i dist/artifacts/fleet-windows-amd64.exe
-        else
-            cp $i dist/artifacts
-        fi
-    fi
-done
+cp bin/fleet* dist/artifacts/
```
StrongMonkey, line 24: You will need to add `-r`

scripts/build
```diff
@@ -6,16 +6,44 @@ source $(dirname $0)/version
...
-CGO_ENABLED=0 go build -ldflags "$LINKFLAGS $OTHER_LINKFLAGS" -o bin/fleetcontroller ./cmd/fleetcontroller
-CGO_ENABLED=0 go build -ldflags "$LINKFLAGS $OTHER_LINKFLAGS" -o bin/fleetagent ./cmd/fleetagent
-if [ "$CROSS" = "true" ] && [ "$ARCH" = "amd64" ]; then
-    GOOS=darwin go build -ldflags "$LINKFLAGS" -o bin/fleet-darwin${SUFFIX}
-    GOOS=windows go build -ldflags "$LINKFLAGS" -o bin/fleet-windows${SUFFIX}
+OTHER_LINKFLAGS="-extldflags -static"
+
+if [ "$CROSS" = "true" ]; then
+
+    # Linux amd64
```
StrongMonkey, line 17: Bonus point: Can you do this in a for loop in bash?

scripts/build
```diff
@@ -6,16 +6,44 @@ source $(dirname $0)/version
 cd $(dirname $0)/..
 
 mkdir -p bin
-if [ "$(uname)" = "Linux" ]; then
```
StrongMonkey, line 9: Don't you need this? Any reason this is removed?

StrongMonkey: nvm, I see you removed for windows build. Should we add this back so that windows/mac command is the same as others?

nickgerace: Well, I don't see an issue with having the variable set even if the host isn't using it. Let's say you set `CROSS` to `true` and the host is a `macOS` host. You technically need the `OTHER` flags in order to perform the cross.

package/Dockerfile-windows.agent
```diff
@@ -0,0 +1,26 @@
...
+    \
+    Write-Host 'Expanding ...'; \
+    Expand-Archive -Force -Path c:\git.zip -DestinationPath c:\git\.; \
+    \
+    Write-Host 'Cleaning ...'; \
+    Remove-Item -Force -Recurse -Path c:\git.zip; \
+    \
+    Write-Host 'Complete.'; \
+    popd;
+RUN $URL = 'https://{0}/fleet/{1}/fleetagent-windows-amd64.exe' -f $env:RELEASES, $env:VERSION; \
```
nickgerace, line 21: @luthermonson: Am I pulling in the `ARG` correctly here? I don't think so. We may just be able to use `-f ${RELEASES}, ${VERSION};`. Thoughts?

nickgerace:
Going to squash all commits after new reviews.

.drone.yml
```diff
@@ -199,17 +343,39 @@ steps:
...
-
+volumes:
+  - name: docker
+    host:
+      path: \\\\.\\pipe\\docker_engine
+depends_on:
+- build
+---
+kind: pipeline
+name: windows-amd64-2004
```
luthermonson, line 354: need to use my clone: false hack to make 2004/20h2 work, looks like this

https://github.com/rancher/rancher/blob/master/.drone.yml#L738-L744

nickgerace: done

.drone.yml
StrongMonkey: Looks like only `dapper ci` needs this volume since it can reuse docker build cache.

scripts/ci
```diff
@@ -7,4 +7,3 @@ cd $(dirname $0)
 ./test
 ./validate
 ./validate-ci
-./package
```
StrongMonkey, line 10: Any reason ./package is removed? If anything is wrong on Dockerfile the build won't be able to catch.

StrongMonkey: On the side note, I think we should add local copy for development per darren's concern. Today if I want to build image locally the curl won't work. And also the ./package should just run local copy for build step so that CI can catch `./package` script.

nickgerace: This needs to be re-added; good catch

package/Dockerfile-windows.agent
nickgerace: Double-checked the docs. What we have is good!

scripts/ci
nickgerace: Re-added

nickgerace:
Switched to only uploading Windows binaries. Sticking to the previous approach for Linux binaries.

nickgerace:
Whoops accidentally closed it there.

luthermonson (approved):

package/Dockerfile-windows.agent
```diff
@@ -0,0 +1,28 @@
+ARG SERVERCORE_VERSION
+FROM mcr.microsoft.com/windows/servercore:${SERVERCORE_VERSION}
```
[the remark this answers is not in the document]
nickgerace: Determined offline that this won't be included in this PR. Opened a separate issue: https://github.com/rancher/rancher/issues/31031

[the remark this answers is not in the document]
nickgerace: The answer though: we could absolutely try this in a multi-stage context. However, we will need to do long term validation, which would block this PR.

charts/fleet/charts/gitjob/values.yaml
```diff
@@ -15,3 +15,12 @@ global:
 
 # comma separated list of domains or ip addresses that will not use the proxy
 noProxy: 127.0.0.0/8,10.0.0.0/8,172.16.0.0/12,192.168.0.0/16,.svc,.cluster.local
+
```
StrongMonkey, line 18: Can you remove this change? I would like to get this updated along with a new tag in gitjob next time so it is easy to track

manifest.tmpl
```diff
@@ -0,0 +1,17 @@
+image: rancher/fleet:{{#if build.tag}}{{build.tag}}{{else}}{{replace "release/" "" build.branch }}-head{{/if}}
+manifests:
+  -
+    image: rancher/fleet:{{#if build.tag}}{{build.tag}}{{else}}{{replace "release/" "" build.branch }}-head{{/if}}-amd64
```
StrongMonkey, line 4: Since your image format will look like `$tag-linux-$arch`, shouldn't the image in manifest follow the same format?

package/Dockerfile
```diff
@@ -1,4 +1,6 @@
 FROM alpine:3.12.3
-COPY bin/fleetcontroller /usr/bin/
+ARG ARCH
+ENV ARCH=$ARCH
+COPY bin/fleetcontroller-linux-${ARCH} /usr/bin/
```
StrongMonkey, line 4: you should probably rename to `fleetcontroller` still so you don't have to change CMD in dockerfile.

package/Dockerfile.agent
```diff
@@ -1,4 +1,6 @@
 FROM alpine:3.12.3
+ARG ARCH
+ENV ARCH=$ARCH
 RUN apk add -U --no-cache git bash
-COPY bin/fleetagent bin/fleet package/log.sh /usr/bin/
+COPY bin/fleetagent-linux-$ARCH bin/fleet-linux-$ARCH package/log.sh /usr/bin/
```
StrongMonkey, line 5: same here. Make sure the name is fleet-agent and fleet otherwise the code will not find binaries.

charts/fleet/charts/gitjob/values.yaml
nickgerace: Yes

manifest.tmpl
nickgerace: Good point; changing this to match the format.

package/Dockerfile
nickgerace: Ah this was actually a mistake and it _should_ be that. Good catch.

package/Dockerfile.agent
nickgerace: Same as above. This _should_ have been the case. Good catch.

StrongMonkey (approved):

Issue #185, opened 2020-12-16
CPU is maxed out after adding five repositories


dnoland1:
Steps to reproduce:
1. Set up Rancher v2.5.3. Using 2 m5a.large (2vCPU, 8GB RAM) on AWS with k3s and RDS/Aurora/MySQL.
2. Launch 5 VMs to be used for downstream clusters. Using t3a.medium instance on AWS.
3. Create 5 custom clusters, test1, test2, test3, test4, test5.
4. Create one node etcd/cp/worker clusters using the VMs created in step 2.
5. Go to Continuous Delivery
6. Create a new Fleet repository. I'm using https://github.com/chrisurwin/fleet-demo.git and main branch. Name it fleet1 and only add test1 cluster.
7. Wait about 10-15 minutes for the workloads to deploy.
8. Repeat step 6-7 four times, but use fleet2,3,4,5 for the repo names and test2,3,4,5 for the clusters.

Results:
After adding each fleet repository, the fleet-controller is consuming about 20% of the CPU. After adding all 5 repos, 100% CPU is being used. Here's the CPU graph:
![image](https://user-images.githubusercontent.com/468462/102383747-bfbb0280-3f80-11eb-9eb3-da5437a29411.png)

repos were added at approximately 7:20, 7:37, 7:52, 8:06, and 8:21
Here's a top sample:
```
top - 16:28:57 up  1:38,  0 users,  load average: 3.21, 3.37, 3.05
Tasks: 145 total,   2 running, 141 sleeping,   0 stopped,   2 zombie
%Cpu(s): 97.3 us,  2.0 sy,  0.0 ni,  0.5 id,  0.0 wa,  0.0 hi,  0.2 si,  0.0 st
MiB Mem :   7710.5 total,    162.0 free,   2955.6 used,   4592.9 buff/cache
MiB Swap:      0.0 total,      0.0 free,      0.0 used.   5757.0 avail Mem

    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
   7438 root      20   0  974180 311956  38492 S 181.1   4.0  77:34.25 fleetcontroller
    940 root      20   0 1874060   1.6g  89568 S  13.0  21.9  20:09.02 k3s-server
```
[...]

PR #204, opened 2020-12-30
Save gvk to bundle to avoid recompute

StrongMonkey:
https://github.com/rancher/fleet/issues/185


pkg/display/gitrepo.go
```diff
@@ -208,48 +196,16 @@ func appendState(states map[key][]fleet.ResourcePerClusterState, key key, state
...
-				} else {
-					key.Namespace = opts.DefaultNamespace
-				}
-			}
-			key.APIVersion, key.Kind = gvk.ToAPIVersionAndKind()
-			bundleResources[key] = []fleet.ResourcePerClusterState{}
+func bundleResources(bundle *fleet.Bundle) map[key][]fleet.ResourcePerClusterState {
+	bundleResources := map[key][]fleet.ResourcePerClusterState{}
+	for _, resourceKey := range bundle.Status.ResourceKey {
+		k := key{
```
ibuildthecloud, line 202: Can the key type just be ResourceKey now?  Can we delete the key type declaration?

pkg/controllers/bundle/controller.go
```diff
@@ -116,10 +130,86 @@ func (h *handler) OnBundleChange(bundle *fleet.Bundle, status fleet.BundleStatus
 		return nil, status, err
 	}
 
+	if err := setResourceKey(&status, bundle, h.isNamespaced, set); err != nil {
```
ibuildthecloud, line 133: I think the set temp variable isn't needed.  Just change set here so `status.ObservedGenerate != bundle.Generation` and just always set observergeneration to generation before return.

pkg/controllers/git/git.go
```diff
@@ -237,14 +236,6 @@ func (h *handler) DeleteOnChange(key string, gitrepo *fleet.GitRepo) (*fleet.Git
 	return nil, nil
 }
 
-func (h *handler) isNamespaced(gvk schema.GroupVersionKind) bool {
-	mapping, err := h.mapper.RESTMapping(gvk.GroupKind(), gvk.Version)
```
ibuildthecloud, line 241: can the h.mapper struct member be deleted from the handler now?

pkg/controllers/git/git.go
```diff
@@ -46,7 +45,6 @@ func Register(ctx context.Context,
 	bundles fleetcontrollers.BundleController,
 	gitRepos fleetcontrollers.GitRepoController) {
 	h := &handler{
-		mapper:              mapper,
```
ibuildthecloud, line 49: The mapper argument should be dropped from Register now

PR #228, opened 2021-01-19
Add image scan feature

StrongMonkey:


pkg/controllers/image/image.go
```diff
@@ -0,0 +1,330 @@
...
+
+func Register(ctx context.Context, core corev1controler.Interface, gitRepos fleetcontrollers.GitRepoController, images fleetcontrollers.ImageScanController) error {
+	h := handler{
+		secretCache:  core.Secret().Cache(),
+		gitrepoCache: gitRepos.Cache(),
+	}
+
+	fleetcontrollers.RegisterImageScanStatusHandler(ctx, images, "ImageScanned", "image-scan", h.onChange)
+	return nil
+}
```
nickgerace, line 59: Does this function need to return an error?

pkg/controllers/image/image.go
```diff
@@ -0,0 +1,330 @@
...
+
+func (h handler) onChange(image *v1alpha1.ImageScan, status v1alpha1.ImageScanStatus) (v1alpha1.ImageScanStatus, error) {
+	if image == nil || image.DeletionTimestamp != nil {
+		return status, nil
+	}
+
+	if image.Spec.Suspend {
+		return status, nil
+	}
```
nickgerace, line 73: nit: You may be able to combine this conditionals.

pkg/controllers/image/image.go
```diff
@@ -0,0 +1,330 @@
...
+
+	canonical := ref.Context().String()
+	if canonical != status.CanonicalImageName {
+		status.CanonicalImageName = canonical
+	}
+
+	if !shouldScan(image) {
+		return status, nil
+	}
```
nickgerace, line 88: nit: This may able to be combined with the conditional on L67.

pkg/bundle/read.go
```diff
@@ -117,6 +119,20 @@ func read(ctx context.Context, name, baseDir string, bundleSpecReader io.Reader,
...
+		if scan.Image == "" {
+			continue
+		}
+		scans = append(scans, &fleet.ImageScan{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: fmt.Sprintf("%s-%s", name, name2.Hex(scan.Image, 5)),
+			},
+			Spec: scan,
+		})
+	}
```
nickgerace, line 134: Will `scans` always have at least one scan? Or, is it possible that scans could remain unitialized?

pkg/controllers/image/image.go
```diff
@@ -0,0 +1,338 @@
...
+
+	gitrepo, err := h.gitrepoCache.Get(image.Namespace, image.Spec.GitRepoName)
+	if err != nil {
+		kstatus.SetError(image, err.Error())
+		return status, err
+	}
+
+	lock.Lock()
+	defer lock.Unlock()
+	// todo: maybe we should preserve the dir
```
nickgerace, line 135: What would be the advantage of preserving the temporary directory? In case of error?

pkg/controllers/image/image.go
```diff
@@ -0,0 +1,338 @@
...
+
+			if version <= latest && !des {
+				latest = version
+			}
+		}
+		return latest, nil
+	default:
+		return semverLatest("*", versions)
+	}
+	return "", errors.New("failed to find the latest version")
```
nickgerace, line 319: Line 319 may be unreachable.

pkg/controllers/image/image.go
```diff
@@ -0,0 +1,338 @@
...
+	for _, version := range versions {
+		if ver, err := semver.NewVersion(version); err == nil {
+			if latestVersion == nil || ver.GreaterThan(latestVersion) {
+				if contraints.Check(ver) {
+					latestVersion = ver
+				}
+			}
+		}
+	}
+	return latestVersion.Original(), nil
```
nickgerace, line 337: Is it possible that `latestVersion` will remain uninitialized? Perhaps there should be an error returned if it's uninitialized.

pkg/controllers/image/image.go
StrongMonkey: This is separated for code readibility.

pkg/controllers/image/image.go
StrongMonkey: So it doesn't have to clone from scatch.

pkg/bundle/read.go
StrongMonkey: It could be empty though.

pkg/controllers/image/image.go
StrongMonkey: Good catch here :+1:

pkg/bundle/read.go
nickgerace: Good point

PR #231, opened 2021-01-21
Ensure deleteHistory is called when a bundle is deleted

boxboatmatt:

The helm deployer is currently never calling`deleteHistory` to delete the helm release history when a bundle is deleted.

### DIFF
--- a/pkg/helmdeployer/deployer.go
+++ b/pkg/helmdeployer/deployer.go
@@ -496,9 +496,14 @@ func (h *helm) delete(bundleID string, options fleet.BundleDeploymentOptions, dr
 		return err
 	}
 
+	// Never uninstall the fleet-agent, just "forget" it
 	if bundleID == "fleet-agent" {
-		// Never uninstall the fleet-agent, just "forget" it
-		return deleteHistory(cfg, bundleID)
+		return nil
+	}
+
+	err = deleteHistory(cfg, bundleID)
+	if err != nil {
+		return err
 	}
 
 	u := action.NewUninstall(&cfg)
--- a/pkg/helmdeployer/deployer.go
+++ b/pkg/helmdeployer/deployer.go
@@ -501,8 +501,7 @@ func (h *helm) delete(bundleID string, options fleet.BundleDeploymentOptions, dr
 		return nil
 	}
 
-	err = deleteHistory(cfg, bundleID)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


boxboatmatt:
@StrongMonkey I am seeing infinite helm release history with my fleet deployment... I think this bug is contributing to it

StrongMonkey:
Thanks I will take a look at your PR. Do you want to create an issue to track this?

pkg/helmdeployer/deployer.go
```diff
@@ -496,9 +496,14 @@ func (h *helm) delete(bundleID string, options fleet.BundleDeploymentOptions, dr
...
 
+	// Never uninstall the fleet-agent, just "forget" it
 	if bundleID == "fleet-agent" {
-		// Never uninstall the fleet-agent, just "forget" it
-		return deleteHistory(cfg, bundleID)
+		return nil
+	}
+
+	err = deleteHistory(cfg, bundleID)
+	if err != nil {
```
nickgerace, line 505: nit
```go
if err = deleteHistory(cfg, bundleID); err != nil {
```

nickgerace (approved):

StrongMonkey:
Is this PR still revelant? The history should be deleted when release is deleted.

boxboatmatt:
Something needs to change with those lines - right now it is only calling `deleteHistory` if the bundle is the `fleet-agent` which is completely opposite of what the comment says to do...

boxboatmatt:
It seems it is supposed to bail out of  the method entirely if the bundle is `fleet-agent` which is what this PR does

StrongMonkey:
https://github.com/rancher/fleet/pull/294 This will probably fix your issue. It appears that if you specify releaseName in fleet.yaml we don't clean up properly.

boxboatmatt:
There is still a logic problem here though, which ostensibly is solved by this PR... if this function is called without this change it only executes if it is the bundle is the `fleet-agent`

StrongMonkey:
The original idea is that if you delete fleet-agent bundle, we don't delete the actual release(instead we delete all the history). If it is other bundle, uninstall will be called and history will also get deleted. So
```
if bundleID == "fleet-agent" {
   deleteHistory()
} else {
   uninstallRelease() # this also deletes history
}
```

The previous issue you were seeing is that uninstallRelease() will never get called(becasue the code didn't use the right release name), and that's why history is not being cleaned up. This is fixed by #294

PR #245, opened 2021-01-26
Update to version 0.3.3 in README.md and mkdocs.yml

miguelpoyatosmora:


mkdocs.yml
```diff
@@ -47,5 +47,5 @@ nav:
 extra:
   fleet:
     apiVersion: fleet.cattle.io/v1alpha1
-    version: v0.3.0
+    version: v0.3.3
     helmversion: 0.3.0
```
nickgerace, line 51: Wondering if we have to worry about `helmversion`. CC: @StrongMonkey

StrongMonkey: yeah we should update this too. @miguelpoyatosmora Can you also update it

miguelpoyatosmora: Oh, thanks! what about this?

StrongMonkey: Thanks! LGTM now

StrongMonkey (approved):

PR #253, opened 2021-01-29
Fix dapper build distribution

nickgerace:
Relevant issue: https://github.com/rancher/rancher/issues/30516

### DIFF
--- a/scripts/build
+++ b/scripts/build
@@ -53,3 +53,6 @@ fi
 if ! [ "$LOCAL" = "false" ]; then
     build-binaries ${OS} ${ARCH}
 fi
+
+mkdir -p dist/artifacts
+cp -r bin/fleet* dist/artifacts/
--- a/scripts/package
+++ b/scripts/package
@@ -20,9 +20,6 @@ function build-image() {
   fi
 }
 
-mkdir -p dist/artifacts
-cp -r bin/fleet* dist/artifacts/
-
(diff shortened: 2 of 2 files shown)


scripts/build
```diff
@@ -53,3 +53,9 @@ fi
 if ! [ "$LOCAL" = "false" ]; then
     build-binaries ${OS} ${ARCH}
 fi
+
+# When only running "dapper build", you may have to copy the binaries to "dist".
+if [ "$COPY_TO_DIST" = "true" ]; then
```
StrongMonkey, line 58: Can this be the default behavior? When you run dapper build it is always going to cp to dist/artifacts

nickgerace: Sure thing, but we should remove it from `scripts/package` then. Thoughts?

StrongMonkey: yep, I think so

nickgerace: Ok great, will do

.drone.yml
```diff
@@ -194,6 +196,8 @@ steps:
 # Update this tag once dapper arm v7 works on drone.
 - name: build
   image: rancher/dapper:v0.4.1
+  environment:
```
StrongMonkey, line 199: Remove this?

StrongMonkey (approved):

luthermonson (approved):

Issue #236, opened 2021-01-22
Deploying the multi-cluster kustomize example results in bundle render error


janeczku:
Steps to reproduce:

1. Label a cluster with "env=dev"
2. Create a GitRepo via the UI pointing to the [Multi-Cluster Kustomize Example folder.](https://github.com/rancher/fleet-examples/tree/master/multi-cluster/kustomize)

```YAML
kind: GitRepo
apiVersion: fleet.cattle.io/v1alpha1
metadata:
  name: kustomize
  namespace: fleet-default
spec:
  repo: https://github.com/rancher/fleet-examples
  paths:
  - multi-cluster/kustomize
  targets:
  - name: dev
    clusterSelector:
      matchLabels:
        env: dev
```

**Expected:**

App is deployed to the cluster labeled "env=dev"

**Actual:**
App is not deployed. Bundle is stuck in error.

> error while running post render on files: Object 'Kind' is missing in '{"resources":["frontend-deployment.yaml","frontend-service.yaml","redis-master-deployment.yaml","redis-master-service.yaml","redis-slave-deployment.yaml","redis-slave-service.yaml"]}'

**Environment:**
Rancher v2.5.5

**Export of failing bundle:**

```
apiVersion: fleet.cattle.io/v1alpha1
kind: Bundle
metadata:
  creationTimestamp: "2021-01-22T13:56:00Z"
  generation: 1
  labels:
    fleet.cattle.io/commit: c6e54d7a56565e52a63de8a2088997b46253c1fb
    fleet.cattle.io/repo-name: test123
  managedFields:
  - apiVersion: fleet.cattle.io/v1alpha1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:labels:
          .: {}
          f:fleet.cattle.io/commit: {}
          f:fleet.cattle.io/repo-name: {}
      f:spec:
        .: {}
        f:namespace: {}
        f:resources: {}
        f:targetRestrictions: {}
        f:targets: {}
    manager: fleet
    operation: Update
    time: "2021-01-22T13:56:00Z"
  - apiVersion: fleet.cattle.io/v1alpha1
    fieldsType: FieldsV1
    fieldsV1:
      f:status:
        .: {}
        f:conditions: {}
        f:display:
          .: {}
          f:readyClusters: {}
        f:maxUnavailable: {}
        f:maxUnavailablePartitions: {}
        f:observedGeneration: {}
        f:summary:
          .: {}
          f:desiredReady: {}
[...]

PR #262, opened 2021-02-04
Ignore the error if file is not a k8s object

StrongMonkey:


StrongMonkey:
https://github.com/rancher/fleet/issues/236

pkg/rawyaml/resources.go
```diff
@@ -21,7 +21,7 @@ func ToObjects(c *chart.Chart) (result []runtime.Object, _ error) {
 		}
 		objs, err := yaml.ToObjects(bytes.NewBuffer(resource.Data))
 		if err != nil {
-			return nil, err
```
ron1, line 24: Thanks in advance for your patience while I familiarize myself with the Fleet codebase.

Is returning an error here helpful in some cases where malformed yaml is unexpected? 

Also, is it expected for processing of a kustomize workload to invoke a function in the rawyaml package?

StrongMonkey: Thanks for looking at it. I am still debating on this change. The problem with that issue is that we are marshalling non-k8s manifest(kustomize.yaml) into k8s object and that raised an error. It should probably need to exclude the file or ignore the error.

pkg/rawyaml/resources.go
```diff
@@ -21,7 +21,7 @@ func ToObjects(c *chart.Chart) (result []runtime.Object, _ error) {
...
+			continue
```
nickgerace, line 24: Reading from https://github.com/rancher/fleet/pull/262/files#r570424566 to get some context, should we at least log the error before continuing?

pkg/rawyaml/resources.go
ron1: Since filename `kustomization.yaml` has special meaning in the gitops space, would it be appropriate to explicitly exclude it rather than ignore all erroneous yaml manifests?

StrongMonkey: Yeah that's the alternative way. I will dicuss with @ibuildthecloud to see what's the best way to solve this

pkg/rawyaml/resources.go
ron1: Any idea why I would not encounter this error running Rancher 2.5.2?

StrongMonkey: @ron1 which version of fleet that you don't see this problem?

Issue #267, opened 2021-02-07
Fleet-agent bundle is in flapping state between `wait Applied` and `modified`


Ellenqs:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**
- Deploy rancher single install.
- Create a downstream cluster.
- Go to system project, find the workload fleet-agent, edit it, e.g. adding an environment variable.
- Go to cluster explorer ->continuous delivery -> bundles

**Result:**
The fleet-agent bundle state switches between `modified` and `waitApplied` state.
<img width="462" alt="Screen Shot 2021-02-06 at 3 23 22 PM" src="https://user-images.githubusercontent.com/75277058/107131956-0695ec00-6890-11eb-8087-38f4bb450c64.png">
<img width="425" alt="Screen Shot 2021-02-06 at 3 23 34 PM" src="https://user-images.githubusercontent.com/75277058/107131959-085faf80-6890-11eb-9f2e-4fcc1cb139cc.png">

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): `v2.5.6-rc4`
- Installation option (single install/HA): single install
- Fleet version: `v0.3.4-rc3`

StrongMonkey:
This is not easy to address due to the nature of fleet.(it uses bundle to manage fleet-agent so it has to make sure the actual state of fleet-agent matches the spec). I would put a requirement to not modify fleet-agent from outside.

sowmyav27:
@StrongMonkey Based on https://github.com/rancher/fleet/issues/267#issuecomment-844377177, should we release note this or have a docs issue logged for this?

StrongMonkey:
@sowmyav27 I think putting a note in doc should be enough.

Jono-SUSE-Rancher:
@sowmyav27 - Did you happen to log a docs issue for this yet, or do you need Sheilagh or myself to do that? Let me know, and if you did, go ahead and close this with the understanding that we're going to document it.

aaronRancher:
Logged a docs issue: https://github.com/rancher/docs/issues/3339

Issue #232, opened 2021-01-21
Infinite Helm Release History


boxboatmatt:
I am seeing infinite helm release history with my fleet install.

Somewhat related to rancher/fleet#233 but they appear to be different bugs. The MaxHistory 5 setting does not appear to be honored

StrongMonkey:
@boxboatmatt I can't reproduce this. I deleted the gitRepo(then bundle) and helm release history(stored as secrets in k8s) is deleted. Doesn't uninstall delete the history? Do you have extra steps to reproduce this?

StrongMonkey:
Also what does your gitrepo(manifest) look like? Is it raw k8s manifest, helm chart or kustomize?

boxboatmatt:
This particular bug isn't about deleting the bundle  - #233 covers that issue. This bug covers the issue where all of my fleet bundles retain all of the release history for each helm upgrade. I had 1000s of k8s secrets - rather than 5 per bundle

StrongMonkey:
@boxboatmatt Sorry commenting on the wrong issue, I will take a look at why setting history to 5 does not take effect.

StrongMonkey:
On the side note can you provide some steps to reproduce #233?

StrongMonkey:
@boxboatmatt Do you want to give this a try https://github.com/rancher/fleet/pull/243? Currently I don't think maxHistory can be set.

boxboatmatt:
I would try it out but there doesn't seem to be a way to permanently override the `fleet-agent` deployment image... It keeps reverting to `v0.3.0`

janeczku:
This is easily reproducable:

1. Patch the fleet-agent deployment resource in the downstream cluster using `kubectl edit`, e.g. adding an environment variable to the pod spec
This causes the fleet-agent bundle to enter into a flapping state between `modified` and `waitApplied`, resulting in approx. 15 Helm releases each minute.
2. Watch the number of helm releases growing to infinity (eventually killing the cluster once the etcd db is filled)

```
> helm -n fleet-system list
NAME            NAMESPACE       REVISION        UPDATED                                 STATUS          CHART                                                                                   APP VERSION
fleet-agent     fleet-system    3685             2021-02-05 18:05:24.368763503 +0000 UTC deployed        fleet-agent-v0.0.0+s-ad50b23019cb9d1ba3ba24283509ec06d3d237fa416bd74ac60c1b69809cb                 
> kubectl get secret -n fleet-system|wc -l
3694
> 
```

[...]

PR #271, opened 2021-02-09
Don't force to redeploy fleet agent

StrongMonkey:
Looks like we are force to reploy agent, which is causing https://github.com/rancher/fleet/issues/267. This can cause rapid CPU increase to fleet-agent as it is trying redeploying fleet-agent with no backoff.


StrongMonkey:
@ibuildthecloud What's the original purpose? Making sure fleet-agent is always update to date? Should we really treat fleet-agent with other bundle(so the default behavior is to redeploy after 5 minutes if state is modified)

StrongMonkey:
Related to https://github.com/rancher/fleet/issues/232

Issue #251, opened 2021-01-29
Document diff.comparePatches


janeczku:
The behavior and the proper use of `diff.comparePatches` need to be documented.

nickgerace:
Closed by https://github.com/rancher/fleet/pull/277

PR #277, opened 2021-02-12
Document diff.comparePatches #251

catherineluse:
https://github.com/rancher/fleet/issues/251

@shpwrck Denise asked me to document https://github.com/rancher/fleet/issues/251. I searched for comparePatches in Slack and found your explanation to Nick about how it works, so I incorporated your phrasing into this doc.

@janeczku Is this sufficient so that a newcomer could understand what they need to do?

Also, should this be in the reference on this page, or is it more appropriate in a troubleshooting section?


docs/gitrepo-add.md
```diff
@@ -128,3 +128,31 @@ stringData:
 
 Note: If you don't add it any server's public key will be trusted and added.
 
+# Troubleshooting
+
+### Fleet deployment stuck in modified state
+
+When you add a Helm chart to Fleet, some of the components are modified, and this causes the "modified" flag in the Fleet environment.
```
ron1, line 135: This is true for kustomize and raw yaml workloads in addition to helm.

docs/gitrepo-add.md
```diff
@@ -128,3 +128,31 @@ stringData:
...
+
+To ignore the modified flag for the differences between the Helm install generated by `fleet.yaml` and the resource in your cluster, add a `diff.comparePatches` to the `fleet.yaml` for your Deployment, as shown in this example:
```
ron1, line 137: If you specifically reference a Deployment resource, you should comparePatch a Deployment resource in the example yaml below such as https://github.com/ibrokethecloud/core-bundles/blob/21f9ab03f89b52ff3b60c52e1cfedcee7e23e8b0/monitoring/fleet.yaml#L45-51

shpwrck:
We can do better. Let me produce a code snippet for you.

docs/gitrepo-add.md
nickgerace: Are you suggesting that we generalize the language?

docs/gitrepo-add.md
nickgerace: I agree with the analysis. Do you have a suggested change here?

docs/gitrepo-add.md
ron1: Yes, no need to specifically mention helm here

catherineluse: Changed to

> When you add a workload to Fleet

docs/gitrepo-add.md
catherineluse: Changed code sample to:


```yaml
defaultNamespace: <namespace name> 
helm:  
  releaseName: <release name>  
  repo: <repo name> 
  chart: <chart name>
diff:  
  comparePatches:  
  - apiVersion: apps/v1Gaurav Mehta, 4 months ago: • Working Repo with common base charts
    kind: Deployment
    operations:
    - {"op":"remove", "path":"/spec/template/spec/hostNetwork"}
    - {"op":"remove", "path":"/spec/template/spec/nodeSelector"}
    - {"op":"remove", "path":"/spec/template/spec/priorityClassName"}
    - {"op":"remove", "path":"/spec/template/spec/tolerations"}
```

nickgerace (approved):

docs/gitrepo-add.md
StrongMonkey: I think this should
> When you deploy bundles to Fleet

Which we should use specific term in Fleet

catherineluse: Changed to

> When you deploy bundles to Fleet

catherineluse:
Added section about how to determine operations:


To determine which operations should be removed, observe the logs from `fleet-agent` on the target cluster. You should see entries similar to the following:

```text
level=error msg="bundle monitoring-monitoring: deployment.apps monitoring/monitoring-monitoring-kube-state-metrics modified {\"spec\":{\"template\":{\"spec\":{\"hostNetwork\":false}}}}"
```

Based on the above log, you can add the following entry to remove the operation:

```json
{"op":"remove", "path":"/spec/template/spec/hostNetwork"}
```

StrongMonkey (approved):

iosifnicolae2:
It would be great for Rancher Fleet to generate and display in the UI the `diff` configuration 😊 .

PR #291, opened 2021-02-22
Use Argo diff

StrongMonkey:


modules/agent/pkg/deployer/monitor.go
```diff
@@ -20,12 +31,153 @@ type DeploymentStatus struct {
...
+		}
+		var isNamespaced bool
+		for _, r := range apiResources.APIResources {
+			if r.Kind != gvk.Kind {
+				continue
+			}
+			if r.Namespaced {
+				isNamespaced = true
+			}
+		}
```
nickgerace, line 73: Should we set `isNamespaced` to `false` if it remains unset before being accessed?

modules/agent/pkg/deployer/normalizers/mutatingwebhook.go
```diff
@@ -0,0 +1,82 @@
...
+                unknown := v1.SideEffectClassUnknown
+                config.SideEffects = &unknown
+            }
+        }
+        if config.TimeoutSeconds == nil {
+            switch m.KubeVersion.Minor {
+            case "19":
+                config.TimeoutSeconds = &[]int32{30}[0]
+            case "20":
+                config.TimeoutSeconds = &[]int32{10}[0]
```
nickgerace, line 47: Do we need a default case?

modules/agent/pkg/deployer/normalizers/mutatingwebhook.go
```diff
@@ -0,0 +1,82 @@
...
+        webhook.Webhooks[i] = config
+    }
+    newObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&webhook)
+    if err != nil {
+        logrus.Error("Object unable to convert from secret")
+        return nil
+    }
+    if webhook.Webhooks != nil {
+        err = unstructured.SetNestedField(un.Object, newObj["webhooks"], "webhooks")
+        if err != nil {
```
nickgerace, line 76: nit:
```go
if err = unstructured.SetNestedField(un.Object, newObj["webhooks"], "webhooks"); err != nil {
```

StrongMonkey:
@nickgerace This is not quite ready for review. I will let you know when it is ready.

modules/agent/pkg/deployer/monitor.go
StrongMonkey: isNamespaced should be default to false

modules/agent/pkg/deployer/normalizers/mutatingwebhook.go
StrongMonkey: this is the wrong implemtation and I have changed

modules/agent/pkg/deployer/manager.go
```diff
@@ -62,7 +74,7 @@ func (m *Manager) Cleanup() error {
 		bundleDeployment, err := m.bundleDeploymentCache.Get(m.fleetNamespace, deployed.BundleID)
 		if apierror.IsNotFound(err) {
 			logrus.Infof("Deleting orphan bundle ID %s, release %s", deployed.BundleID, deployed.ReleaseName)
-			if err := m.deployer.Delete(deployed.BundleID, ""); err != nil {
+			if err := m.deployer.Delete(deployed.BundleID, deployed.ReleaseName); err != nil {
```
StrongMonkey, line 77: This is the place where bundle is deleted and we need to clean up release. If this is empty then we will try to delete release with the bundleId, which could leave orphan releases since bundleID is not the same as releaseName.

StrongMonkey: @ibuildthecloud

StrongMonkey:
https://github.com/rancher/rancher/issues/30696

modules/agent/pkg/deployer/monitor.go
```diff
@@ -20,12 +31,144 @@ type DeploymentStatus struct {
...
+	for _, obj := range objs {
+		result.Add(obj)
+	}
+
+	norms, err := m.normalizers(bd)
+	if err != nil {
+		return plan, err
+	}
+
+	for gvk, objs := range plan.Update {
```
ibuildthecloud, line 54: If you bump to the latest wrangler the sharedclientfactory (which you can obtain from the sharedcachefactory) there is a IsNamespaced method: https://github.com/rancher/lasso/blob/master/pkg/client/sharedclientfactory.go#L30

modules/agent/pkg/deployer/monitor.go
```diff
@@ -20,12 +31,144 @@ type DeploymentStatus struct {
...
+			if r.Kind != gvk.Kind {
+				continue
+			}
+			if r.Namespaced {
+				isNamespaced = true
+			}
+		}
+		var errs []error
+		if isNamespaced {
+			for k, v := range result[gvk] {
```
ibuildthecloud, line 76: Both if and else do that same loop, you can make this code more concise

ibuildthecloud: I'm lost on what you are doing here. Let me look back at this in a bit.

StrongMonkey: This two are not the same. This is to normalize the namespace field to have defaultNamespace or no namespace(if clusterScoped)

modules/agent/pkg/deployer/monitor.go
StrongMonkey: :+1:

modules/agent/pkg/deployer/monitor.go
```diff
@@ -21,33 +29,166 @@ type DeploymentStatus struct {
...
 	if err != nil {
 		return apply.Plan{}, err
 	}
-	return a.DryRun(objs...)
+	plan, err := a.DryRun(objs...)
+	if err != nil {
+		return plan, err
+	}
+
+	desired := objectset.ObjectByGVK{}
```
ibuildthecloud, line 45: This can just be objectset.NewObjectSet(objs...).ObjectsByGVK()

modules/agent/pkg/deployer/monitor.go
```diff
@@ -21,33 +29,166 @@ type DeploymentStatus struct {
...
+	for _, obj := range objs {
+		desired.Add(obj)
+	}
+
+	live := objectset.ObjectByGVK{}
```
ibuildthecloud, line 50: This can be `objectset.NewObjectSet(plan.Objects...).ObjectsByGVK()`

Issue #233, opened 2021-01-21
Helm Release History not cleared when bundle deleted


boxboatmatt:
When a bundle deployment is deleted the helm release history is not being cleared

Addressed by https://github.com/rancher/fleet/pull/231

`pkg/helmdeployer/deployer.go` is never calling `deleteHistory`.

boxboatmatt:
@StrongMonkey I think I saw this issue occurring on my cluster when I updated my `fleet.yaml` to use a different release name... As you noted in #232, a helm uninstall will remove all of the history. I wrote this bug mostly on the basis of the code having a logic bug see https://github.com/rancher/fleet/pull/231.

boxboatmatt:
I have a `GitRepo` configured to scan `/test`, I put a helm chart in `/test/fleet.yaml`. Fleet installs the bundle properly (interestingly I get 2 release secrets created - `.v1` and `.v2`). If I delete the bundle from the Rancher CD UI, the helm release does not get deleted  or uninstalled from the target cluster; however the `bundledeployment` does get deleted from the rancher cluster.

boxboatmatt:
Another observation - though this may be a separate bug. If you update the `fleet.yaml` `helm.releaseName` then the old helm release is not uninstalled.

StrongMonkey:
At this time I don't think we handle the case where you modify the release name. Does the behavior still happen when you don't modify the release name in fleet.yaml?

boxboatmatt:
Yes... if you delete the bundle from rancher the release is not uninstalled from the target.

StrongMonkey:
I still can't reproduce this issue.
The steps I took:
1. Create a git repo https://github.com/rancher/fleet-examples.git and path `multi-cluster/helm-external`.
2. Bundle and helm app are created.
3. Delete git Repo.
4. Releases are deleted.

@boxboatmatt Can you elobarate on the steps and what your manifest looks like?

StrongMonkey:
Basically we found a bug that if you specify releaseName in fleet.yaml, we don't clean up the release properly. For example: 
```
namespace: test
helm:
  # The release name to use. If empty a generated release name will be used
  releaseName: mongo
  chart: "mongodb"
  repo: "https://charts.bitnami.com/bitnami"
  version: ""
  force: false
  timeoutSeconds: 0
  values:
    auth:
      enabled: false
```
[...]

PR #294, opened 2021-02-24
Properly delete releases that have a custom name

ibuildthecloud:


pkg/helmdeployer/deployer.go
```diff
@@ -428,10 +428,23 @@ func (h *helm) Resources(bundleID, resourcesID string) (*deployer.Resources, err
...
+			if deployment.BundleID == bundleID {
+				releaseName = deployment.ReleaseName
+				break
+			}
+		}
+	}
+	if releaseName == "" {
+		// Never found anything to delete
+		return nil
 	}
```
nickgerace, line 446: nit: we could probably move this into the scope above since this logic proceeds the for loop.
```go
if releaseName == "" {
    deployments, err := h.ListDeployments()
    // snip
    for _, deployment := range deployments {
    // snip
    }
    if releaseName == "" {
        return nil
    }
}

jiaqiluo:
https://github.com/rancher/fleet/issues/233

Issue #144, opened 2020-11-05
SOPS encrypted resources


mumblez:
Hi,

I'm having trouble using sops encrypted resources, getting the error from the fleet-agent container:

```
time="2020-11-05T18:25:04Z" level=info msg="getting history for release test-manifests-repo-fleet"
time="2020-11-05T18:25:04Z" level=error msg="error syncing 'cluster-stores-test-01-83b14484851f/test-manifests-repo-fleet': handler bundle-deploy: failed to decrypt with sops: Error getting data key: 0 successful groups required, got 0, requeuing"
```

I have this project installed on the agent cluster - https://github.com/isindir/sops-secrets-operator
The idea is that the client / leaf cluster will decrypt (sops/kms) and generate the real secrets.

I take it fleet is also trying to do this, can you advise on how I can either configure fleet agent with the sops details or bypass sops processing by fleet and allow the client sops operator to handle?

Many Thanks

thirdeye-yusuf:
If it helps this is where it was added - https://github.com/rancher/fleet/commit/9aeda3141bf3e449b4e5baa139b2171c5e72eeb2
Though I don't see an obvious way to configure credentials.

Is it possible we can disable the feature until there's a way to configure it?, e.g. patch out the check that looks for `sops` in the file.

domhdp:
@mumblez Did you work out how to do this? I am trying to do the same thing.

mumblez:
I haven't had time unfortunately, but last I looked into it, in theory, if I fork and undo the changes in this https://github.com/rancher/fleet/commit/9aeda3141bf3e449b4e5baa139b2171c5e72eeb2

Then mess with the build to upload to your own registry and use your new registry in the helm deployment you should be able to get around this.

Though if there was a way to configure with fleet natively that would be fantastic (much preferred)

domhdp:
@mumblez Ok no worries. I think I will configure an IAM role for the fleet ServiceAccount and see if I can get decryption happening. (I am using EKS for the Rancher cluster)

I have decided to not use `isindir/sops-secrets-operator` as I think I would like to have the MAC validation feature of SOPS. I am looking at https://github.com/craftypath/sops-operator but I will see what fleet can do first.

mumblez:
@domhdp It might be a simple case of exposing the right environment variables, e.g. in your case the AWS ones so it knows how to auth with the KMS service, from what I remember AWS would be:

AWS_ACCESS_KEY_ID=...
[...]

PR #298, opened 2021-03-01
disable sops check and decryption

mumblez:
Fixes https://github.com/rancher/fleet/issues/144

Ignores `sops` related resources (allow downstream to handle)

Currently there is no way to configure sops via fleet (to pass in cloud KMS details) so would be good to allow downstream to handle, e.g. with https://github.com/isindir/sops-secrets-operator

### DIFF
--- a/pkg/apis/fleet.cattle.io/v1alpha1/bundle.go
+++ b/pkg/apis/fleet.cattle.io/v1alpha1/bundle.go
@@ -186,6 +186,7 @@ type BundleDeploymentOptions struct {
 	ForceSyncGeneration int64             `json:"forceSyncGeneration,omitempty"`
 	YAML                *YAMLOptions      `json:"yaml,omitempty"`
 	Diff                *DiffOptions      `json:"diff,omitempty"`
+	Decrypt             *bool             `json:"decrypt,omitempty"`
 }
 
 type DiffOptions struct {
--- a/pkg/controllers/manageagent/manageagent.go
+++ b/pkg/controllers/manageagent/manageagent.go
@@ -109,6 +109,9 @@ func (h *handler) getAgentBundle(ns string, cluster *fleet.Cluster) (runtime.Obj
 		Spec: fleet.BundleSpec{
 			BundleDeploymentOptions: fleet.BundleDeploymentOptions{
 				DefaultNamespace: h.systemNamespace,
+				// https://stackoverflow.com/a/28818489 - create reference to bool
+				// Decrypt: true,
+				Decrypt: &[]bool{true}[0],
 				Helm: &fleet.HelmOptions{
 					TakeOwnership: true,
 				},
--- a/pkg/manifest/output.go
+++ b/pkg/manifest/output.go
@@ -13,7 +13,7 @@ import (
 	"go.mozilla.org/sops/v3/decrypt"
 )
 
-func (m *Manifest) ToTarGZ() (io.Reader, error) {
+func (m *Manifest) ToTarGZ(decrypt bool) (io.Reader, error) {
 	buf := &bytes.Buffer{}
 	gz := gzip.NewWriter(buf)
 	w := tar.NewWriter(gz)
@@ -24,9 +24,12 @@ func (m *Manifest) ToTarGZ() (io.Reader, error) {
 			return nil, err
 		}
 
-		bytes, err = decryptSOPS(bytes)
-		if err != nil {
-			return nil, err
+		if decrypt {
+			bytes, err = decryptSOPS(bytes)
+
+			if err != nil {
+				return nil, err
+			}
 		}
 
 		if err := w.WriteHeader(&tar.Header{
--- a/pkg/render/helm.go
+++ b/pkg/render/helm.go
@@ -32,5 +32,12 @@ func ToChart(name string, m *manifest.Manifest, options fleet.BundleDeploymentOp
 		return nil, err
 	}
 
-	return m.ToTarGZ()
+	decrypt := false
+	if options.Decrypt == nil || *options.Decrypt {
+		decrypt = true
+	} else {
+		decrypt = false
+	}
+
+	return m.ToTarGZ(decrypt)
 }
--- a/charts/fleet-crd/templates/crds.yaml
+++ b/charts/fleet-crd/templates/crds.yaml
@@ -24,6 +24,9 @@ spec:
         properties:
           spec:
             properties:
+              decrypt:
+                nullable: true
+                type: boolean
               defaultNamespace:
                 nullable: true
                 type: string
@@ -391,6 +394,9 @@ spec:
                           nullable: true
                           type: object
                       type: object
+                    decrypt:
+                      nullable: true
+                      type: boolean
                     defaultNamespace:
                       nullable: true
                       type: string
@@ -861,6 +867,9 @@ spec:
                 type: string
               options:
                 properties:
+                  decrypt:
+                    nullable: true
+                    type: boolean
                   defaultNamespace:
                     nullable: true
                     type: string
@@ -1004,6 +1013,9 @@ spec:
                 type: string
               stagedOptions:
                 properties:
+                  decrypt:
+                    nullable: true
+                    type: boolean
                   defaultNamespace:
                     nullable: true
                     type: string
--- a/docs/gitrepo-structure.md
+++ b/docs/gitrepo-structure.md
@@ -85,6 +85,10 @@ helm:
 # Default: false
 paused: false
 
+# Decrypt SOPS encrypted resources / fields
+# Default: true
+decrypt: true
+
 rolloutStrategy:
     # A number or percentage of clusters that can be unavailable during an update
     # of a bundle. This follows the same basic approach as a deployment rollout
@@ -266,3 +270,4 @@ For Clusters:
 **Modified**: There are bundles in this cluster that are in Modified state.
 
 **Ready**: Bundles in this cluster have been deployed and all resources are ready.
+
--- a/pkg/apis/fleet.cattle.io/v1alpha1/zz_generated_deepcopy.go
+++ b/pkg/apis/fleet.cattle.io/v1alpha1/zz_generated_deepcopy.go
@@ -183,6 +183,11 @@ func (in *BundleDeploymentOptions) DeepCopyInto(out *BundleDeploymentOptions) {
 		*out = new(DiffOptions)
 		(*in).DeepCopyInto(*out)
 	}
+	if in.Decrypt != nil {
+		in, out := &in.Decrypt, &out.Decrypt
+		*out = new(bool)
+		**out = **in
+	}
 	return
 }
 


pkg/manifest/output.go
```diff
@@ -24,10 +24,11 @@ func (m *Manifest) ToTarGZ() (io.Reader, error) {
 			return nil, err
 		}
 
-		bytes, err = decryptSOPS(bytes)
-		if err != nil {
-			return nil, err
-		}
+		// Remove to allow downstream to handle
```
nickgerace, line 27: I think this would be better suited as a flag in the CRD.
CC: @StrongMonkey @kinarashah 

Perhaps under `BundleDeploymentOptions`, `DecryptSOPS` as `*bool` set to `true` in our defaults?
https://github.com/rancher/fleet/blob/master/pkg/apis/fleet.cattle.io/v1alpha1/bundle.go#L177

nickgerace (changes requested):

pkg/manifest/output.go
mumblez: hi @nickgerace , I added your recommendations, tested and confirmed it works locally but it's failing the tests here on github, not sure why (no access)

nickgerace: No problem @mumblez. Thank you for your quick changes. Before taking a look at the CI, I think there are some additional places we should look at. I'll make comments at those places.

mumblez: sure, np

mumblez: @nickgerace just wondering if you know where the sops decryption is happening by default, is it on the controller (`fleet` chart) or on the agent cluster (`fleet-agent`), or is it `fleet-agent` on the controller server?

mumblez: Hi @nickgerace, not sure if any comments were left, might be some permission issue.
(Note, on the side I'm trying to get sops to work correctly, I managed to self contain the `output.go` and can confirm passing in the right credentials and env vars works, but doing the same to the `fleet` deployment container makes no difference, maybe there's some other job / pod doing it but can't figure it out :( )

nickgerace: Hi @mumblez. Do not worry; still need to re-review. Been a busy week :)

mumblez: np

pkg/controllers/manageagent/manageagent.go
```diff
@@ -104,6 +104,7 @@ func (h *handler) getAgentBundle(ns string) ([]runtime.Object, error) {
 			Spec: fleet.BundleSpec{
 				BundleDeploymentOptions: fleet.BundleDeploymentOptions{
 					DefaultNamespace: h.systemNamespace,
+					Decrypt: true,
```
nickgerace, line 107: We should allow for the user to set this.

pkg/apis/fleet.cattle.io/v1alpha1/bundle.go
```diff
@@ -183,6 +183,7 @@ type BundleDeploymentOptions struct {
 	ForceSyncGeneration int64             `json:"forceSyncGeneration,omitempty"`
 	YAML                *YAMLOptions      `json:"yaml,omitempty"`
 	Diff                *DiffOptions      `json:"diff,omitempty"`
+	Decrypt             bool              `json:"decrypt,omitempty"`
```
nickgerace, line 186: We should use `*bool` since we want to know when it's unset. If `*bool` is `nil`, then we set the value to `true` to maintain existing behavior..

pkg/render/helm.go
```diff
@@ -32,5 +32,5 @@ func ToChart(name string, m *manifest.Manifest, options fleet.BundleDeploymentOp
 		return nil, err
 	}
 
-	return m.ToTarGZ()
+	return m.ToTarGZ(options.Decrypt)
```
nickgerace, line 35: Perhaps we add something above L35 like this...
```go
var decrypt bool
if options.Decrypt == nil {
    decrypt = true
} else {
    decrypt = options.Decrypt
}
```

There may be a more efficient way to do this though.
Regardless, we want to cover the default case: what happens under these three conditions before we execute the `ToTarGZ` method...

1. Unset
2. True
3. False

Here is a playground example: https://play.golang.org/p/TyIFUDuyruG
Here is the plaintext if that link becomes inaccessible:
```go
package main

import (
	"fmt"
)

func example1() {
	var input *bool
	var output bool
	if input == nil {
		output = true
	} else {
		output = *input
	}
	fmt.Printf("%+v\n", output)
}

func example2() {
	var input2 *bool
	var output2 bool

	value := false
	input2 = &value

	if input2 == nil {
		output2 = true
	} else {
		output2 = *input2
	}
	fmt.Printf("%+v\n", output2)
}

func main() {
	example1()
	example2()
}
```

nickgerace (changes requested):
I like where this is going. We need to handle the default case and ensure that the fields are customizable for the user though.

CC: @StrongMonkey @kinarashah 
Should we add this to the chart(s) at all? I think if we are going to maintain default behavior, then we do not need to change the charts; only the docs to show how this new `BundleDeploymentOptions` field works.

pkg/manifest/output.go
nickgerace: At a cursory glance, functions/methods that call `Deploy --> ToChart --> ToTarGZ` from `pkg/helmdeployer/deployer.go` [here](https://github.com/rancher/fleet/blob/master/pkg/helmdeployer/deployer.go#L150) are what we need to look out for. I think putting the field in `BundleDeploymentOptions` is a good place to start.

This method gets called twice:
1. The `agent` deployer will call this method once in `modules/agent/pkg/deployer/manager.go` [here](https://github.com/rancher/fleet/blob/f8008e0437b7c77c08e9a3e840e3d19142fa6537/modules/agent/pkg/deployer/manager.go#L113). I believe this answers the `agent` part of your question.
2. The `helmdeployer` template will call this method once in `pkg/helmdeployer/template.go` [here](https://github.com/rancher/fleet/blob/cfbe1dce4e6a5d836ff949b2eee33dbde4da79ec/pkg/helmdeployer/template.go#L17). **However,** this is where we need to make sure we understand why/when this is called. Let's continue below...

Following the `helmdeployer` path, we trace to this unit test.
We can optionally add the field here. I do not think it is necessary though.
```modules/cli/match/match.go <-- modules/cli/cmds/test.go```

The other trace leads us down this path, which is where we need to look.
```pkg/controllers/bundle/controller.go```
Within this controller, we have where the options are calculated [here](https://github.com/rancher/fleet/blob/ffa88d184cab87c4a53c17c65df45c0d213358c3/pkg/controllers/bundle/controller.go#L156).
Keeping this in mind, we have two ways to go from here.

1. Add a field to the `GitRepo` CRD since that is our user-facing CRD. `Bundle` is an internal unit. We can essentially at a similar `*bool` field, and set to `true` if it is unset.
2. Add value to the Fleet chart(s) for this option. I prefer the first option since affecting the entire chart seems to be unnecessary. The benefit to this approach is that you have a "global" state on how to handle `sops`.

I hope this mini-rant helps :)
What do you think? I'm thinking we should go with option 1 above.

CC: @StrongMonkey @kinarashah

**Comment by nickgerace** [comment]
@mumblez adding my comment here in case it gets lost...
https://github.com/rancher/fleet/pull/298#discussion_r587742678

File: pkg/apis/fleet.cattle.io/v1alpha1/bundle.go (hunk 3 above)
Comment (nickgerace) [reply pkg/apis/fleet.cattle.io/v1alpha1/bundle.go:186]: Actually, if we create a `GitRepo` field of `*bool` type, this can actually be `bool` since we will set its value based on the former.

File: pkg/render/helm.go (hunk 4 above)
Comment (nickgerace) [reply pkg/render/helm.go:35]: Relevant note to consider. This functionality needs to happen, but could be where `GitRepo` CRs are reconciled.
See comments:
- https://github.com/rancher/fleet/pull/298#discussion_r587747643
- https://github.com/rancher/fleet/pull/298#discussion_r587742678

File: pkg/manifest/output.go (hunk 1 above)
Comment (mumblez) [reply pkg/manifest/output.go]: @nickgerace thank you, that makes perfect sense, setting in `GitRepo` is a cleaner solution, I will try to refactor taking your points onboard, will keep you posted.

Comment (mumblez) [reply pkg/manifest/output.go]: Hi @nickgerace 

I think it might be better if I try to pass in the sops credentials via helm vs adding this `decrypt` var to numerous places, in theory there won't be any code or logic changes as by default it will try to decrypt only if `sops` related resources are found.

I assume it's the fleet controller (management server) that is doing the decryption, but when I amend it's helm chart to mount the secret and expose the environment variable it makes no difference, after fleet controller extension there is also fleet agent on the management server, I assume it is this that is doing the decryption?, is the controller installing the agent itself and if so how would you recommend I pass down the sops credentials?

Or is it the downstream fleet agent that is doing the decryption?

Many Thanks

Comment (nickgerace) [reply pkg/manifest/output.go]: >  in theory there won't be any code or logic changes as by default it will try to decrypt only if sops related resources are found

I could be wrong here, but isn't this the current behavior?

```go
func decryptSOPS(data []byte) ([]byte, error) {
	if !bytes.Contains(data, []byte("sops:")) {
		return data, nil
	}
        // --snip--
}
```

As far as your idea goes, we would only need `decrypt` in one place, and have it defined in two CRDs, which I don't think is big overhead. However, I'll answer your questions to help us confirm that your second idea might be a better one :) 

> I assume it's the fleet controller (management server) that is doing the decryption, but when I amend it's helm chart to mount the secret and expose the environment variable it makes no difference, after fleet controller installation there is also fleet agent on the management server (installed by fleet controller I assume), I assume it is this that is doing the decryption?

I believe it is the `fleet-agent` from my previous trace of the codebase. Just curious, for you "allow(ing) downstream to handle" sops, does it matter which pod we are concerned with? I'm curious to the motivation behind this question.

> is the controller installing the agent itself and if so how would you recommend I pass down the sops credentials?
> Or is it the downstream fleet agent that is doing the decryption?

Admittedly, I am unsure. I can get back to you with an exact answer here. Regardless of which agent performs sops decryption, if you're going with this second approach (assuming we still use `GitRepo` as the payload), then we would most likely store the credentials as a `Secret` that eventually becomes a field in the `Bundle` or `BundleDeployment` (or other) CRD.

pkg/apis/fleet.cattle.io/v1alpha1/bundle.go
StrongMonkey: agree, it should be type *bool. so
```
if decrypt == nil || *decrypt {
  decrypt()
}
```

nickgerace: The only case where it doesn't need to be `*bool` though is if GitRepo is `*bool`. Theoretically, we could evaluate the GitRepo field and make this field a `bool`. Or, we could make this field `*bool` as well and copy the GitRepo field value directly to this one.

 I am fine either way.

pkg/manifest/output.go
nickgerace: Alright, I spoke to @StrongMonkey and while your new idea is really good and clean, I think we should go with the original approach of adding a `*bool` to GitRepo. The new approach of passing SOPS credentials via Helm is great, but we are evaluating how we do SOPS decryption overall, and I'd prefer to not overload this PR and send you in multiple directions. Adding the bool field/flag should be sufficient for now, I believe. Does that sound good to you? @mumblez

mumblez: yes, this is the current behaviour (if `sops:` is detected in the resource)

> I believe it is the fleet-agent from my previous trace of the codebase. Just curious, for you "allow(ing) downstream to handle" sops, does it matter which pod we are concerned with? I'm curious to the motivation behind this question.

Am just trying to work out which container is doing the decryption so I can test and verify mounting the credentials and exporting the env var works. (Preferably this happens on the controller vs downstream), then I can work out how to amend the helm charts and configs to supply this. 

Currently we can helm deploy the `fleet-controller` and `fleet-agent` and configure their helm charts, but when `fleet-controller` also sets it is local `fleet-agent`; I'm unsure how to feed it the sops credentials, ideally we can just amend the helm chart / k8s manifest but we can't do that for `fleet-agent` on the control server.

pkg/controllers/manageagent/manageagent.go
nickgerace: This can be resolved. @StrongMonkey: "We don't manage agent bundle in git. There won't be a chance that agent bundle is encrypted"

nickgerace (changes requested):
So, as a summary to our prior discussions, let's continue with the `*bool` field in `BundleDeploymentOptions` approach (configured by `fleet.yaml`). I think once that's in place, this should be good to go.

pkg/manifest/output.go
mumblez: > Alright, I spoke to @StrongMonkey and while your new idea is really good and clean, I think we should go with the original approach of adding a *bool to BundleDeploymentOptions. The new approach of passing SOPS credentials via Helm is great, but we are evaluating how we do SOPS decryption overall, and I'd prefer to not overload this PR and send you in multiple directions. Adding the bool field/flag should be sufficient for now, I believe. Does that sound good to you? @mumblez

Sure, np, I'm just a little stuck on how to access from the gitrepo config, e.g. I can add the value to the `GitRepoSpec` struct in `pkg/apis/fleet.cattle.io/v1alpha1/git.go` but not sure how to access from the functions / bundles, any help appreciated.

nickgerace: Actually, you were correct the first time and I led you astray by accident @mumblez. Apologies for that; long week with release.

The `BundleDeploymentOptions` should be configurable via `fleet.yaml`, like in the docs file you have above. We should not add anything to `GitRepo`, like I initially suggested. Apologies for that.

nickgerace: Also, I see what you mean. I believe both the `fleet-controller` and the _local_ `fleet-agent` perform decryption. I am confirming whether or not the downstream ones do as well, but I believe that _not_ to be the case.

pkg/apis/fleet.cattle.io/v1alpha1/bundle.go
mumblez: awesome, apologies I missed the message, just pushed the change to handle nil and make it a pointer, hope it's alright.

mumblez:
@nickgerace awesome, pushed the changes, hope it's alright.

nickgerace:
Creating a related issue. Will get back to this PR soon

mumblez:
awesome :)

nickgerace:
Hi @mumblez - your PR sparked an internal discussion on handling SOPS for the future in Fleet, so thank you for the PR and important community feedback. The decision that we think may be the best way to go, based on our discussions publicly in this PR and internally, is to use environment variables to pass the boolean variable rather than affect the API (CRDs). This is because we do not want to make a change to the API for a field that may not always exist (e.g. what if Fleet's SOPS story changes over time?).

I've opened a draft PR (https://github.com/rancher/fleet/pull/308/files) and made you a co-author on the first commit (and will on the final commit when I squash potential future commits). This has not been tested yet, nor has the code been validated. It's an early WIP. Apologies if these changes come late. Your effort was not in vain as this PR gave us an overview on where we would like to go. I'll reach out once the other PR is ready. Thank you.

fristedt:
> Hi @mumblez - your PR sparked an internal discussion on handling SOPS for the future in Fleet, so thank you for the PR and important community feedback. The decision that we think may be the best way to go, based on our discussions publicly in this PR and internally, is to use environment variables to pass the boolean variable rather than affect the API (CRDs). This is because we do not want to make a change to the API for a field that may not always exist (e.g. what if Fleet's SOPS story changes over time?).
> 
> I've opened a draft PR (#308) and made you a co-author on the first commit (and will on the final commit when I squash potential future commits). This has not been tested yet, nor has the code been validated. It's an early WIP. Apologies if these changes come late. Your effort was not in vain as this PR gave us an overview on where we would like to go. I'll reach out once the other PR is ready. Thank you.

@nickgerace Is there anything you can share regarding the internal discussion regarding Fleet and SOPS? We really need something like SOPS for our use case and we would prefer not to handle this in the downstream clusters.

nickgerace:
Hi @fristedt - RE: https://github.com/rancher/fleet/pull/298#issuecomment-823179285

We have been discussing a greater strategy for SOPS and how we want to handle it. We've been focused on the v2.5.8 release of Rancher and Fleet v0.3.5 recently, and I hope that we return on the subject. Feel free to discuss and promote your desires for Fleet SOPS support in the [users Slack](https://slack.rancher.io/) in order to help us prioritize this. Thank you for your patience.

StrongMonkey:
@mumblez Are you able to rebase this PR? I think we decided that still go with disabling sops per bundle. Your PR looks good and If you rebase it I will put LGTM.

mumblez:
sure, I've merged the latest release `v0.3.5`, is this ok?

StrongMonkey:
@mumblez You should just do a git rebase on master. After rebase can you squash your commits into one commit?

mumblez:
@StrongMonkey ok, squashed down to 1 commit and merged with master, hope this is alright.

mumblez:
wait, it doesn't look right....

mumblez:
:( I think I messed up and lost all my changes :(

mumblez:
@StrongMonkey ok I cleaned up the commits :) (sorry, had to learn to squash and rebase)

pkg/render/helm.go
```diff
@@ -32,5 +32,12 @@ func ToChart(name string, m *manifest.Manifest, options fleet.BundleDeploymentOp
 		return nil, err
 	}
 
-	return m.ToTarGZ()
+	var decrypt bool
+	if *options.Decrypt == nil || *options.Decrypt {
+		decrypt = true
+	} else {
+		decrypt = false
+	}
```
nickgerace, line 40: For pointer safety, I would prefer that we set it to false by default.
```go
decrypt := false
if *options.Decrypt == nil || *options.Decrypt {
    decrypt = true
}
```

pkg/controllers/manageagent/manageagent.go
```diff
@@ -109,6 +109,9 @@ func (h *handler) getAgentBundle(ns string, cluster *fleet.Cluster) (runtime.Obj
 		Spec: fleet.BundleSpec{
 			BundleDeploymentOptions: fleet.BundleDeploymentOptions{
 				DefaultNamespace: h.systemNamespace,
+				// https://stackoverflow.com/a/28818489 - create reference to bool
+				// Decrypt: true,
+				Decrypt: &[]bool{true}[0],
```
nickgerace, line 114: @StrongMonkey: thoughts here? Looks good to me, but I have not seen this pattern before.

nickgerace (changes requested):
One small change would be great. Thanks for your patience! Apologies that this has taken a long time.

pkg/render/helm.go
mumblez: github is having some issues today, hopefully the change will show up soon (already pushed on my branch)

pkg/controllers/manageagent/manageagent.go
StrongMonkey: Yeah It is fine, I always use this to create pointers too.

nickgerace (approved):

pkg/render/helm.go
```diff
@@ -32,5 +32,12 @@ func ToChart(name string, m *manifest.Manifest, options fleet.BundleDeploymentOp
 		return nil, err
 	}
 
-	return m.ToTarGZ()
+	var decrypt bool
+	if *options.Decrypt == nil || *options.Decrypt {
```
StrongMonkey, line 36: Shouldn't it be 
```
if options.Decrypt == nil || *options.Decrypt {
```

StrongMonkey:
@mumblez Can you fix the CI? I believe there are syntax errors and it is not compiling.

pkg/render/helm.go
mumblez: yes, pushed a fix

mumblez:
@StrongMonkey I've fixed some of the issues but not quite sure what the error is now (I've merged the latest changes from Master again)

mumblez:
@StrongMonkey ok, sorry about that, I think I've got everything passing now, hope it looks ok.

indstri:
Hi!  Any updates on this being included?  It would unblock our org for secrets until/if Fleet supports something native akin to Flux.

strowi:
Whats the state of this after 1 year? 
I am trying to deploy sops-secrets, but can't because fleet will interfere and fail if a SopsSecret resource or sops-encrypted helm-values is present.

manno:
We removed SOPS support:
https://github.com/rancher/fleet/pull/1205

PR #309, opened 2021-03-15
Add cloudprovider tolerations to fleet-agent

StrongMonkey:
https://github.com/rancher/rancher/issues/30471


nickgerace (approved):
LGTM, I don't think there are other places we need to add this

Issue #188, opened 2020-12-19
Add helm.valueFiles string array to fleet.yaml


ron1:
Add `helm.valueFiles` string array to `fleet.yaml` for the following reasons:

1. In-lining Helm values within `fleet.yaml` makes them inaccessible for traditional testing using the helm client binary (Note that using the fleet CLI for this purpose is currently undocumented, is missing a --dry-run option, and has some issues running stand-alone such as https://github.com/rancher/fleet/issues/186)
2. In-lining large values within `fleet.yaml` makes this resource difficult to comprehend
3. Eases migration from Argo-CD `application.yaml` `spec.source.helm.valueFiles` (see https://github.com/argoproj/argo-cd/blob/45b3e48dfc607194e0498a0d654e0df2a624060d/docs/operator-manual/application.yaml#L35)

virtuman:
+1

Absolutely agree with all points and naming convention (over previously suggested values.file in ticket #179)

paulchoi:
+1
Helm has `--set-file` parameter, for large input values. Would like to see something like that in Fleet.

jmatis-bullhorn:
+1

paralika:
+1
It would be helpful.

jinzishuai:
+1

Amos-85:
+1

thecrazyrussian:
+1
slight change to use "ValuesFrom" in the Kubernetes parlance but same concept and implementation.

TristanFAURE:
+1

rsemlal-tas:
+1

izaac:
Fleet `v0.3.5`

I am using a modified version of `multi-cluster/helm` from the `fleet-examples` repo to test the `valuesFile`

```
helm git:(valuesFiles) tree
.
├── Chart.yaml
├── README.md
├── fleet.yaml
├── templates
│   ├── frontend-deployment.yaml
│   ├── frontend-service.yaml
│   ├── redis-master-deployment.yaml
│   ├── redis-master-service.yaml
│   ├── redis-slave-deployment.yaml
│   └── redis-slave-service.yaml
├── values
│   └── override.yaml
└── values.yaml
```

```yaml
namespace: fleet-mc-helm-example
targetCustomizations:
- name: dev
  helm:
    # values:
    #   replication: false
    valuesFile:
    - values/override.yaml
  clusterSelector:
    matchLabels:
      env: dev
```

```
cat values/override.yaml
replication: false
```

[...]

PR #316, opened 2021-03-26
valuesFiles support

ibrokethecloud:
Minor tweak to `fleet apply` logic to support valuesFiles in HelmOptions.

The `fleet apply` will now parse the list in valuesFiles and merge them into the `values` for the HelmOptions.

This allows users to specify values files to better organize their GitRepos rather than having all values inline.

If the same value override is defined in multiple valuesFiles, then the one in the last file will take precedence.

Helm handles setting the same variable multiple times in the same manner.

This allows user to reference valuesFiles in the fleet.yaml

For example:

```
defaultNamespace: loki
helm:
  releaseName: loki
  repo: https://kubernetes-charts.banzaicloud.com
  chart: loki
  values:
    key1: value1
  valuesFiles:
  - values/override.yaml
targetCustomizations:
- name: dev
  helm:
    valuesFiles:
    - values/override2.yaml
  clusterSelector:
    matchLabels:
      name: local
```

When the bundle is generated the valuesFiles are read and appended to the existing override values passed to the helm chart.

The generated bundle would look like..

```
kind: Bundle
metadata:
  labels:
    fleet.cattle.io/commit: 3b641674c670c621e9e1828dab22faffc71d341e
  name: loki-loki
  namespace: fleet-default
spec:
  defaultNamespace: loki
  helm:
    chart: loki
    releaseName: loki
    repo: https://kubernetes-charts.banzaicloud.com
    values:
      key1: value1
      car: honda
      fuel: petrol
      some: value
    valuesFiles:
    - values/override.yaml
  .....
```

Where the contents of the values/override.yaml are

```
some: value
car: honda
fuel: petrol
```

### DIFF
--- a/pkg/bundle/resources.go
+++ b/pkg/bundle/resources.go
@@ -24,6 +24,7 @@ import (
 	"github.com/rancher/fleet/pkg/content"
 	"golang.org/x/sync/errgroup"
 	"golang.org/x/sync/semaphore"
+	v2 "gopkg.in/yaml.v2"
 )
 
 func readResources(ctx context.Context, spec *fleet.BundleSpec, compress bool, base string) ([]fleet.BundleResource, error) {
@@ -34,7 +35,7 @@ func readResources(ctx context.Context, spec *fleet.BundleSpec, compress bool, b
 		return nil, err
 	}
 
-	var chartDirs []*fleet.HelmOptions
+	var chartDirs, parsedChartDirs []*fleet.HelmOptions
 
 	if spec.Helm != nil && spec.Helm.Chart != "" {
 		chartDirs = append(chartDirs, spec.Helm)
@@ -44,9 +45,26 @@ func readResources(ctx context.Context, spec *fleet.BundleSpec, compress bool, b
 		if target.Helm != nil && target.Helm.Chart != "" {
 			chartDirs = append(chartDirs, target.Helm)
 		}
+		if target.Helm != nil && target.Helm.Chart == "" && len(target.Helm.ValuesFiles) != 0 {
+			// generate values //
+			parsedChart, err := parseValueFiles(base, target.Helm)
+			if err != nil {
+				return nil, err
+			}
+			target.Helm.Values = parsedChart.Values
+		}
 	}
 
-	directories, err = addCharts(directories, base, chartDirs)
+	// append helm valuesFiles into values
+	for _, chart := range chartDirs {
+		parsedChart, err := parseValueFiles(base, chart)
+		if err != nil {
+			return nil, err
+		}
+		parsedChartDirs = append(parsedChartDirs, parsedChart)
+	}
+
+	directories, err = addCharts(directories, base, parsedChartDirs)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


pkg/bundle/resources.go
```diff
@@ -306,3 +324,52 @@ func readContent(ctx context.Context, progress *progress.Progress, base, name st
 
 	return files, nil
 }
+
+func parseValueFiles(base string, chart *fleet.HelmOptions) (parsedChart *fleet.HelmOptions, err error) {
```
ron1, line 328: Rename parseValuesFiles to be consistent with field name.

ibrokethecloud:
Related to: https://github.com/rancher/fleet/issues/188

pkg/bundle/resources.go
```diff
@@ -306,3 +324,52 @@ func readContent(ctx context.Context, progress *progress.Progress, base, name st
...
+	if chart.Values != nil {
+		valuesMap = chart.Values.Data
+	}
+	for _, value := range chart.ValuesFiles {
+		valuesByte, err := ioutil.ReadFile(base + "/" + value)
+		if err != nil {
+			return nil, err
+		}
+		tmpMap := make(map[string]interface{})
+		err = v2.Unmarshal(valuesByte, tmpMap)
```
StrongMonkey, line 356: There are existing yaml package from k8s sigs.k8s.io/yaml. Can we just use this without introducing new yaml package?

pkg/bundle/resources.go
```diff
@@ -306,3 +324,52 @@ func readContent(ctx context.Context, progress *progress.Progress, base, name st
...
+			parsedChart.Values = &fleet.GenericMap{}
+		}
+		parsedChart.Values.Data = valuesMap
+	}
+
+	return parsedChart, nil
+}
+
+func generateValues(base string, chart *fleet.HelmOptions) (valuesMap map[string]interface{}, err error) {
+
```
StrongMonkey, line 345: Nit: empty space

pkg/bundle/resources.go
```diff
@@ -34,7 +35,7 @@ func readResources(ctx context.Context, spec *fleet.BundleSpec, compress bool, b
 		return nil, err
 	}
 
-	var chartDirs []*fleet.HelmOptions
+	var chartDirs, parsedChartDirs []*fleet.HelmOptions
```
StrongMonkey, line 38: Why do you need extra `parsedChartDirs`? Can we just use chartDirs, but overriding chartDir.Helm.Values? since you are already modifying/overriding the helm.Values.

ibrokethecloud: tweaked based on feedback.

pkg/bundle/resources.go
ibrokethecloud: cleaned up.

pkg/bundle/resources.go
ibrokethecloud: tweaked the parsing logic to use k8s yaml

pkg/bundle/resources.go
```diff
@@ -37,11 +37,18 @@ func readResources(ctx context.Context, spec *fleet.BundleSpec, compress bool, b
...
 	if spec.Helm != nil && spec.Helm.Chart != "" {
+		if err := parseValueFiles(base, spec.Helm); err != nil {
+			return nil, err
+		}
 		chartDirs = append(chartDirs, spec.Helm)
 	}
 
 	for _, target := range spec.Targets {
-		if target.Helm != nil && target.Helm.Chart != "" {
+		err := parseValueFiles(base, target.Helm)
```
StrongMonkey, line 47: In here you still need to check if target.Helm is nil, otherwise it will panic

ibrokethecloud: Tweaked the logic.

StrongMonkey (approved):
💯

StrongMonkey:
https://github.com/rancher/fleet/issues/188

Issue #187, opened 2020-12-18
fleet.yaml tries to get a helm chart via an ssh url


aceeric:
Define a  gitlab repository containing a `fleet.yaml` like so:
```
helm:
  chart: git@gitlab....:vmware-tanzu/helm-charts/...
```
And reference this repo in a `GitRepo` resource. The `GitRepo` resource also references a `clientSecretName` with the SSH credentials needed to talk to GitLab.

The resulting fleet `Job` has a `step-git-source` container and a `fleet` container. The `step-git-source` container mounts the SSH secret and successfully clones the repo containing the `fleet.yaml`. Then the `fleet` container tries to git glone the helm chart referenced in the `fleet.yaml` and fails.

Notice that the `fleet` container in the `Job` never mounts the SSH secret and so it can't negotiate the handshake with GitLab to get the helm chart.

Is it intended to not support the ssh protocol for getting a helm chart?

ron1:
Similar to issue https://github.com/rancher/fleet/issues/120 but for ssh.

m4ce:
can we also specify a git branch..?

aceeric:
This relates to v0.3.1

ron1:
The GitRepo `helm.chart` value is a go-getter url according to https://fleet.rancher.io/gitrepo-structure/#reference. The go-getter URL format is described here: https://github.com/hashicorp/go-getter#url-format. Yes, this format does allow specification of a ref within git urls.

jackivanov:
Is ssh and git already supported? none of the go-getter urls work for me on 0.3.5

```
time="2021-06-18T17:30:17Z" level=fatal msg="Get \"git::ssh://git@domain.com:7999/test/fleet/nginx/index.yaml\": unsupported protocol scheme \"git\""

time="2021-06-18T17:35:33Z" level=fatal msg="parse \"git@domain.com:7999/test/fleet//nginx/index.yaml\": first path segment in URL cannot contain colon"

time="2021-06-18T17:37:54Z" level=fatal msg="failed to read helm repo from https://domain.com/scm/test/fleet//nginx/index.yaml, error code: 501, response body: Invalid request from host xxxx: Note that Bitbucket does not support Git's legacy HTTP transport protocol."

time="2021-06-18T17:40:15Z" level=fatal msg="Get \"git::https://domain.com/scm/test/fleet//nginx/index.yaml\": unsupported protocol scheme \"git\""

[...]

Issue #120, opened 2020-10-09
Auth for private Helm registries


shane-davidson:
There was nothing in the documentation in regards to using authentication for a helm chart registry.

```
helm:
    secret: secretName
```

Can we add the ability to either add a secret or username/password for private registries

gz#14091
gz#14979

ron1:
Also add ssh auth for charts with ssh go-getter urls.

StrongMonkey:
QA: https://github.com/rancher/fleet/pull/334 docs PR are here for testing.

StrongMonkey:
Available to test in v0.3.5-rc4

bmdepesa:
rancher/rancher:v2.5.8-rc3
fleet v0.3.5-rc4

Tested 3 scenarios
* Create a Helm repo requiring basic auth
* Create a Helm repo with a custom CA and basic auth
* Reference a Helm chart using `ssh` protocol (go-getter URL)

### For the first two scenarios requiring basic auth:
* Reference the helm repo in a fleet bundle:
```
helm:
  releaseName: guestbook
  chart: "guestbook"
  repo: "https://xx.xx.xx.xx.xip.io"
```
* Create secrets in the same namespace as the GitRepo object, supplying the user, password, and CA if needed:
```
kubectl create secret -n fleet-default generic helm-secret \
  --from-literal=username=user1 \
  --from-literal=password=testabc123$%^ \
  --from-file=cacerts=ca.pem
```
* In the GitRepo, for `helmSecretName`, specify the secret
```
kind: GitRepo
apiVersion: fleet.cattle.io/v1alpha1
metadata:
  name: helm
  namespace: fleet-default
spec:
  branch: helmtest
  # bad secret name
  helmSecretName: helm-secret
  repo: https://git-repo-url
  paths:
  - single-cluster/helm
  targets:
    - clusterSelector: {}
```
* Save the GitRepo and the bundle deploys correctly

### For the Helm chart using `ssh`:
* Tested with both `OpenSSH` and `RSA` keys
* Create the secret in the same namespace as the GitRepo:
```
kubectl create secret -n fleet-default generic helm-ssh \
  --from-file=ssh-privatekey=key.pem
```
* Reference the chart in the bundle:
```
helm:
  releaseName: guestbook
  chart: "git@github.com:privategit/privaterepo.git/guestbook-0.0.0.tgz"
  repo: ""
```
* Save the GitRepo 
* The bundle deploys correctly

### Notes / Issues
[...]

PR #321, opened 2021-03-31
Add basic auth support for private helm repo

StrongMonkey:


pkg/bundle/resources.go
```diff
@@ -150,23 +147,9 @@ func chartURL(location *fleet.HelmOptions, auth auth) (string, error) {
...
-					return nil, err
-				}
-				secret, err := c.Core.Secret().Get(client.Namespace, chart.SecretName, metav1.GetOptions{})
-				if err != nil {
-					return nil, err
-				}
-				auth.username = string(secret.Data[corev1.BasicAuthUsernameKey])
-				auth.password = string(secret.Data[corev1.BasicAuthPasswordKey])
-			}
-
```
nickgerace: Should we keep this as a flag for existing users?

StrongMonkey: @nickgerace Yeah, I change the design. Now these are flags `--username-file` `--password-file` and `--cabundle-file`

StrongMonkey:
This PR adds the ability to specify a `helmSecretName` for gitrepo to support http basic auth and custom CABundle for exteral helm repo. The helm secret should have the following data
```
spec:
  username: 
  password:
  cacerts:
```
For QA, I use chartmusuem to test this. For reference

```
chartmuseum --debug --port=8080 --storage="local"  --storage-local-rootdir="./chartstorage" --basic-auth-user=foo --basic-auth-user=foo --basic-auth-pass=bar --tls-cert=/home/daishan/self-sign-certs/SERVER.crt --tls-key=/home/daishan/self-sign-certs/mydomain.com.key
```

StrongMonkey:
https://github.com/rancher/fleet/issues/120

cloudnautique:
@StrongMonkey  is there going to be a docs PR for this as well?

StrongMonkey:
@cloudnautique Yes, after this PR is merged

modules/cli/cmds/apply.go
```diff
@@ -58,6 +62,27 @@ func (a *Apply) Run(cmd *cobra.Command, args []string) error {
...
+		opts.Auth.Password = string(password)
+	}
+	if a.CABundleFile != "" {
+		cabundle, err := ioutil.ReadFile(a.CABundleFile)
+		if err != nil && !os.IsNotExist(err) {
+			return err
+		}
+		opts.Auth.CABundle = cabundle
+	}
+
```
nickgerace, line 85: So if the user provides a file, but the error is not a DNE error, then we ignore the error? Will that always be okay?

Source: https://golang.org/src/os/file.go?s=20199:20241#L641
(in Go 1.16 `ioutil.ReadFile` calls `os.ReadFile`)

StrongMonkey: Yeah this is a concern but this is mostly used internally. The reason I ignore the error is that https://github.com/rancher/fleet/pull/321/files#diff-e45faf37c46946875b36e69c4720af10dbe9f75f5a1d8295b79347149d14205fR573-R579 this is hardcoded so it could be a case where file doesn't exist(we mount the secret into the pod and if key doesn't exist then file won't be there)

nickgerace (approved):

ron1:
@StrongMonkey @nickgerace Would it make sense to expand this pr to also support ssh auth for go-getter urls? Or is that best addressed in a separate pr?

StrongMonkey:
@ron1 not on the scope of this PR. Do you have an example of how that would work? I am not familiar with the context.

ron1:
@StrongMonkey Folks access private helm repos via ssh-auth using go-getter urls just like they do via http basic-auth. This is similar to how they access their private git repos. See https://github.com/rancher/fleet/issues/187 for an example. I was thinking that any work to make private helm repos accessible to fleet would implement both basic-auth and ssh-auth. Of course, a separate pr for "Add ssh-auth support for private helm repo" could be implemented separately and hopefully share some of the code.

StrongMonkey:
@ron1 Let me see if I can add this. We use hashicorp go getter library so hopefully this can be easily added

ron1:
@StrongMonkey Yes, that was my thought as well. Thanks much!

StrongMonkey:
@ron1 @nickgerace ssh support has been added in a separate commit.

pkg/bundle/resources.go
```diff
@@ -332,6 +332,14 @@ func readContent(ctx context.Context, progress *progress.Progress, base, name st
 		}
 		httpGetter.Client.Transport = transport
 	}
+	if auth.SshPrivateKey != nil {
+		if strings.IndexAny(c.Src, "&;") == -1 {
+			c.Src += "?"
+		} else {
+			c.Src += "&"
+		}
```
nickgerace, line 340: I understand how this code works, but why do we need this check? Checked the SSH (rsa256) spec and am curious.

nickgerace (approved):

pkg/bundle/resources.go
StrongMonkey: This is just trying to `sshkey` query parameters to url without parsing it. It is how https://github.com/hashicorp/go-getter#git-git reads it. Parsing the url is tricky as it is git protocol so it requires special. So instead I just detect the query parameters and add it at the end of string.

nickgerace: Good idea. Parsing would have more edge cases, so I agree with the greedy detection and addition.

nickgerace:
Approved with new SSH commit

Issue #152, opened 2020-11-18
Variables/Secrets and Helm Value Templating


mitchellmaler:
We would like the use fleet to deploy multiple helm charts but we want different settings/values per cluster in the group that are determined at deploy time. Is there a way to template the helm values passing in metadata from the cluster group or cluster it is deploying too such as the cluster name, region, etc.

Another thing we are looking for is a way to pull secrets from Hashicorp Vault to pass to the chart when it is being deployed.

gz#14407

flostru:
yeah, same here.

Im thinking of using fleet to get rid of some manual work.
First step is to deploy rancher-monitoring charts to multiple clusters.
Since prometheus doesnt know my cluster names i cant see what actual cluster sends the alert (i wonder how others do that)

We plan to use Fleet/Prometheus to send alerts to OpsGenie, all works well but identification of clusters is a pain.

I want to deploy rancher-monitoring with fleet like this:

```yaml
defaultNamespace: cattle-monitoring-system
paused: false
helm:
  releaseName: rancher-monitoring
  repo: https://charts.rancher.io
  chart: rancher-monitoring
  values:
    alertmanager:
      alertmanagerSpec:
        useExistingSecret: true
        configSecret: alertmanager-rancher-monitoring-alertmanager

targetCustomizations:
  - name: aks
    helm:
      values:
        prometheus:
          prometheusSpec:
            ...
            more specific values
            ...
            externalLabels:
              location: $(LOCATION) # substitute with some clusterSelector
              cluster: $(CLUSTER_NAME) # substitute with "management.cattle.io/cluster-display-name: cluster-display-name"
    clusteSelector:
      matchLabels:
        location: aks

  - name: eks
    helm:
      values:
        prometheus:
          prometheusSpec:
            ...
            more specific values
            ...
            externalLabels:
              location: $(LOCATION) # substitute with some clusterSelector
              cluster: $(CLUSTER_NAME) # substitute with "management.cattle.io/cluster-display-name: cluster-display-name"
    clusterSelector:
      matchLabels:
        location: eks

  - name: onprem
    helm:
      values:
        prometheus:
          prometheusSpec:
            ...
            more specific values
            ...
            externalLabels:
              location: $(LOCATION) # substitute with some clusterSelector
              cluster: $(CLUSTER_NAME) # substitute with "management.cattle.io/cluster-display-name: cluster-display-name"
    clusterSelector:
      matchLabels:
        location: onprem
```
[...]

Issue #193, opened 2020-12-23
Support for using/filling variables at runtime


smnmtzgr:
It would be good to have a mechanism that enables users to use variables in the git-repo which are populated at runtime, e.g. before applying the resources to the k8s-clusters.

This mechanism should enable users to define dynamic variables which are different between clusters. E.g. when using "external-dns" one would need other parameters based on where the k8s-cluster is deployed.

My goald would be to make the git-repo very generic so that I doesnt have to edit it (push new things...) everytime a new cluster is generated.
My preferred workflow should be:
1) create K8S-Cluster with terraform -> already working
2) import it into rancher -> already working
3) add the cluster into a fleet cluster group based on a label -> already working
4) use the defined git-repo for fleet and deploy things like ingress-nginx, external-dns..... per cluster, but partially with different values. 

Other examples and information are discussed here: based on this discussion: https://rancher-users.slack.com/archives/C013SSBKB6U/p1607538982296500?thread_ts=1607538982.296500&cid=C013SSBKB6U

gz#14407

paulchoi:
I would love to be able to fill in the cluster name as a variable.
When deploying a monitoring/logging agent, I like to identify the cluster by its name. Right now I'd need to create a separate overlay per cluster.

kkaempf:
Should be fixed by #1164

PR #325, opened 2021-04-07
Ability to use cluster labels as helm values

ibrokethecloud:
The PR allows users to pass values to helm charts via labels on the cluster objects.

This can include cluster specific information such as name to the downstream helm chart.

fleet already applies cluster labels as helm values to all charts on the downstream cluster.

Once this change is merged, the users will be able to refer these labels directly in the values section.

```
namespace: default
helm:
  releaseName: labels
  values:
    clusterName: global.fleet.clusterLabels.management.cattle.io/cluster-display-name
```

The fleet controller will render the bundle deployment objects after performing the replacement, which can then be picked up by the fleet-agents on the downstream cluster.

This should help address the issue: https://github.com/rancher/fleet/issues/152

### DIFF
--- a/pkg/target/target.go
+++ b/pkg/target/target.go
@@ -278,18 +278,7 @@ func addClusterLabels(opts *fleet.BundleDeploymentOptions, labels map[string]str
 		return
 	}
 
-	prefix := "global.fleet.clusterLabels."
-
-	for key, val := range opts.Helm.Values.Data {
-		valStr, ok := val.(string)
-		if ok && strings.HasPrefix(valStr, prefix) {
-			label := strings.TrimPrefix(valStr, prefix)
-			labelVal, labelPresent := clusterLabels[label]
-			if labelPresent {
-				opts.Helm.Values.Data[key] = labelVal
-			}
-		}
-	}
+	processLabelValues(opts.Helm.Values.Data, clusterLabels)
 
 	opts.Helm.Values.Data = data.MergeMaps(opts.Helm.Values.Data, newValues)
 
@@ -511,4 +500,22 @@ func Summary(targets []*Target) fleet.BundleSummary {
 		bundleSummary.DesiredReady++
 	}
 	return bundleSummary
+}
+
+func processLabelValues(valuesMap map[string]interface{}, clusterLabels map[string]string) {
+	prefix := "global.fleet.clusterLabels."
+	for key, val := range valuesMap {
+		valStr, ok := val.(string)
+		if ok && strings.HasPrefix(valStr, prefix) {
+			label := strings.TrimPrefix(valStr, prefix)
+			labelVal, labelPresent := clusterLabels[label]
+			if labelPresent {
+				valuesMap[key] = labelVal
+			}
+		}
+
+		if valMap, ok := val.(map[string]interface{}); ok {
+			processLabelValues(valMap, clusterLabels)
+		}
+	}
 }
--- a/pkg/target/target.go
+++ b/pkg/target/target.go
@@ -217,7 +217,10 @@ func (m *Manager) Targets(fleetBundle *fleet.Bundle) (result []*Target, _ error)
 			}
 
 			opts := options.Calculate(&fleetBundle.Spec, match.Target)
-			addClusterLabels(&opts, cluster.Labels)
+			err = addClusterLabels(&opts, cluster.Labels)
+			if err != nil {
+				return nil, err
+			}
 
 			deploymentID, err := options.DeploymentID(manifest, opts)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


ibrokethecloud:
Also helps address: https://github.com/rancher/fleet/issues/193

StrongMonkey (commented):
LGTM but we might want to have a debate the name? What would be a good alternative UX? @nickgerace @ibuildthecloud any opinion on this?

pkg/target/target.go
```diff
@@ -278,7 +278,21 @@ func addClusterLabels(opts *fleet.BundleDeploymentOptions, labels map[string]str
...
 
+	prefix := "global.fleet.clusterLabels."
+
+	for key, val := range opts.Helm.Values.Data {
+		valStr, ok := val.(string)
+		if ok && strings.HasPrefix(valStr, prefix) {
+			label := strings.TrimPrefix(valStr, prefix)
+			labelVal, labelPresent := clusterLabels[label]
+			if labelPresent {
+				opts.Helm.Values.Data[key] = labelVal
```
nickgerace, line 289: Are we essentially reformatting the same data that's already been passed in? A bit confused here.

nickgerace:
I agree with the feature, but I think pre-pending these with a required prefix may confuse users. I like that this isn't an API change though, so I'm a bit conflicted as well.

pkg/target/target.go
nickgerace: Resolved from discussion.

ibrokethecloud:
Additional changes to handle nested references to labels in  the values section, like one below:

```
namespace: default
helm:
  releaseName: labels
  values:
    clusterName: global.fleet.clusterLabels.management.cattle.io/cluster-display-name
    someMap:
      nestedMap:
        clusterName: global.fleet.clusterLabels.management.cattle.io/cluster-display-name
    secondMap:
      clusterName: global.fleet.clusterLabels.management.cattle.io/cluster-display-name
    thirdMap:
      nestedMap:
        nestedMap2:
          clusterName: global.fleet.clusterLabels.management.cattle.io/cluster-display-name
```

Rendered BundleDeployment is as follows:

```
    helm:
      releaseName: labels
      values:
        clusterName: one
        global:
          fleet:
            clusterLabels:
              management.cattle.io/cluster-display-name: one
              management.cattle.io/cluster-name: c-vd6lm
        secondMap:
          clusterName: one
        someMap:
          nestedMap:
            clusterName: one
        thirdMap:
          nestedMap:
            nestedMap2:
              clusterName: one
```

ibrokethecloud:
Missing labels are now bubbled up to the bundle

image

cc: @nickgerace @StrongMonkey

pkg/target/target.go
```diff
@@ -267,18 +270,27 @@ func addClusterLabels(opts *fleet.BundleDeploymentOptions, labels map[string]str
...
 	opts.Helm = opts.Helm.DeepCopy()
 	if opts.Helm.Values == nil || opts.Helm.Values.Data == nil {
 		opts.Helm.Values = &fleet.GenericMap{
 			Data: newValues,
 		}
-		return
+		return nil
+	}
+
+	err = processLabelValues(opts.Helm.Values.Data, clusterLabels)
```
StrongMonkey, line 284: Nit
```suggestion
	if err := processLabelValues(opts.Helm.Values.Data, clusterLabels); err != nil {
	  return err
	}
```

pkg/target/target.go
```diff
@@ -267,18 +270,27 @@ func addClusterLabels(opts *fleet.BundleDeploymentOptions, labels map[string]str
...
+
+	if err != nil {
+		return err
 	}
 
 	opts.Helm.Values.Data = data.MergeMaps(opts.Helm.Values.Data, newValues)
+
+	return nil
```
StrongMonkey, line 292: Nit: extra space

StrongMonkey (commented):
LGTM except two small nit

pkg/target/target.go
ibrokethecloud: fixed

pkg/target/target.go
ibrokethecloud: fixed

StrongMonkey (approved):

nickgerace (approved):

StrongMonkey (approved):

StrongMonkey:
@ibrokethecloud Thanks for adding this! Don't forget that we also need docs PR :)

wzrdtales:
so this is good to have already, but this will not go well with helm charts. What no user ever wants is to maintain every single helm chart. As far as I get it, this can not be templated in, into a combined value.

The most prominent example would be a helm ingress url. There I would want to add the clustername as a prefix to the rest of the url. But this doesn't seem to be really supported yet without maintaining a new helm template?

PR #326, opened 2021-04-07
add env vars to agent manifest

StrongMonkey:
https://github.com/rancher/rancher/issues/29993

### DIFF
--- a/modules/cli/match/match.go
+++ b/modules/cli/match/match.go
@@ -22,6 +22,7 @@ type Options struct {
 	BaseDir            string
 	BundleSpec         string
 	BundleFile         string
+	ClusterName        string
 	ClusterGroup       string
 	ClusterLabels      map[string]string
 	ClusterGroupLabels map[string]string
@@ -61,7 +62,7 @@ func Match(ctx context.Context, opts *Options) error {
 	}
 
 	if opts.Target == "" {
-		m := b.Match(map[string]map[string]string{
+		m := b.Match(opts.ClusterName, map[string]map[string]string{
 			opts.ClusterGroup: opts.ClusterGroupLabels,
 		}, opts.ClusterLabels)
 		return printMatch(b, m, opts.Output)
--- a/pkg/controllers/manageagent/manageagent.go
+++ b/pkg/controllers/manageagent/manageagent.go
@@ -9,6 +9,7 @@ import (
 	fleetcontrollers "github.com/rancher/fleet/pkg/generated/controllers/fleet.cattle.io/v1alpha1"
 	"github.com/rancher/wrangler/pkg/apply"
 	corecontrollers "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
+	"github.com/rancher/wrangler/pkg/name"
 	"github.com/rancher/wrangler/pkg/relatedresource"
 	"github.com/rancher/wrangler/pkg/yaml"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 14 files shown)


modules/cli/match/match.go
```diff
@@ -61,7 +61,7 @@ func Match(ctx context.Context, opts *Options) error {
 	}
 
 	if opts.Target == "" {
-		m := b.Match(map[string]map[string]string{
+		m := b.Match("", map[string]map[string]string{
```
nickgerace, line 64: Do we not have the cluster name at this point?

pkg/match/cluster.go
```diff
@@ -15,11 +15,17 @@ func toSelector(labels *metav1.LabelSelector) (labels.Selector, error) {
...
+	if clusterName != "" {
+		t.criteria = append(t.criteria, func(clusterNameTest, _ string, _, _ map[string]string) bool {
+			return clusterName == clusterNameTest
+		})
+	}
+
 	if clusterGroup != "" {
-		t.criteria = append(t.criteria, func(clusterGroupTest string, _, _ map[string]string) bool {
+		t.criteria = append(t.criteria, func(_, clusterGroupTest string, _, _ map[string]string) bool {
 			return clusterGroup == clusterGroupTest
```
nickgerace, line 29: Should we change these "test" strings to constants?

pkg/target/partition.go
```diff
@@ -37,7 +37,7 @@ func manualPartition(rollout *fleet.RolloutStrategy, targets []*Target) ([]Parti
 	targetLoop:
 		for _, target := range targets {
 			for _, cg := range target.ClusterGroups {
-				if matcher.Match(cg.Name, cg.Labels, target.Cluster.Labels) {
+				if matcher.Match(target.Cluster.Name, cg.Name, cg.Labels, target.Cluster.Labels) {
 					partitionTargets = append(partitionTargets, target)
 					continue targetLoop
```
nickgerace, line 42: Ah, so is _this_ why we need the cluster name?

pkg/controllers/manageagent/manageagent.go
```diff
@@ -71,9 +72,14 @@ func (h *handler) OnNamespace(key string, namespace *corev1.Namespace) (*corev1.
...
-	if err != nil {
-		return nil, err
+	var objs []runtime.Object
+
+	for _, cluster := range clusters {
+		bundle, err := h.getAgentBundle(namespace.Name, cluster)
+		if err != nil {
+			return nil, err
+		}
+		objs = append(objs, bundle)
```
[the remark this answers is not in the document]
StrongMonkey: We are changing the behavior so that one bundle is generated per cluster, because bundle would have different env vars per cluster.

modules/cli/match/match.go
StrongMonkey: This is actually coming from CLI. But yeah, I should add an option in CLI

pkg/match/cluster.go
StrongMonkey: What does this mean? these are function parameter names.

pkg/target/partition.go
StrongMonkey: Yeah, I am adding clusterName support because
1. An alternative way to scheduling bundle to just one cluster without dealing with labels.
2. To support fleet-agent bundle per cluster so that a bundle can schedule exact one cluster without dealing with labels

pkg/match/cluster.go
nickgerace: _facepalm_ ugh. I knew I was missing something. Disregard!

kinarashah (approved):

Issue #225, opened 2021-01-14
Release name exceeds max length of 53


atsai1220:
Issue:
![image](https://user-images.githubusercontent.com/14171643/104635006-361d5400-5656-11eb-8525-a99654b8dedc.png)
![image](https://user-images.githubusercontent.com/14171643/104635125-5b11c700-5656-11eb-9925-cbec1a45a58e.png)

It seems like a Bundle is created based on
`git-repo-name`-`paths-in-git-repo`

Renaming folders resolves the issue but this creates a limit on how deep our directory can get.

To resolve
----
1. Update `GitRepo` with shorter paths
2. Delete Bundles
3. Force Update on `GitRepo`

petertang2012:
Ran into same problem here

haswalt:
I think i've hit this similar issues in that the naming messes with our current service discovery, since service names are generated by the helm chart. Fleet sets the Release.name with that format above which isn't expected and I can't see anywhere to override it.

paulchoi:
I ran into this issue even when deploying raw YAMLs and is not a Helm chart.

Now my `fleet.yaml` looks like this, just to get around this issue
```yaml
defaultNamespace: monitoring

# While this is not a Helm chart, a Helm release name is still generated
# And throws an error if the release name is over 53 chars
helm:
  releaseName: fleet-local-ethtool-exporter
```

nickgerace:
Closed by https://github.com/rancher/fleet/issues/432 being resolved

PR #332, opened 2021-04-09
Limit name to 53 characters

StrongMonkey:
https://github.com/rancher/fleet/issues/225


nickgerace (approved):
LGTM, but we should document this

Issue #336, opened 2021-04-12
fleet test subcommand broken in v0.3.5-rc4


ron1:
The ``fleet test`` subcommand appears to be broken in v0.3.5-rc4. See below:

```
$ fleet_linux_amd64 test -t dev
panic: unable to redefine 'n' shorthand in "test" flagset: it's already used for "cluster-name" flag

goroutine 1 [running]:
github.com/spf13/pflag.(*FlagSet).AddFlag(0xc0009cc400, 0xc0007681e0)
	/go/pkg/mod/github.com/spf13/pflag@v1.0.5/flag.go:874 +0x6f0
github.com/spf13/pflag.(*FlagSet).AddFlagSet.func1(0xc0007681e0)
	/go/pkg/mod/github.com/spf13/pflag@v1.0.5/flag.go:887 +0xa8
github.com/spf13/pflag.(*FlagSet).VisitAll(0xc0009cc300, 0xc000b0fb40)
	/go/pkg/mod/github.com/spf13/pflag@v1.0.5/flag.go:290 +0x94
github.com/spf13/pflag.(*FlagSet).AddFlagSet(0xc0009cc400, 0xc0009cc300)
	/go/pkg/mod/github.com/spf13/pflag@v1.0.5/flag.go:885 +0x58
github.com/spf13/cobra.(*Command).mergePersistentFlags(0xc00072c2c0)
	/go/pkg/mod/github.com/spf13/cobra@v1.1.1/command.go:1642 +0x92
github.com/spf13/cobra.stripFlags(0xc0001d2180, 0x2, 0x2, 0xc00072c2c0, 0x0, 0x2, 0x0)
	/go/pkg/mod/github.com/spf13/cobra@v1.1.1/command.go:562 +0x5a
github.com/spf13/cobra.(*Command).Find.func1(0xc00072c2c0, 0xc0001d2180, 0x2, 0x2, 0x4, 0xc0001d2180, 0x2, 0x2)
	/go/pkg/mod/github.com/spf13/cobra@v1.1.1/command.go:621 +0x5c
github.com/spf13/cobra.(*Command).Find.func1(0xc00072d8c0, 0xc00013e050, 0x3, 0x3, 0x18a89ef, 0xc0009f8b00, 0xc0002f5800, 0xa)
	/go/pkg/mod/github.com/spf13/cobra@v1.1.1/command.go:629 +0x11e
[...]

PR #337, opened 2021-04-12
Remove duplicated flag

StrongMonkey:


StrongMonkey:
https://github.com/rancher/fleet/issues/336

modules/cli/cmds/test.go
```diff
@@ -19,7 +19,7 @@ type Test struct {
 	BundleInputArgs
 	Quiet       bool              `usage:"Just print the match and don't print the resources" short:"q"`
 	Group       string            `usage:"Cluster group to match against" short:"g"`
-	ClusterName string            `usage:"Cluster name to match against" short:"n"`
+	ClusterName string            `usage:"Cluster name to match against"`
```
ron1, line 22: I suspect ClusterName will be a popular option. What would you think about using capital 'N' as a short?

nickgerace (approved):

modules/cli/cmds/test.go
StrongMonkey: I am not sure if Capital flag is good idea here. @nickgerace Thoughts?

nickgerace: We could use `-c`, but I agree that capital flags are uncommon enough that we should only use them when necessary.

StrongMonkey: It looks like we already have `-L`. To make it consistent I will rename to it `name` and shortflag it as `-N`

nickgerace (approved):

Issue #341, opened 2021-04-15
JSON unmarshalling error when incorrect or no password is provided for Helm basic auth


bmdepesa:
rancher/rancher:v2.5.8-rc3
fleet v0.3.5-rc4

* Create a Helm repo using basic auth
* Reference the helm repo in a fleet bundle:
```
helm:
  releaseName: guestbook
  chart: "guestbook"
  repo: "https://xx.xx.xx.xx.xip.io"
```
* Create a secret with an incorrect password
```
kubectl create secret -n fleet-default generic helm-secret \
  --from-literal=username=user1 \
  --from-literal=password=badpassword
```
* Reference the secret in the GitRepo
```
kind: GitRepo
apiVersion: fleet.cattle.io/v1alpha1
metadata:
  name: helm
  namespace: fleet-default
spec:
  branch: helmtest
  helmSecretName: helm-secret
  repo: https://git-repo-url
  paths:
  - single-cluster/helm
  targets:
    - clusterSelector: {}
```

The GitRepo goes into an error state, but the error shows an unmarshaling error instead of a failed authentication error.
```
time="2021-04-15T16:46:44Z" level=fatal msg="error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go value of type repo.IndexFile"
```

The same error shows if a `helmSecretName` is not provided in the GitRepo object when the helm repo requires authentication.

bmdepesa:
rancher/rancher:v2.5-head 9e412a3
fleet v0.3.5-rc5

* Create a Helm repo using basic auth
* Reference the helm repo in a fleet bundle
* Create a secret with an incorrect password
* Reference the secret in the GitRepo
* The GitRepo goes into an error state correctly, with the error: 
```
time="2021-04-20T22:26:47Z" level=fatal msg="failed to read helm repo from https://repo.url/index.yaml, error code: 401, response body: <html>\r\n<head><title>401 Authorization Required</title></head>\r\n<body>\r\n<center><h1>401 Authorization Required</h1></center>\r\n<hr><center>nginx/1.19.10</center>\r\n</body>\r\n</html>\r\n"
```

Issue #340, opened 2021-04-15
GitRepo does not show an error state when credential secret doesn't exist when using Helm authentication


bmdepesa:
rancher/rancher:v2.5.8-rc3
fleet v0.3.5-rc4

* Create a Helm repo using basic auth
* Reference the helm repo in a fleet bundle:
```
helm:
  releaseName: guestbook
  chart: "guestbook"
  repo: "https://xx.xx.xx.xx.xip.io"
```
* In the GitRepo, for `helmSecretName`, specify a secret that doesn't exist 
```
kind: GitRepo
apiVersion: fleet.cattle.io/v1alpha1
metadata:
  name: helm
  namespace: fleet-default
spec:
  branch: helmtest
  # bad secret name
  helmSecretName: doesntexist
  repo: https://git-repo-url
  paths:
  - single-cluster/helm
  targets:
    - clusterSelector: {}
```

After saving the GitRepo, the state is in `Active` but never progresses or enters an error state. If `helmSecretName` is omitted entirely from the GitRepo object, the state will go to an error state with an appropriate error.

bmdepesa:
rancher/rancher:v2.5-head 9e412a3
fleet v0.3.5-rc5

* Create a Helm repo using basic auth
* Reference the helm repo in a fleet bundle
* In the GitRepo, for helmSecretName, specify a secret that doesn't exist
* The GitRepo goes into an error state with: `failed to look up helmSecretName, error: secrets "helmsecret" not found`

PR #342, opened 2021-04-16
Handle the error for private helm repo

StrongMonkey:
https://github.com/rancher/fleet/issues/340
https://github.com/rancher/fleet/issues/341

### DIFF
--- a/pkg/bundle/resources.go
+++ b/pkg/bundle/resources.go
@@ -135,6 +135,10 @@ func chartURL(location *fleet.HelmOptions, auth Auth) (string, error) {
 		return "", err
 	}
 
+	if resp.StatusCode != 200 {
+		return "", fmt.Errorf("failed to read helm repo from %s, error code: %v, response body: %s", location.Repo+"index.yaml", resp.StatusCode, bytes)
+	}
+
 	repo := &repo.IndexFile{}
 	if err := yaml.Unmarshal(bytes, repo); err != nil {
 		return "", err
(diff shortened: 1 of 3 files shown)


pkg/bundle/resources.go
```diff
@@ -135,6 +135,10 @@ func chartURL(location *fleet.HelmOptions, auth Auth) (string, error) {
 		return "", err
 	}
 
+	if resp.StatusCode != 200 {
+		return "", fmt.Errorf("failed to read helm repo from %s, error code: %v, response body: %s", location.Repo+"index.yaml", resp.StatusCode, bytes)
```
nickgerace, line 139: Are bytes always type `string`? Or can it be a slice of bytes?

pkg/controllers/controllers.go
```diff
@@ -188,7 +188,8 @@ func Register(ctx context.Context, systemNamespace string, cfg clientcmd.ClientC
 		appCtx.BundleDeployment(),
 		appCtx.GitRepoRestriction().Cache(),
 		appCtx.Bundle(),
-		appCtx.GitRepo())
+		appCtx.GitRepo(),
+		appCtx.Core.Secret().Cache())
```
nickgerace, line 192: Why cache the secret?

StrongMonkey: This avoids api call which cause latency

pkg/bundle/resources.go
StrongMonkey: yeah, doing `%s` with bytes will also work

StrongMonkey: https://golang.org/pkg/fmt/ good catch :+1: I checked the reference and it works with string/slices of bytes

nickgerace (approved):

PR #351, opened 2021-04-26
Redeploy agent when fleet-agent deployment is deleted

StrongMonkey:
Context:
We switched from one fleet-agent bundle per workspace to one fleet bundle per cluster in order to support agentEnvVars per cluster https://github.com/rancher/rancher/issues/29993. This has caused a problem because we have to delete old `fleet-agent` bundle and use the new name `fleet-agent-$clustername`. When deleting the `fleet-agent` bundle, `fleet-agent` deployment will also get deleted in downstream cluster, causing https://github.com/rancher/rancher/issues/32317. We need to deploy the agent if downstream cluster doesn't have the fleet-agent deployment.

### DIFF
--- a/pkg/controllers/manageagent/manageagent.go
+++ b/pkg/controllers/manageagent/manageagent.go
@@ -49,9 +49,9 @@ func Register(ctx context.Context,
 	relatedresource.WatchClusterScoped(ctx, "manage-agent-resolver", h.resolveNS, namespace, clusters)
 }
 
-func (h *handler) resolveNS(namespace, name string, obj runtime.Object) ([]relatedresource.Key, error) {
-	if _, ok := obj.(*fleet.Cluster); ok {
-		if _, err := h.bundleCache.Get(namespace, agentBundleName); err != nil {
+func (h *handler) resolveNS(namespace, _ string, obj runtime.Object) ([]relatedresource.Key, error) {
+	if cluster, ok := obj.(*fleet.Cluster); ok {
+		if _, err := h.bundleCache.Get(namespace, name.SafeConcatName(agentBundleName, cluster.Name)); err != nil {
 			return []relatedresource.Key{{Name: namespace}}, nil
 		}
 	}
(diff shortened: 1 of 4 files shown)


nickgerace (approved):

pkg/controllers/manageagent/manageagent.go
```diff
@@ -49,9 +49,9 @@ func Register(ctx context.Context,
 	relatedresource.WatchClusterScoped(ctx, "manage-agent-resolver", h.resolveNS, namespace, clusters)
 }
 
-func (h *handler) resolveNS(namespace, name string, obj runtime.Object) ([]relatedresource.Key, error) {
-	if _, ok := obj.(*fleet.Cluster); ok {
-		if _, err := h.bundleCache.Get(namespace, agentBundleName); err != nil {
+func (h *handler) resolveNS(namespace, _ string, obj runtime.Object) ([]relatedresource.Key, error) {
+	if cluster, ok := obj.(*fleet.Cluster); ok {
+		if _, err := h.bundleCache.Get(namespace, name.SafeConcatName(agentBundleName, cluster.Name)); err == nil {
```
kinarashah, line 54: I assumed that earlier if we didn't find bundle cache (err !=nil), we'd still return a key that would call agent-resolver and deploy agent. Now we assume agent bundle will always exist?

kinarashah: Oh okay, so now importCluster will always deploy agent vs earlier it would do it only for manager initiated setups?

StrongMonkey: whoops you raised a good question. I originall thought this was a bug. Like when we change the agentEnvs on the cluster, it didn't enqueue the namespace(which should cause resync) because we only check if `err != nil`. Now i think it is only when the bundle is not there then we enqueue the namespace.

kinarashah (approved):

kinarashah (approved):

PR #354, opened 2021-04-28
Add migration logic, drop deployment check

StrongMonkey:
We need to drop the previous deployment check logic, as it is causing chatty connection to downstream cluster when cluster is updated. Instead we will add `agentMigrated` in cluster.status to force upgrade fleet-agent on imported clusters. This is to handle the case where we migrate from one `fleet-agent` bundle per workspace to `fleet-agent` bundle per cluster, and fleet will delete fleet-agent deployment when deleting `fleet-agent` bundle, which will cause a outage(see https://github.com/rancher/rancher/issues/32317). This is fixed in https://github.com/rancher/fleet/pull/351/files#diff-7657d93aa6e7023c8a3230bb999b94845f6b607b146c9bc7e137620032683f0fR481-R485. Once agent is fully migrated to new version, we would not need to worry about this.

So, adding a new status `agentMigrated` to cluster and if status is false then force upgrade to new agent version using import controller. This will only impact upgrade since all the clusters druing the upgrade will be forced to new fleet-agent using import controller, but this is already happening during a normal fleet upgrade.

### DIFF
--- a/charts/fleet-crd/templates/crds.yaml
+++ b/charts/fleet-crd/templates/crds.yaml
@@ -1583,6 +1583,8 @@ spec:
             agentDeployedGeneration:
               nullable: true
               type: integer
+            agentMigrated:
+              type: boolean
             conditions:
               items:
                 properties:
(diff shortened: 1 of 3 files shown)


nickgerace (approved):
Love this approach

nickgerace:
Might need to generate CRDs?

PR #357, opened 2021-05-04
add TargetRestrictions to GitRepo crd

kalinon:
This PR allows you to define target restrictions at the GitRepo level. This allows you to define something like:

```yaml
  targetRestrictions:
  - clusterSelector: {}
  targets:
  - clusterSelector: null
```

Which will allow you to specify the targets in the Bundle level without inheriting the defaults. 

For example if i have 2 cluster groups (prod and dev). This allows me to specify a `fleet.yaml` that will just deploy to 1 group.

```yaml
targets:
  - clusterGroup: prod
```


pkg/controllers/git/git.go
```diff
@@ -97,6 +97,18 @@ func targetsOrDefault(targets []fleet.GitTarget) []fleet.GitTarget {
...
 }
 
+func targetRestrictionsOrDefault(targets []fleet.GitTargetRestriction) []fleet.GitTargetRestriction {
+	if len(targets) == 0 {
+		return []fleet.GitTargetRestriction{
+			{
+				Name:         "default",
+				ClusterGroup: "default",
+			},
+		}
```
nickgerace, line 107: Does this replicate the default behavior today from your testing?

nickgerace (commented):
Adding folks to the review, but the idea is sound with me.

pkg/controllers/git/git.go
kalinon: tbh i have not been able to figure out a good way to test this. unless i missed some documentation, its not clear how to effectively set up a testing env.

nickgerace: Testing ENVs are usually set up by installing a local cluster of some sort (k3d/k3s) and then installing the fleet charts directly. Many people install a local cluster, then install the latest Rancher v2.5.x release to get Fleet in the GUI. With both approaches though, there is only a local cluster (unless you create a downstream cluster).

StrongMonkey:
Today the target in gitrepo also creates targetRestriction in fleet bundle, which scopes the target that fleet.yaml can specify. Will that work for your use cases?

https://github.com/rancher/fleet/blob/9d9fe2dede722999e9c94f518c7419ef7ca6e4d7/pkg/controllers/git/git.go#L103-L115

kalinon:
> Today the target in gitrepo also creates targetRestriction in fleet bundle, which scopes the target that fleet.yaml can specify. Will that work for your use cases?
> 
> https://github.com/rancher/fleet/blob/9d9fe2dede722999e9c94f518c7419ef7ca6e4d7/pkg/controllers/git/git.go#L103-L115

It doesn't, because currently the `GitRepo.targets` is appended to the `targetRestriction`. Which means, you cant actually make the scope smaller, as its additive.

PR #358, opened 2021-05-04
add a fleet.yaml option overrideTargets to exclude inheriting targets

kalinon:
This PR adds a `overrideTargets` option to the `fleet.yaml` file. This bool flag will prevent inheriting the default `targets` from the `GitRepo` definition.


pkg/bundle/read.go
```diff
@@ -153,9 +154,12 @@ func read(ctx context.Context, name, baseDir string, bundleSpecReader io.Reader,
 
 	def.Spec.ForceSyncGeneration = opts.SyncGeneration
 
-	def, err = appendTargets(def, opts.TargetsFile)
-	if err != nil {
-		return nil, err
+	// If override targets is set do not append other targets
+	if !bundle.OverrideTargets && len(def.Spec.Targets) > 0 {
+		def, err = appendTargets(def, opts.TargetsFile)
```
nickgerace, line 159: What happens if `OverrrideTargets` is `false`, but the length is `<= 0`?

kalinon: yeah, i think i was trying to be too fancy. I think if `OverrrideTargets` is `false` append targets. Think the scenario i was worried about was if `OverrrideTargets` is `true` but no targets were defined, we would want to append defaults.

nickgerace: I actually think that's fine! I was asking whether or not that scenario is one we should account for still.

kalinon: I'm not sure. Part of me says that `override` should be authoritative and if you end up with 0 targets, then it should be removed. But a safety net is good as well.

I think my original logic was wrong, i think this would be the `if` statement to handle the safety net scenario.
```go
 	if !bundle.OverrideTargets || (bundle.OverrideTargets && len(def.Spec.Targets) > 0) { }
```

kalinon: Think my last commit resolves this.

kalinon:
@nickgerace I went through and tested this change. Had to make a few tweaks that you can see in https://github.com/rancher/fleet/pull/358/commits/02e0f723e58cf9600c4510d783fd56ddace32f04 to get it working. But its working as intended now. Ive dumped all my CRDs related to my tests. I used a modified version of the `fleet-mc-helm-example` in the examples repo.
[target-override.tar.gz](https://github.com/rancher/fleet/files/6549725/target-override.tar.gz)

kalinon:
Any updates on this?

nickgerace:
Hi @kalinon we are in the middle of release for the next version of Fleet. This PR might be a tad delayed. Apologies for the wait.

If you could make an issue in this repo and link your two PRs to it, I can see about getting it scheduled for the release following `v0.3.6`. Thank you.

kalinon:
> Hi @kalinon we are in the middle of release for the next version of Fleet. This PR might be a tad delayed. Apologies for the wait.
> 
> If you could make an issue in this repo and link your two PRs to it, I can see about getting it scheduled for the release following `v0.3.6`. Thank you.

Thanks @nickgerace i created #462 to do this.

kalinon:
Any update on this? would love to get it in v0.3.7

nickgerace:
Hi @kalinon: we did not plan on releasing v0.3.7 for Rancher v2.6.1, so its existence is to just tighten up the edges on v0.3.6 additions and changes.

However, I'd the parent issue to our "release candidates" column for future scheduling. I hope to get it scheduled and thank you for your patience.

kalinon:
@nickgerace awesome! sounds good to me!

raulcabello:
Superseded by https://github.com/rancher/fleet/pull/1580

PR #359, opened 2021-05-04
add support for fleet.yml file name

kalinon:
Simple enough, this adds a check to see if there is a file named `fleet.yml` if `fleet.yaml` is not found.


pkg/bundle/read.go
```diff
@@ -40,7 +40,21 @@ func Open(ctx context.Context, name, baseDir, file string, opts *Options) (*Bund
...
+			// fleet.yaml does *not* exist
+			file = filepath.Join(baseDir, "fleet.yml")
+			if _, err := os.Stat("/path/to/whatever"); err == nil {
+				// fleet.yml exists
+			} else if os.IsNotExist(err) {
+				// fleet.yaml and fleet.yml does *not* exist
+				return nil, err
+			}
+		}
+
```
nickgerace, line 57: Part of me almost wants to enforce the `.yaml` extension instead... Hm... Thoughts @StrongMonkey @kinarashah ?

kalinon: I don't feel too strongly either way, but it was something that just got me when i kept naming it `fleet.yml`.

StrongMonkey: I would rather use fleet.yaml and not bother a different format like fleet.yml. It is just a convention that user will follow.

kalinon: My only argument is that generally `yml` and `yaml` are interchangeable. Fleet itself will register any `yml` resource files, so i would expect `fleet.yml` to be handled the same way.

skaven81: I support this update, as it can be really frustrating for new users that are used to using `.yml` extensions, to figure out why Fleet isn't behaving as expected, when the only "problem" is that they created `fleet.yml` instead of `fleet.yaml`.

Source: This was me for the first few days trying to learn how to use Fleet.

nickgerace:
This was actually already merged in in by corollary work we were doing for another customer escalation. Thank you so much for the effort in your PR! We usually look to work with community PRs first, so apologies for the miscommunication on our part here. Thank you again.

Merged PR: https://github.com/rancher/fleet/pull/592

PR #381, opened 2021-05-20
Allow to disable gitops part in fleet

StrongMonkey:


nickgerace (approved):
LGTM, both windows and the disable idea. Happy we are adding that.

Issue #384, opened 2021-05-21
Request: Add overlay option from cluster secrets to handle secret strings in the values.yaml


ewokpoacher:
We need to be able to set secret strings on the values.yaml with a staged secret similar to Flux's helm release using  [valuesFrom](https://fluxcd.io/docs/guides/helmreleases/#refer-to-values-in-configmap-and-secret-resources).

For example, Harbor has you set the database password in the values.yaml.  We want to use Sealed Secrets to apply a secret, `secret-values.yaml`, via a separate bundle to the harbor namespace downstream with a key, `values.yaml`, of: 
```
database:
  external:
    password: areallybadpass
```
Then call that overlay at the end of the fleet.yaml with something like this:
```
defaultNamespace: harbor
helm:
  takeOwnership: true
  releaseName: dev
  chart: "harbor"
  repo: "https://helm.goharbor.io"
  version: "1.6.0"
  force: false
  timeoutSeconds: 0
  values:
    stuff: andthangs
    database:
      external:
        password: ThisWillBeReplaced
  valuesFrom:
    - kind: Secret
      name: secret-values.yaml
      valuesKey: values.yaml
```
The overlaying would happen just the same as overlay/values_patch.yaml would work.

The fleet-agent on the downstream cluster would handle the final rendering of the bundle before applying the chart.  This is to both prevent decrypted secrets from being in the bundle and allow each downstream cluster to have its own value for the `secret-values.yaml`.  

An alternative pattern could be to add something like [vals](https://github.com/variantdev/vals) to the downstream fleet-agent and allow vals type syntax in the values tree to reference secrets stored across multiple store types.  Example:
```
values:
    stuff: andthangs
    database:
      external:
        password: refs+vault://mycluster/harbor/database#/password
```
Vals approach could require additional authentication secrets in each cluster's fleet namespace depending on the backend.

Our current workaround is having fleet deploy helm-controller and source-controller then using fleet to apply helm-releases for charts that force this pattern.  We'd like to consolidate to just fleet and drop the other controllers.

StrongMonkey:
Thanks for putting this. We are actually planning to add valuesFrom like flux does.

StrongMonkey:
@izaac You can test this feature using docs here https://github.com/rancher/fleet/pull/414.

izaac:
Rancher version `2.6 master-head (06/18/2021)` _3763e5d_

[...]

PR #388, opened 2021-05-21
Add support for valuesFrom

StrongMonkey:
This PR adds support to define valueFrom in `fleet.yaml`, which allow user to put additional values.yaml into configmap/secrets. This is in case the valuesFile contains sensetive information and can't be exposed in git, user can prepopulate data into secrets in downstream clusters. 

https://github.com/rancher/fleet/issues/384


pkg/helmdeployer/deployer.go
```diff
@@ -352,6 +356,102 @@ func (h *helm) install(bundleID string, manifest *manifest.Manifest, chart *char
...
+				return nil, err
+			}
+			data, ok := secret.Data[key]
+			if !ok {
+				return nil, fmt.Errorf("key %s is missing from secret %s/%s, can't use it in valuesFrom", key, namespace, name)
+			}
+			if err := yaml.Unmarshal(data, &val); err != nil {
+				return nil, err
+			}
+		} else if valuesFrom.ConfigMapKeyRef != nil {
```
nickgerace, line 396: Don't we want to check both the Secret and the ConfigMap? Should this be a separate `if` block?

StrongMonkey: The api is designed to only respect either secretRef or configmapRef. This followed the same valueFrom dedign from k8s pod api.

pkg/helmdeployer/deployer.go
```diff
@@ -352,6 +356,102 @@ func (h *helm) install(bundleID string, manifest *manifest.Manifest, chart *char
...
+			name := valuesFrom.SecretKeyRef.Name
```
kinarashah, line 397: We should read name and namespace form `valuesFrom.ConfigMapKeyRef`?

StrongMonkey: That's a copy paste error :). Will fix

kinarashah (approved):

nickgerace (approved):

PR #390, opened 2021-05-24
Secrets with vals

digiserg:
Parses helm charts with https://github.com/variantdev/vals to inject secrets from any of the supported backends.

[vals](https://github.com/variantdev/vals) is a tool I use all the time to centralize secrets in a single location. I don't know if people will support this change but I personally would like it added.

I've only done the helm deployer as I don't use any other deployment method but it is simple enough it can be quickly add it to all.

Sample usage.

```yaml
helm:
  valuesFiles:
    - secrets.yaml
```

and `secrets.yaml` would contain something like

```
password: ref+vault://secret/rancher/password
```

Make sure you have the right env vars for the [backend](https://github.com/variantdev/vals#suported-backends) configured


nickgerace (commented):
Looks good to me on paper, but would like other reviewers' opinions.

StrongMonkey:
I assume this will require you to add extra mounts or environmental variable into fleet-agent deployment?

digiserg:
Yes, depending on the backend you most certainly will need to. Mostly env vars but some will probably also require secret files such google service account.

philomory:
I would _really_ love to see _exactly_ this, but with the ability to have the Fleet Controller perform the vals retrieval, rather than the agent.

digiserg:
> I would _really_ love to see _exactly_ this, but with the ability to have the Fleet Controller perform the vals retrieval, rather than the agent. Basically the same code as this PR includes, but have it called in e.g. `bundle.Read()`.

Would you elaborate as to why do you think it's better this performed by the controller?

philomory:
> Would you elaborate as to why do you think it's better this performed by the controller?

So, there's a couple of different reasons.

The biggest reason is to avoid a chicken-and-egg problem and dramatically simplify management in some scenarios. If the downstream clusters are responsible for decrypting/accessing secrets, that means they need to either have the decryption keys themselves, or have credentials for accessing the relevant secrets-management system (e.g. AWS KMS, or Vault). Those credentials or decryption keys are, themselves, secrets that need to be managed. If your downstream clusters are running in AWS, you can use IAM to manage access to the secret provider. Other cloud providers have similar solutions. But if your downstream clusters are just running on hardware, you'll have to get each system the credentials it needs to access the secrets management system.

For example, say you are using AWS KMS to handle decryption. Your individual clusters, however, are not in AWS. So, to allow them to decrypt secrets, you need to get IAM credentials onto each downstream cluster. You have two options: use the same IAM credentials for each downstream cluster, or use distinct IAM credentials for each downstream cluster.

If you use the same IAM credentials for each downstream cluster, it's (relatively) easy to manage, but, you'll run into big problems if any downstream cluster is compromised. Now you not only have to rotate the secrets that Fleet deployed, you also need to rotate all the shared IAM credentials, which _aren't_ being managed by Fleet so you probably have to touch each cluster individually or write some sort of ad-hoc script to do it.

If you give each downstream cluster its own set of IAM credentials, you're much safer when a cluster gets compromised; you just need to kick the compromised cluster out, and rotate the Fleet-managed secrets. _But_, managing the initial deployment of individual IAM credentials to every cluster is a lot of extra work. If only there was an individual, cluster-specific secret already _on_ the downstream clusters we could take advantage of...

As it turns out, there _is_ such a secret, it's the ClusterRegistrationToken. Every downstream cluster already has one, and they can't get secrets (or anything else) from Rancher/Fleet without a valid one. Now, I _could_ try to jerry-rig a setup where the downstream clusters asked e.g. Vault for a secret, and authenticated themselves using their Rancher CRT, and then the Vault server would ask Rancher to verify that the CRT was valid... but it's a lot easier to just have the downstream clusters authenticate themselves with Rancher/Fleet using the CRT (this already happens), and cut out the middle-man. Only the Rancher/Fleet server needs access to the secrets-management tool, and the downstream clusters just get the information they're given. Depending on the exact secrets-management solution used, maybe you don't even expose it to the public internet.

Another (rather less important) reason is just that it'd be nice to use Rancher and Fleet for managing multiple downstream clusters in the same manner we're accustomed to using Helm to deploy to an individual cluster. Certainly tools like Sealed Secrets are valuable, but they're not the _only_ way to deploy secrets using Helm. You can also just have your secret in a SOPS-encrypted file in git, and have it decrypted as part of deployment using e.g. `helm-secrets`. Then the downstream cluster doesn't need to worry about it. Having the Fleet Controller handle the decryption/retrieval of secrets would be roughly equivalent to the Helm Secrets workflow, then.

digiserg:
> The biggest reason is to avoid a chicken-and-egg problem and dramatically simplify management in some scenarios. If the downstream clusters are responsible for decrypting/accessing secrets, that means they need to either have the decryption keys themselves, or have credentials for accessing the relevant secrets-management system (e.g. AWS KMS, or Vault). 

This is actually the main reason for me to have them at the agent level. We have several clusters and each of them has its own `vault` instance. Therefore the secrets are only available from the target cluster.

I also very much get your point you the secrets could be local. In fact, we ourselves will easily run into this in some scenarios.

philomory:
I should be clear, I never meant to imply that this functionality should be available in the Fleet Controller _instead of_ in the Fleet Agent; I meant to imply that it should be available in the Fleet Controller _in addition to_ the Fleet Agent. Neither option on it's own is going to be flexible enough to address all of the use-cases that Fleet itself gets used for. Both options together, however, will be. 

Note also that this argument applies equally to other implementations of Fleet-based secrets management, not just the `vals`-based one in this PR. I do happen to think that the `vals`-based solution is a particularly nice one, though.

philomory:
@digiserg Out of curiosity, can I ask why you opted to close out this PR? Has it been superseded by another PR?

digiserg:
I've been doing work on it yesterday and when I tried to resolve merge conflicts I found it too complex. I'm rewriting it today and will send much better PR later on.

nickgerace:
We are looking forward to it! @digiserg

digiserg:
@nickgerace @philomory 

I'd appreciate some feedback before I open the PR: 

https://github.com/digitalis-io/fleet/blob/secretsWithVals/docs/secrets.md

nickgerace:
It might be easier to open the PR so we can review line by line @digiserg

nickgerace:
I believe you can add us in a "Draft" PR too

philomory:
@digiserg I'm not really in a position to comment on the implementation code, but from the linked documentation that sounds excellent, exactly the sort of thing I'm looking for.

PR #403, opened 2021-06-04
updating path to fleet-examples simple

asloan7:
the old path is 404 not found


nickgerace:
Thank you! Good catch.

Issue #372, opened 2021-05-12
serviceAccount in GitRepo does not work: forbidden


toddnni:
Service Account definition seem to break bundle apply. Apply works fine without the service account definition.

For example this

```
apiVersion: fleet.cattle.io/v1alpha1
kind: GitRepo
metadata:
  name: manifests-bug
  namespace: fleet-local
spec:
  paths:
  - single-cluster/manifests
  repo: https://github.com/rancher/fleet-examples
  serviceAccount: fleet-agent
```
That is the traditional example, but notice `serviceAccount` field. We put the default service account there, that should work in any case.

The end result is an error message
```
query: failed to query with labels: secrets is forbidden: User "serviceaccount:fleet-system:fleet-agent" cannot list resource "secrets" in API group "" in the namespace "fleet-manifests-example"
```

Comparing to kubectl `--as`
```
# k get secrets -n fleet-manifests-example --as serviceaccount:fleet-system:fleet-agent
Error from server (Forbidden): secrets is forbidden: User "serviceaccount:fleet-system:fleet-agent" cannot list resource "secrets" in API group "" in the namespace "fleet-manifests-example"
```
But if we change a bit
```
# k get secrets -n fleet-manifests-example --as system:serviceaccount:fleet-system:fleet-agent
No resources found in fleet-manifests-example namespace.
```
Notice, that the example works, and the difference is in `system:serviceaccount:` prefix. 

Based on this, I can make following guess. However, I'm not an expert in the code at all. Could it be that this impersonates with wrong prefix, `serviceaccount:` instead of `system:serviceaccount:`
https://github.com/rancher/fleet/blob/f414eab0e4de0523eca797fa7bf47fccb1997e4a/pkg/helmdeployer/impersonate.go#L71
and this does not hit normally as impersonation is not done, if `serviceAccount` is not defined
https://github.com/rancher/fleet/blob/f414eab0e4de0523eca797fa7bf47fccb1997e4a/pkg/helmdeployer/deployer.go#L255

Running fleet version v0.3.4, but if the this part of code is same in the latest also.

StrongMonkey:
@toddnni Thanks for great analysis. I think it could be a bug.

thehejik:
**Test report:**
Tested on `rancher/rancher:master-head 5aa95f2` on local RKE cluster.

I used the example yaml from summary as it is (inc. `spec.serviceAccount: fleet-agent` and was able to create the fleet repo resource by using `Import YAML` button on top of dashboard.

[...]

PR #410, opened 2021-06-10
Use system prefix for impersonating serviceaccount

StrongMonkey:


StrongMonkey:
From k8s docs:

> Service accounts authenticate with the username system:serviceaccount:(NAMESPACE):(SERVICEACCOUNT), and are assigned to the groups system:serviceaccounts and system:serviceaccounts:(NAMESPACE).

https://github.com/rancher/fleet/issues/372

nickgerace (approved):
LGTM, but should this be configurable?

StrongMonkey:
@nickgerace Which part should be configurable? the serviceaccount is a top-level field from gitrepo and bundle.

nickgerace:
@StrongMonkey curious if the prefix, including `system`, should be configurable.

kinarashah (approved):

StrongMonkey:
@nickgerace Oh, I don't think that's configurable according to the docs.

Issue #378, opened 2021-05-18
Image scan feature


StrongMonkey:
We need to add image scan feature into fleet. Allowing user to scan an image respository and write back the image changes to git repo. This allows user to automatically apply the manifest with new image tag once image is released.

See similar feature: https://fluxcd.io/docs/guides/image-update/

StrongMonkey:
@izaac You can test this featue following the doc here https://github.com/rancher/fleet/pull/413

SheilaghM:
@PennyScissors - please validate this feature for QA.  Daishan left instructions on testing, see below.  He's also back Monday if you have questions.

PennyScissors:
Tested in rancher/rancher:master-head 9e0d1b2

Steps:
1. Create k3s cluster (all default values).
2. Add a Docker Hub secret to the `default` namespace in the k3s cluster for the fleet image scan policy configuration.
3. Create a Git Repo ([what I used](https://github.com/PennyScissors/testing/tree/master/fleet/image-scan)) in the cluster's Continuous Integration and change the target to the k3s cluster (not all clusters in workspace).
4. Verify the deployment gets deployed successfully
5. Bump the image tag and push to Docker Hub.
6. Verify the deployment gets updated to match the new image tag.

Problems:
1. After step 3, I see the following error, but is not consistent. As a workaround, forcing an update on the Git Repo seems to fix it.
```
Time="2021-07-23T17:03:31Z" level=info msg="updated: fleet-default/testing-fleet-image-scan\n" time="2021-07-23T17:03:31Z" level=info msg="updated (scan): fleet-default/imagescan-testing-fleet-image-scan-0\n" time="2021-07-23T17:03:31Z" level=info msg="created (scan): fleet-default/testing-fleet-image-scan\n" time="2021-07-23T17:03:31Z" level=fatal msg="resource name may not be empty" 
```
2. Pushed a new tag for the hello-world image and the change was not detected.
Deployment still using scresporancher/hello-world:1.0.0
![image](https://user-images.githubusercontent.com/8951271/126819730-fc07eed0-0bfa-40a1-9f39-ef557e9fd488.png)
Docker Hub with a new scresporancher/hello-world:1.0.1 image available
![image](https://user-images.githubusercontent.com/8951271/126820104-3f8f07c3-1b02-4612-920b-3791bdd6bd6e.png)

StrongMonkey:
@PennyScissors This should be fixed in v0.3.6-rc7

[...]

PR #413, opened 2021-06-11
Add image scan docs

StrongMonkey:
https://github.com/rancher/fleet/issues/378

### DIFF
--- a/docs/imagescan.md
+++ b/docs/imagescan.md
@@ -1,0 +1,106 @@
+# Image scan
+
+Image scan in fleet allows you to scan your image repository, fetch the desired image and update your git repository, 
+without the need to manually update your manifests.
+
+!!! hint "Experimental"
+    This feature is considered as experimental feature.
+    
+Go to `fleet.yaml` and add the following section.
+
+```yaml
+imageScans:
+# specify the policy to retrieve images, can be semver or alphabetical order 
+- policy: 
+    # if range is specified, it will take the latest image according to semver order in the range
+    # for more details on how to use semver, see https://github.com/Masterminds/semver
+    semver: 
+      range: "*" 
+    # can use ascending or descending order
+    alphabetical:
+      order: asc 
+
+  # specify images to scan
+  image: "your.registry.com/repo/image" 
+
+  # Specify the tag name, it has to be unique in the same bundle
+  tagName: test-scan
+
+  # specify secret to pull image if in private registry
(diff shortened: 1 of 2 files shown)


docs/imagescan.md
```diff
@@ -0,0 +1,105 @@
...
+  # user must properly provide a secret that have write access to git repository
+  clientSecretName: secret 
+  # specify the commit pattern
+  imageScanCommit:
+    authorName: foo
+    authorEmail: foo@bar.com
+    messageTemplate: "update image"
+```
+
+Try pushing a new image tag, for example, `your.registry.com/repo/image:v1.0.3`. Wait for a while and there should be a new commit pushed into your git repository to change tag in deployment.yaml.
```
nickgerace, line 104: Perhaps we should use `<tag>` instead of `v1.0.3`

docs/imagescan.md
```diff
@@ -0,0 +1,105 @@
...
+  template:
+    metadata:
+      labels:
+        app: redis
+        role: slave
+        tier: backend
+    spec:
+      containers:
+      - name: slave
+        image: your.registry.com/repo/image:v1.0.2 # {"$imagescan": "test-scan"}
```
nickgerace, line 62: Same here with the tag, perhaps: `<image>:<tag>`

docs/imagescan.md
```diff
@@ -0,0 +1,105 @@
+# Image scan
+
+Image scan in fleet allows you to scan your image repository, fetch the desired image and update your git repository, 
+without the need to manually update your manifests.
+
+!!! hint "Experimental"
+    This feature is considered as experimental feature.
+    
+Go to fleet.yaml and add the following section.
```
nickgerace, line 9: nit: `fleet.yaml` instead of fleet.yaml

philomory:
This documentation should _probably_ be more explicit about what a semver range is, with an example that isn't just `"*"` (it's worth noting that the SemVer spec doesn't include the word "range" at all, so you're left to guess whose notion of "range" we're using here). It's probably also worth explicitly calling out how prerelease versions are handled; for example, if you write `>= 1.0.0`, then no pre-release SemVer tag will be considered, but if you write `>= 1.0.0-0` then they will be. I had to go digging through the source code to find out that we're using the `Masterminds/semver` library, and then go look at their readme to see how things worked.

nickgerace (approved):

Issue #415, opened 2021-06-14
Fleet sorting files alphabetically, deployment fails due to missing CRDs


dirien:
Hi,

i try to deploy in Rancher the kyverno helm chart. Somehow it looks like that fleet is sorting the resource alphabetically which leads to breaking the deployment as the CRDS are not deployed...

![image](https://user-images.githubusercontent.com/38325136/121922050-e395b000-cd39-11eb-9951-7bc1c5f57b0d.png)

Log from the fleet-agent

```bash

ime="2021-06-14T16:03:13Z" level=info msg="WARNING: This chart or one of its subcharts contains CRDs. Rendering may fail or contain inaccuracies."
--
Mon, Jun 14 2021 6:03:13 pm | time="2021-06-14T16:03:13Z" level=info msg="getting history for release kyverno-kyverno"
Mon, Jun 14 2021 6:03:13 pm | time="2021-06-14T16:03:13Z" level=error msg="error syncing 'cluster-fleet-default-c-fvtf4-e9273a00dc7f/kyverno-kyverno': handler bundle-deploy: unable to build kubernetes objects from release manifest: unable to recognize \"\": no matches for kind \"ClusterPolicy\" in version \"kyverno.io/v1\", requeuing"
```

Thats how the chart repo from kyverno looks like:
![image](https://user-images.githubusercontent.com/38325136/121923104-e3e27b00-cd3a-11eb-9d40-92c7282ce31e.png)
https://github.com/kyverno/kyverno/tree/main/charts/kyverno

Non Fleet deployment is working fine:
![image](https://user-images.githubusercontent.com/38325136/121924444-307a8600-cd3c-11eb-9330-4cf0f890b77c.png)
![image](https://user-images.githubusercontent.com/38325136/121924474-38d2c100-cd3c-11eb-869f-5ae202a54a9e.png)

PR #423, opened 2021-06-17
Fix chart with crd folder not being deployed

StrongMonkey:
We do dry-run before the real install but if CR and CRD both exists in
chart, it will fail since dry-run doesn't install CRD. We need to turn
of validation in order to prevent failure


StrongMonkey:
https://github.com/rancher/fleet/issues/415

PR #446, opened 2021-06-30
Use `cluster.Name` as the context override for `ClientConfig`

epk:
## What

Use `cluster.Name` as the context override for `ClientConfig`

## Why 

This is to accommodate the use case of sharing a single kubeconfig secret between many clusters. 

Currently we cannot share a single kubeconfig as the `ClientConfig` defaults to the `CurrentContext`
```go
// getContextName returns the default, or user-set context name, and a boolean that indicates
// whether the default context name has been overwritten by a user-set flag, or left as its default value
func (config *DirectClientConfig) getContextName() (string, bool) {
	if config.overrides != nil && len(config.overrides.CurrentContext) != 0 {
		return config.overrides.CurrentContext, true
	}
	if len(config.contextName) != 0 {
		return config.contextName, false
	}

	return config.config.CurrentContext, false
}
```

### DIFF
--- a/pkg/controllers/cluster/import.go
+++ b/pkg/controllers/cluster/import.go
@@ -266,8 +266,12 @@ func (i *importHandler) restConfigFromKubeConfig(clusterName string, data []byte
 			if err == nil {
 				raw.Clusters[cluster].CertificateAuthorityData = nil
 			}
+
+			return clientcmd.NewDefaultClientConfig(raw, &clientcmd.ConfigOverrides{CurrentContext: clusterName}).ClientConfig()
 		}
-	} else if raw.Contexts[raw.CurrentContext] != nil {
+	}
+
+	if raw.Contexts[raw.CurrentContext] != nil {
 		cluster := raw.Contexts[raw.CurrentContext].Cluster
 		if raw.Clusters[cluster] != nil {
 			_, err := http.Get(raw.Clusters[cluster].Server)
@@ -277,5 +281,5 @@ func (i *importHandler) restConfigFromKubeConfig(clusterName string, data []byte
 		}
 	}
 
-	return clientcmd.NewDefaultClientConfig(raw, &clientcmd.ConfigOverrides{CurrentContext: clusterName}).ClientConfig()
+	return clientcmd.NewDefaultClientConfig(raw, &clientcmd.ConfigOverrides{}).ClientConfig()
 }
--- a/pkg/controllers/cluster/import.go
+++ b/pkg/controllers/cluster/import.go
@@ -167,7 +167,7 @@ func (i *importHandler) importCluster(cluster *fleet.Cluster, status fleet.Clust
 		apiServerCA = cfg.APIServerCA
 	}
 
-	restConfig, err := i.restConfigFromKubeConfig(secret.Data["value"])
+	restConfig, err := i.restConfigFromKubeConfig(cluster.Name, secret.Data["value"])
 	if err != nil {
 		return status, err
 	}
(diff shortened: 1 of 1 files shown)


pkg/controllers/cluster/import.go
```diff
@@ -259,7 +259,15 @@ func (i *importHandler) restConfigFromKubeConfig(data []byte) (*rest.Config, err
 		return nil, err
 	}
 
-	if raw.Contexts[raw.CurrentContext] != nil {
+	if raw.Contexts[clusterName] != nil {
```
epk, line 262: We could probably error out here if this is nil

pkg/controllers/cluster/import.go
```diff
@@ -259,7 +259,15 @@ func (i *importHandler) restConfigFromKubeConfig(data []byte) (*rest.Config, err
...
+		cluster := raw.Contexts[clusterName].Cluster
+		if raw.Clusters[cluster] != nil {
+			_, err := http.Get(raw.Clusters[cluster].Server)
+			if err == nil {
+				raw.Clusters[cluster].CertificateAuthorityData = nil
+			}
+		}
+	} else if raw.Contexts[raw.CurrentContext] != nil {
```
nickgerace, line 270: Does Fleet maintain backwards compatible behavior even with the context check moved after the new cluster name check?

genevieve: I think this could be backwards compatible if (1) the context entry for a cluster is named differently than the cluster entry _and_ (2) the cluster name provided by the user is also used as the name for a context in the config that points to a _different_ cluster.

nickgerace: Great point @genevieve. Perhaps, it might be a good idea to add checks here for those two scenarios.

nickgerace: Ah that makes sense. Regardless, I think ensuring backwards compatibility is essential for this PR since Fleet auto-updates with Rancher currently.

genevieve: I can swap the if checks so that it continues to default to selecting by the current context?

nickgerace: That would be a good idea, but then we might never use `cluster.Name` as context override. We may want to keep the same if branch order, but add a check.

```go
cluster := raw.Contexts[clusterName].Cluster
if raw.Clusters[cluster] != nil {}
```

We may want to default to the previous behavior if this is `nil`. 

CC: @kinarashah @StrongMonkey

epk: I updated the PR to only use the newer codepath if `raw.Contexts[clusterName] != nil` otherwise it will fallback to the older behaviour for backwards compatibility
```go
// restConfigFromKubeConfig checks kubeconfig data and tries to connect to server. If server is behind public CA, remove CertificateAuthorityData in kubeconfig file.
func (i *importHandler) restConfigFromKubeConfig(clusterName string, data []byte) (*rest.Config, error) {
	clientConfig, err := clientcmd.NewClientConfigFromBytes(data)
	if err != nil {
		return nil, err
	}

	raw, err := clientConfig.RawConfig()
	if err != nil {
		return nil, err
	}

	if raw.Contexts[clusterName] != nil {
		cluster := raw.Contexts[clusterName].Cluster
		if raw.Clusters[cluster] != nil {
			_, err := http.Get(raw.Clusters[cluster].Server)
			if err == nil {
				raw.Clusters[cluster].CertificateAuthorityData = nil
			}
		}

		return clientcmd.NewDefaultClientConfig(raw, &clientcmd.ConfigOverrides{CurrentContext: clusterName}).ClientConfig()
	}

	if raw.Contexts[raw.CurrentContext] != nil {
		cluster := raw.Contexts[raw.CurrentContext].Cluster
		if raw.Clusters[cluster] != nil {
			_, err := http.Get(raw.Clusters[cluster].Server)
			if err == nil {
				raw.Clusters[cluster].CertificateAuthorityData = nil
			}
		}
	}

	return clientcmd.NewDefaultClientConfig(raw, &clientcmd.ConfigOverrides{}).ClientConfig()
}
```

pkg/controllers/cluster/import.go
```diff
@@ -259,6 +259,18 @@ func (i *importHandler) restConfigFromKubeConfig(data []byte) (*rest.Config, err
 		return nil, err
 	}
 
+	if raw.Contexts[clusterName] != nil {
+		cluster := raw.Contexts[clusterName].Cluster
```
StrongMonkey, line 263: can you combine the logic http://fleet.rancher.io/gitrepo-structure/ to this if? We can avoid duplicate code.

Issue #383, opened 2021-05-20
support schedule based rollout


ibrokethecloud:
Users should be able to rollout update to downstream clusters based on a predefined schedule.

This is useful for users running a large number of clusters, but would prefer to have the rollout perform during a specific window.

Also related to: https://github.com/rancher/rancher/issues/30266

shane-davidson:
Our use case for this would include the option for a global schedule or a targeted schedule per cluster. With the targeted option overriding the global value.


```
defaultNamespace: test
  schedule:
    # global schedule details
    range: '03:00 - 04:00'
    weekday: saturday

targetCustomizations:
- name: cluster1
  clusterSelector:
    matchLabels:
      management.cattle.io/cluster-display-name: cluster1
  schedule:
    # cluster specific schedule
    range: '03:00 - 04:00'
    period: daily  ( or weekly or monthly etc )
    
- name: cluster2
  clusterSelector:
    matchLabels:
      management.cattle.io/cluster-display-name: cluster2
  schedule:
    # cluster specific schedule
    weekday: wednesday
    period: monthly
```

philomory:
This would be _huge_ for us.

I'd like to at least advance the idea that this could potentially be something worth controlling on a cluster or cluster-group level, separate from the per-git-repo or per-bundle settings in `fleet.yaml`. That said, even if the schedule has to be repeated in every repo's `fleet.yaml`, that'd still be a huge improvement over not having scheduling control at all.

philomory:
@deniseschannon This was removed from the 2.6 milestone, is there any chance that this could be added back as part of the 2.6.1, 2.6.2 or 2.6.x milestones?

hyunsun:
@deniseschannon @ibrokethecloud Will this be available in the future release?

Athosone:
Hi guys !

Are you still aiming at implementing this feature for the 2.6.X release ?

CiraciNicolo:
It seems that the feature has been postponed many times, there are any news? Do you accept contributions? I would be happy to implement this feature.

PR #450, opened 2021-07-02
Ability to define schedule for bundles

ibrokethecloud:
The PR introduces `schedule` and `scheduleWindow` fields in the `cluster.fleet` and `bundle` crd's.

This allows users to define a cron based schedule and a schedule window at the bundle or cluster level.

If a schedule exists at the cluster level, then the schedule is applied to all bundles in scope for the said cluster. The cluster level schedule takes precedence of bundle level schedule if there is one.

The user can add a schedule and window to the cluster as follows:

```yaml:
kind: Cluster
metadata:
  name: c-kd88w
  namespace: fleet-default
spec:
  clientID: 5n7ddd4ftxmwg5nd4zm4j99kcgdtjz9fq8cbv9pqxxfmqlgcptv65n
  kubeConfigSecret: c-kd88w-kubeconfig
  paused: false
  redeployAgentGeneration: 3
  schedule: 05 03 * * *
  scheduleWindow: 10m
```

Similarly a schedule and window can also be defined at the bundle level too.

```yaml:
namespace: default
schedule: "30 13 * * *"
scheduleWindow: "5m"
helm:
  releaseName: labels
  values:
    clusterName: global.fleet.clusterLabels.management.cattle.io/cluster-display-name
diff:
  comparePatches:
  - apiVersion: networking.k8s.io/v1beta1
    kind: Ingress
    name: labels-fleetlabelsdemo
    namespace: default
    jsonPointers:
    - "/spec"
```

Once the bundle has been processed by the fleet controller, the downstream cluster's fleet agent will honor the scheduling of the said bundle deployment and deploy the same.

The scheduleWindow allows us to handle use cases where the fleet agent may be disconnected for any reason. 

This allows for the downstream cluster to still process the bundle as long as the reconnect time falls within the schedule + scheduleWindow.

If the cluster fails to reconnect, the bundle will get rescheduled.

Related to: https://github.com/rancher/fleet/issues/383


cloudnautique:
@ibrokethecloud is there a concept of a maintenance window? If a user wants to specify that clusters can be updated between midnight and 5 AM for example? Could a combination of wild cards and a Window be used? 

Also, will the fleet yaml bundles allow for target overrides?

ibrokethecloud:
Hi @cloudnautique there are two main items here.. The cron schedule, at which time the downstream cluster will try and process the bundle. Since the bundle deployment is processed by downstream cluster, there are cases where the cluster may be unable to reach the Rancher master for the bundle deployments. In that case the `scheduleWindow` kicks in. The idea is that if the downstream cluster connects back within the Start Time (`schedule`) +  `scheduleWindow` , the bundle will still be processed. This should help manage use cases where there may be a change window, and cluster may be down for other changes. 

The user can specify the `schedule` and `scheduleWindow` at the cluster or bundle level. If a `schedule` and `scheduleWindow` are present on the cluster, then all bundles being applied to this cluster will be applied using this spec. Any bundle level `schedule` and `scheduleWindow` will be overridden. The idea is that the cluster admin would be able to define the schedule and window at the cluster level, and may be we dont want a developer to apply a bundle outside this schedule + window combo.

If there are no cluster specific `schedule` and `scheduleWindow` then the users can specify the same at the bundle level, and again apply target specific overrides for `schedule` and `scheduleWindow` too.

philomory:
Is there any possibility of this PR being reviewed? This functionality was originally scheduled to go into Rancher 2.6 (so, Fleet 0.3.6), but, that milestone was removed from the associated issue, and this PR hasn't had any movement in a few months.

pkg/apis/fleet.cattle.io/v1alpha1/bundle.go
```diff
@@ -31,6 +31,8 @@ var (
 	}
 )
 
+const DefaultWindow = "1h"
```
nickgerace, line 34: We should put this in the Helm chart.

modules/agent/pkg/controllers/bundledeployment/controller.go
```diff
@@ -80,6 +82,55 @@ func (h *handler) Cleanup(key string, bd *fleet.BundleDeployment) (*fleet.Bundle
...
+		after := scheduledRun.Sub(time.Now())
+		h.bdController.EnqueueAfter(bd.Namespace, bd.Name, after)
+		status.ScheduledAt = scheduledRun.Format(time.RFC3339)
+		status.Scheduled = true
+		condition.Cond(fleet.BundleScheduledCondition).SetStatusBool(&status, true)
+		condition.Cond(fleet.BundleDeploymentConditionDeployed).SetStatusBool(&status, false)
+		return status, nil
+	}
+
+	if bd.Spec.Options.Schedule != "" && status.ScheduledAt != "" {
```
nickgerace, line 101: This seems unreachable based on line 86 and 98?

modules/agent/pkg/controllers/bundledeployment/controller.go
```diff
@@ -80,6 +82,55 @@ func (h *handler) Cleanup(key string, bd *fleet.BundleDeployment) (*fleet.Bundle
...
+		}
+
+		windowDuration, err := time.ParseDuration(window)
+		if err != nil {
+			return status, err
+		}
+
+		if err != nil {
+			return status, err
+		}
```
nickgerace, line 118: Duplicate?

modules/agent/pkg/controllers/bundledeployment/controller.go
```diff
@@ -80,6 +82,55 @@ func (h *handler) Cleanup(key string, bd *fleet.BundleDeployment) (*fleet.Bundle
 }
 
 func (h *handler) DeployBundle(bd *fleet.BundleDeployment, status fleet.BundleDeploymentStatus) (fleet.BundleDeploymentStatus, error) {
+
+	if bd.Spec.Options.Schedule != "" && status.ScheduledAt == "" {
```
nickgerace, line 86: What happens if one is unset and the other isn't?

nickgerace:
@StrongMonkey we should get this into next release.

pkg/apis/fleet.cattle.io/v1alpha1/bundle.go
ibrokethecloud: @nickgerace do you want to expose this default Window as an option for the agent controller? similar to checkin interval?

modules/agent/pkg/controllers/bundledeployment/controller.go
ibrokethecloud: yep 🤦

modules/agent/pkg/controllers/bundledeployment/controller.go
ibrokethecloud: There are two ways this can happen:

1. You edit the bundle and remove the Schedule from the bundleSpec. In this scenario, the checks on 86 and 101 will not be met, and the bundle will be deployed immediately.
2. You remove the status.ScheduledAt, this will cause a reconcile of the bundle, and a new ScheduledAt will be applied to bundle, the bundle will be requeued to be applied at the specified time.

ibrokethecloud:
@nickgerace i have rebased changes from master. May need a hand to kick off the failed CI step again.

pkg/apis/fleet.cattle.io/v1alpha1/bundle.go
nickgerace: Yes!

modules/agent/pkg/controllers/bundledeployment/controller.go
nickgerace: Alright, that checks out

modules/agent/pkg/controllers/bundledeployment/controller.go
nickgerace: Alright, that sounds good

modules/agent/pkg/controllers/bundledeployment/controller.go
```diff
@@ -90,6 +92,55 @@ func (h *handler) Cleanup(key string, bd *fleet.BundleDeployment) (*fleet.Bundle
...
+		if bd.Spec.Options.ScheduleWindow != "" {
+			window = bd.Spec.Options.ScheduleWindow
+		}
+
+		windowDuration, err := time.ParseDuration(window)
+		if err != nil {
+			return status, err
+		}
+
+		if err != nil {
```
Athosone, line 126: Hi,

There is a duplicated error check here

kkaempf:
Closing, see #3726

PR #466, opened 2021-07-21
BUG FIX: fix bug where apply would create then try to update a nil resource

kalinon:
when `fleet apply` is used it will error out when making new bundle records. this is because it would create then try to update a nil resource.

Before:
```
$ fleet apply monitoring
INFO[0001] created: fleet-local/monitoring
FATA[0001] resource name may not be empty
```

After:
```
$ fleet apply monitoring
INFO[0001] created: fleet-local/monitoring
INFO[0002] created: fleet-local/monitoring-crd
INFO[0002] created: fleet-local/monitoring-novo
```

### DIFF
--- a/modules/cli/apply/apply.go
+++ b/modules/cli/apply/apply.go
@@ -142,6 +142,7 @@ func Dir(ctx context.Context, client *client.Getter, name, baseDir string, opts 
 		opts = &Options{}
 	}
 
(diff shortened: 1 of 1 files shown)


StrongMonkey:
Should be fixed in https://github.com/rancher/fleet/pull/472.

Issue #437, opened 2021-06-26
Error in deployment after creating git repo


aaronRancher:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
- On Rancher `master-head` `version c7f3cce` single install.
- Log in as admin.
- Provision 2 downstream clusters.
- Verify clusters are available and visible under Continuous Delivery > Clusters
- Navigate to dashboard > Continuous Delivery > Git repo
- Create a git repo with the following fields:
Repository URL: https://github.com/rancher/fleet-examples
Path: multi-cluster/helm
Deploy To: all clusters in the workspace
- Wait for git repo to become available in dashboard.
- Navigate to Ember UI > Local > System workloads

**Result:**
1 pod is available. 1 pod errors with: `Error: fleet-default/testrepo-simple time="2021-06-25T23:56:02Z" level=fatal msg="resource name may not be empty"`
![Screen Shot 2021-06-25 at 5 03 39 PM](https://user-images.githubusercontent.com/57409052/123504857-9b676f00-d610-11eb-816c-196317804c02.png)
This error is seen with both `multi-cluster/helm` and `/simple` path
This error is seen in both Rancher HA and single install

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): `master-head` `version c7f3cce`
- Installation option (single install/HA): single install

nickgerace:
Error reported comes from here: https://github.com/tektoncd/pipeline/tree/v0.16.3/cmd/entrypoint

cbron:
Is this 2.6 only ? Would it have existing in 2.5.

nickgerace:
It is. It was due to recent changes in the agent in v0.3.6-rc1.

aaronRancher:
This issue is seen on Rancher `master-head` `version 9e0d1b2`. Tested with the above steps.
Fleet version `fleet-agent:v0.3.6-rc5`

Errors are seen on pods after git repo creation:
![Screen Shot 2021-07-22 at 4 32 44 PM](https://user-images.githubusercontent.com/57409052/126721533-c6354a1d-6f2d-4aac-9730-3fe2e2140809.png)

thehejik:
Tested on `rancher/rancher:master-head 21ab02c`

[...]

PR #472, opened 2021-07-23
Fix resource empty issue and give better status on imageScan

StrongMonkey:
https://github.com/rancher/fleet/issues/437

https://github.com/rancher/fleet/issues/378


modules/cli/apply/apply.go
```diff
@@ -187,16 +187,16 @@ func save(client *client.Getter, bundle *fleet.Bundle, imageScans ...*fleet.Imag
 		logrus.Infof("created: %s/%s\n", bundle.Namespace, bundle.Name)
 	} else if err != nil {
 		return err
+	} else {
```
StrongMonkey, line 190: @nickgerace I actually found out this. We can't update outside of if block cause it will fail on the first run. Obj doesn't exist so when updating it it will just give error says `resource name may not be empty`.

nickgerace (approved):

PR #474, opened 2021-07-28
Delete old agent resources upon upgrade

nickgerace:
Relevant issue: https://github.com/rancher/rancher/issues/33677

### DIFF
--- a/pkg/controllers/cluster/import.go
+++ b/pkg/controllers/cluster/import.go
@@ -101,44 +101,61 @@ func (i *importHandler) OnChange(key string, cluster *fleet.Cluster) (_ *fleet.C
 }
 
 func (i *importHandler) deleteOldAgent(cluster *fleet.Cluster, kc kubernetes.Interface) error {
-	err := kc.CoreV1().Secrets(i.systemNamespace).Delete(i.ctx, "fleet-agent", metav1.DeleteOptions{})
+	err := kc.CoreV1().Secrets(i.systemNamespace).Delete(i.ctx, config.AgentConfigName, metav1.DeleteOptions{})
 	if err != nil && !apierrors.IsNotFound(err) {
 		return err
 	}
 
-	err = kc.CoreV1().Secrets(i.systemNamespace).Delete(i.ctx, "fleet-agent-bootstrap", metav1.DeleteOptions{})
+	err = kc.CoreV1().Secrets(i.systemNamespace).Delete(i.ctx, config.AgentBootstrapConfigName, metav1.DeleteOptions{})
 	if err != nil && !apierrors.IsNotFound(err) {
 		return err
 	}
 
-	deployment, err := kc.AppsV1().Deployments(i.systemNamespace).Get(i.ctx, "fleet-agent", metav1.GetOptions{})
-	if apierrors.IsNotFound(err) {
-		return nil
-	} else if err != nil {
-		return nil
-	}
-
-	logrus.Infof("Deleted old agent for cluster %s/%s", cluster.Namespace, cluster.Name)
-
-	err = kc.AppsV1().Deployments(i.systemNamespace).Delete(i.ctx, "fleet-agent", metav1.DeleteOptions{})
-	if err != nil {
+	if err := i.deleteOldAgentDeployment(kc, i.systemNamespace, cluster.Namespace, cluster.Name); err != nil {
 		return err
 	}
-
-	pods, err := kc.CoreV1().Pods(i.systemNamespace).List(i.ctx, metav1.ListOptions{
-		LabelSelector: metav1.FormatLabelSelector(deployment.Spec.Selector),
-	})
-	if err != nil {
-		return err
-	}
-
-	for _, pod := range pods.Items {
-		err := kc.CoreV1().Pods(i.systemNamespace).Delete(i.ctx, pod.Name, metav1.DeleteOptions{})
-		if err != nil && !apierrors.IsNotFound(err) {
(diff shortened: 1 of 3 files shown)


pkg/controllers/cluster/import.go
```diff
@@ -139,6 +140,21 @@ func (i *importHandler) deleteOldAgent(cluster *fleet.Cluster, kc kubernetes.Int
...
+			return err
+		}
+		apply, err := apply.NewForConfig(restConfig)
+		if err != nil {
+			return err
+		}
+		// Using apply with nil objects to delete objects in the default namespace.
+		if err := apply.WithSetID(deployer.GetSetID(config.AgentConfigName, config.DefaultNamespace)).ApplyObjects(nil); err != nil {
+			return err
+		}
```
nickgerace, line 155: We might even want to remove this if we want to leave everything else behind. thoughts?

pkg/controllers/cluster/import.go
```diff
@@ -139,6 +140,21 @@ func (i *importHandler) deleteOldAgent(cluster *fleet.Cluster, kc kubernetes.Int
 		}
 	}
 
+	if i.systemNamespace != config.DefaultNamespace {
+		// Attempt to delete the agent in the default namespace if it has changed.
+		if err := kc.AppsV1().Deployments(config.DefaultNamespace).Delete(i.ctx, config.AgentConfigName, metav1.DeleteOptions{}); err != nil {
```
StrongMonkey, line 145: We need to check if deployment or namespace exist and if they don't exist we need to skip. Otherwise this will fail.

pkg/controllers/cluster/import.go
```diff
@@ -139,6 +140,21 @@ func (i *importHandler) deleteOldAgent(cluster *fleet.Cluster, kc kubernetes.Int
...
+			return err
+		}
+		apply, err := apply.NewForConfig(restConfig)
```
StrongMonkey, line 148: Actually we might just need to drop this. I doubt this will actually work and magically clean up everything.

pkg/controllers/cluster/import.go
StrongMonkey: It is better to clean up other resources given the fact that we are already cleaning fleet-agent deployment. The apply is not going to work most likely. We can try deleting it manually.

pkg/controllers/cluster/import.go
nickgerace: Agreed. Dropped

pkg/controllers/cluster/import.go
nickgerace: Good catch. Worked for me because it existed

pkg/controllers/cluster/import.go
```diff
@@ -139,6 +137,20 @@ func (i *importHandler) deleteOldAgent(cluster *fleet.Cluster, kc kubernetes.Int
 		}
 	}
 
+	if i.systemNamespace != config.DefaultNamespace {
+		if _, err := kc.CoreV1().Namespaces().Get(i.ctx, config.DefaultNamespace, metav1.GetOptions{}); err != nil {
+			if !apierrors.IsNotFound(err) {
```
StrongMonkey, line 142: Well...
```
If apierrors.IsNotFound(err) {
 do nothing and return. // don't need to log, this will be annoying in a fresh install
}
```

pkg/controllers/cluster/import.go
```diff
@@ -139,6 +137,20 @@ func (i *importHandler) deleteOldAgent(cluster *fleet.Cluster, kc kubernetes.Int
...
+				logrus.Infof("Deleted old agent for cluster %s/%s", cluster.Namespace, cluster.Name)
+				return nil
+			}
+			return err
+		}
+		if err := kc.AppsV1().Deployments(config.DefaultNamespace).Delete(i.ctx, config.AgentConfigName, metav1.DeleteOptions{}); err != nil {
```
StrongMonkey, line 148: Same for here. If err != nil && errors.isNotFound(err) then return error. Otherwise this will fail if fleet-system exist but no fleet-agent deployment.

StrongMonkey: You can test if this should enough, cover both case if namespace and deployment doesn't exist. If both doesn't exist then we should do nothing, which is the case in fresh install.
```
if err := kc.AppsV1().Deployments(config.DefaultNamespace).Delete(i.ctx, config.AgentConfigName, metav1.DeleteOptions{}); err != nil && errors.isNotFound(err) {
```

nickgerace: I'll return nil in order to avoid the logging message if it's not found.

pkg/controllers/cluster/import.go
nickgerace: Whoops, copy/paste mistake

pkg/controllers/cluster/import.go
```diff
@@ -139,6 +138,22 @@ func (i *importHandler) deleteOldAgent(cluster *fleet.Cluster, kc kubernetes.Int
...
+			if apierrors.IsNotFound(err) {
+				return nil
+			}
+			return err
+		}
+		logrus.Infof("Deleted old agent for cluster (%s/%s) in namespace: %s", cluster.Namespace, cluster.Name, config.DefaultNamespace)
```
nickgerace, line 154: @StrongMonkey I returned `nil` since I only want to log the successful deletion if it occurs without error (even "NotFound")

StrongMonkey (dismissed):

StrongMonkey (approved):

PR #475, opened 2021-07-29
Ensure namespace is not deleted (again) and delete old agent after new one is created

nickgerace:
Relevant issues:
- https://github.com/rancher/rancher/issues/33203
- https://github.com/rancher/rancher/issues/33819

### DIFF
--- a/pkg/controllers/cluster/import.go
+++ b/pkg/controllers/cluster/import.go
@@ -13,15 +13,14 @@ import (
 	fleet "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
 	"github.com/rancher/fleet/pkg/config"
 	fleetcontrollers "github.com/rancher/fleet/pkg/generated/controllers/fleet.cattle.io/v1alpha1"
+	fleetns "github.com/rancher/fleet/pkg/namespace"
 	"github.com/rancher/wrangler/pkg/apply"
 	corecontrollers "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	"github.com/rancher/wrangler/pkg/randomtoken"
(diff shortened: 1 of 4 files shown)


pkg/controllers/manageagent/manageagent.go
```diff
@@ -86,6 +87,7 @@ func (h *handler) OnNamespace(key string, namespace *corev1.Namespace) (*corev1.
 		WithOwner(namespace).
 		WithDefaultNamespace(namespace.Name).
 		WithListerNamespace(namespace.Name).
+		WithNoDeleteGVK(fleetns.GVK()).
```
StrongMonkey, line 90: You probably don't need this. This apply only creates bundle CRD.

pkg/controllers/cluster/import.go
```diff
@@ -254,15 +230,26 @@ func (i *importHandler) importCluster(cluster *fleet.Cluster, status fleet.Clust
 		return status, err
 	}
 
-	if err := i.deleteOldAgent(cluster, kc); err != nil {
```
StrongMonkey, line 257: You can't switch the order here. Otherwise fleet-agent deployment will be deployed then deleted.

StrongMonkey: I would prefer not touching this unless we need to. Darren probably had a reason to clean up agent first then apply with new agent. Keep the same order so that we can avoid any regression.

StrongMonkey (approved):

pkg/controllers/cluster/import.go
nickgerace: Got it. Only thing that has changed is the re-organization of the method itself now

PR #500, opened 2021-08-25
Check agentEnvVars via cluster labels

nickgerace:
Relevant issue: https://github.com/rancher/rancher/issues/34368


pkg/controllers/cluster/controller.go
```diff
@@ -202,3 +213,32 @@ func (h *handler) createNamespace(cluster *fleet.Cluster, status fleet.ClusterSt
...
+		if cluster.Labels == nil {
+			cluster.Labels = make(map[string]string, 1)
+		}
+		cluster.Labels[agentEnvVarsHashLabel] = hashed
+	}
+	return nil
+}
+
+func hashAgentEnvVars(agentEnvVar []v1.EnvVar) (string, error) {
+	if len(agentEnvVar) < 0 {
```
StrongMonkey, line 234: We need to check len(agentEnvVar) == 0. And If there is no agentEnv, hash should be no-op and we shouldn't be writing labels to clusters. This will give minimal impacts on majority of the clusters that don't have agentEnv set, cause json.marshall and hash function are not cheap.

pkg/controllers/cluster/controller.go
```diff
@@ -21,6 +24,8 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+var agentEnvVarsHashLabel = "agent-env-vars/hash"
```
StrongMonkey, line 27: better to use `agent-env-vars.fleet.cattle.io/hash`

pkg/controllers/cluster/controller.go
nickgerace: Few accidental mistakes there. Good catch.

PR #507, opened 2021-08-27
Allow helm values overrides to inject cluster labels via go templates

gregsidelinger:
Signed-off-by: Greg Sidelinger <gate@ilive4code.net>

### DIFF
--- a/docs/gitrepo-structure.md
+++ b/docs/gitrepo-structure.md
@@ -73,8 +73,8 @@ helm:
   # These can now be accessed directly as variables
     variableName: global.fleet.clusterLabels.LABELNAME
   # They can also be access using go template logic
-    variableName: "kubernetes.io/cluster/{{ .LABELNAME }}"
-    variableName: "{{ if eq .LABELNAME \"production\" }}Production Workload{{ else }}Non Prod{{ end }}"
+    variableName: "kubernetes.io/cluster/{{ .global.fleet.clusterLabels.LABELNAME }}"
+    variableName: "{{ if eq .global.fleet.clusterLabels.LABELNAME \"production\" }}Production Workload{{ else }}Non Prod{{ end }}"
 
   # Path to any values files that need to be passed to helm during install
   valuesFiles:
--- a/docs/gitrepo-structure.md
+++ b/docs/gitrepo-structure.md
@@ -72,6 +72,10 @@ helm:
   # All labels on Rancher clusters are available using global.fleet.clusterLabels.LABELNAME
   # These can now be accessed directly as variables
     variableName: global.fleet.clusterLabels.LABELNAME
+  # They can also be access using go template logic
+    variableName: "kubernetes.io/cluster/{{ .LABELNAME }}"
+    variableName: "{{ if eq .LABELNAME \"production\" }}Production Workload{{ else }}Non Prod{{ end }}"
+
   # Path to any values files that need to be passed to helm during install
   valuesFiles:
     - values1.yaml
--- a/pkg/target/target.go
+++ b/pkg/target/target.go
@@ -512,13 +512,13 @@ func Summary(targets []*Target) fleet.BundleSummary {
 
 func processLabelValues(valuesMap map[string]interface{}, clusterLabels map[string]string) error {
 	prefix := "global.fleet.clusterLabels."
-        scopedClusterLables := map[string]interface{}{
-                "global": map[string]interface{}{
-                        "fleet": map[string]interface{}{
-                                "clusterLabels": clusterLabels,
-                        },
-                },
-        }
+	scopedClusterLables := map[string]interface{}{
+		"global": map[string]interface{}{
+			"fleet": map[string]interface{}{
+				"clusterLabels": clusterLabels,
+			},
+		},
+	}
 	for key, val := range valuesMap {
 		valStr, ok := val.(string)
 		if ok && strings.HasPrefix(valStr, prefix) {
--- a/pkg/target/target.go
+++ b/pkg/target/target.go
@@ -525,7 +525,7 @@ func processLabelValues(valuesMap map[string]interface{}, clusterLabels map[stri
 		} else {
 			switch val.(type) {
 			case string:
-				values_template,_ := template.New("clusterLabels").Option("missingkey=error").Parse(valStr)
+				values_template, _ := template.New("clusterLabels").Option("missingkey=error").Parse(valStr)
 				var tpl bytes.Buffer
 				err := values_template.Execute(&tpl, clusterLabels)
 				if err == nil {
--- a/pkg/target/target.go
+++ b/pkg/target/target.go
@@ -219,7 +219,7 @@ func (m *Manager) Targets(fleetBundle *fleet.Bundle) (result []*Target, _ error)
 			}
 
 			opts := options.Calculate(&fleetBundle.Spec, match.Target)
-			err = addClusterLabels(&opts, cluster.Labels)
+			err = addClusterValues(&opts, cluster.Labels, cluster.Annotations)
 			if err != nil {
 				return nil, err
 			}
@@ -247,21 +247,23 @@ func (m *Manager) Targets(fleetBundle *fleet.Bundle) (result []*Target, _ error)
 	return result, m.foldInDeployments(fleetBundle, result)
 }
 
-func addClusterLabels(opts *fleet.BundleDeploymentOptions, labels map[string]string) (err error) {
+func addClusterValues(opts *fleet.BundleDeploymentOptions, labels map[string]string, annotations map[string]string) (err error) {
 	clusterLabels := yaml.CleanAnnotationsForExport(labels)
+	clusterAnnotations := yaml.CleanAnnotationsForExport(annotations)
 	for k, v := range labels {
 		if strings.HasPrefix(k, "fleet.cattle.io/") || strings.HasPrefix(k, "management.cattle.io/") {
 			clusterLabels[k] = v
 		}
 	}
-	if len(clusterLabels) == 0 {
+	if len(clusterLabels) == 0 && len(clusterAnnotations) == 0 {
 		return
 	}
 
 	newValues := map[string]interface{}{
 		"global": map[string]interface{}{
 			"fleet": map[string]interface{}{
-				"clusterLabels": clusterLabels,
+				"clusterLabels":      clusterLabels,
+				"clusterAnnotations": clusterAnnotations,
 			},
 		},
 	}
@@ -283,7 +285,7 @@ func addClusterLabels(opts *fleet.BundleDeploymentOptions, labels map[string]str
 		return nil
 	}
 
-	if err := processLabelValues(opts.Helm.Values.Data, clusterLabels); err != nil {
+	if err := processValues(opts.Helm.Values.Data, clusterLabels, clusterAnnotations); err != nil {
 		return err
 	}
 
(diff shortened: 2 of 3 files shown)


mattmattox:
This would be really nice for when you want to expose a service on all clusters.
Example: longhorn.ClusterName.example.com

Currently, you can use variables like global.fleet.clusterLabels.management.cattle.io/cluster-display-name in your values but you can't append anything to the variable.

docs/gitrepo-structure.md
```diff
@@ -72,6 +72,10 @@ helm:
   # All labels on Rancher clusters are available using global.fleet.clusterLabels.LABELNAME
   # These can now be accessed directly as variables
     variableName: global.fleet.clusterLabels.LABELNAME
+  # They can also be access using go template logic
```
nickgerace, line 75: Small grammatical fixes:
```
In addition, they can be accessed using Go template logic
```

pkg/target/target_test.go
```diff
@@ -103,4 +106,32 @@ func TestProcessLabelValues(t *testing.T) {
...
+	}
+
+	templateLogic, ok := bundle.Helm.Values.Data["templateLogic"]
+	if !ok {
+		t.Fatal("key templateLogic not found")
+	}
+
+	if templateLogic != "Non Prod" {
+		t.Fatal("unable to assert correct templateLogic")
+	}
```
nickgerace, line 135: nit: let's remove blank lines between the same template comparisons

pkg/target/target.go
```diff
@@ -520,6 +529,18 @@ func processLabelValues(valuesMap map[string]interface{}, clusterLabels map[stri
 			} else {
 				return fmt.Errorf("invalid_label_reference %s in key %s", valStr, key)
 			}
+		} else {
+			switch val.(type) {
```
nickgerace, line 533: We already gather `valStr` on line 523. Perhaps, we can refactor this code to use it instead. I'd prefer that we handle this scenario once.

nickgerace (changes requested):

pkg/target/target.go
gregsidelinger: I actually have an updated PR I was going to send in that changes how a lot of this stuff is done and does not have any extra string conversions in that area of the code.

After playing around with the configmap stuff in 0.3.6 I thought it would be fun to allow the value substitution to come from any overridden values like the ones imported from a configmap, not just the ones in helm.vlaues from the fleet.yaml.  So I moved this code into the helm install function to do that.  I also added basic cluster annotation support on top of the cluster one to deal with values that can not be stored in an label. It's a bigger PR but adds a lot more flexibility that has been preventing us from using fleet for more deployments.

gregsidelinger:
https://github.com/rancher/fleet/pull/528 supersedes this PR

nickgerace:
@gregsidelinger why are these PRs separate? I'd personally prefer one PR with discrete commits (or squashed, if small enough) in order to reduce context switching. However, it's probable that there's a perspective here that I am missing.

gregsidelinger:
@nickgerace Because I'm not 100% sure if there are any major design issues with where/how I did things in #528.  It was a much bigger change than this small PR which has more lines of testing than actual code.   I would say look at #528 and if nothing major jumps out with the approach I will close this one and focus on it.

nickgerace:
Thanks @gregsidelinger for the clarification. I'll pause my review here and jump to the other one.

gregsidelinger:
So after testing this branch and the one in #528 I think this is the PR that should be merged.  I updated this PR with annotation support and added a few more test cases.  As well as trapping errors from calling the go-templating that causes panics.  Docs could get some for TLC as accessing labels and annotations with dashes in their name requires using index which I added a test case for.  But I'm not sure if all of this should end up in the github-structure docs as that is starting to look full and out of place if there are too many examples of using go tempaltes.

The only real difference between this one and #528 is this supports only fleet.yaml values and valuesfiles.  Configmaps and secrets in the destination cluster are not processed, which may have been overkill.

xMAC94x:
Hi @gregsidelinger  thanks for this PR, its exactly what i am looking for. I noticed there is a comment on your PR that need to be resolved before merging.
Fyi, i found out that even Yaml Anchors don't work in current release fleet version ( https://helm.sh/docs/chart_template_guide/yaml_techniques/#yaml-anchors )

pkg/target/target.go
```diff
@@ -508,31 +512,61 @@ func Summary(targets []*Target) fleet.BundleSummary {
...
+			if strings.HasPrefix(valStr, labelPrefix) {
+				label := strings.TrimPrefix(valStr, labelPrefix)
+				labelVal, labelPresent := clusterLabels[label]
+				if labelPresent {
+					valuesMap[key] = labelVal
+				} else {
+					return fmt.Errorf("invalid_label_reference %s in key %s", valStr, key)
+				}
+			} else if strings.HasPrefix(valStr, annotationPrefix) {
+				annoation := strings.TrimPrefix(valStr, annotationPrefix)
```
aiyengar2, line 539:
```suggestion
				annotation := strings.TrimPrefix(valStr, annotationPrefix)
```

nit: typo

pkg/target/target.go
```diff
@@ -508,31 +512,61 @@ func Summary(targets []*Target) fleet.BundleSummary {
...
+				annoationVal, annoationPresent := clusterLabels[annoation]
```
aiyengar2, line 540:
```suggestion
				annotationVal, annotationPresent := clusterLabels[annotation]
```

nit: typo

pkg/target/target.go
```diff
@@ -508,31 +512,61 @@ func Summary(targets []*Target) fleet.BundleSummary {
...
+				if annoationPresent {
+					valuesMap[key] = annoationVal
+				} else {
+					return fmt.Errorf("invalid_annotation_reference %s in key %s", valStr, key)
+				}
 			} else {
-				return fmt.Errorf("invalid_label_reference %s in key %s", valStr, key)
+				valuesTemplate, _ := template.New("values").Option("missingkey=error").Parse(valStr)
```
aiyengar2, line 547: If we are reusing the same template with multiple calls to `.Parse`, `template.New("values").Option("missingkey=error")` should be allocated once in the top of this file instead of allocated per override.

aiyengar2: we can put this into a variable like `tpl := template.New("values").Option("missingkey=error")` that is used in following calls

pkg/target/target.go
```diff
@@ -508,31 +512,61 @@ func Summary(targets []*Target) fleet.BundleSummary {
...
+				var tpl bytes.Buffer
+				defer func() {
+					if r := recover(); r != nil {
+						err = fmt.Errorf("Failed to process template substitution for key '%s' with value '%s': [%v]", key, valStr, err)
+					}
+				}()
```
aiyengar2, line 553: why do we need this? Is there a part of our code that panics?

pkg/target/target_test.go
```diff
@@ -94,13 +106,47 @@ func TestProcessLabelValues(t *testing.T) {
...
+	err = processValues(bundle.Helm.Values.Data, clusterLabels, clusterAnnotations)
+	if err == nil {
+		t.Fatalf("failed to error on {{ .missingValue }}")
+	}
+
+	bundle.Helm.Values.Data["error"] = "{{ .missingValue-panic }}"
+	err = processValues(bundle.Helm.Values.Data, clusterLabels, clusterAnnotations)
+	if err == nil {
+		t.Fatalf("failed to error on {{ .missingValue-panic }}")
+	}
```
aiyengar2, line 151: How are these different tests?

aiyengar2 (changes requested):
Thank you for your PR @gregsidelinger! Apologies for the late re-review, had some comments on this approach though.

docs/gitrepo-structure.md
```diff
@@ -72,6 +72,10 @@ helm:
...
+    variableName: "kubernetes.io/cluster/{{ .global.fleet.clusterLabels.LABELNAME }}"
+    variableName: "{{ if eq .global.fleet.clusterLabels.LABELNAME \"production\" }}Production Workload{{ else }}Non Prod{{ end }}"
```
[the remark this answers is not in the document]
ibuildthecloud: That is an excellent point. I would think we could make this behavior opt in. So something like add parseValuesValues: true (that's a bad name fyi).  Then if somebody wants to templatize something that has templates in it they would escape the template characters.

pkg/target/target.go
gregsidelinger: yes, a bad go template can actually raise a panic.

pkg/target/target_test.go
gregsidelinger: One is designed to raise a panic from the go templating.  I ran into a panic at one point in time so I wrote a test for it.  If I recall correctly the `-` in the variable name actually causes the issue.

docs/gitrepo-structure.md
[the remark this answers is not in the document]
gregsidelinger: Yes, that is true.  An early version of the code actually would leave any lines alone that returned an error from the go templating call allowing fleet to still handle values files with go templating strings in them.  This however would not return errors to the user if they did something like this `variableName: "kubernetes.io/cluster/{{ .global.fleet.clusterLabels.MISSING_LABEL }}"` so I changed it as I forgot why I did it in the first place till you mentioned it.


Assuming I leave all clusterLabels prefixed with .global.fleet. a simple approach would be to only call to go templating if the string contained `{{`, `}}` and `.global.fleet.` which should in theory be safe. for most use cases.   Or only support the templating on stuff in fleet.yaml and not values files.

pkg/target/target.go
xMAC94x: is the panic introduced with this change or is it already in master and you just added the logging ?

gregsidelinger:
@xMAC94x @ibuildthecloud @aiyengar2 
Design question.   I spend some time over the weekend adding a boolean to turn on the templating to deal with charts that support the templating this would break and fixing the other issues pointed out.  I also spend some time on PR https://github.com/rancher/fleet/pull/671 too and I like @rajiteh approach better.  The flexibility of adding templateContext to the clusters seems better than relying upon labels and annotations to inject all data.  And personally I like his format for accessing values inside the templating instead of the .global.fleet prefix I used.
```
	values := map[string]interface{}{
		"ClusterNamespace":   cluster.Namespace,
		"ClusterName":        cluster.Name,
		"ClusterLabels":      clusterLabels,
		"ClusterAnnotations": clusterAnnotations,
		"Values":             templateContext,
	}
```

So from a long term design perspective which approach is preferred?

xMAC94x:
Agree, a special syntax is prob a better solution than put everything in labels, this would also get around of some restrictions as label values cannot contain all UTF-8 characters, right ?

docs/gitrepo-structure.md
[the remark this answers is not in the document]
gregsidelinger: I was thinking of `helm.preProcessValues` as the setting name to turn the templating on.  I actually submitted a patch to @rajiteh for his PR #671 instead of more updates to my code.  Granted I do have all the changes requested for this PR sitting locally but I think his approach is better you are should all review his PR.

prachidamle:
@gregsidelinger Thank you for this PR and also the follow up design questions. I took a look at this PR and the conversations before where the problem of "how do we support templating  for the charts that already have Go templates to be parsed by Helm engine".  I see that you tried out the setting "helm.preProcessValues" method to turn on the Fleet templating when required. 

So I think when the setting "helm.preProcessValues" is "true" it would indicate that Fleet needs to parse the templated values before passing down to Helm. Is that correct?

How would this make Fleet ignore processing of the templated charts intended to be processed by Helm engine? Are for such charts the setting should always be off?

rajiteh:
@prachidamle author of PR #671 here. I have adopted the same config except inversed the config key. Essentially the config value and its default is `helm.disablePreprocess: false`, so the user could override it to `true` to disable templating. This will be particularly useful if they're sending gotpls as values which is the case for things like prometheus alertrules and etc. 

However, a user could still utilize fleet-level helm templates and retain downstream templating ability by wrapping it in an escape block ``{{`{{ escaped_for_downstream }}`}}``

prachidamle:
@rajiteh so essentially when the gotpls are  wrapped in an escape block {{`{{ escaped_for_downstream }}`}}, the fleet template parsing will unescape it for further parsing via helm?

rajiteh:
Yes. The template string ``{{`{{ hello }}`}}`` will get templated out by gotpl (which is what ths proposed solutions are utilizing) as `{{ hello }}`

SheilaghM:
@gregsidelinger @rajiteh - are we good to close this PR in favor of #671 and pursue that avenue?

gregsidelinger:
> @gregsidelinger @rajiteh - are we good to close this PR in favor of #671 and pursue that avenue?

I'm fine with focusing on #671.  I did submit a PR to @rajiteh over the weekend to address the things brought up so far in the review so hopefully he is ok to merge those changes in.

SheilaghM:
Thanks @gregsidelinger.  I'm going to close this ticket then so we focus on #671.

PR #511, opened 2021-08-30
Update alpine base image versions to `3.14.2`

jakefhyde:
Updated alpine to `3.14.2`


nickgerace:
I'll re-review once v0.3.6 is confirmed to be the Fleet version for Rancher v2.6. That will more than likely happen, but will make sure.

nickgerace (approved):

PR #516, opened 2021-09-01
Docs readd valuesFrom

bashofmann:
Reverts https://github.com/rancher/fleet/commit/d45ae281f1bce8331f128fb5c68138ec8cbba3fd

### DIFF
--- a/docs/gitrepo-structure.md
+++ b/docs/gitrepo-structure.md
@@ -187,7 +187,7 @@ targetCustomizations:
 
 These examples showcase the style and format for using `valuesFrom`.
 
-Example `ConfigMap`:
+Example [ConfigMap](https://kubernetes.io/docs/concepts/configuration/configmap/):
 
 ```yaml
 apiVersion: v1
@@ -202,7 +202,7 @@ data:
     serviceType: NodePort
 ```
 
-Example `Secret`:
+Example [Secret](https://kubernetes.io/docs/concepts/configuration/secret/):
 
 ```yaml
 apiVersion: v1
--- a/docs/gitrepo-structure.md
+++ b/docs/gitrepo-structure.md
@@ -185,22 +185,36 @@ targetCustomizations:
 
 ### Using ValuesFrom
 
-The examples below include filler `values` data for a generic Helm chart.
 These examples showcase the style and format for using `valuesFrom`.
 
-Recommended `ConfigMap` contents for the `data` field:
+Example `ConfigMap`:
 
 ```yaml
-values.yaml: |-
-  replication: true
-  replicas: 2
-  serviceType: NodePort
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: configmap-values
+  namespace: default
+data:  
+  values.yaml: |-
+    replication: true
+    replicas: 2
+    serviceType: NodePort
 ```
 
-Recommended `Secret` contents for the `data` field (to be converted into bytes):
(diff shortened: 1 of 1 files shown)


StrongMonkey (dismissed):

docs/gitrepo-structure.md
```diff
@@ -172,6 +183,26 @@ targetCustomizations:
...
+values.yaml: |-
+  replication: true
+  replicas: 2
+  serviceType: NodePort
+```
+
+Recommended `Secret` contents for the `data` field (to be converted into bytes):
+
+```yaml
+replication: true\nreplicas: 2\nserviceType: NodePort
```
nickgerace, line 203: I vaguely remember this not actually being the syntax. In the craziness for v2.6, I forgot if it's the case.
@bashofmann @StrongMonkey should this include raw JSON instead, no newlines, etc.?

bashofmann: I just reverted the reverted commit. I'll test it tomorrow, to make sure that it's right :-)

nickgerace: Thanks @bashofmann. That would be my mistake, not yours, if it was wrong. Please let me know when you get the chance!

bashofmann: The examples were basically fine. I changed them a bit though to include a full ConfigMap or Secret resource to make it less ambiguous.
Also, the data values of a Secret are not "converted into bytes", but just base64 encoded. I made this easier to read by using `stringData` in the example instead.

nickgerace: @bashofmann I used "converted to bytes" to not get users confused that we base64 encode their secret data. We _do_ perform a byte conversion from a Golang "string" object rather than perform the base64 encoding outselves.

docs/gitrepo-structure.md
```diff
@@ -172,6 +183,40 @@ targetCustomizations:
...
+
+Example `Secret`:
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: secret-values
+  namespace: default
+stringData:
```
nickgerace, line 213: We directly index `data` rather than `stringData`: https://github.com/rancher/fleet/blob/a1e504052b5fcd2e5112973acbe8024873cdff10/pkg/helmdeployer/deployer.go#L619

Having said that, I _do_ see a reason to allow the user to provide either or.

(side note: think that would a good idea? @StrongMonkey)

nickgerace (changes requested):

docs/gitrepo-structure.md
nickgerace: @bashofmann explained to me offline:

> You can specify the data and/or the stringData field when creating a configuration file for a Secret. The data and the stringData fields are optional. The values for all keys in the data field have to be base64-encoded strings. If the conversion to base64 string is not desirable, you can choose to specify the stringData field instead, which accepts arbitrary strings as values.
> The keys of data and stringData must consist of alphanumeric characters, -, _ or .. All key-value pairs in the stringData field are internally merged into the data field. If a key appears in both the data and the stringData field, the value specified in the stringData field takes precedence.
>
> - https://kubernetes.io/docs/concepts/configuration/secret/#overview-of-secrets

nickgerace (approved):

StrongMonkey (approved):

nickgerace:
https://github.com/rancher/fleet/issues/384

PR #530, opened 2021-09-18
Allow two agents to run on the same cluster

ibuildthecloud:
The change opens up the ability to run two or more agents pointed to
different fleet-controllers. The specific use case where this is useful
is for fleet managing a cluster that is running fleet to manage other
clusters. More concretely we have this with Rancher managing Rancher.
In this situation you want two fleet agents, one that is running for
fleet-local and one that is run for the upstream fleet-controller.

### DIFF
--- a/cmd/fleetagent/main.go
+++ b/cmd/fleetagent/main.go
@@ -21,6 +21,7 @@ var (
 type FleetAgent struct {
 	Kubeconfig      string `usage:"kubeconfig file"`
 	Namespace       string `usage:"namespace to watch" env:"NAMESPACE"`
+	AgentScope      string `usage:"An identifier used to scope the agent bundleID names, typically the same as namespace" env:"AGENT_SCOPE"`
 	Simulators      int    `usage:"Numbers of simulators to run"`
 	CheckinInterval string `usage:"How often to post cluster status" env:"CHECKIN_INTERVAL"`
 }
@@ -47,7 +48,7 @@ func (a *FleetAgent) Run(cmd *cobra.Command, args []string) error {
 	if a.Simulators > 0 {
 		return simulator.Simulate(cmd.Context(), a.Simulators, a.Kubeconfig, a.Namespace, "default", opts)
 	}
-	if err := agent.Start(cmd.Context(), a.Kubeconfig, a.Namespace, &opts); err != nil {
+	if err := agent.Start(cmd.Context(), a.Kubeconfig, a.Namespace, a.AgentScope, &opts); err != nil {
 		return err
 	}
 	<-cmd.Context().Done()
--- a/modules/agent/pkg/controllers/bundledeployment/controller.go
+++ b/modules/agent/pkg/controllers/bundledeployment/controller.go
@@ -13,9 +13,13 @@ import (
 	fleet "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
 	fleetcontrollers "github.com/rancher/fleet/pkg/generated/controllers/fleet.cattle.io/v1alpha1"
 	"github.com/rancher/wrangler/pkg/condition"
+	"github.com/rancher/wrangler/pkg/merr"
 	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/dynamic"
 )
 
 type handler struct {
@@ -25,10 +29,14 @@ type handler struct {
 	trigger       *trigger.Trigger
 	deployManager *deployer.Manager
 	bdController  fleetcontrollers.BundleDeploymentController
+	restMapper    meta.RESTMapper
+	dynamic       dynamic.Interface
 }
 
 func Register(ctx context.Context,
 	trigger *trigger.Trigger,
+	restMapper meta.RESTMapper,
+	dynamic dynamic.Interface,
 	deployManager *deployer.Manager,
 	bdController fleetcontrollers.BundleDeploymentController) {
 
@@ -37,6 +45,8 @@ func Register(ctx context.Context,
 		trigger:       trigger,
 		deployManager: deployManager,
 		bdController:  bdController,
+		restMapper:    restMapper,
+		dynamic:       dynamic,
 	}
 
 	fleetcontrollers.RegisterBundleDeploymentStatusHandler(ctx,
@@ -146,8 +156,12 @@ func (h *handler) Trigger(key string, bd *fleet.BundleDeployment) (*fleet.Bundle
 	return bd, nil
 }
 
+func isAgent(bd *fleet.BundleDeployment) bool {
+	return strings.HasPrefix(bd.Name, "fleet-agent")
+}
+
 func shouldRedeploy(bd *fleet.BundleDeployment) bool {
-	if strings.HasPrefix(bd.Name, "fleet-agent") {
+	if isAgent(bd) {
 		return true
 	}
 	if bd.Spec.Options.ForceSyncGeneration <= 0 {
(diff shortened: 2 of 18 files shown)


cmd/fleetagent/main.go
```diff
@@ -21,6 +21,7 @@ var (
 type FleetAgent struct {
 	Kubeconfig      string `usage:"kubeconfig file"`
 	Namespace       string `usage:"namespace to watch" env:"NAMESPACE"`
+	AgentScope      string `usage:"An identifier used to scope the agent bundleID names, typically the same as namespace" env:"AGENT_SCOPE"`
```
nickgerace, line 24: Why "scope" over "ID"? I'm confused about what "scope" means in this context.

modules/agent/pkg/controllers/bundledeployment/controller.go
```diff
@@ -159,6 +172,28 @@ func shouldRedeploy(bd *fleet.BundleDeployment) bool {
 	return *bd.Status.SyncGeneration != bd.Spec.Options.ForceSyncGeneration
 }
 
+func (h *handler) cleanupOldAgent(bd *fleet.BundleDeployment, modified []fleet.ModifiedStatus) error {
+	if !isAgent(bd) {
+		return nil
+	}
```
nickgerace, line 178: nit: We should considering adding a NOP trace log.

modules/agent/pkg/controllers/bundledeployment/controller.go
```diff
@@ -159,6 +172,28 @@ func shouldRedeploy(bd *fleet.BundleDeployment) bool {
...
+			mapping, err := h.restMapper.RESTMapping(gvk.GroupKind(), gvk.Version)
+			if err != nil {
+				return err
+			}
+
+			logrus.Infof("Removing old agent resource %s/%s, %s", modified.Namespace, modified.Name, gvk)
+			err = h.dynamic.Resource(mapping.Resource).Namespace(modified.Namespace).Delete(h.ctx, modified.Name, metav1.DeleteOptions{})
+			if err != nil {
+				return err
+			}
```
nickgerace, line 191: Perhaps, we should collect errors and `continue`. Then, we can return the slice of errors.

pkg/controllers/cluster/import.go
```diff
@@ -213,7 +230,13 @@ func (i *importHandler) importCluster(cluster *fleet.Cluster, status fleet.Clust
 	}
 
 	output := &bytes.Buffer{}
-	err = agentmanifest.AgentManifest(i.ctx, i.systemNamespace, i.systemNamespace, &client.Getter{Namespace: cluster.Namespace}, output, token.Name, &agentmanifest.Options{
+	agentNamespace := i.systemNamespace
```
StrongMonkey, line 233: Is this the whole idea that for local cluster rancher will set spec.agentNamespace to `cattle-fleet-local-system` so that fleet-agent will be installed from there and other fleet-agent goes `cattle-fleet-system`?

pkg/helmdeployer/deployer.go
```diff
@@ -64,9 +65,10 @@ type helm struct {
 	template            bool
 	defaultNamespace    string
 	labelPrefix         string
+	labelSuffix         string
```
StrongMonkey, line 68: hmm what's the purpose of labelSuffix?

modules/agent/pkg/controllers/bundledeployment/controller.go
```diff
@@ -159,6 +172,28 @@ func shouldRedeploy(bd *fleet.BundleDeployment) bool {
 	return *bd.Status.SyncGeneration != bd.Spec.Options.ForceSyncGeneration
 }
 
+func (h *handler) cleanupOldAgent(bd *fleet.BundleDeployment, modified []fleet.ModifiedStatus) error {
```
nickgerace, line 175: nit: we should probably rename "modified" to "modifiedStatuses" to avoid confusion with the below

modules/agent/pkg/controllers/bundledeployment/controller.go
```diff
@@ -159,6 +172,28 @@ func shouldRedeploy(bd *fleet.BundleDeployment) bool {
...
+	for _, modified := range modified {
+		if modified.Delete {
+			gvk := schema.FromAPIVersionAndKind(modified.APIVersion, modified.Kind)
+			mapping, err := h.restMapper.RESTMapping(gvk.GroupKind(), gvk.Version)
+			if err != nil {
+				return err
+			}
+
+			logrus.Infof("Removing old agent resource %s/%s, %s", modified.Namespace, modified.Name, gvk)
+			err = h.dynamic.Resource(mapping.Resource).Namespace(modified.Namespace).Delete(h.ctx, modified.Name, metav1.DeleteOptions{})
```
nickgerace, line 188: Wrangler apply with a setID or owner reference isn't ideal in this scenario?

modules/agent/pkg/register/register.go
```diff
@@ -187,13 +186,8 @@ func createClusterSecret(ctx context.Context, clusterID string, k8s corecontroll
 		newToken := newSecret.Data[Token]
 		clusterNamespace := newSecret.Data[ClusterNamespace]
 		clusterName := newSecret.Data[ClusterName]
-		systemNamespace := string(newSecret.Data[SystemNamespace])
 		deploymentNamespace := newSecret.Data[DeploymentNamespace]
 
-		if !cfg.IgnoreAgentNamespaceCheck && systemNamespace != secret.Namespace {
-			return nil, fmt.Errorf("fleet-agent must be installed in the namespace %s", systemNamespace)
-		}
```
nickgerace, line 195: Never checking the namespace seems fine to me, though I wonder what was the motivation for its original addition.

pkg/helmdeployer/deployer.go
StrongMonkey: I see, it is for differentiating fleet-agent in local cluster so it will be using different setID to track bundle..

pkg/controllers/cluster/import.go
```diff
@@ -230,7 +260,19 @@ func (i *importHandler) importCluster(cluster *fleet.Cluster, status fleet.Clust
...
+		if cluster.Status.Agent.Namespace != "" {
+			if err := i.deleteOldAgent(cluster, kc, cluster.Status.Agent.Namespace); err != nil {
+				return status, err
+			}
+		}
+	}
+
+	if err := i.deleteOldAgent(cluster, kc, agentNamespace); err != nil {
 		return status, err
 	}
```
nickgerace, line 279: Maybe we should consider deleting the bundle within the `deleteOldAgent` method for readability.

pkg/helmdeployer/deployer.go
nickgerace: Seems like it's to scope the correct SetID for the specific agent

ibuildthecloud: When the agent runs the apply code to find resources based on the setID which assumes the bundleID is globally unique for the cluster.  If I have two agents running with the same bundleID they stop on each other.  The labelSuffix is added to the setID to make them unique per agent. For upgrade we only append the labelSuffix for agents for clusters that have AgentNamespace set.

cmd/fleetagent/main.go
ibuildthecloud: AgentScope == labelSuffix, so read the comment above about what labelSuffix does.

modules/agent/pkg/register/register.go
ibuildthecloud: It was that 90% of the plumbing was there to support this scenario of not having the namespaces matching, but it wasn't complete, so we ensured the namespaces matched.  Now in theory it should work and we should be implicitly testing it in rancher.

pkg/controllers/cluster/import.go
ibuildthecloud: I'm afraid of always deleting the bundle as this will only apply to the agent for the local cluster, not all clusters under management.

modules/agent/pkg/controllers/bundledeployment/controller.go
ibuildthecloud: Nah, we know exactly what we want to delete, so just use the dynamic client.

modules/agent/pkg/controllers/bundledeployment/controller.go
ibuildthecloud: done

modules/agent/pkg/controllers/bundledeployment/controller.go
ibuildthecloud: done

modules/agent/pkg/controllers/bundledeployment/controller.go
ibuildthecloud: logging a noop would be too much.  I moved the if out of the method to make it less weird.

pkg/controllers/cluster/import.go
ibuildthecloud: yep.  Two gold stars for you :)

modules/agent/pkg/register/register.go
nickgerace: That makes sense to me. Thanks.

pkg/controllers/cluster/import.go
nickgerace: I'm also thinking we could lock it behind a `bool`, but I'm good with the way it's currently written then.

nickgerace:
Relevant issue: https://github.com/rancher/fleet/issues/531

nickgerace:
@ibuildthecloud what if you create a GitRepo CR in the Rancher cluster managing a Rancher cluster? Does the FleetWorkspace of that downstream Rancher cluster matter?

ibuildthecloud:
No, it doesn't matter. The gitrepos are still in fleet-local (and bundles). The only thing that should be changing is the namespace of where the agent runs for local.

StrongMonkey (approved):
LGTM

nickgerace (approved):
YOLO MERGE 🚀

PR #552, opened 2021-09-29
Improve bundle dependency matching

ibrokethecloud:
Currently bundle's can have dependency based on name of other bundles. This PR attempts to improve this.

The following changes are bundled with this PR:

* Ability to use bundle labels for defining dependencies: This makes it easy to handle multiple bundles via matching labels in place of naming bundles in advance.
* Ability to define custom labels on your bundles
* Ability to define a custom bundle name, rather than combination of bundleName-Path when used directly or GitRepo-Path when consumed via continuous delivery.

The PR also changes the behavior of a bundle when a dependency doesnt exist. At present the bundle is deployed if no bundle matching the dependency is found. Post this PR, the bundle will requeue until a bundle matching the dependencies is found. This helps avoid use cases when bundles are applied out of order but have explicit dependencies available in them.

A few sample bundle fleet.yaml's are as follows:

bundle 1
```yaml
namespace: default
customBundleName: demo1
additionalBundleLabels:
  app: demo1
targets:
- clusterSelector: {}  
```

bundle 2
```yaml
namespace: default
customBundleName: demo2
additionalBundleLabels:
  app: demo2
dependsOn:
  - bundleSelector:
      matchLabels:
        app: demo1
targets:
- clusterSelector: {} 
```


StrongMonkey:
Can we just put something like this?
```
namespace: default
name: demo2
labels:
  app: demo2
dependsOn:
  - bundleSelector:
      matchLabels:
        app: demo1
targets:
- clusterSelector: {} 
```

So that you can actually set `name` and `labels` of the bundles without introducing fields on bundle api.

ibrokethecloud:
@StrongMonkey thanks for the feedback. I have changed the logic to handle name and labels via an internal struct rather than handling it via the BundleSpec.

sample fleet.yaml will look as follows

```yaml
namespace: default
name: demo1
labels:
  app: demo1
targets:
- clusterSelector: {}  
```

This will result in a bundle named `demo1` with the additional labels specified in the labels field.

modules/agent/pkg/controllers/bundledeployment/controller.go
```diff
@@ -111,12 +111,17 @@ func (h *handler) DeployBundle(bd *fleet.BundleDeployment, status fleet.BundleDe
...
+		ls := &metav1.LabelSelector{}
+
+		if depend.BundleSelector != nil {
+			ls = depend.BundleSelector
+		}
+
+		if depend.Name != "" {
+			ls = metav1.AddLabelToSelector(ls, "fleet.cattle.io/bundle-name", depend.Name)
+			ls = metav1.AddLabelToSelector(ls, "fleet.cattle.io/bundle-namespace", bundleNamespace)
 		}
```
nickgerace, line 123: Will line 117's "depend.BundleSelector" get overriden by lines 121 and 122? It appears not to be the case, but just checking.

modules/agent/pkg/controllers/bundledeployment/controller.go
```diff
@@ -125,12 +130,17 @@ func (h *handler) checkDependency(bd *fleet.BundleDeployment) (string, bool, err
...
+			return fmt.Sprintf("no bundles matching labels %s in namespace %s", selector.String(), bundleNamespace), false, nil
+		}
+
 		for _, bd := range bds {
 			c := condition.Cond("Ready")
 			if c.IsTrue(bd) {
 				continue
 			} else {
-				return fmt.Sprintf("%s/%s", bundleNamespace, depend.Name), false, nil
+				return fmt.Sprintf("dependent bundles matching labels %s in namespace %s not ready", selector.String(), bundleNamespace), false, nil
```
nickgerace, line 143: Let's collect errors and return an error slice instead

pkg/bundle/read.go
```diff
@@ -174,6 +180,14 @@ func read(ctx context.Context, name, baseDir string, bundleSpecReader io.Reader,
 		def.Labels[k] = v
 	}
 
+	// apply additional labels from spec
+	for k, v := range bundle.Labels {
+		if def.Labels == nil {
+			def.Labels = map[string]string{}
```
nickgerace, line 186: nit: use `make` for map creation here

pkg/apis/fleet.cattle.io/v1alpha1/bundle.go
```diff
@@ -67,7 +67,8 @@ type BundleSpec struct {
 }
 
 type BundleRef struct {
-	Name string `json:"name,omitempty"`
+	Name           string                `json:"name,omitempty"`
+	BundleSelector *metav1.LabelSelector `json:"bundleSelector,omitempty"`
```
StrongMonkey, line 71: I would just use `selector` as the field name, bundle seems redundant.

modules/agent/pkg/controllers/bundledeployment/controller.go
ibrokethecloud: yeah the labels for name and namespace will be appended to the the provided BundleSelector

pkg/apis/fleet.cattle.io/v1alpha1/bundle.go
ibrokethecloud: fixed up

modules/agent/pkg/controllers/bundledeployment/controller.go
ibrokethecloud: i have collected the dependent bundleNames and returned them, allowing a custom error that wraps around all dependent bundle names.

modules/agent/pkg/controllers/bundledeployment/controller.go
```diff
@@ -108,33 +113,51 @@ func (h *handler) DeployBundle(bd *fleet.BundleDeployment, status fleet.BundleDe
...
-			return "", false, err
-		}
-		for _, bd := range bds {
-			c := condition.Cond("Ready")
-			if c.IsTrue(bd) {
-				continue
-			} else {
-				return fmt.Sprintf("%s/%s", bundleNamespace, depend.Name), false, nil
+		// skip empty BundleRef definitions. Possible if there is a typo in the yaml
+		if depend != (fleet.BundleRef{}) {
```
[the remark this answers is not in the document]
ibrokethecloud: fixed i up

StrongMonkey (approved):
:100:  LGTM. We can add doc examples after this is merged.

nickgerace (approved):

Issue #226, opened 2021-01-15
ClusterGroup status is not updated after removing cluster from group


janeczku:
**Steps to reproduce:**

1. Register two clusters
2. Create a Cluster Group using a common label attached to both clusters
3. Create a GitRepo resource targeting the cluster group
4. After everything is in-sync, remove the common label from one of the cluster such that the group selector only matches a single cluster

**Expected:**

`status.clusterCount` field of the ClusterGroup resource is updated to `1`.

**Actual:**

`status.clusterCount` field of the ClusterGroup resource remains `2` for another 5-10 minutes. Only then it changes to `1`.

**Other information:**

When adding a cluster to a group instead of removing, the status is reflected immediately.

nickgerace:
Keeping in "Review" until a new Fleet RC is available in Rancher.

nickgerace:
Available on Rancher v2.6-head in Fleet v0.3.8-rc1

davidnuzik:
**My checks PASSED**
I filed a new issue, unrelated to the fix, which is mostly likely a frontend issue by my guess. The Cluster Group object updates in the UI and only shows 1 cluster, but on the Git Repos page it will consistently still show 2 clusters ready which is not true. https://github.com/rancher/dashboard/issues/4564
 
**Reproduction Steps:**

Not required.

-----
 
**Validation Environment:**

Rancher version: v2.6-head `cc2b8be` 11/12/2021
Rancher cluster type: single-node docker install
Docker version: 20.10
 
Downstream cluster type: n/a (is not relevant)
Downstream K8s version: n/a (is not relevant)
 
**Validation steps:**

1. Create two clusters, each with a cluster label of `fleet=true` wait for them to become Active.
2. Create a cluster group that selects for the label.
3. Create a GitRepo resource targeting (deploying to) the cluster group. Wait for both clusters to become ready in the UI.
4. After everything is in sync and ready, remove the cluster label from one of the clusters (the cluster group selector should only match a single cluster now).
5. Confirmed that once I remove the label from one of the clusters, the cluster group is reconciled -- meaning it's updated and now only lists 1 cluster in the group. 
Note: I did notice, though, that on the GitRepos page, it still mentions 2 clusters ready, which is not true. I filed a separate bug here for this: https://github.com/rancher/dashboard/issues/4564

PR #554, opened 2021-09-30
Minor tweak to reconcile ClusterGroups when cluster labels are removed

ibrokethecloud:
At present the logic in fleet-controller seems to reconcile ClusterGroups only if labels are added to a cluster.

This ignores the use case when a label is removed from the cluster resulting in cluster not being removed from associated ClusterGroup.

The PR introduces minor logic to trigger reconcile of ClusterGroups if the label is removed from a cluster. 

Related to: https://github.com/rancher/fleet/issues/226

### DIFF
--- a/pkg/controllers/clustergroup/controller.go
+++ b/pkg/controllers/clustergroup/controller.go
@@ -69,6 +69,14 @@ func (h *handler) OnClusterChange(key string, cluster *fleet.Cluster) (*fleet.Cl
 		if sel.Matches(labels.Set(cluster.Labels)) {
 			h.clusterGroups.Enqueue(cg.Namespace, cg.Name)
 		}
+		// if cluster is removed from CG, need to reconcile if ClusterCount doesnt match
+		clusters, err := h.clusterCache.List(cg.Namespace, sel)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


StrongMonkey (approved):
LGTM. We need to hold on merging this until 2.6.2 is out.

nickgerace (approved):

pkg/controllers/clustergroup/controller.go
```diff
@@ -69,6 +69,14 @@ func (h *handler) OnClusterChange(key string, cluster *fleet.Cluster) (*fleet.Cl
 		if sel.Matches(labels.Set(cluster.Labels)) {
 			h.clusterGroups.Enqueue(cg.Namespace, cg.Name)
 		}
+		// if cluster is removed from CG, need to reconcile if ClusterCount doesnt match
+		clusters, err := h.clusterCache.List(cg.Namespace, sel)
+		if err != nil {
+			logrus.Errorf("error fetching clusters in clustergroup %s%s: %v", cg.Namespace, cg.Name, err)
+		}
```
nickgerace, line 76: @StrongMonkey wonder if we should account for "IsNotFound" here...

StrongMonkey: List is not going to return notFound error I think

nickgerace (approved):

PR #558, opened 2021-10-11
Ability to auto-reapply bundles

ibrokethecloud:
The PR introduces a new boolean field in BundleDeploymentOptions `alwaysReapply`

Users can use this flag to trigger re-application of modified bundled on downstream clusters.

When this field is set, the fleet-agent will trigger redeployment of the modified bundle at a fixed interval to avoid overloading the api server.


modules/agent/pkg/controllers/bundledeployment/controller.go
```diff
@@ -252,3 +265,24 @@ func readyError(status fleet.BundleDeploymentStatus) error {
...
+			return ok, err
+		}
+		if reapplyAfter.Before(time.Now()) {
+			status.ReapplyAfter = ""
+			ok = true
+		}
+
+	}
+
+	return ok, nil
```
nickgerace, line 287: I understand your code here, but for readability, could we use literal values instead? `false` over unset `ok`?

pkg/apis/fleet.cattle.io/v1alpha1/bundle.go
```diff
@@ -191,6 +191,7 @@ type BundleDeploymentOptions struct {
 	ForceSyncGeneration int64             `json:"forceSyncGeneration,omitempty"`
 	YAML                *YAMLOptions      `json:"yaml,omitempty"`
 	Diff                *DiffOptions      `json:"diff,omitempty"`
+	AlwaysReapply       bool              `json:"alwaysReapply,omitempty"remo`
```
StrongMonkey, line 194: I think `resync` is probably a good field name.... thoughts? @nickgerace

pkg/apis/fleet.cattle.io/v1alpha1/bundle.go
```diff
@@ -282,6 +283,7 @@ type BundleDeploymentStatus struct {
 	ModifiedStatus      []ModifiedStatus                    `json:"modifiedStatus,omitempty"`
 	Display             BundleDeploymentDisplay             `json:"display,omitempty"`
 	SyncGeneration      *int64                              `json:"syncGeneration,omitempty"`
+	ReapplyAfter        string                              `json:"reapplyAfter,omitempty"`
```
StrongMonkey, line 286: What is this used for? To prevent the bundle being reapplied in 5 minutes?

StrongMonkey: You can use metav1.Time to avoid parsing... and also the best pratice is to record the last time it was reapplied and check if lastAppliedTime is within 5 minutes...

ibrokethecloud: yeah the field is to record the last time it was applied. I make the changes.

nickgerace:
@ibrokethecloud I'd be comfortable with some form of the above feedback from @skaven81 if you have the cycles. Apply delay does not seem as useful to me if we have max backoff and max retries, though I do see the use if the idea is not to hammer an external API (though, I'd argue you have other problems in that case.

I would still prefer to have "resync" exist as a boolean though since we want an intuitive solution here. While us here understand that `-1` means "infinite", not all users will. I'd say maintaining resync and adding options may be the best path forward.

skaven81:
This is awesome, and exactly what I was looking for.  :+1:

leodotcloud (dismissed):
LGTM. Thanks a lot Gaurav for taking care of this.

leodotcloud:
Ping @StrongMonkey @nickgerace ... can you please review and merge this PR?

modules/agent/pkg/controllers/bundledeployment/controller.go
```diff
@@ -103,6 +109,11 @@ func (h *handler) DeployBundle(bd *fleet.BundleDeployment, status fleet.BundleDe
 	if err != nil {
 		return status, err
 	}
+
+	if status.LastApply == nil {
```
StrongMonkey, line 113: Does this only get set once? If it is not nil then it will not be set. Also we might want to set it when it actuallu deploy bundle, since `h.deployManager.Deploy(bd)` could be a no-op.

pkg/apis/fleet.cattle.io/v1alpha1/bundle.go
```diff
@@ -346,3 +350,9 @@ type Content struct {
 
 	Content []byte `json:"content,omitempty"`
 }
+
+type ResyncPolicy struct {
+	MaxRetries   int    `json:"maxRetries,omitempty"`
+	ResyncDelay  string `json:"resyncDelay,omitempty"`
```
StrongMonkey, line 356: Can you make these two fields as `*metav1.Duration` so it can be validated and parsed at api levels? Then you don't have to deal with it in controller code.

modules/agent/pkg/controllers/bundledeployment/controller.go
```diff
@@ -252,3 +279,72 @@ func readyError(status fleet.BundleDeploymentStatus) error {
...
+			if bd.Spec.Options.ResyncPolicy != nil {
+				policy = *bd.Spec.Options.ResyncPolicy
+			}
+
+			if bd.Spec.StagedOptions.ResyncPolicy != nil {
+				policy = *bd.Spec.StagedOptions.ResyncPolicy
+			}
+
+			if policy.ResyncDelay != "" {
+				requeueAfter, err = time.ParseDuration(policy.ResyncDelay)
```
StrongMonkey, line 307: If you make `policy.ResyncDelay` as `metav1.Duration`, you don't have to parse at here. It can be done at api level.

modules/agent/pkg/controllers/bundledeployment/controller.go
```diff
@@ -252,3 +279,72 @@ func readyError(status fleet.BundleDeploymentStatus) error {
 
 	return errors.New(msg)
 }
+
+func hasAutoReapply(bd *fleet.BundleDeployment, status *fleet.BundleDeploymentStatus) (ok bool, requeueAfter time.Duration, err error) {
```
StrongMonkey, line 283: Idealy if you make this method only return `bool` it is much more readable. You don't have to return `requeueAfter` as it can be computed as outside of this method.

modules/agent/pkg/controllers/bundledeployment/controller.go
```diff
@@ -160,17 +171,29 @@ func isAgent(bd *fleet.BundleDeployment) bool {
 	return strings.HasPrefix(bd.Name, "fleet-agent")
 }
 
-func shouldRedeploy(bd *fleet.BundleDeployment) bool {
+func shouldRedeploy(bd *fleet.BundleDeployment, status *fleet.BundleDeploymentStatus) (ok bool, duration time.Duration, err error) {
```
StrongMonkey, line 174: Same. You can make it only return `bool` to improve readability. `ApplyInterval` and err(error can be avoided by changing the api) can be computed/avoided.

pkg/apis/fleet.cattle.io/v1alpha1/bundle.go
```diff
@@ -191,6 +191,8 @@ type BundleDeploymentOptions struct {
 	ForceSyncGeneration int64             `json:"forceSyncGeneration,omitempty"`
 	YAML                *YAMLOptions      `json:"yaml,omitempty"`
 	Diff                *DiffOptions      `json:"diff,omitempty"`
+	Resync              bool              `json:"resync,omitempty"`
```
StrongMonkey, line 194: Do we still need `Resync`? Or we can check if `ResyncPolicy != nil` to decide whether user want to resync or not?

modules/agent/pkg/controllers/bundledeployment/controller.go
```diff
@@ -252,3 +279,72 @@ func readyError(status fleet.BundleDeploymentStatus) error {
...
+
+func hasAutoReapply(bd *fleet.BundleDeployment, status *fleet.BundleDeploymentStatus) (ok bool, requeueAfter time.Duration, err error) {
+	requeueAfter = DefaultReapplyInterval
+	if bd.Spec.Options.Resync || bd.Spec.StagedOptions.Resync {
+		if bd.Spec.Options.ResyncPolicy == nil && bd.Spec.StagedOptions.ResyncPolicy == nil {
+			if status.LastApply == nil {
+				return false, requeueAfter, nil
+			}
+			if status.LastApply.Add(DefaultReapplyInterval).Before(time.Now()) {
+				status.LastApply = nil
```
StrongMonkey, line 291: Seems weird to me that we are resetting `lastApply` field  during `shouldDeploy` method, as it is just checking whether you should redeploy or not. My suggestion to separate the logic so that:

1. You only set `status.LastApplied` and `status.ResyncCounter` after you deployed the bundle. This idealy should live in https://github.com/rancher/fleet/blob/72de385c179ea1c139e53cec813a54e34eb1ac64/modules/agent/pkg/controllers/bundledeployment/controller.go#L112 after the bundle is **actually** applied. There are cases where it is a no-op deploy, so we might want to update it only when it is being applied.
2. `shouldRedeploy` method checks against `lastApplied` and `ResyncCounter` and decide whether it needs to redeploy the bundle based on its own policy. It should not write anything to the status as it is just checking the bundleDeployment.

pkg/apis/fleet.cattle.io/v1alpha1/bundle.go
StrongMonkey: I see the point that nick made. It should be fine.

modules/agent/pkg/controllers/bundledeployment/controller.go
ibrokethecloud: changed the logic to use different in releaseName to identify if a new release has been performed and use the same to calculate LastApply timestamp.

pkg/apis/fleet.cattle.io/v1alpha1/bundle.go
ibrokethecloud: done. Thanks for the pointer.

modules/agent/pkg/controllers/bundledeployment/controller.go
ibrokethecloud: fixed up.

modules/agent/pkg/controllers/bundledeployment/controller.go
ibrokethecloud: done

modules/agent/pkg/controllers/bundledeployment/controller.go
ibrokethecloud: tweaked the logic to ensure status updates are only happening in the DeployBundle and MonitorBundle.

also changed how lastApplied is handled.

nickgerace:
@ibrokethecloud is this ready for re-review?

modules/agent/pkg/controllers/bundledeployment/controller.go
ibrokethecloud: done

modules/agent/pkg/controllers/bundledeployment/controller.go
```diff
@@ -252,3 +277,113 @@ func readyError(status fleet.BundleDeploymentStatus) error {
 
 	return errors.New(msg)
 }
+
+func hasAutoReapply(bd *fleet.BundleDeployment, status *fleet.BundleDeploymentStatus) bool {
+	if bd.Spec.Options.Resync || bd.Spec.StagedOptions.Resync {
```
StrongMonkey, line 282: The same condition has been checked before at https://github.com/rancher/fleet/pull/558/files#diff-8b83e3ec81ef037af3e68026d07535637ca1d84af565be9bcda91135c4b80714R177

modules/agent/pkg/controllers/bundledeployment/controller.go
```diff
@@ -213,8 +232,14 @@ func (h *handler) MonitorBundle(bd *fleet.BundleDeployment, status fleet.BundleD
 	readyError := readyError(status)
 	condition.Cond(fleet.BundleDeploymentConditionReady).SetError(&status, "", readyError)
 	if len(status.ModifiedStatus) > 0 {
-		h.bdController.EnqueueAfter(bd.Namespace, bd.Name, 5*time.Minute)
-		if shouldRedeploy(bd) {
+		duration := customResyncDuration(bd)
+		ok := shouldRedeploy(bd, &status)
+		if ok {
```
StrongMonkey, line 237: There are two `if ok` block, maybe can be combined together?

modules/agent/pkg/controllers/bundledeployment/controller.go
```diff
@@ -252,3 +277,113 @@ func readyError(status fleet.BundleDeploymentStatus) error {
 
 	return errors.New(msg)
 }
+
+func hasAutoReapply(bd *fleet.BundleDeployment, status *fleet.BundleDeploymentStatus) bool {
+	if bd.Spec.Options.Resync || bd.Spec.StagedOptions.Resync {
+		if bd.Spec.Options.ResyncPolicy == nil && bd.Spec.StagedOptions.ResyncPolicy == nil {
+			if status.LastApply == nil {
+				return false
```
StrongMonkey, line 285: This doesn't look correct.... if lastApply is nil and resync is true then we return false? Look like it will never be deployed since that's the initail state..

ibrokethecloud: It will be deployed.

Right now there are two handles for BundleDeployments, DeployBundle and MonitorBundle.

The DeployBundle will deploy the bundle the first time and set the lastApply timestamp. I just wanted MonitorBundle to ignore reapply / re-queueing the bundle until the first deploy has happened.

modules/agent/pkg/controllers/bundledeployment/controller.go
ibrokethecloud: good spot. fixed it up.

modules/agent/pkg/controllers/bundledeployment/controller.go
```diff
@@ -252,3 +275,111 @@ func readyError(status fleet.BundleDeploymentStatus) error {
...
+
+		if bd.Spec.StagedOptions.ResyncPolicy != nil {
+			policy = bd.Spec.StagedOptions.ResyncPolicy
+		}
+
+		if policy.MaxRetries == 0 {
+			return false
+		}
+
+		if policy.MaxRetries == DefaultMaxRetries && status.LastApply.Add(policy.ResyncDelay.Duration).Before(time.Now()) {
```
StrongMonkey, line 301: Hmm why do we even compare `policy.MaxRetries == DefaultMaxRetries` here?

modules/agent/pkg/controllers/bundledeployment/controller.go
```diff
@@ -213,8 +232,12 @@ func (h *handler) MonitorBundle(bd *fleet.BundleDeployment, status fleet.BundleD
...
 	condition.Cond(fleet.BundleDeploymentConditionReady).SetError(&status, "", readyError)
 	if len(status.ModifiedStatus) > 0 {
-		h.bdController.EnqueueAfter(bd.Namespace, bd.Name, 5*time.Minute)
-		if shouldRedeploy(bd) {
+		duration := customResyncDuration(bd)
+		h.bdController.EnqueueAfter(bd.Namespace, bd.Name, duration)
+		ok := shouldRedeploy(bd, &status)
+		if ok {
+			// manage resync counter
+			status.ResyncCounter = manageResyncCounter(bd)
```
StrongMonkey, line 240: Can this method be done in `DeployBundle` method? Basically we only set `ResyncCounter` after deploying it to possibly avoid the situation where this is set twice.

modules/agent/pkg/controllers/bundledeployment/controller.go
ibrokethecloud: DefaultMaxRetries was a specific case where you specify MaxRetries as `-1` which means its an infinite loop where the objects will be queued at the custom reapply interval.

modules/agent/pkg/controllers/bundledeployment/controller.go
ibrokethecloud: Does make sense. I have moved it into `DeployBundle`

Jared314:
This feature would be brilliant and is our only blocker for adopting Fleet. Is this going to be merged?

manno:
We are planning to correct drift instead of reapplying on an interval. We might revisit this in the future.

PR #561, opened 2021-10-18
Fix comparePatches running only last operation

### LINKED ISSUE #371: comparePatches running only last operation
I'm having an issue with `diff.comparePatches`. it seems only the last operation is being applied:

```yaml
diff:
  comparePatches:
  - apiVersion: v1
    kind: Secret
    name: wildcard-production-tls
    namespace: proxy
    operations:
    - {"op":"remove", "path":"/data"}
    - {"op":"remove", "path":"/metadata/annotations"}
```

This gives the repo as Modified with the following patch:
```{"data":{"tls.crt":"","tls.key":""}}```
If I invert the order on the operations, this is the patch:
```{"metadata":{"annotations":{"reflector.v1.k8s.emberstack.com/reflected-at":"","reflector.v1.k8s.emberstack.com/reflected-version":""}}}```

**Versions**
Rancher 2.5.8 - Fleet 0.3.5

I'm not very familiar with Go, but it seems this line overrides the last operation for a given object:
https://github.com/rancher/fleet/blob/5a578473735455c3c7c7e8aceac1b931a3a60dce/modules/agent/pkg/deployer/normalizers/jsonpatch.go#L23

YohanValette-DXC:
closes #371 

This is a proposal and should be tested.
The goal is to allow JSON patch for each of the operations such as describe in https://fleet.rancher.io/bundle-diffs/

This PR adds the index element in the key object so that not only the last operation of the list is run, but also all of them.

### DIFF
--- a/modules/agent/pkg/deployer/monitor.go
+++ b/modules/agent/pkg/deployer/monitor.go
@@ -126,7 +126,7 @@ func (m *Manager) normalizers(live objectset.ObjectByGVK, bd *fleet.BundleDeploy
 				key := objectset.ObjectKey{
 					Name:      patch.Name,
 					Namespace: patch.Namespace,
-					Index: i
+					Index: i,
 				}
 				jsonPatchNorm.Add(gvk, key, patchData)
 			}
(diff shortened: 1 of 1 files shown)


aiyengar2:
@YohanValette-DXC thank you for your contribution! 

However, I'm closing this proposal out in favor of https://github.com/rancher/fleet/pull/717 which will add support for tracking multiple JSON patches on the same resource without modifying the object key used; instead, each object key will support multiple JSON patches.

Issue #432, opened 2021-06-23
Failed to create BundleDeployment with long path within repo (63 characters limit)


petertang2012:
I am seeing this error message from bundles page when I use a long `Paths` for my Git Repo cofig:

> failed to create `a-long-bundle-name-longer-than-63-characters` fleet.cattle.io/v1alpha1, Kind=BundleDeployment for bundle `a-long-bundle-name-longer-than-63-characters` : BundleDeployment.fleet.cattle.io `a-long-bundle-name-longer-than-63-characters` is invalid: metadata.labels: Invalid value: `a-long-bundle-name-longer-than-63-characters` : must be no more than 63 characters

It seems the `Paths` value is extracted as a whole along with `Git Repo name` to construct `metadata.labels` value for BundleDeployment's manifest but with the length being limited at 63 characters I am not able to use a long `Path`.

Workaround is just to use a `shallow` folder for storing configurations but it would be great if I have the freedom to use any arbitrary path for Fleet deployment.

The issue was observed with Fleet `0.3.500` and Rancher `2.5.8`

keifgwinn:
I ran into this as well on starting to use fleet with my existing repo structure.

sgapanovich:
## Steps to reproduce ##
### Test Environment: ###
Rancher version: v2.6.1
Rancher cluster type: HA
Docker version: 20.10
Downstream cluster type: ec2 3 node cluster
Fleet version: v0.3.7

----------

### Test Steps: ###

Log in as an admin
1. Navigate to Main Menu -> Continuous Delivery
2. In Namespace fleet drop down select "fleet-default"
3. Go to "Git Repos"
4. Click "Create"
  * name: `itstest`
  * Repository URL: `https://github.com/sgapanovich/fleet-examples`
  * Branch Name: `test`
    * click "Add Path"
  * Path: `/this-is-text-to-test-sixty-three-characters-in-path/helm`
  * Click "Create"
5. In the left menu click "Advanced"
6. Click "Bundles"
  * you might need to refresh the page to see a new bundle "created"

Issue:
[...]

PR #567, opened 2021-10-19
Limit bundle name to 63 characters

StrongMonkey:
https://github.com/rancher/fleet/issues/432

### DIFF
--- a/modules/cli/apply/apply.go
+++ b/modules/cli/apply/apply.go
@@ -14,6 +14,7 @@ import (
 	"github.com/rancher/fleet/modules/cli/pkg/client"
 	fleet "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
 	"github.com/rancher/fleet/pkg/bundle"
+	name2 "github.com/rancher/wrangler/pkg/name"
 	"github.com/rancher/wrangler/pkg/yaml"
 	"github.com/sirupsen/logrus"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


modules/cli/apply/apply.go
```diff
@@ -134,7 +135,7 @@ func readBundle(ctx context.Context, name, baseDir string, opts *Options) (*bund
 func createName(name, baseDir string) string {
 	path := strings.ToLower(filepath.Join(name, baseDir))
 	path = disallowedChars.ReplaceAllString(path, "-")
-	return multiDash.ReplaceAllString(path, "-")
+	return name2.Limit(multiDash.ReplaceAllString(path, "-"), 63)
```
nickgerace, line 138: Will this just cut off the rest? What is there's two files in that path that share the same 63 char prefix? We might have a collision.

nickgerace: Bumping this @StrongMonkey if you have a chance

StrongMonkey: it will cut off also with a hash. So in your case it will have two different hashes in the end. (foobarbrah1, foobarbrah2)
foobar-hash1
foobar-hash2

StrongMonkey: For your reference 
```
func Limit(s string, count int) string {
	if len(s) < count {
		return s
	}
	return fmt.Sprintf("%s-%s", s[:count-6], Hex(s, 5))
}
```

nickgerace: Ahhh I missed that. Gotcha.

nickgerace (approved):

PR #587, opened 2021-11-01
Add FOSSA to Drone Pipeline

oxr463:
This adds a fossa analyze and test step to the drone pipeline.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -14,6 +14,26 @@ steps:
   volumes:
   - name: docker
     path: /var/run/docker.sock
+- name: fossa
+  image: registry.suse.com/suse/sle15:15.3
+  failure: ignore
+  environment:
+    FOSSA_API_KEY:
+      from_secret: FOSSA_API_KEY
+  commands:
+    - zypper -n install curl unzip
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -14,6 +14,26 @@ steps:
   volumes:
   - name: docker
     path: /var/run/docker.sock
+- name: fossa
+  image: registry.suse.com/suse/sle15:15.3.17.8.1
```
nickgerace, line 18: This will need to be `registry.suse.com/suse/sle15:15.3` as discussed offline.

nickgerace (changes requested):

nickgerace (approved):

nickgerace:
@StrongMonkey I think we are good here with the image tag fixed. Thoughts?

StrongMonkey:
Looks good to me. Have we run this locally to see if it will succeed? I don't expect this to succeed though.

oxr463:
> Looks good to me. Have we run this locally to see if it will succeed? I don't expect this to succeed though.

I've tested it locally without any issues. Please let me know if you need any assistance.

StrongMonkey:
@oxr463 Thanks, I am worried about breaking the pipeline without knowing how to fix it. Given it has been tested, LGTM!

StrongMonkey (approved):

PR #592, opened 2021-11-04
Allow for fleet.yml

nickgerace:
Primary issue (contains child issues): https://github.com/rancher/fleet/issues/593

### DIFF
--- a/pkg/bundle/read.go
+++ b/pkg/bundle/read.go
@@ -13,6 +13,7 @@ import (
 	"strconv"
 
 	fleet "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
+	"github.com/rancher/fleet/pkg/bundleyaml"
 	name1 "github.com/rancher/wrangler/pkg/name"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/yaml"
@@ -43,14 +44,14 @@ func Open(ctx context.Context, name, baseDir, file string, opts *Options) (*Bund
 	)
 
 	if file == "" {
-		file = filepath.Join(baseDir, "fleet.yaml")
-		if f, err := os.Open(file); os.IsNotExist(err) {
-			in = bytes.NewBufferString("{}")
-		} else if err != nil {
+		if file, err := setupIOReader(baseDir); err != nil {
 			return nil, err
+		} else if file != nil {
+			in = file
+			defer file.Close()
 		} else {
-			in = f
-			defer f.Close()
+			// Create a new buffer if opening both files resulted in "IsNotExist" errors.
+			in = bytes.NewBufferString("{}")
 		}
 	} else {
 		f, err := os.Open(filepath.Join(baseDir, file))
@@ -62,6 +63,27 @@ func Open(ctx context.Context, name, baseDir, file string, opts *Options) (*Bund
 	}
 
 	return Read(ctx, name, baseDir, in, opts)
+}
+
+// Try accessing the documented, primary fleet.yaml extension first. If that returns an "IsNotExist" error, then we
+// try the fallback extension. If we receive "IsNotExist" errors for both file extensions, then we return a "nil" file
+// and a "nil" error. If either return a non-"IsNotExist" error, then we return the error immediately.
+func setupIOReader(baseDir string) (*os.File, error) {
+	if file, err := os.Open(bundleyaml.GetFleetYamlPath(baseDir, false)); err != nil && !os.IsNotExist(err) {
+		return nil, err
+	} else if err == nil {
+		// File must be closed in the parent function.
+		return file, nil
+	}
+
+	if file, err := os.Open(bundleyaml.GetFleetYamlPath(baseDir, true)); err != nil && !os.IsNotExist(err) {
+		return nil, err
+	} else if err == nil {
+		// File must be closed in the parent function.
+		return file, nil
+	}
+
+	return nil, nil
 }
 
 func Read(ctx context.Context, name, baseDir string, bundleSpecReader io.Reader, opts *Options) (*Bundle, error) {
--- a/pkg/bundleyaml/bundleyaml.go
+++ b/pkg/bundleyaml/bundleyaml.go
@@ -1,0 +1,39 @@
+package bundleyaml
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+const (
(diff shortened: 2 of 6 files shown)


pkg/bundleyaml/bundleyaml.go
```diff
@@ -0,0 +1,39 @@
+package bundleyaml
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+var (
```
genexpr, line 9: The smallest of nitpicks, but these two could be constants.

pkg/apis/fleet.cattle.io/v1alpha1/zz_generated_deepcopy.go
```diff
@@ -1,3 +1,4 @@
+//go:build !ignore_autogenerated
```
StrongMonkey, line 1: What's this?

pkg/bundle/read.go
```diff
@@ -43,14 +44,20 @@ func Open(ctx context.Context, name, baseDir, file string, opts *Options) (*Bund
 	)
 
 	if file == "" {
-		file = filepath.Join(baseDir, "fleet.yaml")
-		if f, err := os.Open(file); os.IsNotExist(err) {
-			in = bytes.NewBufferString("{}")
+		if openedFile, err := os.Open(bundleyaml.GetFleetYamlPath(baseDir, false)); os.IsNotExist(err) {
```
StrongMonkey, line 47: Nit: For readbility might be good to wrap in a function and cal it in a loop, like
```
for _, v := range []string{"fleet.yaml". "fleet.yml" {
  in, err := getReader(v)
  if err != nil {
   blabla
  }
  break
}
```

pkg/apis/fleet.cattle.io/v1alpha1/zz_generated_deepcopy.go
nickgerace: go 1.17

nickgerace: I'll remove it since Fleet uses 1.16

pkg/bundle/read.go
nickgerace: Good point. Will fix

pkg/bundleyaml/bundleyaml_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+		// Test both the primary extension and the fallback extension.
+		for _, fullPath := range []string{GetFleetYamlPath(path, false), GetFleetYamlPath(path, true)} {
+			a.True(IsFleetYaml(filepath.Base(fullPath)))
+			a.True(IsFleetYamlSuffix(fullPath))
+		}
+	}
+
+	// Test expected failure payloads.
+	for _, fullPath := range []string{"fleet.yaaaaaaaaaml", "", ".", "weakmonkey.yaml", "../fleet.yaaaaml"} {
```
StrongMonkey, line 25: :monkey:

pkg/bundle/read.go
nickgerace: Alright, I won't use a loop because I want to keep the file close defer statement there, but I have a new refactoring idea I'll push in a moment. Will resolve this once pushed

pkg/apis/fleet.cattle.io/v1alpha1/zz_generated_deepcopy.go
StrongMonkey: ah.. Might be good to bump to 1.17

pkg/bundle/read.go
```diff
@@ -64,6 +64,27 @@ func Open(ctx context.Context, name, baseDir, file string, opts *Options) (*Bund
...
 }
 
+// Try accessing the documented, primary fleet.yaml extension first. If that returns a "IsNotExist" error, then we
+// try the fallback extension. If we receive "IsNotExist" errors for both file extensions, then we return a "nil" file
+// and a "nil" error. If either return a non-"IsNotExist" error, then we return the error immediately.
+func setupIOReader(baseDir string) (*os.File, error) {
+	if file, err := os.Open(bundleyaml.GetFleetYamlPath(baseDir, false)); err != nil && !os.IsNotExist(err) {
+		return nil, err
+	} else if err == nil {
+		defer file.Close()
```
StrongMonkey, line 74: You will close file once this method is returned, so the reader can't be read anymore after it has been returned

nickgerace: Good catch

pkg/bundleyaml/bundleyaml.go
```diff
@@ -0,0 +1,39 @@
...
+
+func IsFleetYaml(fileName string) bool {
+	if fileName == fleetYaml || fileName == fallbackFleetYaml {
+		return true
+	}
+	return false
+}
+
+func IsFleetYamlSuffix(filePath string) bool {
+	return strings.HasSuffix(filePath, "/"+fleetYaml) || strings.HasSuffix(filePath, "/"+fallbackFleetYaml)
```
genexpr, line 38: I barely know anything about Fleet, but is it ever run on Windows? If yes, should this path handling be platform-agnostic?

pkg/bundleyaml/bundleyaml.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+func GetFleetYamlPath(baseDir string, useFallbackFileExtension bool) string {
+	if useFallbackFileExtension {
+		return filepath.Join(baseDir, fallbackFleetYaml)
+	}
+	return filepath.Join(baseDir, fleetYaml)
+}
+
+func IsFleetYaml(fileName string) bool {
```
genexpr, line 30: Is this function used outside of tests?

pkg/bundleyaml/bundleyaml.go
genexpr: Looks like `path/filepath` has this:
```func HasPrefix(p, prefix string) bool```

nickgerace: You're absolutely right that we should likely not be wrangling file paths as strings. Path separators should be handled by a library. However, I'd prefer for this PR to retain existing logic since there are _multiple_ places where this would need to be change to keep logic consistent. Thus, I've opened this issue: https://github.com/rancher/fleet/issues/599

Screen Shot 2021-11-08 at 1 46 53 PM
s

genexpr (approved):

StrongMonkey (approved):

PR #647, opened 2021-11-23
Added new advanced users section, updated mkdocs.yml

### LINKED ISSUE #646: Create new section for Advanced Users
- Moving single and multi-cluster install pages underneath new header of Advanced Users. These sections will eventually be deprecated, but some users are still using them, so we will move them away from the regular Fleet user docs.

jtravee:
Closes #646 

- New Advanced Users section added; will continue to update this as Fleet standalone issues arise that need support.

### DIFF
--- a/docs/advanced-users.md
+++ b/docs/advanced-users.md
@@ -1,6 +1,6 @@
 # Advanced Users
 
-Note that using Fleet outside of Rancher is highly discouraged for any users who do not need to perform advanced actions. However, there are some advanced use cases that may need to be performed outside of Rancher, and this section will highlight these cases. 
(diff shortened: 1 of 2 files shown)


docs/advanced-users.md
```diff
@@ -0,0 +1,10 @@
+# Advanced Users
+
+Note that using Fleet outside of Rancher is highly discouraged for any users who do not need to perform advanced actions. However, there are some advanced use cases that may need to be performed outside of Rancher, and this section will highlight these cases.
```
nickgerace, line 3: Perhaps, we should add a note saying something along the lines of the following: "many advanced use cases are covered under what's known as Standalone Fleet, or Fleet without Rancher"

StrongMonkey: That's a good point, I will leave @jtravee to address

StrongMonkey (dismissed):
LGTM

docs/advanced-users.md
jtravee: Reworded, thanks!

StrongMonkey (approved):

PR #648, opened 2021-11-24
Added bundle lifecycle in examples, added link from concepts page

### LINKED ISSUE #627: Showcase the lifecycle of a bundle (via GitRepo CR) from creation to deployment (and updates upon new commits)
We should answer:
- how does a connection happen from token creation to downstream agent bundle deployment?
- what happens when a new commit is pushed? how does Fleet see that change and how does that affect the bundle?
- what are the pods that the bundle information is passed through to the eventual target?

This can likely be easier to manage if we base this off of the [multi-cluster/helm](https://github.com/rancher/fleet-examples/tree/master/multi-cluster/helm) example and use it as the example payload.

jtravee:
Closes #627

References: 
https://fleet.rancher.io/examples/ - added lifecycle under examples page
https://fleet.rancher.io/concepts/ - added link to new section

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -1,4 +1,16 @@
 # Examples
 
-Examples using raw Kubernetes YAML, Helm charts, Kustomize and combinations
+### Lifecycle of a Fleet Bundle
+
+To demonstrate the lifecycle of a Fleet bundle, we will use multi-cluster/helm as a case study.
+
+1. User creates a [GitRepo](./gitrepo-add.md#create-gitrepo-instance) that points to the [multi-cluster/helm](https://github.com/rancher/fleet-examples/tree/master/multi-cluster/helm) repository.
+2. The Gitjob controller will sync changes from the GitRepo and detect changes from the poll/webhook. With every commit change, the Gitjob controller will create a job that clones the git repository, reads content from the repo such as fleet.yaml and other manifests, and creates the Fleet [bundle](./cluster-bundles-state.md#bundles).
+3. The Fleet-controller then syncs changes from the bundle. According to the targets, the Fleet-controller will create `BundleDeployment` resources, which are a combination of a bundle and a target cluster.
+4. The Fleet-agent will then pull the `BundleDeployment` from the Fleet controlplane. The agent deploys a real application and configuration as a [Helm chart](https://helm.sh/docs/intro/install/) from the `BundleDeployment` into the downstream clusters.
+5. The Fleet-agent will continue to monitor the application bundle and report statuses back in the following order: bundledeployment > bundle > GitRepo > cluster.
+
+### Additional Examples
+
+Examples using raw Kubernetes YAML, Helm charts, Kustomize, and combinations
 of the three are in the [Fleet Examples repo](https://github.com/rancher/fleet-examples/).
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -10,7 +10,7 @@ To demonstrate the lifecycle of a Fleet bundle, we will use [multi-cluster/helm]
 >**Note:** The job pod with the image name `rancher/tekton-utils` will be under the same namespace as the GitRepo.
 
 3. The `fleet-controller` then syncs changes from the bundle. According to the targets, the `fleet-controller` will create `BundleDeployment` resources, which are a combination of a bundle and a target cluster.
-4. The `fleet-agent` will then pull the `BundleDeployment` from the Fleet controlplane. The agent deploys a real application and configuration as a [Helm chart](https://helm.sh/docs/intro/install/) from the `BundleDeployment` into the downstream clusters.
+4. The `fleet-agent` will then pull the `BundleDeployment` from the Fleet controlplane. The agent deploys bundle manifests as a [Helm chart](https://helm.sh/docs/intro/install/) from the `BundleDeployment` into the downstream clusters.
 5. The `fleet-agent` will continue to monitor the application bundle and report statuses back in the following order: bundledeployment > bundle > GitRepo > cluster.
 
 ### Additional Examples
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -5,7 +5,7 @@
 To demonstrate the lifecycle of a Fleet bundle, we will use [multi-cluster/helm](https://github.com/rancher/fleet-examples/tree/master/multi-cluster/helm) as a case study.
 
 1. User will create a [GitRepo](./gitrepo-add.md#create-gitrepo-instance) that points to the multi-cluster/helm repository.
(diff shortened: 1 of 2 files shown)


docs/examples.md
```diff
@@ -1,4 +1,16 @@
...
+
+1. User creates a [GitRepo](./gitrepo-add.md#create-gitrepo-instance) that points to the [multi-cluster/helm](https://github.com/rancher/fleet-examples/tree/master/multi-cluster/helm) repository.
+2. The Gitjob controller will sync changes from the GitRepo and detect changes from the poll/webhook. With every commit change, the Gitjob controller will create a job that clones the git repository, reads content from the repo such as fleet.yaml and other manifests, and creates the Fleet [bundle](./cluster-bundles-state.md#bundles).
+3. The Fleet-controller then syncs changes from the bundle. According to the targets, the Fleet-controller will create `BundleDeployment` resources, which are a combination of a bundle and a target cluster.
+4. The Fleet-agent will then pull the `BundleDeployment` from the Fleet controlplane. The agent deploys a real application and configuration as a [Helm chart](https://helm.sh/docs/intro/install/) from the `BundleDeployment` into the downstream clusters.
+5. The Fleet-agent will continue to monitor the application bundle and report statuses back in the following order: bundledeployment > bundle > GitRepo > cluster.
```
nickgerace, line 11: Question to both @jtravee @StrongMonkey: I think we should standardize on one of the following for Fleet components:

- Fleet-controller
- `fleet-controller`
- fleet-controller
- fleet controller
- Fleet controller
- `fleet` Controller
- `fleet` controller

Personally speaking, I like calling things as a user would see them in a cluster. In that case, that would be `fleet-controller`. However, I could also see a reason to call it the Fleet controller or the controller in Fleet based on the context. Thoughts?

docs/examples.md
```diff
@@ -1,4 +1,16 @@
 # Examples
 
-Examples using raw Kubernetes YAML, Helm charts, Kustomize and combinations
+### Lifecycle of a Fleet Bundle
+
+To demonstrate the lifecycle of a Fleet bundle, we will use multi-cluster/helm as a case study.
+
+1. User creates a [GitRepo](./gitrepo-add.md#create-gitrepo-instance) that points to the [multi-cluster/helm](https://github.com/rancher/fleet-examples/tree/master/multi-cluster/helm) repository.
+2. The Gitjob controller will sync changes from the GitRepo and detect changes from the poll/webhook. With every commit change, the Gitjob controller will create a job that clones the git repository, reads content from the repo such as fleet.yaml and other manifests, and creates the Fleet [bundle](./cluster-bundles-state.md#bundles).
```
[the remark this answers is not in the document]
StrongMonkey: Yes, but the name is generated. Maybe we can call the job with image name so it is more specific.

docs/examples.md
StrongMonkey: :exploding_head:

docs/examples.md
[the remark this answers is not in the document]
StrongMonkey: I am thinking of this... @jtravee Maybe we can add something like this in the end
> With every commit change, the Gitjob controller will create a job that clones the git repository, reads content from the repo such as fleet.yaml and other manifests, and creates the Fleet [bundle](./cluster-bundles-state.md#bundles). You can found the job pod under the same namespace of gitRepo with image name `rancher/tekton-utils`.

docs/examples.md
StrongMonkey: I will leave that to @jtravee since she is the expert here.

jtravee: I agree with `fleet-controller` as the standard!

docs/examples.md
```diff
@@ -1,4 +1,16 @@
 # Examples
 
-Examples using raw Kubernetes YAML, Helm charts, Kustomize and combinations
+### Lifecycle of a Fleet Bundle
+
+To demonstrate the lifecycle of a Fleet bundle, we will use multi-cluster/helm as a case study.
+
+1. User creates a [GitRepo](./gitrepo-add.md#create-gitrepo-instance) that points to the [multi-cluster/helm](https://github.com/rancher/fleet-examples/tree/master/multi-cluster/helm) repository.
```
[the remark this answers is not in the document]
jtravee: Fixed per feedback.

docs/examples.md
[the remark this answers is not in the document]
jtravee: Added.

docs/examples.md
```diff
@@ -1,4 +1,19 @@
...
+
+1. User will create a [GitRepo](./gitrepo-add.md#create-gitrepo-instance) that points to the multi-cluster/helm repository.
+2. The `gitjob-controller` will sync changes from the GitRepo and detect changes from the poll/webhook. With every commit change, the `gitjob-controller` will create a job that clones the git repository, reads content from the repo such as `fleet.yaml` and other manifests, and creates the Fleet [bundle](./cluster-bundles-state.md#bundles).
+
+>**Note:** The job pod with the image name `rancher/tekton-utils` will be under the same namespace as the GitRepo.
+
+3. The `fleet-controller` then syncs changes from the bundle. According to the targets, the `fleet-controller` will create `BundleDeployment` resources, which are a combination of a bundle and a target cluster.
+4. The `fleet-agent` will then pull the `BundleDeployment` from the Fleet controlplane. The agent deploys a real application and configuration as a [Helm chart](https://helm.sh/docs/intro/install/) from the `BundleDeployment` into the downstream clusters.
```
StrongMonkey, line 13: > The agent deploys a real application and configuration
Now I feel like I feel bad about words...

> The agent deploys bundle manifests as a helm chart

jtravee: Sounds good, revised!

docs/examples.md
```diff
@@ -1,4 +1,19 @@
 # Examples
 
-Examples using raw Kubernetes YAML, Helm charts, Kustomize and combinations
+### Lifecycle of a Fleet Bundle
+
+To demonstrate the lifecycle of a Fleet bundle, we will use [multi-cluster/helm](https://github.com/rancher/fleet-examples/tree/master/multi-cluster/helm) as a case study.
+
+1. User will create a [GitRepo](./gitrepo-add.md#create-gitrepo-instance) that points to the multi-cluster/helm repository.
+2. The `gitjob-controller` will sync changes from the GitRepo and detect changes from the poll/webhook. With every commit change, the `gitjob-controller` will create a job that clones the git repository, reads content from the repo such as `fleet.yaml` and other manifests, and creates the Fleet [bundle](./cluster-bundles-state.md#bundles).
```
StrongMonkey, line 8: > from the poll/webhook

by polling or webhook event.

@jtravee I think this is better, what do you think? And also can we add this link https://fleet.rancher.io/webhook/ to webhook event?

jtravee: Revised per feedback.

StrongMonkey (approved):

PR #649, opened 2021-11-25
Create label-all-opened-issues.yml to add team labels to any issue that is opened or reopened

deniseschannon:
Created based on https://docs.github.com/en/actions/managing-issues-and-pull-requests/adding-labels-to-issues

Tested already in https://github.com/rancher/cis-operator/pull/109

### DIFF
--- a/.github/workflows/label-all-opened-issues.yml
+++ b/.github/workflows/label-all-opened-issues.yml
@@ -16,7 +16,7 @@ jobs:
       issues: write
     steps:
       - name: Label issues
-        uses: andymckay/labeler@5c59dabdfd4dd5bd9c6e6d255b01b9d764af4414
+        uses: andymckay/labeler@e6c4322d0397f3240f0e7e30a33b5c5df2d39e90
         with:
           add-labels: "team/area3"
           repo-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/label-all-new-issues.yml
```diff
@@ -0,0 +1,21 @@
...
+    types:
+      - opened
+jobs:
+  label_issues:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - name: Label issues
+        uses: andymckay/labeler@5c59dabdfd4dd5bd9c6e6d255b01b9d764af4414
```
nickgerace, line 18: Is there a reason why we want this commit over the latest tag? https://github.com/andymckay/labeler/releases/latest

nickgerace (commented):
One comment, but generally LGTM. Go do thanksgiving stuff @deniseschannon 🦃

.github/workflows/label-all-new-issues.yml
deniseschannon: I picked out what was in the recommended doc of how to add labels, though I realize now that we should do reopened as well.

nickgerace:
Is this ready for re-review @deniseschannon?

nickgerace (approved):

.github/workflows/label-all-new-issues.yml
StrongMonkey: It is best practice for security since you know which version of the code the action runs. If using latest and the action repo is  hacked, we could have issue losing our token/access to fleet repo. See https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

Issue #650, opened 2021-11-26
ImageScan feature does not update manifests in git if `paths` field is an empty list


belgaied2:
By following step by step the fleet [documentation about the ImageScan feature](https://fleet.rancher.io/imagescan/) in Fleet, the manifest will never be updated but will not throw any error.

Indeed, the following GitRepo definition:
```yaml
kind: GitRepo
apiVersion: fleet.cattle.io/v1alpha1
metadata:
  name: my-repo
  namespace: fleet-local
spec:
  # change this to be your own repo
  repo: https://github.com/rancher/fleet-examples 
  # define how long it will sync all the images and decide to apply change
  imageScanInterval: 5m 
  # user must properly provide a secret that have write access to git repository
  clientSecretName: secret 
  # specify the commit pattern
  imageScanCommit:
    authorName: foo
    authorEmail: foo@bar.com
    messageTemplate: "update image"
```
does not contain the `paths` field and the ImageScan code [here ](https://github.com/rancher/fleet/blob/c73432776a595f7c5f5b87a7f5653e778e477fa3/pkg/controllers/image/image.go#L240)iterates on the paths to update and commit the changes in manifests.

It is absolutely imaginable that a user would want to deploy manifests in the root of the GIT repo and would not fill the `paths` field.

kkaempf:
stale, closing

PR #651, opened 2021-11-26
Adding the case where paths field is empty when updating manifests

belgaied2:
This PR aims at integrating an overseen case in the ImageScan feature, when the `paths` field is empty. 

The PR adds the default value for the `paths` array, which is `[1]string{"/"}`  in an intermediate `paths` variable that is then iterated on.

This is linked to the issue #650

### DIFF
--- a/pkg/controllers/image/image.go
+++ b/pkg/controllers/image/image.go
@@ -237,7 +237,14 @@ func (h handler) onChangeGitRepo(gitrepo *v1alpha1.GitRepo, status v1alpha1.GitR
 		return status, err
 	}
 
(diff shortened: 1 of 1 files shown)


StrongMonkey (approved):
LGTM. Thanks @belgaied2 

cc @aiyengar2 @prachidamle We can merge this after 0.3.8 since it is already past code freeze of v2.6.3

belgaied2:
@MKlimuszka This Pull request solves a bug in Fleet when using the imageScan feature when the paths field is empty. it is blocked since now 6 months!

prachidamle (approved):
LGTM, thanks for catching this!

Issue #658, opened 2021-11-30
[Refactor] kube client smoke tests should be consolidated


nickgerace:
In an internal knowledge transfer for Fleet development, it was evident that the kube client connection smoke tests were difficult to understand without an explicitly named function or comment. We should consolidate them to one function.

nickgerace:
Closing since this does not require QA validation and https://github.com/rancher/fleet/pull/659 has been merged.

PR #659, opened 2021-11-30
Consolidate kube client tests and remove unused client test

nickgerace:
> "In an internal knowledge transfer for Fleet development, it was evident that the kube client connection smoke tests were difficult to understand without an explicitly named function or comment. We should consolidate them to one function."

Relevant issue: https://github.com/rancher/fleet/issues/658

### DIFF
--- a/modules/cli/agentmanifest/agent.go
+++ b/modules/cli/agentmanifest/agent.go
@@ -25,7 +25,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/watch"
-	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 )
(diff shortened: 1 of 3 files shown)


modules/cli/agentmanifest/agent.go
```diff
@@ -78,24 +77,6 @@ func insecurePing(host string) {
...
-	client, err := kubernetes.NewForConfig(restConfig)
-	if err != nil {
-		return fmt.Errorf("failed to test build client from kubeconfig: %w", err)
-	}
-	_, err = client.Discovery().ServerVersion()
-	if err != nil {
-		return fmt.Errorf("failed to test connection to %s: %w", host, err)
-	}
-	return nil
-}
```
nickgerace, line 97: **To reviewers:** in the search for kube client connection tests, this function was found and is unused by anything else.

pkg/connection/connection.go
```diff
@@ -0,0 +1,10 @@
+package connection
+
+import (
+	"k8s.io/client-go/kubernetes"
+)
+
+func TestKubeClientConnection(client *kubernetes.Clientset) error {
```
StrongMonkey, line 7: Nit: in general I think golang discourage function with test prefix because that's their conversion for test method.

StrongMonkey (approved):

pkg/connection/connection.go
nickgerace: You're right. Oversight

StrongMonkey (approved):

PR #665, opened 2021-12-02
Added more troubleshooting tips for Bundle, GitRepo error states

### LINKED ISSUE #629: Add all known Bundle, GitRepo, and BundleDeployment states to docs with descriptions
We should also include troubleshooting for associated states (e.g. "what should I do if I see this state? Wait and go make coffee? Re-deploy with a field changed?")

jtravee:
Closes #629

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -78,7 +78,7 @@ For users who want to deploy to the local cluster as well, they may move the clu
 
 **Result**: The cluster will be migrated to `fleet-default`.
 
-## Additional Solutions for Other Fleet Errors
+## **Additional Solutions for Other Fleet Errors**
 
 ### Fleet fails with bad response code: 403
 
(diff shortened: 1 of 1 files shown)


docs/troubleshooting.md
```diff
@@ -153,3 +152,15 @@ Based on the above log, you can add the following entry to remove the operation:
...
 {"op":"remove", "path":"/spec/template/spec/hostNetwork"}
 ```
+
+### `GitRepo` or `Bundle` stuck in modified state
+
+**Modified** means that there is a mismatch between the actual state and the desired state, the source of truth, which lives in the git repository.
+
+1. Check the [bundle diffs documentation](./bundle-diffs.md) for more information. 
+
+1. You can also force update the `gitrepo` to perform a manual resync. Select **GitRepo** on the left navigation bar, then select **Force Update**.
```
StrongMonkey, line 162: The number seems not correct, should it be 2?

StrongMonkey (approved):

Issue #375, opened 2021-05-13
Global labels inside string


shane-davidson:
I have an issue where I need to use the global labels inside strings.

An example would be the following:
```
defaultNamespace: testing
helm:
  values:
    ingress:
      tls:
        - hosts:
            - {{ global.fleet.clusterLabels.management.cattle.io/cluster-display-name }}.example.com
```

Most public helm charts do not expose something like a subdomain options for the hosts setting. eg.

```
helm:
  values:
    subdomain: global.fleet.clusterLabels.management.cattle.io/cluster-display-name
```
This makes it impossible to effectively use the cluster name unless I create a label manually. my current solution is this:
```
defaultNamespace: testing
helm:
  values:
    ingress:
      enabled: true

targetCustomizations:
- name: test1
  helm:
    values:
      ingress:
        tls:
          - hosts:
              - test1.example.com
  clusterSelector:
    matchLabels:
      management.cattle.io/cluster-display-name: test1
- name: test2
  helm:
    values:
      ingress:
        tls:
          - hosts:
              - test2.example.com
  clusterSelector:
    matchLabels:
      management.cattle.io/cluster-display-name: test2

```
This is a manual task to update for every cluster added.

If fleet took the global labels and then passed them through to golang/sprig templates you would have a lot more flexibility in the fleet file and where you can use the values.

Another example of templates include creating a range on targetCustomizations
```
global:
  clusters:
    - one
    - two
    - three
    - testing
    - prod
    
targetCustomizations:
{{- range $.clusters }}
- name: {{ . }}
  clusterSelector:
    matchLabels:
      management.cattle.io/cluster-display-name: {{ . }}
  kustomize:
    dir: overlays/{{ . }}
{{- end }}  
```

manno:
I believe this is now implemented with https://github.com/rancher/fleet/pull/671

Issue #355, opened 2021-04-29
would it be possible to concatenate label value with global labels now?


virtuman:
Now that rancher/fleet#325 and rancher/fleet#152 are marked as resolved, would this be possible to achieve for non-helm-based resources:

we are adding a series of ExternalSecrets on each new cluster that is spun up. Unfortunately we have not yet found a way to register new cluster in hashicorp vault, that remains a manual step, but we would like for the externalSecrets to be formed properly to reflect the cluster and the role names that are based on cluster name.

so essentualy, we need to concatenate string with value from cluster-display-name label, such as - this is PSEUDO-CODE-EXAMPLE not intended to work:
I cluster name is:
```
global: 
  fleet: 
    clusterLabels.management.cattle.io/cluster-display-name: gke_some-cluster-external
```

and I want to add something like "**-role**" to the cluster name as a value to a helm attribute:
```
apiVersion: 'kubernetes-client.io/v1'
kind: ExternalSecret
metadata:
  name: jaeger-operator-jaeger
spec:
  backendType: vault
  vaultMountPoint: $(global.fleet.clusterLabels.management.cattle.io/cluster-display-name)-external-secrets
  vaultRole: $(global.fleet.clusterLabels.management.cattle.io/cluster-display-name)-role
  kvVersion: 2
  data:
  - name: ES_USERNAME
    key: infra-secrets/data/elastic-cloud
    property: ES_USERNAME
```

to achieve something like this as a final result that gets applied to the cluster (see **vaultMountPoint** and **vaultRole** keys ):
```
apiVersion: 'kubernetes-client.io/v1'
kind: ExternalSecret
metadata:
  name: jaeger-operator-jaeger
spec:
  backendType: vault
  vaultMountPoint: gke_some-cluster-external-secrets
  vaultRole: gke_some-cluster-role
  kvVersion: 2
  data:
  - name: ES_USERNAME
    key: infra-secrets/data/elastic-cloud
    property: ES_USERNAME
```

StrongMonkey:
That's more like a feature to add to allow templating on raw k8s yaml, but it might be easier to convert your externalSecret into a simple helm chart

virtuman:
no go for this feature? i've actually found quite a few apps that follow this exact same pattern with replacing encapsulated values only

virtuman:
[...]

PR #671, opened 2021-12-04
Sprig Templating for Helm Values with Inputs from Cluster Resource

rajiteh:
This PR introduces the ability for fleet.yaml's `helm.values` object to contain go template strings in either the keys or values. The context of the go template also includes the cluster labels (for convenient migration from the existing `global.fleet.clusterLabels.FOO` users.

Cluster specific template variables can be added to `Cluster` CR under a new optional key in `spec.templateContext`

Related issues/PRs
- rancher/fleet#507 but adds the per-cluster custom template context
- rancher/fleet#375 
- rancher/fleet#355

### DIFF
--- a/charts/fleet-crd/templates/crds.yaml
+++ b/charts/fleet-crd/templates/crds.yaml
@@ -117,6 +117,8 @@ spec:
                   chart:
                     nullable: true
                     type: string
+                  disablePreProcess:
+                    type: boolean
                   force:
                     type: boolean
                   maxHistory:
@@ -488,6 +490,8 @@ spec:
                         chart:
                           nullable: true
                           type: string
+                        disablePreProcess:
+                          type: boolean
                         force:
                           type: boolean
                         maxHistory:
@@ -997,6 +1001,8 @@ spec:
                       chart:
                         nullable: true
                         type: string
+                      disablePreProcess:
+                        type: boolean
                       force:
                         type: boolean
                       maxHistory:
@@ -1144,6 +1150,8 @@ spec:
                       chart:
                         nullable: true
                         type: string
+                      disablePreProcess:
+                        type: boolean
                       force:
                         type: boolean
                       maxHistory:
@@ -1784,6 +1792,10 @@ spec:
                 type: string
               redeployAgentGeneration:
                 type: integer
+              templateValues:
+                nullable: true
+                type: object
+                x-kubernetes-preserve-unknown-fields: true
             type: object
           status:
             properties:
--- a/pkg/apis/fleet.cattle.io/v1alpha1/bundle.go
+++ b/pkg/apis/fleet.cattle.io/v1alpha1/bundle.go
@@ -237,6 +237,9 @@ type HelmOptions struct {
 
 	// Atomic sets the --atomic flag when Helm is performing an upgrade
 	Atomic bool `json:"atomic,omitempty"`
+
+	// DisablePreProcess disables template processing in values
+	DisablePreProcess bool `json:"disablePreProcess,omitempty"`
 }
 
 // Define helm values that can come from configmap, secret or external. Credit: https://github.com/fluxcd/helm-operator/blob/0cfea875b5d44bea995abe7324819432070dfbdc/pkg/apis/helm.fluxcd.io/v1/types_helmrelease.go#L439
--- a/pkg/target/target.go
+++ b/pkg/target/target.go
@@ -6,10 +6,12 @@
 package target
 
 import (
+	"bytes"
 	"fmt"
 	"sort"
 	"strconv"
 	"strings"
+	"text/template"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -21,7 +23,6 @@ import (
 	"github.com/rancher/fleet/pkg/options"
 	"github.com/rancher/fleet/pkg/summary"
 
-	"github.com/rancher/wrangler/pkg/data"
 	corecontrollers "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	"github.com/rancher/wrangler/pkg/name"
 	"github.com/rancher/wrangler/pkg/yaml"
@@ -30,6 +31,8 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/sets"
+
+	"github.com/Masterminds/sprig/v3"
 )
 
 var (
@@ -38,6 +41,8 @@ var (
 	defAutoPartitionSize        = intstr.FromString("25%")
 	defMaxUnavailablePartitions = intstr.FromInt(0)
 )
+
+const maxTemplateRecursionDepth = 50
 
 type Manager struct {
 	clusters                    fleetcontrollers.ClusterCache
@@ -263,7 +268,7 @@ func (m *Manager) Targets(bundle *fleet.Bundle, manifest *manifest.Manifest) ([]
 			}
 
 			opts := options.Merge(bundle.Spec.BundleDeploymentOptions, target.BundleDeploymentOptions)
-			err = addClusterLabels(&opts, cluster.Labels)
+			err = preprocessHelmValues(&opts, cluster)
 			if err != nil {
 				return nil, err
 			}
@@ -290,9 +295,11 @@ func (m *Manager) Targets(bundle *fleet.Bundle, manifest *manifest.Manifest) ([]
 	return targets, m.foldInDeployments(bundle, targets)
 }
 
-func addClusterLabels(opts *fleet.BundleDeploymentOptions, labels map[string]string) (err error) {
-	clusterLabels := yaml.CleanAnnotationsForExport(labels)
-	for k, v := range labels {
+func preprocessHelmValues(opts *fleet.BundleDeploymentOptions, cluster *fleet.Cluster) (err error) {
+	clusterLabels := yaml.CleanAnnotationsForExport(cluster.Labels)
+	clusterAnnotations := yaml.CleanAnnotationsForExport(cluster.Annotations)
+
+	for k, v := range cluster.Labels {
 		if strings.HasPrefix(k, "fleet.cattle.io/") || strings.HasPrefix(k, "management.cattle.io/") {
 			clusterLabels[k] = v
 		}
@@ -301,27 +308,15 @@ func addClusterLabels(opts *fleet.BundleDeploymentOptions, labels map[string]str
 		return
 	}
 
-	newValues := map[string]interface{}{
-		"global": map[string]interface{}{
-			"fleet": map[string]interface{}{
-				"clusterLabels": clusterLabels,
-			},
-		},
-	}
-
 	if opts.Helm == nil {
-		opts.Helm = &fleet.HelmOptions{
-			Values: &fleet.GenericMap{
-				Data: newValues,
-			},
-		}
+		opts.Helm = &fleet.HelmOptions{}
 		return nil
 	}
 
 	opts.Helm = opts.Helm.DeepCopy()
 	if opts.Helm.Values == nil || opts.Helm.Values.Data == nil {
 		opts.Helm.Values = &fleet.GenericMap{
-			Data: newValues,
+			Data: map[string]interface{}{},
 		}
 		return nil
 	}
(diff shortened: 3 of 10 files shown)


pkg/target/target.go
```diff
@@ -508,6 +532,42 @@ func Summary(targets []*Target) fleet.BundleSummary {
...
+
+func processTemplateValues(valuesMap map[string]interface{}, templateContext map[string]interface{}) (map[string]interface{}, error) {
+	var tplResult bytes.Buffer
+
+	valuesMapStr, err := k8syaml.Marshal(valuesMap)
+	if err != nil {
+		return nil, err
+	}
+
+	tpl, err := template.New("values").Funcs(tplFuncMap()).Option("missingkey=error").Parse(string(valuesMapStr))
```
gregsidelinger, line 554: So I'm getting an error from the test suite when I started adding some longer lines to the test suite for variable substitution.

Here is the tested bundleYamlWithTemplate I was playing around with.

```
const bundleYamlWithTemplate = `namespace: default
helm:
  releaseName: labels
  values:
    clusterName: "{{ .ClusterLabels.name }}"
    fromAnnotation: "{{ .ClusterAnnotations.testAnnotation }}"
    clusterNamespace: "{{ .ClusterNamespace }}"
    fleetClusterName: "{{ .ClusterName }}"
    workingLongClusterName: kubernets.io/cluster/{{ index .ClusterLabels "p-w" }}
    brokenLongClusterName: kubernets.io/cluster/{{ index .ClusterLabels "not-working-for-some-reason" }}
    customStruct:
      - name: "{{ .Values.topLevel }}"
        key1: value1
        key2: value2
      - element2: "{{ .Values.nested.secondTier.thirdTier }}"
      - "element3_{{ .ClusterLabels.envType }}": "{{ .ClusterLabels.name }}"
    funcs:
      upper: "{{ .Values.topLevel | upper }}_test"
      join: '{{ .Values.list | join "," }}'
diff:
  comparePatches:
  - apiVersion: networking.k8s.io/v1
    kind: Ingress
    name: labels-fleetlabelsdemo
    namespace: default
    operations:
    - op: remove
      path: /spec/rules/0/host
`
```

And here is the error I'm getting from my `brokenLongClusterName: kubernets.io/cluster/{{ index .ClusterLabels "not-working-for-some-reason" }}` line.
```
--- FAIL: TestProcessTemplateValues (0.00s)
    /home/gate/src/fleet/pkg/target/target_test.go:179: error during label processing template: values:1: unclosed action
FAIL
```
I've done a bit of digging and it appears to have something to do with the k8syam.Marshal conversion to a string.  New lines are getting inserted into the line. So `brokenLongClusterName: kubernets.io/cluster/{{ index .ClusterLabels "not-working-for-some-reason" }}` will end up like `brokenLongClusterName: kubernets.io/cluster/{{ index .ClusterLabels "not-working-for-some-reason"\n }}`.  I have not been able to explain why the new lines are getting inserted.
```

Comment (gregsidelinger) [reply pkg/target/target.go:554]: https://github.com/kubernetes-sigs/yaml/issues/68 has details about what is happening with lines over 80 chars.  I can confirm using go-yaml v3  instead of the k8s sigs yaml library resolves the issue.

Comment (rajiteh) [reply pkg/target/target.go:554]: That's an interesting catch. I will update the library and see if that will fix it. However, fundamentally, this issue is raised by the fact that I serialize the entire document and re-parse it as template string. Looking back at it, I can't help it but feel a little bit unhappy about this approach because a carefully crafted template variable could fundamentally change the yaml structure of the templated output. Perhaps a better approach would be to iterate on each individual key values and template them as strings instead. WDYT?

Comment (gregsidelinger) [reply pkg/target/target.go:554]: I really liked the idea of treating all of the values as a big template so you can start to do some fun if/then logic and not try to cram it all on one line.   But this 80 char line break in the yaml libraries breaks that.  My current use case is better variable substitute and not doing crazy logic inside the values so I'm happy with each line getting process separately like I'm doing in #507.

Comment (rajiteh) [reply pkg/target/target.go:554]: Indeed. That's what convinced me to consider it. However, an alternative approach would be to define a dedicated field in fleet.yaml like `templatesValues:` and let the user define it as a multi-line string. This way it is implied that the entire structure is templated as a whole.

Comment (rajiteh) [reply pkg/target/target.go]: Interestingly enough, I couldn't reproduce the new line behaviour on my local system (go 1.17.6) -- but I switched to goyamlv3 just to be safe. I also refactored the templating function to be a lot more "structure" safe, as in only templating within the context of a specific string rather than the entire object.

**Comment by rajiteh** [comment]
@gregsidelinger in case you're interested I am planning to propose an enhancement on top of this templating system to introduce secret management via vals. I have a PR here with a basic implementation, docs pending. IMO this gives an elegant way to template and as well as resolve secrets per release/cluster basis.

 https://github.com/rajiteh/fleet/pull/3/files

File: docs/gitrepo-structure.md [hunk 2]
```diff
@@ -90,14 +100,16 @@ helm:
...
-      namespace: default 
+      namespace: default
       key: values.yaml
     secretKeyRef:
       name: secret-values
       namespace: default
       key: values.yaml
   # Override immutable resources. This could be dangerous.
   force: false
+  # Disable go template pre-prosessing on the fleet values
```
prachidamle, line 111: nit: typo pre-processing

pkg/target/target.go
```diff
@@ -303,7 +307,14 @@ func addClusterLabels(opts *fleet.BundleDeploymentOptions, labels map[string]str
...
 
-	opts.Helm.Values.Data = data.MergeMaps(opts.Helm.Values.Data, newValues)
+	if opts.Helm.DisablePreProcess == false {
+		opts.Helm.Values.Data, err = processTemplateValues(opts.Helm.Values.Data, values)
+		if err != nil {
+			return err
+		}
+		logrus.Debugf("preProcess completed for %v", opts.Helm.ReleaseName)
+	}
+
```
prachidamle, line 317: When opts.Helm.DisablePreProcess = true, what will be set to opts.Helm.Values.Data? Seems like originally we were setting it using `opts.Helm.Values.Data = data.MergeMaps(opts.Helm.Values.Data, newValues)`

pkg/target/target_test.go
```diff
@@ -104,3 +104,194 @@ func TestProcessLabelValues(t *testing.T) {
...
+	join, ok := funcs["join"]
+	if !ok {
+		t.Fatal("key join not found")
+	}
+
+	if join.(string) != "alpha,beta,omega" {
+		t.Fatal("join func was not right")
+	}
+
+}
```
prachidamle, line 297: Can we also add a test to demonstrate the usage of Helm.DisablePreProcess: true usecase?

prachidamle:
@rajiteh I am not clear on how would the Cluster specific template variables be added to a Cluster CR under spec.templateContext? Is the idea that the cluster Admins need to add this data to the target clusters? I am wondering if this templateContext data added to the Fleet target cluster CRs will get overwritten when synced with the management.cattle.io.cluster objects in Rancher.

pkg/target/target.go
gregsidelinger: opts.Helm.Values.Data is left alone in that case. It will either be a copy of the original from the deep copy or an empty map. https://github.com/rancher/fleet/blob/8af675f78e02838aed87fd2ddd16298a5e1960e8/pkg/target/target.go#L298-L304

But as I stare at it now that is dropping the global.fleet.clusterLabels from the helm values and should be added back.

rajiteh:
> @rajiteh I am not clear on how would the Cluster specific template variables be added to a Cluster CR under spec.templateContext? Is the idea that the cluster Admins need to add this data to the target clusters? I am wondering if this templateContext data added to the Fleet target cluster CRs will get overwritten when synced with the management.cattle.io.cluster objects in Rancher.

Yes, the idea is that admins will add these values. I can't speak to Rancher use case unfortunately as I'm only using fleet standalone. 

PS: I will address your feedback over the weekend!

gregsidelinger:
I have not seen rancher touch the templateContext on the fleet cluster in my setup.  Once set rancher left everything alone when it synced labels from the management cluster.

pkg/target/target_test.go
rajiteh: Added.

docs/gitrepo-structure.md
rajiteh: Fixed

rajiteh:
@prachidamle hello! sorry for the delay, please have a look when you get a chance. I believe I address all the requested efeedback.

k0da:
Hello, any updates?

thardeck:
@rajiteh is there something you would miss in the implementation of #1035? It would be nice if you can comment there so we can find an implementation which covers most requirements. Thanks in advance.

rajiteh:
FYI: I have updated the fork with the latest code from master! Love to get the ball rolling on this!! :)

charts/fleet-crd/templates/crds.yaml
```diff
@@ -1784,6 +1792,10 @@ spec:
                 type: string
               redeployAgentGeneration:
                 type: integer
+              templateContext:
```
manno, line 1795: I'd prefer it, if the name included "values" somehow, because it is the context for templating helm's `values:`.
Just not `valuesValues`.

pkg/apis/fleet.cattle.io/v1alpha1/bundle.go
```diff
@@ -237,6 +237,9 @@ type HelmOptions struct {
 
 	// Atomic sets the --atomic flag when Helm is performing an upgrade
 	Atomic bool `json:"atomic,omitempty"`
+
+	// DisablePreprocess deactivates fleet YAML templating
+	DisablePreprocess bool `json:"disablePreprocess,omitempty"`
```
manno, line 242: This seems to make everything more complicated. Why would we want to disable it, for performance, security?

pkg/target/target.go
manno: Seems like this is resolved?

pkg/target/target.go
```diff
@@ -565,6 +580,81 @@ func Summary(targets []*Target) fleet.BundleSummary {
...
+	}
+	compiledYaml, ok := tplResult.(map[string]interface{})
+	if !ok {
+		return nil, fmt.Errorf("templated result was expected to be map[string]interface{}, got %T", tplResult)
+	}
+
+	return compiledYaml, nil
+}
+
+func templateSubstitutions(src interface{}, templateContext map[string]interface{}, tplFn *template.Template) (interface{}, error) {
```
manno, line 608: Without knowing too much about sprig, do we have to walk the values and apply the template to every value?
Could we just treat the values as one big template? E.g. marshap It would allow keys to be templated, but otherwise?

manno (commented):
Adding templating for helm values in fleet.yaml seems to be popular.
There is a "small" solution, which limits templating to just cluster labels, in PR #1035 and this one, which adds a new field to the cluster resource.

We will need to discuss this further before we make a decision.

pkg/target/target.go
rajiteh: sprig is simply providing an additional set of functions inside the templating environment, behind the scenes we're just  using `text/template` to do the actual templating. I had an iteration where I did exactly what you suggested where I would simply serialize the entire `values:` key and pass it through the template engine, however I reverted to explicit templating of keys and values because I worried about the security implications. A carefully crafted template could alter the structure of the object once templating is finished. In this approach, we're ensuring the that all templating is scoped to that particular field. One drawback with this approach right now is that it only let's you template input strings to output strings. No way to output `null`, `integer`, or `float` yaml types.

charts/fleet-crd/templates/crds.yaml
rajiteh: We could call it `templateValues` ?

pkg/target/target.go
manno: Hm, I see. If we let the template instructions from the cluster resource operate on the whole existing helm.Data struct, one could for example overwrite existing keys by duplication and create new keys.

It's just that the code would look much simpler without the recursion. From a threat model point of view, operators can edit cluster resources, but users can create bundles (for those clusters). So I don't see too much of a problem with permissions? Actually, in such a scenario, where users can't change cluster resources,  users would have to ask the operator to include their cluster specific templates.

I'm more concerned with general failure handling, it's easy to get complex templates wrong the first time, but I guess that would just lead to no deployment.

manno:
@rajiteh Let's get this merged. Do you want to rename the context variable and rebase this PR against current master, or should we try to schedule this?

rajiteh:
@manno I made all the changes and fixed the build!

manno:
> @manno I made all the changes and fixed the build!
> 
> PS: Rebasing seems to be a little painful, we could squash the commits instead perhaps?

Yes, that would be great.

pkg/target/target_test.go
```diff
@@ -104,3 +107,375 @@ func TestProcessLabelValues(t *testing.T) {
...
+	} else {
+		if field != expectedValue {
+			t.Fatalf("key %s was not the expected value. Expected: '%s' Actual: '%s'", key, field, expectedValue)
+		}
+	}
+
+}
+
+func TestRecursionDepthForTemplating(t *testing.T) {
+	var bundleYaml = `namespace: default
```
rajiteh, line 455: @manno Added a constraint on the maximum recursion depth

manno (approved):

thardeck:
@rajiteh There is still a small linting issue: https://github.com/rancher/fleet/actions/runs/3576792840/jobs/6029393464
```
  Error: `getClusterAndBundle` - `clusterYaml` always receives `clusterYamlWithTemplateValues` (`"apiVersion: fleet.cattle.io/v1alpha1\nkind: Cluster\nmetadata:\n  na...`) (unparam)
```

manno (approved):

rajiteh:
@manno the relevant documentation for this PR can be found here: https://github.com/rancher/fleet-docs/pull/17

manno:
Documentation is live under https://fleet.rancher.io/next/gitrepo-add

@thardeck created a new alpha release which includes the feature: https://github.com/rancher/fleet/releases/tag/v0.6.0-alpha4

Issue #719, opened 2022-02-24
Bundle name is dynamic from gitrepo config upon edit prevents re-deployment


Jackbennett:
I made a gitRrepo in the UI with a path `example/` for the namespace `example-demo-master`. The `example/fleet.yaml` only has `kustomize: dir: ./deployed/site`. I edited the gitRepo to change the path to `example/deployed/site/apply` which is just k8s yaml, this is kustomize build output. expecting to no longer need the fleet.yaml. This now errors to apply;

```
rendered manifests contain a resource that already exists. Unable to continue with install: Deployment "assets" in namespace "example-demo-master" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "example-example-deployed-example-4c31c": current value is "example-demo-example-demo"
```

helm-release-name not being linked to the gitrepo is weird. Shouldn't that just use the gitrepo CRD name which has to be unique anyway due to k8s? As far as I can tell this name reconstruction isn't documented. It seems like I should at least be able to set the release name in the gitRepo yaml.

Now editing this back to the original path I can't clear that error. I had to delete the newly created bundle with the longer name.

Jackbennett:
Thanks to slack I didn't realize `helm.releaseName` was inside `fleet.yaml` which could help. I'm not clear why this info needs coupling inside the git repo instead of being in the CRD.

mak3r:
In prior versions of fleet, I was able to create a git repo with configuration files (Deployment, Service, Ingress, etc.) but no fleet.yaml required. I could, at a later time, add a fleet.yaml with an overlays directory and patches, etc and fleet would pick up those changes with no issue. Now it seems that this change forces me to include a fleet .yaml with `targetCustomizations.helm.releaseName` defined. This change is not at all clear how it is to be used in practice and if it causes breaking changes for existing process as described.

mak3r:
Following up with more detail.

Basically, if I add a new `targetCustomization` to my fleet.yaml it appears that I need to know something about the inner workings of fleet which generates an annotation and uses that to validate the bundle ownership. 

GitRepo.yaml`
```
apiVersion: fleet.cattle.io/v1alpha1
kind: GitRepo
metadata:
  name: hello
spec:
  branch: main
  paths:
    - /live
  repo: https://github.com/mak3r/fleet-demo-src
  targets:
    - clusterSelector:
        matchExpressions:
          - key: provider.cattle.io
            operator: NotIn
            values:
[...]

PR #720, opened 2022-02-25
Explain Bundle release tracking in docs

Jackbennett:
Following comments in a [slack thread](https://rancher-users.slack.com/archives/C013SSBKB6U/p1645696945259349) I found gauravm comments very help but did not find it documented.

Comments;

>You can define a custom name for your bundle so it doesnt use the generated name…

> if there is no name in the fleet.yaml, then bundle naming will follow the standard naming convention of GitRepo + Path

This was to get to the bottom of #719


ibrokethecloud:
Hi @Jackbennett, thanks a lot for your PR.

The custom bundle name will be applied if the `name` field exists in the fleet.yaml

For example as shown below. 
```yaml
namespace: default
name: demo1
labels:
  app: demo1
targets:
- clusterSelector: {} 
```

The releaseName details will only change the helm release name and not modify the bundle name at all.

Can we please also add a similar explanation as releaseName to the [git-repo-structure](https://github.com/rancher/fleet/blob/master/docs/gitrepo-structure.md#reference)

PR #738, opened 2022-04-08
Move fleet-content and flet-agent-cred rolebinding to use specific serviceAccount

prachidamle:
Previously they were created statically to use all serviceAccounts.


prachidamle:
closing this as this was a trial to test an issue.

PR #758, opened 2022-05-05
Do not git ignore the charts/fleet folder

manno:
This also affects tools like ripgrep, which would not report any findings in the main helm chart.

Note: There is a check for a dirty repo in `scripts/validate-ci`. If some part of the build system creates a fleet binary in a sub-director, then we need to adapt this PR. However all binaries seem to have suffixes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,4 @@
 *.swp
 .idea
 *.DS_Store
-fleet
+/fleet


thardeck (approved):

aiyengar2 (approved):
LGTM, thanks for catching this bug!

prachidamle (approved):
LGTM!

PR #764, opened 2022-05-12
[WIP] bundle remove on git repo bundle delete

prachidamle:


charts/fleet-crd/templates/crds.yaml
```diff
@@ -2091,6 +2091,8 @@ spec:
               pollingInterval:
                 nullable: true
                 type: string
+              pruneBundles:
+                type: boolean
```
aiyengar2, line 2095: as discussed offline, we should move this into an option in the `fleet apply` command and by default we should be pruning. If we want, we can provide this as an option like `--keep-bundles`

modules/cli/apply/apply.go
```diff
@@ -69,12 +70,15 @@ func Apply(ctx context.Context, client *client.Getter, name string, baseDirs []s
...
 
 	foundBundle := false
+
 	for i, baseDir := range baseDirs {
 		matches, err := globDirs(baseDir)
 		if err != nil {
 			return fmt.Errorf("invalid path glob %s: %w", baseDir, err)
 		}
 		for _, baseDir := range matches {
+			logrus.Infof("*** baseDir %v", baseDir)
```
aiyengar2, line 80: need to remove log lines

modules/cli/apply/apply.go
```diff
@@ -90,18 +94,50 @@ func Apply(ctx context.Context, client *client.Getter, name string, baseDirs []s
...
+				for _, bundle := range bundles.Items {
+					if ok := gitRepoBundlesMap[bundle.Name]; !ok {
+						logrus.Infof("Bundle to be deleted %v %v", bundle.Namespace, bundle.Name)
+						err := c.Fleet.Bundle().Delete(bundle.Namespace, bundle.Name, nil)
+						if err != nil {
+							return err
+						}
+					}
+				}
+			}
```
aiyengar2, line 140: we could consider refactoring this whole logic into using wrangler.apply, but as spoken offline I'm not sure how much work that would be

aiyengar2 (changes requested):

modules/cli/apply/apply.go
prachidamle: yeah, that would help by not having two separate create and delete calls via API; however yeah maybe this can be added as an improvement later

modules/cli/apply/apply.go
prachidamle: yes added them for test; would remove them

charts/fleet-crd/templates/crds.yaml
prachidamle: I do not think we can just add the option to `fleet apply` cli without backing up with the API change - all the existing options of the apply command are API fields.
I think we can add prune by default for now.

modules/cli/apply/apply.go
```diff
@@ -90,18 +92,39 @@ func Apply(ctx context.Context, client *client.Getter, name string, baseDirs []s
...
+
+			for _, bundle := range bundles.Items {
+				if ok := gitRepoBundlesMap[bundle.Name]; !ok {
+					logrus.Debugf("Bundle to be deleted since it is not found in gitrepo %v anymore %v %v", name, bundle.Namespace, bundle.Name)
+					err := c.Fleet.Bundle().Delete(bundle.Namespace, bundle.Name, nil)
+					if err != nil {
+						return err
+					}
+				}
+			}
```
aiyengar2, line 127: One small change: if an output is provided, I think we should only log something to stderr and perform no action. If an output is not provided, then this looks fine.

i.e. we should follow the logic in `Dir` for whether or not to run the `save` command:
https://github.com/rancher/fleet/blob/17c46aca876eabd32f1564251b24e3bdbf4e43ef/modules/cli/apply/apply.go#L192-L196

Can we also wrap the code here in a `prune` helper function, similar to `save`?

aiyengar2 (changes requested):
one comment, otherwise LGTM.

We should ensure that we only perform the deletion when an output is not provided (e.g. it is not a dry-run), not when an output is provided (which is a dry run).

modules/cli/apply/apply.go
prachidamle: yes, added a helper function that figures out the bundles to prune and then either deletes them or logs them to output depending on opts.Output provided

aiyengar2 (approved):
LGTM

PR #774, opened 2022-05-17
Docs: use correct artifact version

moio:
It seems to me that the current link is not correct because the same version number is used both for fleet and the agent helm chart.

Current docs say:
```sh
helm -n fleet-system install --create-namespace --wait \
    ${CLUSTER_LABELS} \
    --values values.yaml \
    fleet-agent https://github.com/rancher/fleet/releases/download/v0.3.9/fleet-agent-0.3.3.tgz
```

But:

```
% curl https://github.com/rancher/fleet/releases/download/v0.3.9/fleet-agent-0.3.3.tgz
Not Found
```

It is my understanding the proposed fix should work!

### DIFF
--- a/docs/agent-initiated.md
+++ b/docs/agent-initiated.md
@@ -116,7 +116,7 @@ Finally, install the agent using Helm.
 helm -n fleet-system install --create-namespace --wait \
     --set clientID="${CLUSTER_CLIENT_ID}" \
     --values values.yaml \
-    fleet-agent https://github.com/rancher/fleet/releases/download/{{fleet.version}}/fleet-agent-{{fleet.helmversion}}.tgz
+    fleet-agent https://github.com/rancher/fleet/releases/download/{{fleet.version}}/fleet-agent-{{fleet.version}}.tgz
 ```
 
 The agent should now be deployed.  You can check that status of the fleet pods by running the below commands.


thardeck (approved):

aiyengar2 (approved):
TLDR: Yeah I agree with this fix. 

The reason why this was listed as helmVersion is because we hard-code the version to be 0.0.0 in the committed chart:

https://github.com/rancher/fleet/blob/042190df8ebbf8f1d470c34e0ce6bcf697729b26/charts/fleet/Chart.yaml#L20

And then patch it based off the value of the `HELM_VERSION` env var in `scripts/package-helm`:

https://github.com/rancher/fleet/blob/042190df8ebbf8f1d470c34e0ce6bcf697729b26/scripts/package-helm#L16-L19

But in an actual release, we always just default to using the TAG value anyways, which is where the download link would also be found:

https://github.com/rancher/fleet/blob/042190df8ebbf8f1d470c34e0ce6bcf697729b26/scripts/version#L26-L27

Therefore, I'd agree it's safe to change the docs to say that these values are the same, unless / until we change the scripting to support multiple versions in one release (no plans to do so atm though, which is why we are safe to proceed with this change).

prachidamle (approved):

moio:
Thanks for the detailed explanation and reviews!

PR #776, opened 2022-05-18
Fix linter issues

thardeck:
Follow up of #757, which should be merged before. It fixes the mentioned issues or ignores them if they are not valid for us.

Regarding [enforcing TLS 1.2](https://github.com/rancher/fleet/commit/5638998e520a9343d25403df7b4bf55537f75044), it was introduced 2008, it is also the Minimum in modern browsers like Firefox without know issues, so it should not be a problem.
Nevertheless we should document it in the next release notes though if some customer has a very old https setup.

The [debug endpoint](https://github.com/rancher/fleet/commit/d7bb9c4673a420198d837eb71178e8464d7c9995) should not be a security issue because it only listens on localhost but it should in my opinion still not be enabled by default. I have not changed the code by now to not change expected behavior/use cases but we should discuss to only activate it in a debug mode for example.

The last two commits ([1](https://github.com/rancher/fleet/commit/09acbdcea410b545f9f4dd37eb917260958281f8), [2](https://github.com/rancher/fleet/commit/7cccb746038469473e82f63e3148470cf91acf09)) change the error handling in a few places. By not ignoring them anymore it might be possible that unexpected issues appear for some users, which did not happen before for the same setup.
We should still handle errors properly so I would treat these potential issues as bugs or a use case which should have not worked before because of missing files for example.

### DIFF
--- a/pkg/bundle/resources.go
+++ b/pkg/bundle/resources.go
@@ -360,6 +360,10 @@ func readContent(ctx context.Context, progress *progress.Progress, base, name st
 	}
 
 	err = filepath.Walk(temp, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+
 		if info.IsDir() {
 			if strings.HasPrefix(filepath.Base(path), ".") {
 				return filepath.SkipDir
--- a/pkg/bundle/resources.go
+++ b/pkg/bundle/resources.go
@@ -338,7 +338,7 @@ func readContent(ctx context.Context, progress *progress.Progress, base, name st
 		httpGetter.Client.Transport = transport
 	}
 	if auth.SSHPrivateKey != nil {
-		if strings.IndexAny(c.Src, "?") == -1 {
+		if !strings.ContainsAny(c.Src, "?") {
 			c.Src += "?"
 		} else {
 			c.Src += "&"
--- a/pkg/bundle/resources.go
+++ b/pkg/bundle/resources.go
@@ -238,7 +238,7 @@ func readDirectories(ctx context.Context, compress bool, directories ...director
 		dir := dir
 		eg.Go(func() error {
 			defer sem.Release(1)
-			resources, err := readDirectory(ctx, p, compress, dir.prefix, dir.base, dir.path, dir.auth)
+			resources, err := readDirectory(ctx, compress, dir.prefix, dir.base, dir.path, dir.auth)
 			if err != nil {
 				return err
 			}
@@ -258,10 +258,10 @@ func readDirectories(ctx context.Context, compress bool, directories ...director
 	return result, eg.Wait()
 }
 
-func readDirectory(ctx context.Context, progress *progress.Progress, compress bool, prefix, base, name string, auth Auth) ([]fleet.BundleResource, error) {
+func readDirectory(ctx context.Context, compress bool, prefix, base, name string, auth Auth) ([]fleet.BundleResource, error) {
 	var resources []fleet.BundleResource
 
-	files, err := readContent(ctx, progress, base, name, auth)
+	files, err := readContent(ctx, base, name, auth)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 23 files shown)


manno (approved):
Already reviewed previously: https://github.com/manno/fleet/pull/2
Content is identical.

pkg/controllers/bundle/controller.go
```diff
@@ -101,7 +101,10 @@ func (h *handler) OnClusterChange(_ string, cluster *fleet.Cluster) (*fleet.Clus
 		}
 		for _, bundleDeployment := range bundleDeployments {
 			logrus.Debugf("cleaning up bundleDeployment %v in namespace %v not matching the cluster: %v", bundleDeployment.Name, bundleDeployment.Namespace, cluster.Name)
-			h.bundleDeployments.Delete(bundleDeployment.Namespace, bundleDeployment.Name, nil)
+			err := h.bundleDeployments.Delete(bundleDeployment.Namespace, bundleDeployment.Name, nil)
+			if err != nil {
+				return nil, err
+			}
```
prachidamle, line 107: Returning an error here will stop the workflow of handling the cluster change event and in turn will not let the bundle change events enqueued.
We should instead log the error and let the workflow proceed.

pkg/bundle/resources.go
```diff
@@ -360,6 +362,10 @@ func readContent(ctx context.Context, progress *progress.Progress, base, name st
 	}
 
 	err = filepath.Walk(temp, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
```
manno, line 365: @prachidamle can you comment on this, too? 

I was under the impression that this is a case, where we want to stop, but I can't be sure: 
https://github.com/manno/fleet/pull/2#discussion_r870022658

pkg/controllers/bundle/controller.go
thardeck: Thanks, logged as debug message.

pkg/bundle/resources.go
prachidamle: yes, we can error out here and that will catch any problems if the dir is inaccessible. We are returning error already from this method when the file cannot be read, so adding this should be fine.

manno (approved):

pkg/controllers/clusterregistration/controller.go
```diff
@@ -169,7 +169,7 @@ func (h *handler) OnSecretChange(key string, secret *v1.Secret) (*v1.Secret, err
 		return secret, nil
 	}
 
-	if time.Now().Sub(secret.CreationTimestamp.Time) > deleteSecretAfter {
+	if time.Since(secret.CreationTimestamp.Time) > deleteSecretAfter {
```
prachidamle, line 172: What is the linter error here? Does it provide any error code?

pkg/controllers/clusterregistrationtoken/handler.go
```diff
@@ -195,6 +195,6 @@ func (h *handler) deleteExpired(token *fleet.ClusterRegistrationToken) (bool, er
 		return true, h.clusterRegistrationTokens.Delete(token.Namespace, token.Name, nil)
 	}
 
-	h.clusterRegistrationTokens.EnqueueAfter(token.Namespace, token.Name, expire.Sub(time.Now()))
+	h.clusterRegistrationTokens.EnqueueAfter(token.Namespace, token.Name, time.Until(time.Now()))
```
prachidamle, line 198: Same comment as above - what is the linter error for using Sub() method on time package?

prachidamle (approved):
Some minor comments pending to answer; But rest looks good. LGTM

pkg/controllers/clusterregistration/controller.go
manno: You can run the linter locally like this:
`golangci-lint run --config=.golangci-full.json`

The change resulted from the gosimple linter: 

```
pkg/controllers/clusterregistration/controller.go:172:5: S1012: should use `time.Since` instead of `time.Now().Sub` (gosimple)
	if time.Now().Sub(secret.CreationTimestamp.Time) > deleteSecretAfter {
```

pkg/controllers/clusterregistrationtoken/handler.go
manno:
```
pkg/controllers/clusterregistrationtoken/handler.go:198:72: S1024: should use time.Until instead of t.Sub(time.Now()) (gosimple)
	h.clusterRegistrationTokens.EnqueueAfter(token.Namespace, token.Name, expire.Sub(time.Now()))
```

https://staticcheck.io/docs/checks/#S1024

PR #777, opened 2022-05-18
Remove helm chart for fleet-agent

manno:
I wasn't able to find any usage of the fleet-agent helm chart within the fleet repo. I think this chart is unused and can be deleted.
The fleet-agent helm chart that is found in the cluster after installing the fleet chart is created here: https://github.com/rancher/fleet/blob/master/pkg/agent/manifest.go#L42

The main difference between the two, the generated chart and the one in the repo, is the `command:` key. The repo chart allows for an optional debug flag and does not rely on the Docker image's entrypoint.

If the chart is actually used, we should document that in a chart/README.md. We might still want to remove it from the release, if it is only there to aid in debugging.


thardeck (approved):

aiyengar2:
Standalone Fleet users use this for agent-based registration https://github.com/rancher/fleet/blob/master/docs/agent-initiated.md.

aiyengar2 (changes requested):
Standalone Fleet users use this for agent-based registration https://github.com/rancher/fleet/blob/master/docs/agent-initiated.md; this can't be removed.

PR #778, opened 2022-05-18
Add e2e workflows for CI

manno:
This PR consists of two new Github Actions workflows:

* Deploys Rancher latest and registers a second cluster, then updates fleet to the version under test. Finally applies the multi cluster examples from fleet-examples and does some basic checks on the results.
* Deploys Fleet standalone from checkout and runs the single cluster examples from fleet-examples. 

The workflows can be run manually. They're also run every night to make sure the master branch and external dependencies are still ok. They are also run for every PR.

Notes:
* here is a repo with my helper scripts for running tests locally: https://github.com/manno/fleet-dev-tools/
* the expectations and test cases could be more sophisticated
* setting up rancher without the gui is a bit flaky and much more involved than I'd like it to be. Maybe someone more experienced can help to simplify that (in a separate PR)
* Fixing https://github.com/rancher/rancher/issues/37574 would allow us to use the newest CLI in test setup
* the rancher workflow needs a container image registry to work around flakiness, currently using ttl.sh, an internal registry (on the runner) would be better


mattfarina:
@brandonsuse @bmdepesa Can you please take a look at this?

It's failing because of GHCR related things. Before we set any of that up, can you please take a look at it.

.github/assets/register-downstream-clusters.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+url="${1-172.18.0.1.omg.howdoi.website}"
+cluster="${2-k3d-k3s-second}"
+
+# hardcoded from .github/assets/token-ci.yml
+token="token-ci:zfllcbdr4677rkj4hmlr8rsmljg87l7874882928khlfs2pmmcq7l5"
```
izaac, line 9: just curious if is it safe to have this token like this. Im guessing the environment is ephemeral, am I correct?

.github/assets/setup-latest-rancher.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+url="${url-172.18.0.1.omg.howdoi.website}"
```
izaac, line 5: is this an internal domain resolvable in CI?

.github/workflows/standalone-fleet-ci.yml
```diff
@@ -0,0 +1,152 @@
...
+      #   uses: mxschmitt/action-tmate@v3
+      #   timeout-minutes: 15
+      #   with:
+      #     limit-access-to-actor: true
+      -
+        name: Dump failed environment
+        if: failure()
+        run: |
+          mkdir -p tmp
+          kubectl get -A pod,secret,service,ingress -o json > tmp/cluster.json
```
izaac, line 135: how is this dumped information accessible after a failure?

e2e/multi-cluster/multi_cluster_test.go
```diff
@@ -0,0 +1,101 @@
...
+		Expect(err).ToNot(HaveOccurred(), out)
+	})
+
+	AfterEach(func() {
+		out, err := k.Delete("-f", testenv.AssetPath(asset))
+		Expect(err).ToNot(HaveOccurred(), out)
+
+	})
+
+	When("creating gitrepo resource", func() {
```
izaac, line 34: Would it be possible to have an  `When("edit gitrepo resource"),`?

For example creating `multi-cluster/helm` (the one that contains a helm chart) Editing the fleet cluster label from `dev` to `test` or viceversa you'd end with different resources (pod numbers, iirc please check the `multi-cluster/helm`). I think in my opinion will complete the E2E check on CI but im not sure if that would require much effort in this framework.

izaac (commented):
A few comments, I am not familiar with the CI infra used here or the ginkgo lib but I've focused more on the test definitions and coverage.

.github/workflows/rancher-fleet-ci.yml
```diff
@@ -0,0 +1,224 @@
...
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      -
+        name: Docker meta - fleet
+        id: meta-fleet
+        uses: docker/metadata-action@v4
+        with:
+          images: |
+            ghcr.io/manno/fleet
```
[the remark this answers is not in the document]
manno: Yes, this workflow needs a registry before it can be green. I asked for github packages in the org, but didn't receive a clear response yet. While Github container registry would be preferred, any registry should do. However, I don't have permissions to set up secrets in the repo, so I can't set up anything that requires them.

As long as we don't make this workflow a blocker for PRs, we can merge it before it actually works, set up the registry and make a small PR to replace the 'ghrc.io/manno' repo.

.github/assets/setup-latest-rancher.sh
manno: This will expand to either `$url`, or if not set to '172.18.0.1.omg.howdoi.website' which is a 'magic' DNS domain, that will resolve to 172.18.0.1. 172.18.0.1 is the IP of the first k3d cluster, which exposes port 80 and 443.
As far as I know, we still control 'omg.howdoi.website'. Any other magic DNS is fine, too.

e2e/multi-cluster/multi_cluster_test.go
manno: Yes, that would be a good future addition. That would check this https://github.com/rancher/fleet-examples/blob/master/multi-cluster/helm/fleet.yaml#L12-L14 functionality, right?

There is some work on another multi-cluster setup (standalone via terraform), I hope this will allow us to spin up three k3ds in total, so we can have two downstream clusters with different labels.

.github/assets/register-downstream-clusters.sh
manno: Yes, it's all within the Github actions runner. If we self-host in the future, we will need to deal with isolation and clean up.

.github/workflows/rancher-fleet-ci.yml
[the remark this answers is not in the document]
mattfarina: I'm going to suggest this uses GHCR. The tests are running in GitHub Actions. Using the container registry to push/pull from would be best to stay in the same vendor/setup. They can optimize for this.

The images uses here aren't for distribution. So we don't need them to be generally available.

e2e/single-cluster/single_cluster_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+	When("creating a gitrepo resource", func() {
+		Context("containing a helm chart", func() {
+			BeforeEach(func() {
+				asset = "single-cluster/helm.yaml"
+			})
+
+			It("deploys the helm chart", func() {
+				Eventually(func() string {
+					out, _ := env.Kubectl.Namespace("fleet-helm-example").Get("pods")
```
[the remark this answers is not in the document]
mattfarina: Let's get small slices of testing in per pull request. We don't need to solve for everything in one. We should be [agile](https://www.merriam-webster.com/dictionary/agile) in making changes.

Future changes can add more testing.

e2e/multi-cluster/multi_cluster_test.go
```diff
@@ -0,0 +1,101 @@
...
+		Context("containing a helm chart", func() {
+			BeforeEach(func() {
+				asset = "multi-cluster/helm.yaml"
+			})
+
+			It("deploys the helm chart in the downstream cluster", func() {
```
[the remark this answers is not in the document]
mattfarina: I would expect additional testing could be done in a future PR.

manno:
After discussing with @aiyengar2 we decided to put the multi-cluster tests in a separate PR. This PR no longer needs any registry and does not rely on Rancher for setup of fleet.

.github/workflows/standalone-fleet-ci.yml
```diff
@@ -0,0 +1,119 @@
+name: E2E Standalone Fleet
+
+on:
+  schedule:
+    - cron:  '0 5 30 * *'
```
aiyengar2, line 5: why do we need a cron schedule?

.github/workflows/standalone-fleet-ci.yml
```diff
@@ -0,0 +1,119 @@
...
+      -
+        name: Install Ginkgo CLI
+        run: go install github.com/onsi/ginkgo/v2/ginkgo@v2.1
+      -
+        name: Build fleet binaries
+        run: |
+          go build -o bin/fleetcontroller-linux-$GOARCH ./cmd/fleetcontroller
+
+          go build -o "bin/fleet-linux-$GOARCH"
+          go build -o "bin/fleetagent-linux-$GOARCH" ./cmd/fleetagent
```
aiyengar2, line 55: we should be using `./scripts/build`: https://github.com/rancher/fleet/blob/master/scripts/build

This would ensure the link flags are provided and would mimic our build process: https://github.com/rancher/fleet/blob/d45a36738d276ef9caf84aecc850b69d66b6c2cf/scripts/build#L10-L13

.github/workflows/standalone-fleet-ci.yml
```diff
@@ -0,0 +1,119 @@
...
+      -
+        name: Build Docker images
+        run: |
+          docker build -f package/Dockerfile -t rancher/fleet:dev --build-arg="ARCH=$GOARCH" .
+          docker build -f package/Dockerfile.agent -t rancher/fleet-agent:dev --build-arg="ARCH=$GOARCH" .
```
aiyengar2, line 60: we should use `./scripts/package`: https://github.com/rancher/fleet/blob/master/scripts/package

aiyengar2: probably should also run `./scripts/package-helm` in this step to inject the right versions into the chart: https://github.com/rancher/fleet/blob/d45a36738d276ef9caf84aecc850b69d66b6c2cf/scripts/package-helm#L1-L42

.github/workflows/standalone-fleet-ci.yml
```diff
@@ -0,0 +1,119 @@
...
+            --network "nw01"
+      -
+        name: K3d import images
+        run: |
+          k3d image import rancher/fleet:dev rancher/fleet-agent:dev -m direct
+      -
+        name: Set up tmate debug session
+        if: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.debug_enabled }}
+        uses: mxschmitt/action-tmate@v3
+        timeout-minutes: 15
```
aiyengar2, line 80: should this also be defined in `env` at the top like the Go version?

.github/workflows/standalone-fleet-ci.yml
```diff
@@ -0,0 +1,119 @@
...
+        with:
+          limit-access-to-actor: true
+      -
+        name: Deploy standalone fleet
+        run: |
+          helm -n fleet-system install --create-namespace --wait fleet-crd charts/fleet-crd
+          helm -n fleet-system upgrade --install --create-namespace --wait fleet charts/fleet
```
aiyengar2, line 87: how do we define what version of helm is used here? Is it defined above in the file?

aiyengar2: same question for kubectl below, base image is just `ubuntu-latest` right?

.github/workflows/standalone-fleet-ci.yml
```diff
@@ -0,0 +1,119 @@
+name: E2E Standalone Fleet
+
+on:
+  schedule:
+    - cron:  '0 5 30 * *'
+  workflow_dispatch:
+    inputs:
+      debug_enabled:
+        description: 'Run the build with tmate debugging enabled'
```
aiyengar2, line 9: nit:
```suggestion
        description: 'Enable debugging via tmate'
```

.github/workflows/standalone-fleet-ci.yml
```diff
@@ -0,0 +1,119 @@
+name: E2E Standalone Fleet
+
+on:
+  schedule:
+    - cron:  '0 5 30 * *'
+  workflow_dispatch:
+    inputs:
+      debug_enabled:
```
aiyengar2, line 8: nit:

can we change the name of this or is this a fixed name? if possible to change, we should make this `debug` or `enable_debug` instead of `debug_enabled`

.github/workflows/standalone-fleet-ci.yml
```diff
@@ -0,0 +1,119 @@
...
+            ~/.cache/go-build
+            ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+      -
+        name: Install Ginkgo CLI
+        run: go install github.com/onsi/ginkgo/v2/ginkgo@v2.1
+      -
+        name: Build fleet binaries
```
aiyengar2, line 50: nit:

Can we capitalize all of these workflow names?

```suggestion
        name: Build Fleet Binaries
```

.github/workflows/standalone-fleet-ci.yml
```diff
@@ -0,0 +1,119 @@
...
+      -
+        name: Dump failed environment
+        if: failure()
+        run: |
+          mkdir -p tmp
+          kubectl get -A pod,secret,service,ingress -o json > tmp/cluster.json
+          kubectl get -A events > tmp/events.log
+          helm list -A > tmp/helm.log
+          kubectl logs -n fleet-system -l app=fleet-controller > tmp/fleetcontroller.log
+          kubectl logs -n fleet-system -l app=fleet-agent > tmp/fleetagent.log
```
aiyengar2, line 106: Should we also get all Fleet CRs here as well? GitRepos, Clusters, ClusterGroups, Bundles, and BundleDeployments would be nice.

e2e/assets/single-cluster/helm-kustomize.yaml
```diff
@@ -0,0 +1,8 @@
+kind: GitRepo
+apiVersion: fleet.cattle.io/v1alpha1
+metadata:
+  name: helm-kustomize
```
aiyengar2, line 4: namespace needs to be provided, i.e. `fleet-default`

aiyengar2: should we also change the name here to be `single-cluster-helm-kustomize`? Same comment for the other GitRepo

e2e/assets/single-cluster/helm-kustomize.yaml
```diff
@@ -0,0 +1,8 @@
+kind: GitRepo
+apiVersion: fleet.cattle.io/v1alpha1
```
aiyengar2, line 2: nit: apiVersion and kind are flipped, I usually see apiVersion come first in manifests

e2e/assets/single-cluster/helm-multi-chart.yaml
```diff
@@ -0,0 +1,8 @@
+kind: GitRepo
+apiVersion: fleet.cattle.io/v1alpha1
+metadata:
+  name: helm
```
aiyengar2, line 4: should this have been `helm-multi-chart`? Also as indicated above, maybe this should be `single-cluster-helm-multi-chart` or `sc-helm-multi-chart`

e2e/assets/single-cluster/helm.yaml
```diff
@@ -0,0 +1,15 @@
...
+  repo: https://github.com/rancher/fleet-examples
+  paths:
+  - single-cluster/helm
+  targets:
+  - clusterSelector:
+      matchExpressions:
+      - key: provider.cattle.io
+        operator: NotIn
+        values:
+        - harvester
```
aiyengar2, line 15: don't think this is required in standalone fleet testing

.github/workflows/standalone-fleet-ci.yml
aiyengar2: Not sure if you will cover this in a followup PR where you test multi-cluster already

e2e/single-cluster/single_cluster_test.go
```diff
@@ -0,0 +1,102 @@
...
+		Context("containing a kustomize manifest", func() {
+			BeforeEach(func() {
+				asset = "single-cluster/kustomize.yaml"
+			})
+
+			It("runs kustomize and deploys the manifest", func() {
+				Eventually(func() string {
+					out, _ := env.Kubectl.Namespace("fleet-kustomize-example").Get("pods")
+					return out
+				}, testenv.Timeout).Should(ContainSubstring("frontend-"))
```
aiyengar2, line 68: nit: perhaps this should be encoded in a helper function?

aiyengar2 (approved):
@manno thank you for the contribution and sorry for the delay! While I've left a ton of comments here, I'm going to approve this PR since I don't see any red flags.

However, if you could review and respond to my comments before merging, that would be great. For any comment starting with `nit:`, if you believe your approach is better you can feel free to resolve the comment without responding.

.github/workflows/standalone-fleet-ci.yml
```diff
@@ -0,0 +1,119 @@
...
+
+on:
+  schedule:
+    - cron:  '0 5 30 * *'
+  workflow_dispatch:
+    inputs:
+      debug_enabled:
+        description: 'Run the build with tmate debugging enabled'
+        required: false
+        default: "false"
```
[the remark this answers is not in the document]
manno: I think this is only a problem, when you want to trigger one workflow from another. The title of the doc you linked is "GitHub Action for Dispatching Workflows". 
Here we use it to add a "Run workflow" button to the actions page, but the PR needs to be merged before the button appears.

Also, the same setup works in Epinio (e.g. https://github.com/epinio/epinio/blob/main/.github/workflows/aks-letsencrypt.yml#L13-L29).

.github/workflows/standalone-fleet-ci.yml
manno: It will give us a signal on flakiness of the tests and the health of the main branch. Especially in times of low activity, it's good to see that the main branch is still green. It's also a good baseline to judge a PR's status by, if nightly is red and a PR is green, that creates doubt. Finally, it can indicate problems with dependencies in the workflow.

.github/workflows/standalone-fleet-ci.yml
manno: I don't like to call scripts from workflows, especially if they were built for another purpose or have too many arguments and if clauses.

For tests we don't want those linker flags. If this was a QA test for a release candidate, sure, we might want to set a version, strip debug info and symbols from the binary and such.

This script is also used by Drone, so any changes we might want to do will affect the release process of fleet. I wanted to avoid  that for now.

.github/workflows/standalone-fleet-ci.yml
manno: We don't need the versions and the workflow actually depends on the default "dev" tag.

However, if `k3d image import` turns out to be too unreliable (sometimes it blocks on Github Actions, mabye because of `-m direct`) I want to replace the `docker build` lines with a Github action and use a local or ephemeral image registry, instead. This is already necessary for the multi-cluster workflow.

e2e/assets/single-cluster/helm.yaml
manno: Right. On the other hand it's good to see that this gets deployed nevertheless on a single-cluster.

.github/workflows/standalone-fleet-ci.yml
manno: The github runner images come with a lot of tools: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md

If this breaks, we will have to fix it.

.github/workflows/standalone-fleet-ci.yml
manno: Hm yes, I copied that from the tmate action docs. `enable_debug` is much nicer, but it doesn't really enable debug.. it enables tmate for debugging. It's not putting any fleet component into debug mode, or run the workflow with debug. Maybe just `enable_tmate`?

.github/workflows/standalone-fleet-ci.yml
manno: Yes, it's weird that some are capitalized and some are not. However this is a loosing battle ;) The very first title, coming directly from Github, is "Set up job".

.github/workflows/standalone-fleet-ci.yml
manno: Yes, good idea.

e2e/assets/single-cluster/helm-kustomize.yaml
manno: If we change the name, we should also change it in fleet-examples.

PR #789, opened 2022-06-07
Docs: cluster registration fixes

moio:
While developing [code to set up fleet automatically](https://github.com/moio/fleetaform) I found out a couple of steps missing from the documentation when deploying Fleet standalone.

I hope this helps 😃

### DIFF
--- a/docs/cluster-tokens.md
+++ b/docs/cluster-tokens.md
@@ -49,15 +49,26 @@ while ! kubectl --namespace=clusters  get secret new-token; do sleep 5; done
 
 ## Obtaining Token Value (Agent values.yaml)
 
-The token value is the contents of a `values.yaml` file that is expected to be passed to `helm install`
-to install the Fleet agent on a downstream cluster.  The token is stored in a Kubernetes secret referenced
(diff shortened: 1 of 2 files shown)


docs/cluster-tokens.md
```diff
@@ -39,17 +39,36 @@ spec:
...
 
+Such value is contained in the `values` field of the `Secret` mentioned above. To obtain the YAML content for the
+above example one can run the following one-liner:
 ```shell
-kubectl -n clusters get secret new-token -o 'jsonpath={.data.values}' | base64 --decode > values.yaml
+kubectl --namespace clusters get secret new-token -o 'jsonpath={.data.values}' | base64 --decode > values.yaml
 ```
 
-This `values.yaml` file can now be used repeatedly by clusters to register until the TTL expires.
+Note the `values.yaml` file is incomplete and will need manual adjustment before being used. Contents will be similar
```
manno, line 61: I noticed that, too, but I was under the impression that these values are not needed, if you specify them on the helm install command line (https://github.com/manno/fleet-dev-tools/blob/main/bin/setup-fleet-k3d-second#L38-L46).

manno (approved):

PR #791, opened 2022-06-09
E2E multi cluster workflow

manno:
This build upon #778 and adds the remaining multi-cluster examples in a second workflow. It also contains a fix for the single-cluster workflow, which broke because of an update to fleet-examples.

### DIFF
--- a/.github/workflows/standalone-fleet-ci.yml
+++ b/.github/workflows/standalone-fleet-ci.yml
@@ -72,7 +72,7 @@ jobs:
       -
         name: K3d Import Images
         run: |
-          k3d image import rancher/fleet:dev rancher/fleet-agent:dev -m direct
+          k3d image import rancher/fleet:dev rancher/fleet-agent:dev
       -
         name: Set Up Tmate Debug Session
         if: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.enable_tmate }}
--- a/.github/workflows/standalone-fleet-ci.yml
+++ b/.github/workflows/standalone-fleet-ci.yml
@@ -5,8 +5,8 @@ on:
     - cron:  '0 5 30 * *'
   workflow_dispatch:
     inputs:
-      debug_enabled:
-        description: 'Run the build with tmate debugging enabled'
+      enable_tmate:
+        description: 'Enable debugging via tmate'
         required: false
         default: "false"
   pull_request:
@@ -47,19 +47,19 @@ jobs:
         name: Install Ginkgo CLI
         run: go install github.com/onsi/ginkgo/v2/ginkgo@v2.1
       -
-        name: Build fleet binaries
+        name: Build Fleet Binaries
         run: |
           go build -o bin/fleetcontroller-linux-$GOARCH ./cmd/fleetcontroller
 
           go build -o "bin/fleet-linux-$GOARCH"
           go build -o "bin/fleetagent-linux-$GOARCH" ./cmd/fleetagent
       -
-        name: Build Docker images
+        name: Build Docker Images
         run: |
           docker build -f package/Dockerfile -t rancher/fleet:dev --build-arg="ARCH=$GOARCH" .
           docker build -f package/Dockerfile.agent -t rancher/fleet-agent:dev --build-arg="ARCH=$GOARCH" .
       -
-        name: Set up k3d cluster
+        name: Set Up k3d Cluster
         uses: AbsaOSS/k3d-action@v2
         # k3d will automatically create a network named k3d-test-cluster-1 with the range 172.18.0.0/16
         with:
@@ -70,18 +70,18 @@ jobs:
             --agents 3
             --network "nw01"
       -
-        name: K3d import images
+        name: K3d Import Images
         run: |
           k3d image import rancher/fleet:dev rancher/fleet-agent:dev -m direct
       -
-        name: Set up tmate debug session
-        if: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.debug_enabled }}
+        name: Set Up Tmate Debug Session
+        if: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.enable_tmate }}
         uses: mxschmitt/action-tmate@v3
         timeout-minutes: 15
         with:
           limit-access-to-actor: true
       -
-        name: Deploy standalone fleet
+        name: Deploy Standalone Fleet
         run: |
           helm -n fleet-system install --create-namespace --wait fleet-crd charts/fleet-crd
           helm -n fleet-system upgrade --install --create-namespace --wait fleet charts/fleet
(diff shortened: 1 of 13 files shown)


.github/workflows/standalone-fleet-ci.yml
```diff
@@ -47,19 +47,19 @@ jobs:
...
           go build -o "bin/fleetagent-linux-$GOARCH" ./cmd/fleetagent
       -
-        name: Build Docker images
+        name: Build Docker Images
         run: |
           docker build -f package/Dockerfile -t rancher/fleet:dev --build-arg="ARCH=$GOARCH" .
           docker build -f package/Dockerfile.agent -t rancher/fleet-agent:dev --build-arg="ARCH=$GOARCH" .
       -
-        name: Set up k3d cluster
+        name: Set Up k3d Cluster
```
aiyengar2, line 62: nit:
```suggestion
        name: Provision k3d Cluster
```
```suggestion
        name: Set Up k3d Cluster
```

.github/workflows/standalone-fleet-ci.yml
```diff
@@ -70,36 +70,37 @@ jobs:
             --agents 3
             --network "nw01"
       -
-        name: K3d import images
+        name: K3d Import Images
```
aiyengar2, line 73: nit: consistent naming of `k3d` and format of naming workflows with `<verb>...`
```suggestion
        name: Import Images Into k3d
```
```suggestion
        name: K3d Import Images
```

aiyengar2, line 73: nit: consistent naming of `k3d` and format of naming workflows with `<verb>...`
```suggestion
        name: Import Images Into k3d
```
```suggestion
        name: K3d Import Images
```

.github/workflows/standalone-fleet-ci.yml
```diff
@@ -70,36 +70,37 @@ jobs:
...
-        if: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.debug_enabled }}
+        name: Set Up Tmate Debug Session
+        if: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.enable_tmate }}
         uses: mxschmitt/action-tmate@v3
         timeout-minutes: 15
         with:
           limit-access-to-actor: true
       -
-        name: Deploy standalone fleet
+        name: Deploy Standalone Fleet
```
aiyengar2, line 84: nit: don't think we need to distinguish standlone fleet, we're either deploying Fleet or deploying Rancher
```suggestion
        name: Deploy Fleet
```

aiyengar2, line 84: nit: don't think we need to distinguish standlone fleet, we're either deploying Fleet or deploying Rancher
```suggestion
        name: Deploy Fleet
```

.github/workflows/standalone-fleet-multicluster-ci.yml
```diff
@@ -0,0 +1,179 @@
...
+        with:
+          cluster-name: "downstream"
+          args: >-
+            -p "81:80@agent:0:direct"
+            -p "444:443@agent:0:direct"
+            --api-port 6644
+            --agents 1
+            --network "nw01"
+      -
+        name: K3d Import Images
```
aiyengar2, line 84: same nit comments as to the above workflow

aiyengar2, line 84: same nit comments as to the above workflow

.github/workflows/standalone-fleet-multicluster-ci.yml
```diff
@@ -0,0 +1,179 @@
...
+          kubectl config use-context k3d-upstream
+          helm -n fleet-system install --create-namespace --wait fleet-crd charts/fleet-crd
+          helm -n fleet-system upgrade --install --create-namespace --wait fleet charts/fleet
+          kubectl -n fleet-system rollout status deploy/fleet-controller
+          { grep -q -m 1 "fleet-agent"; kill $!; } < <(kubectl get deployment -n fleet-system -w)
+          kubectl -n fleet-system rollout status deploy/fleet-agent
+      -
+        name: Deploy and Register Downstream Fleet
+        run: |
+          kubectl apply -f - <<EOF
```
aiyengar2, line 107: does this need `kubectl config use-context k3d-upstream`?

aiyengar2, line 107: does this need `kubectl config use-context k3d-upstream`?

.github/workflows/standalone-fleet-multicluster-ci.yml
```diff
@@ -0,0 +1,179 @@
...
+          ca=$(kubectl get -n default secret "$name" -o go-template='{{index .data "ca.crt" | base64decode}}')
+
+          kubectl config use-context k3d-downstream
+          helm -n fleet-system upgrade --install --create-namespace --wait fleet-agent charts/fleet-agent \
+            --set-string labels.env=dev \
+            --set apiServerCA="$ca" \
+            --set apiServerURL="https://172.18.0.1.omg.howdoi.website:6443" \
+            --set clusterNamespace="fleet-local" \
+            --set systemRegistrationNamespace="fleet-clusters-system" \
+            --set token="$token"
```
aiyengar2, line 129: I feel like non-trivial scripts like this should be placed into a common set of scripts under `./scripts`; this will also prevent us from duplicating the same script logic in both workflows.

Perhaps we should make these bash scripts under `./scripts/bootstrap/*.sh`?

aiyengar2, line 129: I feel like non-trivial scripts like this should be placed into a common set of scripts under `./scripts`. Perhaps we should make these bash scripts under `./scripts/bootstrap/*.sh`?

aiyengar2 (changes requested):

.github/workflows/standalone-fleet-multicluster-ci.yml
manno: No, the context doesn't change in between the steps.

.github/workflows/standalone-fleet-multicluster-ci.yml
manno: On the other hand the script is trivial, no parameters, no forks, as long as it has one purpose and is not in the common script folder.
Duplication is less maintenance than the wrong abstraction. 

I previously put the "setup-rancher" scripts into `.github/assets`, because I didn't want to suggest that these scripts should be used by anyone but the workflows.
I do not intend to build re-usable scripts. I don’t want to make the journey easier for devs and their local test setups (not in this PR I mean ;) ). I want to avoid parameters and conditionals.

So that’s why the workflows are mostly self-contained, have one user / environment (github actions runner) and don’t use shared scripts. As long as we don’t have re-use in maybe three places, I’m happy with copy and paste. We might decide to use terraform instead and all of this can be deleted soon. 

I have https://github.com/manno/fleet-dev-tools/tree/main/bin and use that myself for local testing, but I'm not ready to turn these into general-purpose scripts and maintain them in the main repo.

aiyengar2:
> @aiyengar2 I changed the action names and capitalized everything. However I'm wondering if we do the right thing here. These read much more like descriptions, than titles ([#778 (comment)](https://github.com/rancher/fleet/pull/778#discussion_r893250354)).
> 
> The very first "name", coming directly from Github, is "Set up job". Screenshot 2022-06-14 at 14 21 56

Thanks @manno! I think we should be consistent with what GitHub already uses to style the workflow descriptions. This looks clean to me.

aiyengar2 (approved):

PR #793, opened 2022-06-09
Add Workflow for Integration Tests

manno:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,77 @@
...
+        with:
+          path: |
+            ~/.cache/go-build
+            ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+      -
+        name: unit-test
+        run: go test -shuffle=on $(go list ./... | grep -v /e2e | grep -v /integration)
```
aiyengar2, line 42: This is already run by Drone CI, we shouldn't need to run it here.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,77 @@
...
+            ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+      -
+        name: Install Ginkgo CLI
+        run: go install github.com/onsi/ginkgo/v2/ginkgo@v2.1
+      -
+        name: Test
+        run: ginkgo integration
```
aiyengar2, line 77: Since we don't need a live cluster to test `fleet apply`, I think this should just be a `./scripts/integration` script that is defined similarly to `./scripts/test` and called in `./scripts/ci`:

https://github.com/rancher/fleet/blob/976fd120fb15eb0d310540af436b52e1f5817821/scripts/ci#L7

Installing Ginkgo CLI can be done in the [`Dockerfile.dapper`](https://github.com/rancher/fleet/blob/976fd120fb15eb0d310540af436b52e1f5817821/Dockerfile.dapper) itself and we can avoid creating another GH workflow for this since it will just be part of running `ci` itself.

aiyengar2 (changes requested):

manno:
Hm, integration test is such a overloaded term (﻿https://martinfowler.com/bliki/IntegrationTest.html). But unit tests depending on directories from fleet-examples is not good. So maybe naming can be improved by calling this a "feature" test?

The tests seem not extensive enough and we test the actual CLI over in the fleet-examples repo now. I'm closing this draft for now. This will be easier once https://github.com/rancher/fleet/issues/809 is done and we have a better understanding (from bugs?) about what should be tested (cluster targeting, repo->bundle conversion, ...).

PR #794, opened 2022-06-09
Fix values.yaml order of acceptance

timofey-drozhzhin:
`helm.valueFiles` is misspelled. The correct spelling is `helm.valuesFiles` (note the `s`).


mattfarina:
Thanks for catching the oops.

PR #796, opened 2022-06-13
Switch docker image to bci-base

manno:
The bci-micro image would suffice for the controller, however the agent image is
also used for debugging and needs ssh and git installed.
As we don't use the controller without the agent, using the same base
image reduces the download size.


olblak (approved):

mattfarina:
We might be able to change to a smaller version of the BCI. I'm still getting details so I'm holding off another day on the approval.

manno:
> We might be able to change to a smaller version of the BCI. I'm still getting details so I'm holding off another day on the approval.

As long as dependabot is able to bump a version like this: [15.3.17.17.13](https://registry.suse.com/static/bci/bci-base/index.html)

manno:
We suspect this will create issues with ARMv7 once we push a tag.

package/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM alpine:3.16.0
+FROM registry.suse.com/bci/bci-base:15.3
```
mattfarina, line 1: Can we get the full image tag. The current one is `15.3.17.17.13`.

There are two reasons I'm asking for this...
1. We'll know the exact version of the BCI being used at any given time. This gives us visibility (e.g., if we ever need to troubleshoot)
2. There won't be a different version between dev and production. Let's say we do our work and testing to one version. Then, before release there is a new version of 15.3. The release would be built on that instead of the tested version. While I would not expect issues, this provides a little more control.

mattfarina (changes requested):
After we create the release branch that will be used with the next Rancher release we can update the image to use 15.4. I created #844 to track that.

manno:
> Could we also extend this PR a bit to:
> 
>     1. `fleet` image
>        
>        * Instead of `USER 1000`, can we create a user like `useradd -m -U fleet` and then `USER fleet`? It's a best practice to actually create the user instead just running with a random UID. https://github.com/rancher/fleet/blob/cf66c2ec52150eb97260acd52af80e40dd0b04c7/package/Dockerfile#L5

Well, it's not really random, by tradition this is the first user ID for non-system users. (`UID_MIN` in `/etc/login.defs`).

But yes, that user id doesn't exist neither in alpine nor bci, we need to create it. 

>     2. `fleet-agent` image
>        
>        * It creates the user, but it misses the `USER fleet-apply` directive. 

👍 

Are we sure no one relies on the exact value of 1000?

package/Dockerfile
manno: Like I said before, that version is not a valid semver and I am afraid that dependabot can't update it. Do you know if that works in other projects?

macedogm: Dependabot does work with the long version. I tested it - https://github.com/rancher/shell/pull/28.

manno: I'm not sure what problems that could be, given that we could replace the OS with a `scratch` image.

manno:
> > Are we sure no one relies on the exact value of 1000?
> 
> Not 100% sure, because I never deeply tested if this would break it or not. However, relying on a fixed UID would not be ideal. So if it breaks, better break it now before shipping 2.6.6.

Reading https://kubernetes.io/docs/concepts/security/pod-security-policy/#users-and-groups sounds like we shouldn use a numeric value in the `USER` directive?

I'm not sure why we put a user in the Dockerfile at all, I'd modify the deployment to have a securityContext instead: https://github.com/epinio/helm-charts/blob/bd2308024c8d9218fe3bf0fd6a71814e52102f69/chart/epinio/templates/server.yaml#L311-L314

mattfarina (approved):

Issue #798, opened 2022-06-17
Bump to Go 1.18


manno:
Fleet is using an old version of Go, assuming there is no technical reason to stay on the old version:

* Let's bump to the latest version. 
* Use a minor version in the Dockerfile, so we don't need to bump for every patch version.
* We also need to update golangci-lint to be compatible.

I propose to remove golangci-lint from Drone CI (scripts/validate-ci), 
* because it is outdated
* it's has deprecated linters in it's configuration and only a few
* installation is manual/scripted as opposed to using a popular github action
* we want to use drone for building only

manno:
We are waiting for the bump in rancher/gitjob, ranchger/wrangler-cli

manno:
gitjob is more difficult to update, extracted to https://github.com/rancher/fleet/issues/836

PR #800, opened 2022-06-17
Remove linter from Drone CI

manno:
Refers to #798


mattfarina:
Note, the linter needs to be updated so that it works with the latest version of Go. There were changes in Go and, as I understand it, older versions of linters have issues with some newer Go.

Removing from Drone means we only need to have linting in one place. Pulling it out of Drone should speed up those runs and let the work happen in parallel.

mattfarina (approved):
I'll do a PR to handle `gofmt` so that this can unblock Go 1.18 work.

I manually testing the lint run and `make validate`.

Issue #809, opened 2022-06-20
Extract unit tests from Drone CI into github workflow


manno:
We want to get rid of the testing in Drone CI.

PR #810, opened 2022-06-20
Add unit tests  to Github Actions

manno:
Refers to #809

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,42 @@
+name: CI
+
+on:
+  schedule:
+    - cron:  '0 5 * * *'
+  pull_request:
+    paths-ignore:
+    - 'docs/**'
+    - 'charts/**'
+    - 'scripts/**'
+    - '*.md'
+
+env:
+  GOARCH: amd64
+  CGO_ENABLED: 0
+  SETUP_GO_VERSION: '^1.18'
+
+jobs:
+  unit-test:
+    runs-on: ubuntu-latest
+
+    steps:
+      -
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      -
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ env.SETUP_GO_VERSION }}
+      -
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cache/go-build
+            ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+      -
+        name: unit-test
+        run: go test -shuffle=on $(go list ./... | grep -v /e2e)
--- a/scripts/ci
+++ b/scripts/ci
@@ -4,7 +4,6 @@ set -e
 cd $(dirname $0)
 
 ./build
-./test
 ./validate
 ./validate-ci
 ./package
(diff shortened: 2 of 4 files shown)


olblak:
@manno @mattfarina  
Do you have a document that explain when we should use Drone and when we should switch to Github Action?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,42 @@
...
+        with:
+          path: |
+            ~/.cache/go-build
+            ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+      -
+        name: unit-test
+        run: go test -shuffle=on $(go list ./... | grep -v /e2e)
```
olblak, line 42: What's your motivation to get rid of the `test` script? 
It's usually a better approach to call small scripts from a CI environment than copy pasting everything in a specific CI config file.
It's easier to test locally or to migrate between CI environment.

So I would prefer something like
```suggestion
        run: ./scripts/test
```

manno: `scripts/test` is used by the Drone CI. I'd rather have the workflows be self-contained and not extract trivial scripting into shell scripts. 
Testing locally or using a different CI are similar use cases, but in practice it turns out they are only similar, not equal. When you try to support multiple use cases (gha, drone, local dev on mac, local dev on windows, ...) you end up with complicated shell scripts.

For myself I have https://github.com/manno/fleet-dev-tools/tree/main/bin, where I could put a script for unit testing, if it weren't so simple.
Again, it is not my intent to improve the tooling for local development with this PR. The value of the unit tests is really low, too. In any case, local development scripts would need to be separate from CI scripts (and especially the dapper scripts in `scripts`).

manno: @olblak  I'm not sure how to proceed, after finding out about https://github.com/rancher/build-tekton I'm afraid to delete the test script.
That  repo has a test script, but no go files? Maybe dapper needs that somehow? In which case I'd leave the test script in and just remove it from the other scripts?

manno:
> @manno @mattfarina Do you have a document that explain when we should use Drone and when we should switch to Github Action?

I discussed with Matt and we want to use Drone only for building the release.

mattfarina:
I'm sorry to say this just occurred to me. Right now tests run before artifacts are released (built, pushed to object storage, or container images are pushed). If they fail the release assets won't be published. Removing them here means we could have failing tests and still put out release assets.

Should we keep the tests as a blocker on release? Failed tests would stop it.

.github/workflows/ci.yml
olblak: > Should we keep the tests as a blocker on release? Failed tests would stop it.

I would keep them as an extra protection but as long as it doesn't slow down (too much) the release process. 
But to be honest I don't know how reliable they are

olblak:
> Should we keep the tests as a blocker on release? Failed tests would stop it.

I would keep them as an extra protection but as long as it doesn't slow down (too much) the release process. 
But to be honest I don't know how reliable they are

.github/workflows/ci.yml
olblak: Personally I would keep the script and call it from GitHub action. It's less disruptive

manno:
We decided to not modify the dapper scripts and just use this to add unit tests. They do run fast anyhow.

olblak (approved):

thardeck (approved):

Issue #803, opened 2022-06-17
Add "Dependabot" Config


manno:
We want to bump dependencies more often, we need tooling to create PRs automatically:

* e.g. Dependabot config for go modules (maybe like https://github.com/epinio/epinio/blob/main/.github/dependabot.yaml)
* something to bump versions of used container images (dependabot can handle simple base image bumps)

Note: Apparently dependabot is already enabled in fleet: https://github.com/rancher/fleet/pull/788

- [x] https://github.com/rancher/fleet
- [x] ~~https://github.com/rancher/wrangler~~ PR opened
- [x] https://github.com/rancher/wrangler-cli
- [x] https://github.com/rancher/build-tekton
- [x] https://github.com/rancher/gitjob

PR #811, opened 2022-06-21
Add dependabot config

manno:
Check gomod, container images and github actions monthly.

Refers to #803

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -4,21 +4,21 @@ updates:
     package-ecosystem: "gomod"
     directory: "/"
     schedule:
-      interval: "monthly"
+      interval: "weekly"
     labels:
       - "dependencies"
   -
     package-ecosystem: "gomod"
     directory: "/pkg/apis"
     schedule:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yaml
```diff
@@ -0,0 +1,30 @@
+version: 2
+updates:
+  -
+    package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "monthly"
```
olblak, line 7: I think as a first step, I would go weekly to see how much updates are needed on a regular basis. 
Then based on  the amount of pullrequest you switch to monthly if needed

manno: I was wondering if we have the bandwidth, but yes let's do it!

olblak (approved):
I would run Dependabot on a weekly basis for everything  including (docker and github-actions) and switch to monthly basis if it gets too noisy.
 
But it's a personal taste so don't consider this as blocker

mattfarina (approved):

PR #828, opened 2022-06-21
Adding gofmt back as a linter

mattfarina:
Refers to https://github.com/rancher/fleet/issues/798

### DIFF
--- a/.golangci.json
+++ b/.golangci.json
@@ -44,7 +44,8 @@
       "gosimple",
       "typecheck",
       "unused",
-      "funlen"
+      "funlen",
+      "gofmt"
     ]
   },
   "issues": {


manno:
I'm under the impression that goimports does the same, quick check seems to confirm that. 

After messing with the indentation:

    example/pkg/app/subcommand.go:13: File is not `goimports`-ed (goimports)
		Long: "Add some long description",


    example/pkg/app/subcommand.go:13: File is not `gofmt`-ed with `-s` (gofmt)
		Long: "Add some long description",


According to the goimports docs:

> In addition to fixing imports, goimports also formats your code in the same style as gofmt so it can be used as a replacement for your editor's gofmt-on-save hook.  (https://pkg.go.dev/golang.org/x/tools/cmd/goimports?utm_source=godoc)


We also have vet and revive (the golint successor) in the default set. Do you have a special case where we miss a fmt rule?

mattfarina:
Good call on goimports. Closing this one.

PR #850, opened 2022-06-27
Changes to Fleet to support K8s 1.24 

prachidamle:
https://github.com/rancher/rancher/issues/37885

Fleet does not work in k8s 1.24 cluster because there is no auto-generated secret-based service account token anymore in K8s 1.24. Prior to 1.24, every serviceAccount created was auto-backed by K8s by a Secret token. Now with 1.24, there will be no auto population of the Secret token for serviceaccounts.

Thus we need to create the ServiceAccount Secrets ourselves in the controllers when needed for all the serviceAccounts used in Fleet.

We need to fix this issue to support the goal of running rancher in k8s 1.24

Changes include:
- If a ServiceAccount does not have a Secret token, then Fleet will create a secret of type SecretTypeServiceAccountToken   and wait for K8s to populate it
- This change is needed for fleet-controller bootstrap and clusterRegistration usecases.

### DIFF
--- a/pkg/controllers/clusterregistration/controller.go
+++ b/pkg/controllers/clusterregistration/controller.go
@@ -12,6 +12,7 @@ import (
 	"github.com/rancher/fleet/pkg/config"
 	fleetcontrollers "github.com/rancher/fleet/pkg/generated/controllers/fleet.cattle.io/v1alpha1"
 	"github.com/rancher/fleet/pkg/registration"
+	secretutil "github.com/rancher/fleet/pkg/secret"
 	"github.com/rancher/wrangler/pkg/apply"
 	corecontrollers "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	rbaccontrollers "github.com/rancher/wrangler/pkg/generated/controllers/rbac/v1"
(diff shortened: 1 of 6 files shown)


olblak:
It is possible to have a bit of context to understand why this pullrequest is needed?

pkg/controllers/clusterregistration/controller.go
```diff
@@ -98,7 +99,7 @@ func saToClusterRegistration(namespace, name string, obj runtime.Object) ([]rela
 	if sa, ok := obj.(*v1.ServiceAccount); ok {
 		ns := sa.Annotations[fleet.ClusterRegistrationNamespaceAnnotation]
 		name := sa.Annotations[fleet.ClusterRegistrationAnnotation]
-		if ns != "" && name != "" && len(sa.Secrets) > 0 {
+		if ns != "" && name != "" {
```
prachidamle, line 102: About this change: There is no way to grab the token of the SA here since no ctx is passed to use the cache/api lookup. But as long as the SA has non-empty ClusterRegistrationAnnotation, it should be ok to enqueue that SA to be checked by ClusterRegistration controllers.

pkg/controllers/clusterregistrationtoken/handler.go
```diff
@@ -81,6 +86,25 @@ func (h *handler) OnChange(token *fleet.ClusterRegistrationToken, status fleet.C
 		if err != nil {
 			return nil, status, err
 		}
+	} else if len(sa.Secrets) == 0 {
+		secretCreated, err := secretutil.GetServiceAccountTokenSecret(sa, h.secretsController)
+		if err != nil {
+			return nil, status, err
+		}
+		secretReloaded, err := h.secretsCache.Get(token.Namespace, secretCreated.Name)
```
kinarashah, line 94: `GetServiceAccountTokenSecret` waits for token and returns the secret, why do we need to reload or/and check of empty data token again?

prachidamle: This is a check to ensure we can proceed with clusterRegistration workflow, if the secret is empty, registration workflow halts and needs a restart for fleet-controller to move forward. So making sure we proceed only token is well set.

kinarashah (approved):

PR #853, opened 2022-07-04
Bump go modules

### LINKED ISSUE #745: Update kustomize version that fleet uses
Currently the Kustomize version used in Fleet  is somewhere between v4.0.5 and v4.1.3 (based on api and kyaml dependencies)

We need to investigate and ugrade  Kustomize to the latest possible version within Fleet.

manno:
This bumps all the go modules at once, due to the many, very broad, 'replace' directives I couldn't find a smaller set.

I also had to copy the 'diff' package from argoproj. The modules have not been updated upstream and we want a newer k8s.io/api. I think copying is better than forking, at least temporary. According to https://blog.argoproj.io/argo-cd-v1-6-democratizing-gitops-with-gitops-engine-5a17cfc87d62 the diff package is meant to be a publicly reusable library. We could switch back to the import once gitops-engine is updated, however we also import from argo-cd, which we should avoid I think.
Looking at a snyk scan of our go modules, we import too much from argo-cd. So maybe importing the gitops-engine and copying/vendoring the argo-cd files, in the future?

Since argo-cd and gitops-engine use the server scheme from k8s.io/kubernetes we need to work around 'unknown revision 0.0.0' errors, by including replace statements in the go.mod file: kubernetes/issues/79384

Vendoring the argoproj files helps with updating k8s, but is not enough
to get rid of the replace statements: https://github.com/argoproj/gitops-engine/issues/56

Fixes #745


mattfarina (commented):
@manno I've started to review this one. I paired with Arvind for a short bit to discuss it.

At first we considered the case of using k8s 1.23 packages since there is a version skew of +/-1 so that would provide k8s 1.24 support. But, wrangler has already updated to k8s 1.24 so that wouldn't work. vendoring part of argo works for the time being.

I'm still reviewing so I'll come back to this. Just wanted to give you an update.

mattfarina:
@manno can you please create a PR to cherry-pick this into the release branch?

PR #854, opened 2022-07-06
Agent, controller deployments run non-root, read-only filesystem

### LINKED ISSUE #845: Run controller deployments as non-root with read-only filesystem
Add security contexts to runAsNonRoot and switch the filesystem to read-only.

The UIDs have to match the user/group that were created in the the container
images.

Example:
https://github.com/epinio/helm-charts/blob/bd2308024c8d9218fe3bf0fd6a71814e52102f69/chart/epinio/templates/server.yaml#L308-L314

TODO:

- [ ] check for filesystem writes (ssh known_hosts?)
- [ ] test on public providers (gke/aks/eks)

manno:
Add security contexts to runAsNonRoot and switch the filesystem to read-only.

The UIDs have to match the user/group that were created in the the container
images.

- [x] possible unhandled write access to /tmp from [image.go](https://github.com/rancher/fleet/blob/master/pkg/controllers/image/image.go#L212)

fixes #845


mattfarina:
Please test in EKS, AKS, and GKE to make sure the agent can get a service account properly (and functions properly).

manno:
Do we need to make the id (1000) configurable via the helm chart, so users can avoid conflicts with existing user ids on their nodes?

manno:
Tested fine on GKE

moio:
> Do we need to make the id (1000) configurable via the helm chart, so users can avoid conflicts with existing user ids on their nodes?

The way I understand it is that user IDs are remapped by the container runtime anyway, so the choice is completely free from constraints coming from nodes.

PR #857, opened 2022-07-07
BUG: race condition in readContent at pkg/bundle/resources.go

nabam:
Shared global variable `getter.Getters` is updated in concurrently running goroutines. This causes corruption of authorization configuration for http getter when charts form various sources are used in the same git repository.

### DIFF
--- a/pkg/bundle/resources.go
+++ b/pkg/bundle/resources.go
@@ -306,35 +306,7 @@ func readContent(ctx context.Context, base, name string, auth Auth) (map[string]
 		return nil, err
 	}
 
-	// copy getter.Getters before changing
-	getters := map[string]getter.Getter{}
-	for k, v := range getter.Getters {
-		getters[k] = v
-	}
 	src := name
-
-	httpGetter := &getter.HttpGetter{
-		Client: &http.Client{},
-	}
-
-	if auth.Username != "" && auth.Password != "" {
-		header := http.Header{}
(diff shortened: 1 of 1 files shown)


nabam:
@StrongMonkey

pkg/bundle/resources.go
```diff
@@ -306,13 +306,19 @@ func readContent(ctx context.Context, base, name string, auth Auth) (map[string]
...
+	}
+
 	c := getter.Client{
 		Ctx:     ctx,
 		Src:     name,
 		Dst:     temp,
 		Pwd:     base,
 		Mode:    getter.ClientModeDir,
-		Getters: getter.Getters,
+		Getters: getters,
```
mattfarina, line 321: Further down in the code, the original `getter.Getters` is being updated. That should use the new `getters` instead. See...

https://github.com/rancher/fleet/blob/bd0fd515a2823c68b0ce0e5eab71558811fadd0c/pkg/bundle/resources.go#L350-L351

Can you please update all the uses of `getter.Getters`.

mattfarina (changes requested):
Thanks for the pull request. I can see the collision. I found one place where there is still a problem in the code.

pkg/bundle/resources.go
nabam: this is `c.Getters` that already refer to copied getters map, but I'll change it anyway for the sake of readability

manno (approved):

PR #859, opened 2022-07-08
Create pull request template

olblak:
To encourage contributors to provide at least a minimal set of information when they create pullrequest, I propose the following pullrequest template

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,23 @@
+<!-- Specify the issue ID that this pullrequest is solving -->
+Fix #XXX
+
+<!-- Describe the changes introduced by this pull request -->
+
+## Test
+
+To test this pull request, you can run the following commands:
+
+```shell
+cp <to_package_directory>
+go test
+```
+
+## Additional Information
+
+### Tradeoff
+
+<!-- Please describe, if any, the tradeoffs that you found acceptable in this pull request -->
+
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,23 @@
...
+Fix #XXX
+
+<!-- Describe the changes introduced by this pull request -->
+
+## Test
+
+To test this pull request, you can run the following commands:
+
+```shell
+cp <to_package_directory>
```
mattfarina, line 11: `cp` -> `cd`

.github/pull_request_template.md
```diff
@@ -0,0 +1,23 @@
...
+go test
```
mattfarina, line 12: should it be `go test ./...` so that all the nested directories are tested?

mattfarina (changes requested):

.github/pull_request_template.md
manno: That won't work, because you need a cluster and the fleet-examples repo for some tests. The unit tests provide ~4% coverage at the moment. They don't help with verifying a PR. 

I'd remove the testing part from the template. We only have the workflows at the moment.

We could mention `golangci-lint run`.

olblak: I can propose instead

manno (approved):

PR #864, opened 2022-07-09
Implement Helm Chart Capability Check

### LINKED ISSUE #399: Helm installation fails because of capability check
Hello,

Trying to install traefik on rancher v2.5.8 (fleet v0.3.5) fails in a `Processed` state when the supplied values requires a capability check.

The fleet.yaml looks like this:
```
defaultNamespace: traefik-experimental
helm:
  releaseName: traefik-experimental
  chart: https://internal-registry.example.com/repository/helm-proxy-traefik/traefik-9.18.2.tgz
  values:
    deployment:
      replicas: 2
    podDisruptionBudget:
      enabled: true
      minAvailable: 1
    globalArguments: []
    service:
      type: NodePort
    logs:
      general:
        level: INFO
      access:
        enabled: true
    ingressRoute:
      dashboard:
        annotations:
          kubernetes.io/ingress.class: experimental
    ingressClass:
      enabled: true
 ```
 
 The error message when looking at the bundle in the UI is this:
 ```
 template: traefik/templates/ingressclass.yaml:7:8: executing "traefik/templates/ingressclass.yaml" at <fail "\n\n ERROR: You must have atleast networking.k8s.io/v1beta1 to use ingressClass">: error calling fail: ERROR: You must have atleast networking.k8s.io/v1beta1 to use ingressClass
 ```
 
I have verified that I can install it manually using `helm install` so it is was not a problem with the version of the underlying kubernetes cluster version.
 
 The problematic part is this:
 ```
     ingressClass:
      enabled: true
 ```
 
 If I remove that part it works with fleet as well, and what is actually happening when setting it can be seen here:
 https://github.com/traefik/traefik-helm-chart/blob/v9.18.2/traefik/templates/ingressclass.yaml:
 ```
 {{- if and .Values.ingressClass.enabled (semverCompare ">=2.3.0" (default .Chart.AppVersion .Values.image.tag)) -}}
  {{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/IngressClass" }}
apiVersion: networking.k8s.io/v1
  {{- else if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1/IngressClass" }}
apiVersion: networking.k8s.io/v1beta1
  {{- else }}
    {{- fail "\n\n ERROR: You must have atleast networking.k8s.io/v1beta1 to use ingressClass" }}
  {{- end }}
```
 
Looking at the helm code in fleet I have noticed the use of DefaultCapabilities here:
https://github.com/rancher/fleet/blob/d21bb02b29ef15bdc689d65a2b2ac918d0d0cae9/pkg/helmdeployer/template.go#L27

Which, if I am understanding the go.mod file correctly in fleet v0.3.5, results in this data that only appears to include Core V1 ("v1") based on the comment:
https://github.com/rancher/helm/blob/v3.3.3-fleet1/pkg/chartutil/capabilities.go#L27-L41

IngressClass is not part of core v1 (https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#ingressclass-v1-networking-k8s-io) and I am wondering if this might be the reason the check fails.

If I understand it correctly there is support in the helm library to fetch capabilities when calling Run() at https://github.com/rancher/fleet/blob/d21bb02b29ef15bdc689d65a2b2ac918d0d0cae9/pkg/helmdeployer/deployer.go#L336 but I am not sure if this is involved at the time the error above is thrown.

romejoe:
This merge request attempts to resolve #399. A replication of Helm's Capabilities API was added to the Cluster's Status. When agents start, they do the same capabilities and version checks that Helm does. These collected capabilities are then passed back to the fleet manager by the status object. This allows the helmdeployer on the fleet manager to have access to the correct version information and capabilities of each cluster when evaluating bundles. This capabilities information is also passed along to the agent's helmdeployer.

Some changes that I concerned about though are:
- I am not sure if the clusters.UpdateStatus reliably updates the status in controller.go. I had some trouble getting that to reliably  transmit the capabilities portion of the status.
- I added a continue to the target processing of bundles in controller.go because it appears that an empty target that was being processed in my testing.

We tested this with the fleet manager running on v1.20.4(rke1) and a single agent running on 1.24.1(rke2) and Rancher 2.6.7-rc1.


romejoe:
I realize that this pull request will not solve the issue and won't work. It works on my machine because I made some assumptions without realizing. I am actively trying to fix the issues I found.

romejoe:
Just pushed up some small fixes because I realized the build and ci were failing...

modules/agent/pkg/controllers/cluster/cluster.go
```diff
@@ -33,14 +33,7 @@ type handler struct {
...
 
-func Register(ctx context.Context,
-	agentNamespace string,
-	clusterNamespace string,
-	clusterName string,
-	checkinInterval time.Duration,
-	nodes corecontrollers.NodeCache,
-	clusters fleetcontrollers.ClusterClient) {
-
+func Register(ctx context.Context, agentNamespace string, clusterNamespace string, clusterName string, checkinInterval time.Duration, nodes corecontrollers.NodeCache, clusters fleetcontrollers.ClusterClient) {
```
olblak, line 36: Out of curiosity, is there any reason to change this function arguments definition?

manno:
Hey just a heads up, the `charts/fleet-crd/templates/crds.yaml` is actually generated from `pkg/crd/crds.go` (as I just found out while working on https://github.com/rancher/fleet/pull/918/files#diff-927d423e0c49728f25e29bb70ac2e0d993b9be8cebd4cfbaea474e18e34f3982).

romejoe:
@manno thanks for the info, I thought I got the crds updated, but I'll double check that. I also see that the merge has a conflict with the upstream crds. I'm going to try to merge the upstream changes into my branch and resolve the conflicts in the morning.

PR #866, opened 2022-07-11
Fix downgrade of dev version by Rancher

manno:


mattfarina (approved):
This should work. If you look at [the code where Rancher Manager runs fleet,](https://github.com/rancher/rancher/blob/release/v2.6/pkg/controllers/dashboard/fleetcharts/controller.go) it has a min version. Just need to be higher than that.

Thanks to @prachidamle for pointing this out to me.

PR #868, opened 2022-07-12
Cherry-Picks for v0.3.10-rc.2

mattfarina:
This cherry picks the following PRs to the release branch so we can use them in the next RC.

- #865
- #853
- #867
- #866

Note, updating the go modules has removed the known vulnerabilities found in the scans with the exception of one that does not apply to fleet but is found in k8s and is unfixed (for quite some time).


prachidamle:
The rc as per current naming convention should be v0.3.10-rc2

mattfarina:
@prachidamle I see how we are doing pre-releases. We might change that in the future as the more semantic way to do it is `rc.2` with the `.`. This is how [the semver spec shares it as an example](https://semver.org/#spec-item-9). This is how GitHub suggests it. Also, go (and other tool) semver handling looks for the number to be separate when it does pre-release comparisons. See an example (in the go source) [here](https://cs.opensource.google/go/x/mod/+/refs/tags/v0.5.1:semver/semver.go;drc=86c51ed26bb44749b7d60a57bab0e7524656fe8a;l=333).

prachidamle:
Issue reference for this PR is https://github.com/rancher/rancher/issues/38248

PR #886, opened 2022-07-22
Bump GitJob to v0.1.29 in release branch

thardeck:


mattfarina (changes requested):
gitjob is also being imported and used as a Go package. Can you update it in the Go modules as well.

mattfarina (approved):

Issue #395, opened 2021-05-27
Update namespace from fleet-system to cattle-fleet-system


kinarashah:
We don't want to continue using `fleet-system` as the default namespace for fleet components, this could cause conflict with user workloads running in the same namespace. We don't want to make it configurable either, we'd be updating it to `cattle-fleet-system` instead. 

Currently, this is set by `catalog.cattle.io/namespace` on fleet and fleet crd charts. 

Make sure upgrade works.

kinarashah:
@StrongMonkey mentioned that the `catalog.cattle.io` annotation comes into play when UI installs the chart. Rancher itself installs fleet via controller, so need to update the release namespace there too https://github.com/rancher/rancher/blob/ea531a0970c0726fa9789210a868ceaa95612e08/pkg/controllers/dashboard/fleetcharts/controller.go#L16

kinarashah:
Yeah, update both the places so we can get away completely from fleet-system name and make sure upgrade doesn't break.

sowmyav27:
Issue on upgrade: https://github.com/rancher/rancher/issues/33163

izaac:
Issue opened for Existing Downstream clusters with bundles installed: https://github.com/rancher/fleet/issues/422

izaac:
Needs a new Fleet RC to check the last issue: https://github.com/rancher/fleet/issues/422

izaac:
Validated and Closed: https://github.com/rancher/fleet/issues/422
Small issue for follow up: https://github.com/rancher/fleet/issues/442

PR #894, opened 2022-07-29
Consistenly use cattle-fleet-system as namespace

### LINKED ISSUE #754: Documentation mentions "old" namespaces, but not new ones (e.g. `cattle-fleet-system`)
Apparently the namespaces for fleet changed for 2.6: https://github.com/rancher/fleet/issues/395

Still when looking trough the documentation, there is not mention of e.g. `cattle-fleet-system`.

I believe the documentation has to be updated accordingly. Examples of places which may need to be updated

- https://fleet.rancher.io/namespaces
- https://fleet.rancher.io/troubleshooting/#check-the-status-of-the-fleet-controller
- https://fleet.rancher.io/uninstall/

There may be more

thardeck:
throughout the documentation.

To be consistent and prevent issues we also need to change the fleet-agent chart, because https://github.com/rancher/fleet/blob/master/pkg/namespace/util.go#L18-L24 generates `systemRegistrationNamespace` by search and replace.

So if someone updates their fleet standalone installation, which uses the old name fleet-system, it is necessary to pass systemNamespace and systemRegistrationNamespace with the old names:

```
helm -n fleet-system upgrade --install ...
  --set systemRegistrationNamespace="fleet-clusters-system" \
  ...
```

Fixes #754

### DIFF
--- a/.github/workflows/e2e-multicluster-ci.yml
+++ b/.github/workflows/e2e-multicluster-ci.yml
@@ -116,7 +116,7 @@ jobs:
           ca=$(kubectl get -n default secret "$name" -o go-template='{{index .data "ca.crt" | base64decode}}')
 
           kubectl config use-context k3d-downstream
-          helm -n fleet-system upgrade --install --create-namespace --wait fleet-agent charts/fleet-agent \
+          helm -n cattle-fleet-system upgrade --install --create-namespace --wait fleet-agent charts/fleet-agent \
             --set-string labels.env=dev \
             --set apiServerCA="$ca" \
             --set apiServerURL="https://172.18.0.1.omg.howdoi.website:6443" \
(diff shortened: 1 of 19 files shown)


mattfarina (commented):
Should the `e2e-multicluster-ci.yml` and `e2e-ci.yml` jobs be updated as well? So the tests that `helm install` and `helm upgrade` are using the same namespaces we have documented.

aiyengar2 (dismissed):
LGTM

docs/troubleshooting.md
```diff
@@ -21,7 +21,7 @@ Go to each downstream cluster and run the following command for the local cluste
 # Downstream cluster
 $ kubectl logs -f $fleet-agent-pod-name -n cattle-fleet-system
 # Local cluster
-$ kubectl logs -f $fleet-agent-pod-name -n cattle-local-fleet-system
+$ kubectl logs -f $fleet-agent-pod-name -n cattle-local-cattle-fleet-system
```
manno, line 24: This seems wrong? Not sure if the old namespace was a good choice either.
"cattle-fleet-system" should be right. 

While we are at it, I prefer `kubectl logs -l app=fleet-agent -n cattle-fleet-system`
 instead of letting the user guess the pod name.

thardeck: Good catch, it was a search and replace issue, thanks.

.github/workflows/e2e-multicluster-ci.yml
```diff
@@ -120,12 +120,11 @@ jobs:
           ca=$(kubectl get -n default secret "$name" -o go-template='{{index .data "ca.crt" | base64decode}}')
 
           kubectl config use-context k3d-downstream
-          helm -n fleet-system upgrade --install --create-namespace --wait fleet-agent charts/fleet-agent \
+          helm -n cattle-fleet-system upgrade --install --create-namespace --wait fleet-agent charts/fleet-agent \
             --set-string labels.env=dev \
             --set apiServerCA="$ca" \
             --set apiServerURL="https://172.18.0.1.omg.howdoi.website:6443" \
             --set clusterNamespace="fleet-local" \
-            --set systemRegistrationNamespace="fleet-clusters-system" \
```
manno, line 128: This is being deleted, because the default is fine?

thardeck: I have changed the default so no additional user interaction for new clusters is necessary.

charts/fleet-agent/values.yaml
```diff
@@ -25,11 +25,11 @@ clientID: ""
 clusterNamespace: ""
 
 # The namespace containing the clusters registration secrets
-systemRegistrationNamespace: fleet-clusters-system
+systemRegistrationNamespace: cattle-fleet-clusters-system
```
manno, line 28: This was necessary, because https://github.com/rancher/fleet/blob/master/pkg/namespace/util.go#L18-L24 generates this name. This is not a user facing namespace?

So if someone updates their fleet standalone installation they must pass `systemRegistrationNamespace` and `systemNamespace` and it should work.

manno (approved):
I like that this brings us in line with the rename in Rancher https://github.com/rancher/fleet/issues/395

However, this will affect fleet standalone users, so we should have some warning in the release notes.

manno (approved):

manno (approved):

PR #919, opened 2022-08-18
Update glob helm version docs

manno:


docs/gitrepo-add.md
```diff
@@ -1,17 +1,17 @@
 # Registering
 
 ## Proper namespace
-Git repos are added to the Fleet manager using the `GitRepo` custom resource type. The `GitRepo` type is namespaced. By default, Rancher will create two Fleet workspaces: **fleet-default** and **fleet-local**.
```
thardeck, line 4: It is ok here, but I think it is better to put cosmetic changes like the removal of trailing white-spaces in a separate commit to make the changes easier to verify/read.

thardeck (dismissed):

docs/gitrepo-add.md
manno: Yes, good point.

thardeck (approved):

PR #922, opened 2022-08-19
Bump helm to 3.9.3

thardeck:


scripts/version
```diff
@@ -4,13 +4,17 @@ if [ -n "$(git status --porcelain --untracked-files=no)" ]; then
...
 if [[ -z "$DIRTY" && -n "$GIT_TAG" ]]; then
     VERSION=$GIT_TAG
 else
-    VERSION="${COMMIT}${DIRTY}"
+    # Make sure tags are available in the CI environment
+    readarray -t TAGS < <(git tag -l)
+    if [[ ${#TAGS[@]} == 0 ]]; then
+        git fetch -t
+    fi
+    VERSION="$(git describe --tags)"
```
mattfarina, line 17: See the comment on the gitjob repo. This doesn't handle dirty situations anymore. The comment on that related PR suggests how to handle it.

mattfarina (changes requested):

scripts/version
thardeck: Should we get rid of https://github.com/rancher/gitjob/blob/master/scripts/version#L27-L31 ? It does not make much sense with the proper handling of the version.

mattfarina (approved):

scripts/version
mattfarina: Nope. I'm not entirely sure what this was for. But, if things are dirty it's "dev". I think this is all for local workflows.

PR #930, opened 2022-08-24
Skip pruneBundlesNotFoundInRepo in output mode

manno:
Before https://github.com/rancher/fleet/pull/764 `fleet apply -o` would
result in a dry run, that did not reach out to a cluster.
The prune PR added pruned objects to the output, but needs to list the
bundles to do so.

The tests in fleet-examples use the output mode and run without any
cluster. They [fail for fleet 0.3.10](https://github.com/rancher/fleet-examples/runs/7975883869?check_suite_focus=true).
My feeling is we don't want to require a cluster and printing to-be-pruned bundles in dry run was not the main intent of the previous PR.

After this PR, if fleet apply is in output (= dry run) mode, it won't
contact a cluster. However it can no longer output which bundles it
would prune.

I'd prefer it, if in the future dry run was more explicitly chosen, instead of relying on the output argument. Both features are independent of each other.

### DIFF
--- a/modules/cli/apply/apply.go
+++ b/modules/cli/apply/apply.go
@@ -106,9 +106,11 @@ func Apply(ctx context.Context, client *client.Getter, name string, baseDirs []s
 				return err
 			}
 
-			err = pruneBundlesNotFoundInRepo(name, gitRepoBundlesMap, client, opts)
-			if err != nil {
-				return err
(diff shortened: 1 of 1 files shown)


thardeck (approved):
As discussed with @aiyengar2 and @prachidamle, lgtm.

Issue #920, opened 2022-08-18
No bundles matching labels


dtrouillet:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When I deploy a GitRepo with two paths and one bundle as dependsOn on the second paths, he is not created.

The GitRepo Yaml with Status
 
```
apiVersion: fleet.cattle.io/v1alpha1
kind: GitRepo
metadata:
  annotations:
    meta.helm.sh/release-name: git-fleet-downstream-rancher-tt
    meta.helm.sh/release-namespace: default
    objectset.rio.cattle.io/id: default-git-fleet-downstream-rancher-tt-cattle-fleet-loca-9504d
  creationTimestamp: "2022-08-18T09:41:39Z"
  generation: 1
  labels:
    app.kubernetes.io/managed-by: Helm
    objectset.rio.cattle.io/hash: 7f842e5714b3f77a963bebcb07b609b23e787367
  managedFields:
  - apiVersion: fleet.cattle.io/v1alpha1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          .: {}
          f:meta.helm.sh/release-name: {}
          f:meta.helm.sh/release-namespace: {}
          f:objectset.rio.cattle.io/id: {}
        f:labels:
          .: {}
          f:app.kubernetes.io/managed-by: {}
          f:objectset.rio.cattle.io/hash: {}
      f:spec:
        .: {}
        f:branch: {}
        f:clientSecretName: {}
        f:insecureSkipTLSVerify: {}
        f:paths: {}
        f:repo: {}
        f:targets: {}
    manager: fleetagent
    operation: Update
    time: "2022-08-18T09:41:39Z"
  - apiVersion: fleet.cattle.io/v1alpha1
    fieldsType: FieldsV1
    fieldsV1:
      f:status:
        .: {}
        f:commit: {}
        f:conditions: {}
        f:desiredReadyClusters: {}
        f:display:
          .: {}
          f:error: {}
          f:message: {}
          f:readyBundleDeployments: {}
          f:state: {}
        f:gitJobStatus: {}
        f:lastSyncedImageScanTime: {}
        f:observedGeneration: {}
        f:readyClusters: {}
        f:resourceCounts:
          .: {}
          f:desiredReady: {}
          f:missing: {}
          f:modified: {}
          f:notReady: {}
          f:orphaned: {}
          f:ready: {}
          f:unknown: {}
[...]

PR #933, opened 2022-08-25
Fix gitrepos with multiple paths

manno:
Fixes https://github.com/rancher/rancher/issues/38672

Fleet 0.3.10 introduced bundle pruning. After debugging this for a while, it seems like the pruning code was too aggressive. When constructing multiple bundles from the same git repo, the loop would prune previously created bundles.

```
% cd fleet-examples
% fleet-0.3.10 apply --debug  --debug-level 9 multiple-paths single-cluster/manifests single-cluster/helm
INFO[0000] updated: fleet-local/multiple-paths-single-cluster-manifests
DEBU[0000] Bundle to be deleted since it is not found in gitrepo multiple-paths anymore fleet-local multiple-paths-single-cluster-helm
INFO[0000] created: fleet-local/multiple-paths-single-cluster-helm
DEBU[0000] Bundle to be deleted since it is not found in gitrepo multiple-paths anymore fleet-local multiple-paths-single-cluster-manifests
```
Afterwards one bundle would remain, but it doesn't have the 'fleet.cattle.io/repo-name' label anymore.

This PR moves pruning out of the loop, towards the end, after all bundles have been detected in all paths.

This also fixes https://github.com/rancher/fleet/issues/920, as the labels now remain intact.

### DIFF
--- a/docs/gitrepo-add.md
+++ b/docs/gitrepo-add.md
@@ -75,12 +75,13 @@ spec:
   # insecureSkipTLSVerify: true
   #
   # A git repo can read multiple paths in a repo at once.
-  # The below field is expected to be an array of paths and will include subdirs.
+  # The below field is expected to be an array of paths and
+  # supports path globbing (ex: some/*/path)
   #
   # Example:
   # paths:
   # - single-path
-  # - multiple-paths
+  # - multiple-paths/*
   paths:
   - simple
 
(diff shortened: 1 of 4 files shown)


thardeck (commented):
I would add the description from the pr to the commit so it is easier to follow later.

olblak (approved):

thardeck (approved):
lgtm

PR #944, opened 2022-08-29
feat: fixes #689 by supporting default values in case label does not exist

### LINKED ISSUE #689: invalid_label_reference for magic labels
We're seeing this error message in `fleet-controller` across multiple bundles after upgrading to Rancher v2.6.2.

This didn't use to happen and none of our objects have changed in Git.

```
level=error msg="error syncing 'fleet-local/velero-fleet-velero-config': handler bundle: invalid_label_reference global.fleet.clusterLabels.management.cattle.io/cluster-display-name in key clustername, requeuing"
```

Our `fleet.yaml`
``` yaml
helm:
  values:
    clustername: global.fleet.clusterLabels.management.cattle.io/cluster-display-name
```

Shaked:
Fix #689 

This pull request enables the usage of a default value for non existing labels in `fleet.yaml`. 

Currently in order to use `global.fleet.clusterLabels.someLabel`, you'd have to set `someLabel` in all of the clusters who work with that GitRepo. It is quite a hassle when many clusters exist and introduces a possibility for errors when there's no default value enabled. 

Instead, this PR introduces a new capability  - setting a default value by using a comma, e.g: `global.fleet.clusterLabels.someLabel, TheDefaultValueHere`.

## Test

```shell
cd pkg/target
go test
```

You could also test this by using a fleet.yaml, such as the one in the `target_test.go` file.


mattfarina:
@Shaked thanks for the pull request. It makes sense. We'll talk about it and circle back with you.

thardeck:
@Shaked thanks for your pr. Could you rebase against current master and fix the linting issue? Thanks in advance.

manno:
Apparently this is about using cluster labels for helm values. From the docs:

    values:
      any-custom: value
      # All labels on Rancher clusters are available using global.fleet.clusterLabels.LABELNAME
      # These can now be accessed directly as variables
      variableName: global.fleet.clusterLabels.LABELNAME

Looking at the source code, it seems there is additional, maybe undocumented functionality, which adds all cluster labels in a separate struct:

```
values:
  global:
    fleet:
      clusterLabels: {}
```


Now, using the former form, with the magic string `global.fleet.clusterLabels.LABELNAME`, fleet fails to deploy any bundle, if one of the clusters doesn't have that label. 

I'm hesitant to merge this PR, because I dislike putting two meanings in one field. `global.fleet.clusterLabels.LABELNAME, defaultValue` seems complicated. 
Is a default really necessary, wouldn't an empty string be closer to a users expectation?

Instead of returning an error, couldn't we just set an empty string in https://github.com/rancher/fleet/blob/master/pkg/target/target.go#L547 ?

Shaked:
Hey @mattfarina 

Sorry, I missed your message. I can do that. 

Hey @manno 

> I'm hesitant to merge this PR, because I dislike putting two meanings in one field. global.fleet.clusterLabels.LABELNAME, defaultValue seems complicated.

I am open for other suggestions. 

> Is a default really necessary, wouldn't an empty string be closer to a users expectation?

Unfortunately not. Empty string doesn't necessarily reflect what the chart is expecting for.

manno:
> Unfortunately not. Empty string doesn't necessarily reflect what the chart is expecting for.

I see, so the charts needs "a" value in that field.

There is also https://github.com/rancher/fleet/pull/671, a big PR which adds sprig templating, which could probably be used to achieve the same.

Shaked:
@manno 

> There is also https://github.com/rancher/fleet/pull/671, a big PR which adds sprig templating, which could probably be used to achieve the same.

#671 seems like a great PR. However, I am not sure that it will be as simple. Imagine that you only need to set default values for N amount of `global.fleet.clusterLabels.*` labels, how would that work? And will it be readable when N > 10 or 20? (@rajiteh maybe can also join this conversion).

manno:
> @manno
> 
> > There is also #671, a big PR which adds sprig templating, which could probably be used to achieve the same.
> 
> #671 seems like a great PR. However, I am not sure that it will be as simple. Imagine that you only need to set default values for N amount of `global.fleet.clusterLabels.*` labels, how would that work? And will it be readable when N > 10 or 20? (@rajiteh maybe can also join this conversion).

If there was templating for `fleet.yaml`, I thought  one could use template functions to provide defaults: 

```
      sprig: |-
          {{ default .ClusterLabels.env "empty" }}
      goIf: |-
          {{if .ClusterLabels.env}} {{.ClusterLabels.env}} {{else}} default {{end}}
```

We'll have to discuss this in our weekly meeting.

rajiteh:
@Shaked @manno 
Since we do have most of the sprig functions available to us, `default` should get the same job done without additional syntactic magic needing to be introduced to fleet yaml parsing. 

Part of the reason I proposed #671 was because I didn't feel strongly about the string matching nature of the current templating implementation. I fear it will be prone to mistakes.

manno:
> #671 seems like a great PR. However, I am not sure that it will be as simple. Imagine that you only need to set default values for N amount of `global.fleet.clusterLabels.*` labels, how would that work? And will it be readable when N > 10 or 20? (@rajiteh maybe can also join this conversion).

If there was templating for `fleet.yaml`, I thought  one could use template functions to provide defaults: 

```
      sprig: |-
          {{ default .ClusterLabels.env "empty" }}
      goIf: |-
          {{if .ClusterLabels.env}} {{.ClusterLabels.env}} {{else}} default {{end}}
```

We'll have to discuss this in our weekly meeting.

Shaked:
@manno @rajiteh 

I am definitely up for a better, more readable solution. It is just that templating is not readable as a definition file. 

It might be my personal opinion here but that dollar sign and curly braces make everything unreadable in scale - imagine having 40+ different labels. 

If we do end up with templating, maybe these magic values could be retrieved with a function, something alone the line: `fleet(clusterLabels.someLabel, TheDefaultValueHere)`. 

What do you think?

k0da:
> @manno @rajiteh
> 
> I am definitely up for a better, more readable solution. It is just that templating is not readable as a definition file.
> 
> It might be my personal opinion here but that dollar sign and curly braces make everything unreadable in scale - imagine having 40+ different labels.
> 
> If we do end up with templating, maybe these magic values could be retrieved with a function, something alone the line: `fleet(clusterLabels.someLabel, TheDefaultValueHere)`.
> 
> What do you think?

Yes, with template you can implement custom functions to satisfy usecase. I'm awaiting for template for a long time, without it it simply doesn't scale as labels are limited in length and characters you might use.

thardeck:
@Shaked  many thanks for your pr. We have recently merged the template pr #671 which should be able to cover this use case
Please let us know if it works as expected for your use case.
I will cut another Fleet alpha release to be able to easily test it.

kkaempf:
considered fixed, closing.

PR #951, opened 2022-09-02
Add missing doc for targetCustomizations.clusterName selector

bashofmann:
Add missing doc for targetCustomizations.clusterName selector

## Additional Information

See https://github.com/rancher/fleet/blob/master/pkg/apis/fleet.cattle.io/v1alpha1/bundle.go#L98

### DIFF
--- a/docs/gitrepo-structure.md
+++ b/docs/gitrepo-structure.md
@@ -173,6 +173,8 @@ targetCustomizations:
   clusterSelector:
     matchLabels:
       env: prod
+  # A selector used to match a specific cluster by name.    
+  clusterName: dev-cluster    
   # A selector used to match cluster groups.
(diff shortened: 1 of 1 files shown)


mattfarina:
@bashofmann CI is failing because it's trying to do something with `arm` and there are no BCI images for 32-bit ARM. We dropped that. This is because your base for the work is fairly old. Could you rabase from the upstreams master branch so you're up to date? That would fix CI.

bashofmann:
Done. I could have sworn that I did a `git pull` before I created the branch, but I guess I forgot about it in between two calls 😊 .

olblak (approved):

olblak:
This is just a markdown change, I am gonna force merge without waiting for e2e test and UT to be green

Issue #911, opened 2022-08-10
Fleet rejects files that have valid semver names (e.g. file-x.y.z+000.tgz)


coffeycan:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Bundle with dependencies throws an error:

Error while running post render on files: illegal name 'grafana-edge-dashboard-0.1.0+000.tgz' in file creation

### Expected Behavior

Fleet makes the repo with dependencies ready for deployment

### Steps To Reproduce

Create helm chart git repo with dependency on a subchart with version 1.0.0+000 
Run helm dependency update on a clone of the git repo
add / commit / push the charts subdir with dependent charts

### Environment

```markdown
- Architecture:arm64
- Fleet Version:Rancher 2.6.5
- Cluster:
  - Provider:K3s
  - Options:Single node Rancher install
  - Kubernetes Version:1.21
```

### Anything else?

This seems related to Issue rancher/fleet#156 , however is especially troublesome as it requires re-tooling all the sub-charts that use a valid semver versioning scheme...

p-se:
I've been able to reproduce the issue and it is unfortunately _not_ fixed by #964.

kkaempf:
Fixed in 0.6.0

Issue #156, opened 2020-11-19
Allow use of '+' character in kustomize directory name


ron1:
Use of a `+` character in a kustomize directory name causes `fleet test` to generate the following error:
```
FATA[0000] error while running post render on files: illegal name 'dev+test' in directory creation
```

Steps to reproduce the problem:
1. Apply [patch](https://github.com/ron1/fleet-examples/commit/4cc4e9252d0f61c5b82b7f41e6daf4cdf6ae2cfb) to fleet-examples
2. Change into patched `fleet-examples/multi-cluster/kustomize` directory
3. Execute command `fleet test -t dev`

Expected result: yaml manifests are displayed
Actual result: the above error is displayed

hyunsun:
I also ran into the same problem with files containing "@" character in the name, which is unavoidable in YANG modeling.

Mo9a7i:
no fix yet ?

PR #964, opened 2022-09-09
Fix long bundle names

### LINKED ISSUE #523: fleet-agent cleanup continually deleting releases
The cleanup loop in `fleet-agent` isn't able to properly identify releases with bundles. For a given release, it looks for the bundle deployment associated to that release. If the release has a different name than fleet-agent expects (based on the bundle deployment) then fleet-agent deletes the release.

The problem is that then a release with the exact same name is created and the next time the cleanup loop runs, fleet-agent will delete the release again.

For example, a bundle deployment with name `mcc-anupamafinalrcrke2-managed-system-upgrade` creates a release with name `mcc-anupamafinalrcrke2-managed-system-upgrade-c-407d2`. 

It seems that [this function](https://github.com/rancher/fleet/blob/master/modules/agent/pkg/deployer/manager.go#L42) should look at `status.release` as well when trying to determine the name of the release from a bundle deployment.

manno:
Fix #523

The bundle name is used in several places:
* pruning bundles when running "fleet apply"
* finding bundles when fleet controller runs CleanUp
* as a default for the helm release name
* as the name of the bundledeployment (in the clusters namespace)

The bundle name usually consists of the "gitrepo" name and the path in which the bundle was found.

Helm allows for release names of up to 53 characters. When creating the bundle name from git "repoName" + "path to bundle", fleet now limits the lenght to 53. It will also replace all disallowed characters.

If the user specifies a custom release name in the helm options (via fleet.yaml), that will be validated, too. JSON schema was added to the CRD.

### Tradeoff

Old bundle names should remain stable, so they are not reinstalled. Unless the old names are too long, however in that case they are in danger of a collision or being falsely targeted by fleet controller's cleanup.


modules/cli/apply/apply.go
```diff
@@ -167,10 +174,46 @@ func readBundle(ctx context.Context, name, baseDir string, opts *Options) (*bund
...
+	}
+
+	// clean up new name
+	str = strings.ToLower(str)
+	str = disallowedChars.ReplaceAllLiteralString(str, "-")
+	str = multiDash.ReplaceAllString(str, "-")
+	str = strings.TrimLeft(str, "-")
+	str = strings.TrimRight(str, "-")
+
+	// shorten name to 53 characters, the limit for helm release names
```
thardeck, line 208: I would extend the comment that the Limit function of wrangler adds a checksum to prevent collisions itself.

thardeck (approved):
LGTM besides the nitpick, but it is not mandatory.

thardeck:
And thanks for the many additional comments to improve code readability. :bouquet:

manno:
I think this also fixes https://github.com/rancher/fleet/issues/196, https://github.com/rancher/fleet/issues/911 and https://github.com/rancher/fleet/issues/156

PR #968, opened 2022-09-12
Remove dead code (controllermanifest)

manno:
Remove dead code in controllermanifest package.

Also remove the cryptic note.


moio:
+1

to my understanding cfbe1dce4e6a5d836ff949b2eee33dbde4da79ec is the commit which removed the last reference.

Apparently the Fleet CLI had subcommands to generate Helm charts for installation before that point - afterwards the charts are just maintained in the repo and need no in-app generation.

thardeck (approved):

PR #975, opened 2022-09-15
Add basic support for OCI registry based charts

thardeck:
until go-getter has native support we are using Helm for now to download the charts from OCI registries.

Authentication for the OCI registry is not supported yet but should be possible via Helm.


manno (approved):
You could add info about the `version:` field to the docs, but doesn't have to be in this PR.

manno (approved):

PR #983, opened 2022-09-16
Adds priorityClassName to deployments.

KevinJoiner:
Related to https://github.com/rancher/rancher/issues/37927

The purpose of this PR is to add the ability to set a priorityClassName for fleet pods in the rancher cluster.

### Test
- Create a new Kubernetes cluster.
- Helm install with the following options
 ` --set priorityClassName=system-cluster-critical --set gitjob.priorityClassName=system-cluster-critical`
 - Describe fleet and gitjob pods using `kubectl`
 - Verify that both pods have "system-cluster-critical" set as their priorityClassName in the pod spec

### DIFF
--- a/charts/fleet/charts/gitjob/templates/deployment.yaml
+++ b/charts/fleet/charts/gitjob/templates/deployment.yaml
@@ -43,3 +43,6 @@ spec:
 {{- if .Values.tolerations }}
 {{ toYaml .Values.tolerations | indent 8 }}
 {{- end }}
+      {{- if .Values.priorityClassName }}
+      priorityClassName: "{{.Values.priorityClassName}}"
+      {{- end }}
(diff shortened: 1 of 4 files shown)


cbron (approved):
lgtm, wait for a fleet review to merge. This should go into 2.7, but since it's backwards compatible if it does end up going into 2.6 that should be fine as well.

manno:
Would be interested to know why this is needed.

> We need to ensure Rancher pods are always scheduled ahead of other pods in the management cluster.

Is there a situation where it's a problem if fleet comes up.. later?

KevinJoiner:
@manno we care less about if fleet comes up later than if fleet comes up at all. If a user runs other pods with a priority level set on a node with limited resources, we may never be able to run fleet because its priority defaults to 0.

manno:
## Issue: https://github.com/rancher/rancher/issues/37927
 
## Problem

Fleet might not start if other pods have higher priority

## Solution

Add `priorityClassName` to helm's values.
 
## Testing

Deploy with a priority class, make sure it appears on fleet and gitjob.

Issue #948, opened 2022-08-31
Feature Request: Cluster specific registry that the agent image is pulled from


mattfarina:
### Is your feature request related to a problem?

Yes. Rancher has the ability to specify cluster specific registries. These are useful in a variety of ways. When fleet is used with Rancher the cluster specific registry is not passed on for fleet to use. This is a problem when Rancher relies on fleet and it doesn't work with the cluster specific registry.

### Solution you'd like

1. To have the fleet Cluster CRD extended to add a registry
2. When fleet installs an agent in that cluster the registry on the Cluster CR is used for the image to install

### Alternatives you've considered

N/A

### Anything else?

This is a sub-task of the epic at [#712](https://github.com/rancher/fleet/issues/712)

This should not be closed until Rancher is updated to use the change. The change, I believe, needs to be made at [here](https://github.com/rancher/rancher/blob/3e5fb4a537731df7ed96cc9ddae6a3902a57c8b1/pkg/controllers/provisioningv2/fleetcluster/fleetcluster.go#L118-L129). I think you can use the function [`GetPrivateRepoURLFromCluster`](https://github.com/rancher/rancher/blob/3e5fb4a537731df7ed96cc9ddae6a3902a57c8b1/pkg/provisioningv2/image/resolve.go#L34) to get the private repo from the cluster instance.

mattfarina:
This is half implemented. Fleet now supports cluster specific registries via the [privateRepoUrl](https://github.com/rancher/fleet/blob/d33b1f3cf3e827ec808b5eedd29e5ca6a4d70380/charts/fleet-crd/templates/crds.yaml#L1782) (also [here](https://github.com/rancher/fleet/blob/d33b1f3cf3e827ec808b5eedd29e5ca6a4d70380/pkg/apis/fleet.cattle.io/v1alpha1/target.go#L70)) property.

The [PR for Rancher to leverage this](https://github.com/rancher/rancher/pull/38950) did not make it in for 2.7.0 so we are targeting 2.7.1.

mattfarina:
Rancher Manager now works with credentials for private repos. We need to support them. This was fixed in https://github.com/rancher/rancher/pull/39059. It was a bug because the credentials were input but not being used properly.

To get the name of the secret we should be able to use the function `GetPrivateRepoSecretFromCluster`. It's in the same package as `GetPrivateRepoURLFromCluster`. It's new.

[...]

PR #984, opened 2022-09-16
Support cluster specific registry

manno:
Refers to #948


thardeck (approved):
Codewise it looks good to me. Thanks for the refactoring, much easier to comprehend now.

Issue #760, opened 2022-05-06
fleet-agent has high CPU usage when guest POD is in Terminating


w13915984028:
Environment:

```
KVM based 2-node Harvester (v1.0.1) cluster, one control node, one compute node

Fleet-controller,agent:
images:
docker.io/rancher/fleet-agent:v0.3.9
docker.io/rancher/fleet:v0.3.9 (fleet-controller)

charts:
/charts/rancher-monitoring-crd-100.1.0+up19.0.3.tgz
/charts/rancher-monitoring-100.1.0+up19.0.3.tgz

They are integrted in Harvester:
https://github.com/harvester/harvester/releases
The latest ISO:
https://releases.rancher.com/harvester/v1.0.2-rc1/harvester-v1.0.2-rc1-amd64.iso

The installation guide (KVM/vmware based VM is fine to hold a sinlge-node cluster) :
https://docs.harvesterhci.io/v1.0/install/iso-install/
```

Steps:
1. Install a new cluster with 2 nodes
2. Power off the KVM VM which runs the compute node 
3. kubectl shows a couple of PODs are in Terminating

```
cattle-system               rancher-598f5dbc8-htf4m                                  1/1     Terminating   1          32m
harvester-system            harvester-697fb5d565-5822h                               1/1     Terminating   1          32m
harvester-system            harvester-network-controller-manager-99677f8fd-l456c     1/1     Terminating   1          32m
harvester-system            harvester-webhook-584f99698c-kj8sc                       1/1     Terminating   1          32m
kube-system                 rke2-coredns-rke2-coredns-687554ff58-kj6ld               1/1     Terminating   1          32m
longhorn-system             instance-manager-e-b624ea55                              1/1     Terminating   0          12m
longhorn-system             instance-manager-r-0d880b77                              1/1     Terminating   0          12m
```

```
cattle-system               rancher-598f5dbc8-htf4m                                  1/1     Terminating   1          42m
harvester-system            harvester-697fb5d565-5822h                               1/1     Terminating   1          42m
harvester-system            harvester-network-controller-manager-99677f8fd-l456c     1/1     Terminating   1          42m
harvester-system            harvester-webhook-584f99698c-kj8sc                       1/1     Terminating   1          42m
kube-system                 rke2-coredns-rke2-coredns-687554ff58-kj6ld               1/1     Terminating   1          42m
longhorn-system             instance-manager-e-b624ea55                              1/1     Terminating   0          22m
[...]

PR #990, opened 2022-09-19
Remove ignored fields from status struct, to avoid loop

manno:
The fields are ignored in the JSON struct, however it would still trigger a status update, which would trigger another MonitorBundle call. The reason behind that is, that the structs are different, even though the marshalled JSON is not.

Refers to https://github.com/rancher/fleet/issues/760

### DIFF
--- a/modules/agent/pkg/controllers/bundledeployment/controller.go
+++ b/modules/agent/pkg/controllers/bundledeployment/controller.go
@@ -212,6 +212,15 @@ func (h *handler) cleanupOldAgent(modifiedStatuses []fleet.ModifiedStatus) error
 	return merr.NewErrors(errs...)
 }
 
+// removePrivateFields removes fields from the status, which won't be marshalled to JSON.
+// They would however trigger a status update in apply
+func removePrivateFields(s1 *fleet.BundleDeploymentStatus) {
+	for id := range s1.NonReadyStatus {
+		s1.NonReadyStatus[id].Summary.Relationships = nil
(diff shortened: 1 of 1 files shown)


thardeck (approved):

manno:
## Issue: #760 

## Problem

Fleet would run into a dead loop and degrade performance.
 
## Solution

We exclude fields from the status, which were not marshalled to JSON. However they did affect the result of `deepEqual` and trigger an update.
By removing these fields before checking for equality, we make sure the status doesn't get updated if it is unchanged.
 
## Testing

Try to produce the error condition experienced by harvester team: https://github.com/w13915984028/harvester-develop-summary/blob/main/fleet-agent-high-cpu-usage.md

1.  Install a new cluster with 2 nodes
2. Power off the KVM VM which runs the compute node

You should see an error like this in the status of the bundledeployment:

```
  - apiVersion: apps/v1
    kind: Deployment
    name: redis-master
    namespace: fleet-helm-example
    summary:
      message:
      - Deployment does not have minimum availability.
      - 'Available: 0/1'
      state: updating
      transitioning: true
```
 
3. Observe the cluster has high load
4. Update to a fixed version of fleet and see the cpu load go down


## Engineering Testing
### Manual Testing

The fix was confirmed to work by the Harvester team.

### Automated Testing

## QA Testing Considerations

### Regressions Considerations

Nothing was removed from the status message, so I don't expect any regressions.

PR #995, opened 2022-09-22
Fix token expiry bug

manno:
The clusterregistrationtoken has an expiry time. Due to an error the time was always reset to the current time. The fleet-controller had high cpu and kept reconciling the token.

### DIFF
--- a/pkg/controllers/clusterregistrationtoken/handler.go
+++ b/pkg/controllers/clusterregistrationtoken/handler.go
@@ -223,6 +223,6 @@ func (h *handler) deleteExpired(token *fleet.ClusterRegistrationToken) (bool, er
 		return true, h.clusterRegistrationTokens.Delete(token.Namespace, token.Name, nil)
 	}
 
-	h.clusterRegistrationTokens.EnqueueAfter(token.Namespace, token.Name, time.Until(time.Now()))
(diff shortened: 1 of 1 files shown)


thardeck (approved):

olblak (approved):

manno:
## Issue: 

No issue, this was found during development.
 
## Problem

The presence of a clusterregistrationtokens would lead to a higher than normal cpu load of the fleet-controller pod.
Fleet constantly re-checks if an expired token should be deleted. 

This bug was introduced while applying linter fixes. 

## Solution

Switch back to the old code, use the correct expiry time when enqueuing events.
 
## Testing

Create a clusterregistrationtoken and observe the load

```
kubectl apply -f - <<EOF
apiVersion: "fleet.cattle.io/v1alpha1"
kind: ClusterRegistrationToken
metadata:
  name: second-token
  namespace: fleet-local
spec:
  ttl: 12h
EOF
```


## Engineering Testing
### Manual Testing

Creating a token, observing cpu in k9s.

### Automated Testing

## QA Testing Considerations
 
### Regressions Considerations

izaac:
Reproduced on fleet `v0.4.0`

High CPU usage after adding a `ClusterRegistrationToken`

Screen Shot 2022-10-21 at 1 10 35 PM


This is not reproducible on fleet `v0.5.0-rc3`

Screen Shot 2022-10-21 at 1 11 37 PM

Issue #385, opened 2021-05-21
Fleet RBAC


Jasstkn:
Hello! I'm trying to configure restricted access to deployment via Fleet. 
Case: user has an owner permissions in the specific cluster, e.g. sandbox. But they can't see the Continuous Delivery tab with this role. How can I grant specific access to the Fleet API that user will be able to deploy only to allowed cluster?

abelnieva:
same issue here

markus-obstmayer:
I have also the same issue.

tonyjsolano:
I would like to know about this as well

jleni:
Any news about this?

Hryhorii-Tatsyi:
same issue

Baldro:
same issue also for us

ArtourK:
This is a showstopper for us! 
We can't provide Admin rights to the users in Rancher, that need to manage fleet for some cluster.

markus-obstmayer:
same here. It´s also a showstopper for use.

ADustyOldMuffin:
I'd also like to jump in and say this is a huge issue, we use Rancher to let consumers of our clusters deploy apps and the fact they can't see the status of them is terrible.

ADustyOldMuffin:
A quick update, if you're using Rancher you can grant this by using `GlobalRole`'s

Example.
```
apiVersion: management.cattle.io/v3
builtin: false
description: Used to view and access Continuous Delivery in Rancher.
displayName: FleetAccess
kind: GlobalRole
newUserDefault: false
metadata:
  generateName: gr-
rules:
- apiGroups:
  - fleet.cattle.io
  resources:
  - bundles
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - fleet.cattle.io
  nonResourceURLs: []
  resourceNames: []
  resources:
  - bundledeployments
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - fleet.cattle.io
  nonResourceURLs: []
  resourceNames: []
  resources:
  - clusters
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - fleet.cattle.io
  nonResourceURLs: []
  resourceNames: []
  resources:
  - gitrepos
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - management.cattle.io
  nonResourceURLs: []
  resourceNames: []
  resources:
  - fleetworkspaces
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - fleet.cattle.io
[...]

PR #1008, opened 2022-09-27
Add allowedTargetNamespaces to GitRepoRestriction

manno:
Add another field to GitRepoRestrictions, to limit GitRepo deployments to specific targetNamespaces.
When a GitRepoRestrictions has `allowedTargetNamespaces`, all created Gitrepos in the same namespace will be evaluated against the list of allowed namespaces. The deployment will fail if cluster-wide resources are included, or the GitRepo uses a disallowed targetNamespace.

Refers to #385

### DIFF
--- a/charts/fleet-crd/templates/crds.yaml
+++ b/charts/fleet-crd/templates/crds.yaml
@@ -2545,6 +2545,12 @@ spec:
               type: string
             nullable: true
             type: array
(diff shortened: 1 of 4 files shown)


thardeck (approved):

manno:
QA template in issue: https://github.com/rancher/fleet/issues/385#issuecomment-1342856297

PR #1010, opened 2022-09-30
Update quickstart.md

jsburckhardt:
the path is correct, but the binaries don't have a 'v' e.g. `0.3.11` instead of `v0.3.11`. I'm hard coding the version, but is better to update it in your  cd. Cheers

Fix #XXX


mattfarina:
The fleet docs have been moved to https://github.com/rancher/fleet-docs. We're going to be removing them from the tree here (see #1014). This problem seems to exist in the new site which is something I think we can work out over there. See https://fleet.rancher.io/quickstart.

I filed an issue to fix the things over at https://github.com/rancher/fleet-docs/issues/5

PR #1011, opened 2022-09-30
Helm Target Customization Repo/Version Override

### LINKED ISSUE #699: TargetCustomizations -> Helm -> Version not taken in account
First, I like the idea behind fleet a lot. Good job so far!

Having the possibility to update the helm chart version for targeted cluster makes a lot of sense.
One use-case is to test the release upgrade on one cluster and then if this works ok to do the upgrade for all other clusters.

I can notice that  pkg/options/calculate.go takes a lots of parameters in consideration but not the "Version". 

If you think that this can work and this file is the only place that needs to be changed I can create a PR.

### LINKED ISSUE #899: [Question] Helm chart version override in fleet.yaml
Is it possible to override the version of a helm chart with a target customization? I have the following fleet.yaml file and the cluster that has the version change is still using the global version. Reading the documentation on fleet.yaml makes it seem like this is possible, but looking through the code, it seemed to me that it wasn't implemented to support that. I noticed some additional handling  [here](https://github.com/rancher/fleet/blob/4a771da9d42bc1fade1e4f41a9dce7ce85eccf12/pkg/bundle/resources.go#L55), but adding the repo and chart properties to the helm target customization caused the deploy to fail before generating any bundles.

```yaml
defaultNamespace: app
namespace: app

helm: 
  repo: https://registry.company.com/chartrepo/dev
  chart: app
  version: 1.0.0
  values:
    resources:
      enable: true
targetCustomizations:
  - name: cluster-a
    helm:
      valuesFiles:
        - cluster-a/values.yaml
    yaml:
      overlays:
        - cluster-a
    clusterSelector:
      matchLabels:
        cluster-name: cluster-a

  - name: cluster-b
    helm:
      version: 1.1.0
      valuesFiles:
        - cluster-b/values.yaml
    yaml:
      overlays:
        - cluster-b
    clusterSelector:
      matchLabels:
        cluster-name: cluster-b
  
```

romejoe:
Fix #699
Fix #899

This commit fixes an issue stopping versions and repos in Helm specs from being overridden in target customizations. It applies root repos and versions to customizations where appropriate to allow targets to  properly resolve the correct target.

## Test

To test this pull request, I used the following configurations:
[https://raw.githubusercontent.com/romejoe/fleet-test/main/multi-chart-test/fleet.yaml](https://raw.githubusercontent.com/romejoe/fleet-test/main/multi-chart-test/fleet.yaml)
[https://raw.githubusercontent.com/romejoe/fleet-test/main/version-test/fleet-config/fleet.yaml](https://raw.githubusercontent.com/romejoe/fleet-test/main/version-test/fleet-config/fleet.yaml)

The first test verifies that if the entire helm reference(repo, chart and version) are specified in a target customization, the proper chart is deployed to those targets.

The second test verifies that if only a version is specified, the updated version will be used. It also verifies that if the chart is overridden, the appropriate chart is still used.

### Tradeoff
The only potential trade off is that if the user specifies a repo, chart and version in the root helm spec and then only overrides the chart property in a customization, the target customization will treat the chart as a relative path. The only time this could raise an issue is if the user changes the chart in the customization with the intent of selecting a different chart within the same repo.  This is probably out of scope though. The user is still able to do that, they just have to supply at least a repo or version as well in the customization.

### DIFF
--- a/pkg/bundle/read.go
+++ b/pkg/bundle/read.go
@@ -186,6 +186,8 @@ func read(ctx context.Context, name, baseDir string, bundleSpecReader io.Reader,
 
 	setTargetNames(&bundle.BundleSpec)
 
+	propagateHelmChartProperties(&bundle.BundleSpec)
+
 	resources, err := readResources(ctx, &bundle.BundleSpec, opts.Compress, baseDir, opts.Auth)
 	if err != nil {
 		return nil, err
@@ -245,6 +247,29 @@ func read(ctx context.Context, name, baseDir string, bundleSpecReader io.Reader,
 	}
 
 	return New(def, scans...)
+}
+
+// propagateHelmChartProperties propagates root Helm chart properties to the child targets.
+func propagateHelmChartProperties(spec *fleet.BundleSpec) {
+	// Check if there is anything to propagate
+	if spec.Helm == nil {
+		return
+	}
+	for _, target := range spec.Targets {
+		if target.Helm == nil {
+			// This target has nothing to propagate to
+			continue
+		}
+		if target.Helm.Repo == "" {
+			target.Helm.Repo = spec.Helm.Repo
+		}
+		if target.Helm.Chart == "" {
+			target.Helm.Chart = spec.Helm.Chart
+		}
+		if target.Helm.Version == "" {
+			target.Helm.Version = spec.Helm.Version
+		}
+	}
 }
 
 func appendTargets(def *fleet.Bundle, targetsFile string) (*fleet.Bundle, error) {
--- a/pkg/bundle/resources.go
+++ b/pkg/bundle/resources.go
@@ -37,6 +37,8 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
+var hasOCIURL = regexp.MustCompile(`^oci:\/\/`)
+
 func readResources(ctx context.Context, spec *fleet.BundleSpec, compress bool, base string, auth Auth) ([]fleet.BundleResource, error) {
 	var directories []directory
 
@@ -103,6 +105,11 @@ type Auth struct {
 }
 
 func chartURL(location *fleet.HelmOptions, auth Auth) (string, error) {
+	// repos are not supported in case of OCI Charts
+	if hasOCIURL.MatchString(location.Chart) {
+		return location.Chart, nil
+	}
+
 	if location.Repo == "" {
 		return location.Chart, nil
 	}
@@ -314,11 +321,7 @@ func readContent(ctx context.Context, base, name, version string, auth Auth) (ma
 	// go-getter does not support downloading OCI registry based files yet
 	// until this is implemented we use Helm to download charts from OCI based registries
 	// and provide the downloaded file to go-getter locally
-	hasOCIURL, err := regexp.MatchString(`^oci:\/\/`, name)
-	if err != nil {
-		return nil, err
-	}
-	if hasOCIURL {
+	if hasOCIURL.MatchString(name) {
 		src, err = downloadOCIChart(name, version, temp, auth)
 		if err != nil {
 			return nil, err
--- a/e2e/assets/multi-cluster/helm-target-customizations.yaml
+++ b/e2e/assets/multi-cluster/helm-target-customizations.yaml
@@ -1,0 +1,14 @@
+kind: GitRepo
+apiVersion: fleet.cattle.io/v1alpha1
+metadata:
+  name: helm-target-customizations
+spec:
+  repo: https://github.com/rancher/fleet-examples
+  branch: add_target_customizations_test
+  paths:
+  - multi-cluster/helm-target-customizations/
+  targets:
+  - name: test
+    clusterSelector:
+      matchLabels:
+        env: test
--- a/e2e/multi-cluster/multi_cluster_test.go
+++ b/e2e/multi-cluster/multi_cluster_test.go
@@ -19,6 +19,24 @@ var _ = Describe("Multi Cluster Examples", func() {
 		kd    kubectl.Command
 	)
 
+	helmVersion := func(repo string) (string, error) {
+		out, err := k.Get("bundledeployments", "-A", "-l", "fleet.cattle.io/repo-name="+repo, "-o=jsonpath={.items[*].spec.options.helm.version}")
+		if err != nil {
+			return "", err
+		}
+
+		return out, nil
+	}
+
+	helmRepo := func(repo string) (string, error) {
+		out, err := k.Get("bundledeployments", "-A", "-l", "fleet.cattle.io/repo-name="+repo, "-o=jsonpath={.items[*].spec.options.helm.repo}")
+		if err != nil {
+			return "", err
+		}
+
+		return out, nil
+	}
+
 	BeforeEach(func() {
 		k = env.Kubectl.Context(env.Fleet).Namespace(env.Namespace)
 		kd = env.Kubectl.Context(env.Downstream)
@@ -109,5 +127,31 @@ var _ = Describe("Multi Cluster Examples", func() {
 			})
 		})
 
+		Context("containing an external helm chart with targetCustomizations", func() {
+			BeforeEach(func() {
+				asset = "multi-cluster/helm-target-customizations.yaml"
+			})
+
+			It("can replace the chart version and url", func() {
+				expectedVersion := "0.0.36"
+
+				// Verify bundledeployment changes
+				Eventually(func() string {
+					out, _ := helmVersion("helm-target-customizations")
+					return out
+				}).Should(ContainSubstring(expectedVersion))
+
+				Eventually(func() string {
+					out, _ := helmRepo("helm-target-customizations")
+					return out
+				}).Should(ContainSubstring("https://charts.truecharts.org///"))
+
+				// Verify actual deployment downstream
+				Eventually(func() string {
+					out, _ := kd.Get("deployments", "-A", "-l", "helm.sh/chart=radicale-"+expectedVersion)
+					return out
+				}).Should(ContainSubstring("radicale"))
+			})
+		})
 	})
 })
--- a/pkg/options/calculate.go
+++ b/pkg/options/calculate.go
@@ -67,8 +67,14 @@ func merge(base, next fleet.BundleDeploymentOptions) fleet.BundleDeploymentOptio
 		if next.Helm.ValuesFrom != nil {
 			result.Helm.ValuesFrom = append(result.Helm.ValuesFrom, next.Helm.ValuesFrom...)
 		}
+		if next.Helm.Repo != "" {
+			result.Helm.Repo = next.Helm.Repo
+		}
 		if next.Helm.Chart != "" {
 			result.Helm.Chart = next.Helm.Chart
+		}
+		if next.Helm.Version != "" {
+			result.Helm.Version = next.Helm.Version
 		}
 		if next.Helm.ReleaseName != "" {
 			result.Helm.ReleaseName = next.Helm.ReleaseName


thardeck:
Thanks for your pr. We try to have automated tests for every new feature, so I have created one for you this time: https://github.com/romejoe/fleet/pull/1

The tested example fleet.yaml is available in the [fleet-example repo](https://github.com/rancher/fleet-examples/blob/add_target_customizations_test/multi-cluster/helm-target-customizations/fleet.yaml).

I will review your code shortly.

romejoe:
@thardeck ok, thanks! I didn't realize that. I'll remember that for the future and I'll update my other pr to make sure it is properly tested.

romejoe:
@thardeck Thanks! That's super helpful! Up till now I've just been testing on my home lab with separate RKE2 installs on separate vms...

pkg/bundle/read.go
```diff
@@ -247,6 +250,39 @@ func read(ctx context.Context, name, baseDir string, bundleSpecReader io.Reader,
...
+				if target.Helm.Repo == "" {
+					target.Helm.Repo = spec.Helm.Repo
+				}
+				if target.Helm.Version == "" {
+					target.Helm.Version = spec.Helm.Version
+				}
+			}
+
+			if target.Helm.Chart == "" {
+				target.Helm.Chart = spec.Helm.Chart
```
thardeck, line 276: This step should be obsolete because it is already handled in `calculate.go`.

pkg/bundle/read.go
```diff
@@ -247,6 +250,39 @@ func read(ctx context.Context, name, baseDir string, bundleSpecReader io.Reader,
...
 }
 
+// fixAndApplyHelmSpec checks if the spec's Helm repo, if there is one, ends with a '/' and appends one if not. See chartURL
+// This function also propagates root Helm chart properties to the child targets and performs the repo check.
+// Note: The repo and version properties of child targets are only propagated if either property is overridden. This
+// allows child targets to specify a path to a chart, to specify a url to the chart(e.g. oci://), etc.
+func fixAndApplyHelmSpec(spec *fleet.BundleSpec) {
+	if spec.Helm != nil && spec.Helm.Chart != "" {
+		if spec.Helm.Repo != "" && !strings.HasSuffix(spec.Helm.Repo, "/") {
+			spec.Helm.Repo = spec.Helm.Repo + "/"
```
thardeck, line 260: `spec.Helm.Repo` should already have a `/` added by `chartURL`, so I suppose only `target.Helm.Repo` would need that logic, but maybe we should fix the requirement for a `/`  instead if it is really mandatory for the rest of the fleet code.

thardeck (changes requested):
After reading through `fixAndApplyHelmSpec`, it seems the only requirement might be adding the `/` for `target.Helm.Repo`, if Fleet can not properly handle Helm repos without a `/` in the end.
But in this case we should fix the handling in Fleet instead of just adding a `/`.

I can remove the function `fixAndApplyHelmSpec` without my tests for the version and repo targetCustomizations to fail because the changes are already propagated through your code additions to `calculate.go`.

If a relevant `targetCustomization` use case is not covered by the `merge` function could you explain it a little more so we can add another test for it?

pkg/bundle/read.go
```diff
@@ -186,6 +187,8 @@ func read(ctx context.Context, name, baseDir string, bundleSpecReader io.Reader,
 
 	setTargetNames(&bundle.BundleSpec)
 
+	fixAndApplyHelmSpec(&bundle.BundleSpec)
```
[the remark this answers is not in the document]
romejoe: I agree, I like that new function name. I'll make that change, if that function ends up staying.

pkg/bundle/read.go
```diff
@@ -247,6 +250,39 @@ func read(ctx context.Context, name, baseDir string, bundleSpecReader io.Reader,
 	return New(def, scans...)
 }
 
+// fixAndApplyHelmSpec checks if the spec's Helm repo, if there is one, ends with a '/' and appends one if not. See chartURL
+// This function also propagates root Helm chart properties to the child targets and performs the repo check.
+// Note: The repo and version properties of child targets are only propagated if either property is overridden. This
+// allows child targets to specify a path to a chart, to specify a url to the chart(e.g. oci://), etc.
+func fixAndApplyHelmSpec(spec *fleet.BundleSpec) {
+	if spec.Helm != nil && spec.Helm.Chart != "" {
```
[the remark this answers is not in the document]
romejoe: I agree. That function was originally more complicated and that was my attempt to simplify it a bit. I think my brain was just a bit fried at 4am when I wrapped these changes up and I missed that.

pkg/bundle/read.go
romejoe: I tried to use this code and accompioning `if` at the end of the for loop to standardize the repo option with '/' suffix throughout Fleet, so that the logic for verifying the repo ended in '/' for `spec.Helm.Repo` and `target.Helm.Repo` was all in one place. Although that can definitely be made more clear. If that stays, it should probably be extracted to a separate code block or function.

Would the user be required in the `fleet.yaml` to always end the repo in a '/'?

pkg/bundle/read.go
romejoe: I think my logic here was by doing the merge here, any place that needed the helm options would be guaranteed to have the finalized helm options. Looking at it now, I think I agree that it's obsolete. I think the above if statement just needs to be added to `calculate.go`

pkg/bundle/read.go
```diff
@@ -247,6 +250,39 @@ func read(ctx context.Context, name, baseDir string, bundleSpecReader io.Reader,
...
+		if spec.Helm.Repo != "" && !strings.HasSuffix(spec.Helm.Repo, "/") {
+			spec.Helm.Repo = spec.Helm.Repo + "/"
+		}
+	}
+
+	for _, target := range spec.Targets {
+		if target.Helm != nil && spec.Helm != nil {
+			if target.Helm.Repo != "" || target.Helm.Version != "" {
```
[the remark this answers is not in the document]
romejoe: I think this block of code was trying to handle a use case like the first target customization below:
```
namespace: fleet-mc-helm-external-example
helm:
  repo:  https://github.com/rancher/fleet-examples/releases/download/example/
  chart: guestbook
  version: 0.0.0
targetCustomizations:
- name: dev
  helm:
    chart: "./guestbook-dev/"
  clusterSelector:
    matchLabels:
      env: dev

- name: test
  helm:
    chart: oci://docker.github.com/rancher/fleet-examples/guestbook
    version: 0.1.0-test
  clusterSelector:
    matchLabels:
      env: test
```
I'm not sure if the dev customization is a valid, but I think that's what I was trying to account for.

On second look though, I am noticing that the code above will cause issues with the second target customization. This `if` and the code in `calculate.go` will generate a `HelmOptions` like the below:
```
helm:
  repo:  https://github.com/rancher/fleet-examples/releases/download/example/
  chart: oci://docker.github.com/rancher/fleet-examples/guestbook
  version: 0.1.0-test
```
I'm not entirely sure how this will be handled.

romejoe:
@thardeck Ok, I'll try to recall the use case that I encountered that caused me to add the `fixAndApplyHelmSpec`. I'll try to recall the exact target customization case. I'm planning to write up some more robust test cases to identify the weird edge cases that I ran into after work today.

Also, I noticed an issue with the way that some of this code handles OCI registry based charts and I think it may break. I'm going to do some additional testing tonight to make sure there are no issues. I'm also going to incorporate your other suggestions.

pkg/bundle/read.go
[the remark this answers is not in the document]
romejoe: The issue I was trying to address was this: If the user changes the Helm "reference" type(standard repository, OCI repository, local path chart, URI) in a customization, how should options be merged(because a straight merge may override too much and cause issues). I'm not sure but doing the merge of options in `calculate.go` may not provide enough content.

I think I have a way to handle this better than how I did here and I'll implement it later today.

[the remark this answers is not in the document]
thardeck: Personally I would prefer no magic. If the user wants defaults to be overwritten, s/he should do so in the `targetCustomizations` for all preferences they want to change, because they know what they want to achieve and what the default is.
If Fleet has some logic to reset some of the defaults and in other cases not it would be very confusing.

pkg/bundle/read.go
thardeck: No need to add it, it was already there. The `Chart` `targetCustomization` was already supported by fleet, but `Version` and `Repo` not, which you have added in this pr.

pkg/bundle/read.go
thardeck: > Would the user be required in the `fleet.yaml` to always end the repo in a '/'?

No, I do not think so. I do not know why this is added all the time by chartURL, maybe for concatenating, but in this case it should be handled there and not just be added by default.
Unless Helm would require it, which is not the case to my knowledge.

thardeck:
> @thardeck I made the requested changes and reworked the propagation to be, hopefully, a lot clearer. I added a number of configurations [here](https://github.com/romejoe/fleet-examples/tree/add_target_customizations_test/multi-cluster/helm-target-customizations) to try to cover more possible examples that could break things. I still need to update the E2E tests to take these situations into account though. I plan to do some more thorough tests in the morning and update tests later tonight. I think the changes in `calculate.go` may break things, but I will know more after these tests.

Thanks @romejoe, let me know what you find out and take your time - you shared quite a lot of use cases.
I think in most cases Fleet should be able to handle the customizations without the `propagateHelmChartProperties` function, since `targetCustomizations` is handled by `calculate.go`.
I also would guess users to expect matching `targetCustomizations` to overwrite existing values or set empty ones or do we have documentation which states something different?

Also thanks for the test cases but this would be quite a lot for automatic testing.
Usually one or two sanity tests should be enough and of course if one use case did not work in the past and needs extra code, then this could be tested too. But there is no need to cover many different combinations which are working just fine already.

CI is run for every pr, every test takes time to be loaded, applied and so on and on top resources are limited on GithubActions so the goal should be to test the code but not several working combinations for the same code path.

romejoe:
After looking through more of the code base and documentation, I agree that my changes to `read.go` are most likely not necessary. I think I'm over thinking and complicating things a lot... Sorry for wasting time going down this rabbit hole, but I greatly appreciate the feedback!

I moved this to be a draft because the current E2E tests failed. I'll clean things up a bit and fix what's broken before I move it back to a normal pr. Hopefully in the next day or two.

Also, thanks for the clarification on the tests, that saves me a great deal of time that would have been unnecessary.

thardeck:
> I moved this to be a draft because the current E2E tests failed. I'll clean things up a bit and fix what's broken before I move it back to a normal pr.

Btw. we have recently added some dev tools to make it easier to run the tests without the need for an external repo: https://github.com/rancher/fleet/blob/master/dev/README.md

pkg/bundle/resources.go
```diff
@@ -103,6 +103,15 @@ type Auth struct {
...
 func chartURL(location *fleet.HelmOptions, auth Auth) (string, error) {
+
+	hasOCIURL, err := regexp.MatchString(`^oci:\/\/`, location.Chart)
+	if err != nil {
+		return "", err
+	}
+	if hasOCIURL {
+		return location.Chart, nil
+	}
+
```
thardeck, line 114: What do you try to achieve here? `Repo` should be empty in OCI urls as in my [example](https://github.com/rancher/fleet-examples/blob/master/single-cluster/helm-oci/fleet.yaml) so this method should return the location.Chart already without additional logic.

And if there is an issue in OCI chart handling, I would put it at least in a separate commit or better in a new pr - especially if it is not related.

pkg/bundle/read.go
```diff
@@ -247,6 +251,76 @@ func read(ctx context.Context, name, baseDir string, bundleSpecReader io.Reader,
...
+			continue
+		}
+		if !strings.HasSuffix(target.Helm.Repo, "/") {
+			target.Helm.Repo = target.Helm.Repo + "/"
+		}
+	}
+}
+
+// propagateHelmChartProperties propagates root Helm chart properties to the child targets.
+func propagateHelmChartProperties(spec *fleet.BundleSpec, baseDir string) {
```
thardeck, line 274: This method should not be needed, because all three cases are handled by calculate.go already or did I miss something?

thardeck (changes requested):
Could you provide the use case[s] which do not work without your function `ensureValidHelmRepo`?

To my knowledge no additional logic should be needed besides the `calculate.go` changes.

romejoe:
Yeah, I can provide some use cases. I did a few tests last night, I'll write up everything I found in a few minutes.

Also, the new dev scripts were extremely helpful in testing everything.

romejoe:
@thardeck it took a bit longer than I expected to collect the results of my testing but I have them all below and I reran my tests a few minutes ago. Also, if the line numbers below aren't correct, I can update them.

With the e2e multi cluster tests pass with only the change to `calculate.go`.

However, I deployed the gitrepo manually after the tests and in the downstream cluster the following release was deployed:
```
NAME      NAMESPACE                                   REVISION  UPDATED                                 STATUS          CHART                                                                          APP VERSION
radicale  fleet-mc-helm-target-customizations-example 2         2022-10-20 15:29:41.644893852 +0000 UTC deployed        helm-target-customizations-multi-cluster-helm-t-3b15b-v0.0.0+git-b59ee4315897             
```
None of the expected resources were deployed either. The bundle deployment had the expected repo and version though. It appears that `bundles/resources.go:63` is not picking up the new chart in the customization.  

Sanity checking myself with [this config](https://github.com/romejoe/fleet-examples/blob/add_target_customizations_test/multi-cluster/helm-target-customizations/cm-version/fleet.yaml), only one chart, the original, is found and included in the bundle. It looks like if the `propogateHelmChartProperties` isn't used, the check to add a new chart to `chartDirs` at `bundle/resources.go:63` and the if statement at `bundle/resources.go:187` can be mistakenly false(for example, in the above config where only version if overriden in the customization). 

Adding `propogateHelmChartProperties`, the e2e multi cluster tests pass and the generated bundle has the resources from both charts as expected. The downstream cluster also has the correct downstream deployments:
```
NAME      NAMESPACE                                    REVISION  UPDATED                                 STATUS    CHART            APP VERSION
radicale  fleet-mc-helm-target-customizations-example  2         2022-10-20 17:34:44.027904292 +0000 UTC deployed  radicale-0.0.36  3.1.8.0    
```

To further sanity check myself, I ran through the rest of the configs [here](https://github.com/romejoe/fleet-examples/blob/add_target_customizations_test/multi-cluster/helm-target-customizations/). Currently, the expected result was for the `kubernetes-dashboard` chart at version `5.11.0` to be deployed in each namespace and that's the output I'm getting. I think there are still a few use edge cases that those configs aren't covering though.

Note: The change in `chartUrl` to check for oci urls is to prevent the code that follows to try to use the oci url as a chart name.

There are still a few changes that can be made if `propagateHelmChartProperties` is used. There are a if statements that I think can be simplified if customization `HelmOptions` always have the chart details. The loop at `bundle/resource.go:195` can also be optimized to prevent charts from being fetched multiple times. The changes to `calculate.go` are probably also redundant.

Generating a temporary `HelmOptions` in `readResources` in `bundle/resources.go` that includes the chart information from the base chart and the customization so that the chart can be included in the bundle may also be an option. I think the changes to `calculate.go` would still be needed though.

pkg/bundle/resources.go
romejoe: I don't think this is really an issue with the handling of OCI charts. The change in chartUrl to check for OCI urls is to prevent the code that follows to try to use the oci url as a chart name. Normally, chartUrl wouldn't be called because the function calling `chartUrl` has a check for `spec.Repo != ""` that prevents the call if it's an OCI chart. I think the only way a user would hit this currently would be if they supplied a Repo and an oci/go-getter chart.

The addition of `propagateHelmChartProperties` causes the calling function's check to be true in some cases. Instead of this change, I think there is a change that can be made to the calling check to make sure that chartUrl isn't called in that case.

`propagateHelmChartProperties` can also be updated to prevent those propagations from hitting this case, unless the user explicitly supplies a Repo and Chart. I think something similar to the current commented body of  `propagateHelmChartProperties`'s loop would solve this problem.

thardeck:
@romejoe sorry for my late reply and many thanks for the testing and your reproduction steps. I have added a test for the actual downstream deployment and cleaned up the comments: https://github.com/romejoe/fleet/pull/2

romejoe:
@thardeck awesome thanks! Yeah, I've been meaning to clean up my code but haven't had the opportunity to... I merged in your changes and updated this pr

pkg/bundle/read.go
thardeck: The function is needed for the changes to be propagated to the downstream deployments.

pkg/bundle/resources.go
thardeck: You are right, it is better to prevent potential user issues by handling this automatically.

thardeck (approved):

manno (approved):
Maybe rebase and squash some commits? I didn't read through the whole discussion, but my feeling is we need more clarity (comments, docs) on what value overwrites what, in the future. Maybe some renaming/code comments for read/calculate and their role in the deployment process.

thardeck:
@romejoe could you run `git rebase -i` against current master and squash your commits? I would squash all regular code changes in one commit including my hasOCIURL changes. Make sure to add a short description in the commit message why your function is needed, probably referencing this pr because of the broader context.
And on top I would merge my two commits with the tests to one, but that is optional.

romejoe:
@thardeck I just squashed the commits down to 2 (a code one and a testing one). Hopefully I did the rebase correctly. I still have the old branch in case additional changes need to be made or in case I did the rebase wrong.

thardeck (approved):

thardeck:
Many thanks, merging.

PR #1019, opened 2022-10-10
Add test for building binaries/images on several platforms

manno:
This uses goreleaser just to build images and binaries. Nothing is actually released by the workflow. It replicates
the settings from scripts/build and scripts/package. Windows builds are not included in the workflow.

The changes to the Dockerfiles are necessary, because goreleaser automatically provides the right archtitecture binary to in it's build dir. However that is incompatible with the existing Drone release pipeline and `scripts/build`. By using a variable we can support both at the cost of readability.

### DIFF
--- a/.github/workflows/release-images-test.yml
+++ b/.github/workflows/release-images-test.yml
@@ -1,0 +1,49 @@
+name: Release Images Test
+
+on:
+  pull_request:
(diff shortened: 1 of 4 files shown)


manno:
We can revisit this if the drone PR check gives us any trouble. Currently it's just slow.

PR #50, opened 2020-08-15
Major update

ibuildthecloud:
* Direct git integration
* Cluster auto import from cluster-api
* More CLI subcommand
* More flexible cluster registration

PR #1035, opened 2022-10-19
Go Templating for Helm Values with Cluster Labels as Inputs

k0da:
This commit adds simple go template capabilities to be able to construct a liitle bit more complex helm values out of cluster labels

Signed-off-by: Dinar Valeev <dinar.valeev@absa.africa>

Fix #XXX

### Potential improvement

Allow to construct role arn for service account annotation during helm install
```yaml
helm:
  values:
     serviceAccount:
        annotations:
            eks.amazonaws.com/role-arn: 'arn:aws:iam::{{.accountId}}:role/{{.accountName}}-role'
```

### DIFF
--- a/pkg/target/target.go
+++ b/pkg/target/target.go
@@ -6,10 +6,13 @@
 package target
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
 	"sort"
 	"strconv"
 	"strings"
+	"text/template"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -330,6 +333,10 @@ func addClusterLabels(opts *fleet.BundleDeploymentOptions, labels map[string]str
 		return err
 	}
 
+	if err := processValuesTemplate(opts.Helm.Values.Data, clusterLabels); err != nil {
+		return err
+	}
+
 	opts.Helm.Values.Data = data.MergeMaps(opts.Helm.Values.Data, newValues)
 	return nil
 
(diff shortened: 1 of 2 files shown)


gregsidelinger:
There is already a PR for this, https://github.com/rancher/fleet/pull/671  and https://github.com/rancher/fleet/pull/50 which I closed as #671 was much better.

k0da:
> There is already a PR for this, #671 and #50 which I closed as #671 was much better.

Yes, but I see no activities there for over half a year.

k0da:
```yaml
targetCustomizations:
- name: eks
  helm:
    values:
      serviceAccount:
        create: true
        annotations:
          foo: '{{ .ClusterType }}-is-awesome'
```
Rancher

thardeck:
Hey @k0da, thanks for your pr.
For the records there is some discussion regarding this and similar topics ongoing in #944. Feel free to add your ideas/opinions there.

thardeck:
> There is already a PR for this, #671 and #50 which I closed as #671 was much better.

@gregsidelinger Would this pr cover all your requirements?

rajiteh:
Thanks for the poke @thardeck! 

So what doesn't really fit my use case in this implementation is the fact that we are still restricting templating to cluster labels. K8s labels have strict formatting and length requirements that might not fit the kinds of data someone might want to expose to fleet YAMLs. In my implementation I introduced a `.spec.templateContext` object to the cluster CR to work around that.

Few more notes:

- IMO it's a good idea to namespace the templating context in case we want to introduce additional "scopes" for templating. It also keeps the fleet yaml more readable as you're able to know exactly where a variable is coming from. So instead of `{{.accountId}}` I prefer something like `{{clusterLabels.accountId}}`. We can also expand on that to expose more metadata, say annotations, or variables coming from other sources.

- While using standard `text/template` is a decent solution, it falls short when you try to do advanced things with templates, so I would very much prefer bringing [sprig](https://masterminds.github.io/sprig/) library in so  templating as a  feature can be more useful. Besides most folk in the space are already used to helm/helmfile based templating which uses sprig extensively, and it will be missed.

FWIW I've been using my fork #671 in prod for a while now and I think it meets all our needs, would love to have it merged in which I think will satisfy most templating usecases.

k0da:
@rajiteh thanks for your input.

Indeed .Labels.accountId. looks nicer and more readable (as it exposes a source of the data). I have the change prepared, can push any time if we agree we need that 💯 

re: annotations. Our main usecase for that templating is that not everything can be carried over `Labels` (in my case aws::arn) so we split data into carry-able format. My intend it to introduce change in less intrusive fashion, whitout API break.

k0da:
I can also add annotations, but IMO this increasing  a scope of this PR massively.

thardeck:
@k0da many thanks for your pr. We have recently merged one of the previous template prs #671.
Closing because the functionality should be covered by the merged pr.
Please let us know if it works as expected for your use case.
I will cut another Fleet alpha release to be able to easily test it.

PR #1043, opened 2022-10-21
Add an empty json tag to nested API structures

alexander-demicev:
We are using fleet API structures in [elemental operator](https://github.com/rancher/elemental-operator/) and I'm trying to set up CRD generation using the controller runtime tool https://book.kubebuilder.io/reference/controller-gen.html. The generation fails because the tool doesn't know how to handle nested structures without json tag so an empty tag should work.

### DIFF
--- a/pkg/apis/fleet.cattle.io/v1alpha1/bundle.go
+++ b/pkg/apis/fleet.cattle.io/v1alpha1/bundle.go
@@ -56,7 +56,7 @@ type BundleNamespaceMapping struct {
 }
 
 type BundleSpec struct {
-	BundleDeploymentOptions
+	BundleDeploymentOptions `json:""` //nolint
 
 	Paused             bool                      `json:"paused,omitempty"`
 	RolloutStrategy    *RolloutStrategy          `json:"rolloutStrategy,omitempty"`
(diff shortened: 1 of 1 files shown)


manno:
Indeed that seems to work.
I hope you're not trying to generate the CRDs, but just a client?

Because our CRDs are a bit different from the structs, e.g. we add json schema validation and display fields: https://github.com/rancher/fleet/blob/master/pkg/crd/crds.go#L92-L105

alexander-demicev:
Actually, we are generating CRDs from this. I plan to decouple our API from fleet in the future so we don't have nested structures.

manno:
> Actually, we are generating CRDs from this. I plan to decouple our API from fleet in the future so we don't have nested structures.

Hm, not sure I understand this. You could download the "real" CRD from https://github.com/rancher/fleet/releases?
Oh, are you just using a subset of fleet? Searching your operator, it looks like you use the `Bundle`?

alexander-demicev:
@manno https://github.com/rancher/elemental-operator/blob/main/pkg/apis/elemental.cattle.io/v1beta1/os.go#L56 here is an example, we are including fleet structures in our API, that later cause problems with generating CRDs using controller runtime tools.

kkaempf:
This changes Fleet CRDs and thus considered too risky to merge.

PR #1052, opened 2022-10-24
Make syncing of unchanged objects optional

moio:
Extends #1041 a bit by making the parameter configurable (for test purposes only).

Plan is to have this tested by few users and revert this commit if results match expectations.

### Tradeoff

Ideally this parameter should not be configurable, but it is necessary to test it in isolation on user premises before taking a final decision on all users.

### Testing

I tested this manually.

### DIFF
--- a/cmd/fleetcontroller/main.go
+++ b/cmd/fleetcontroller/main.go
@@ -22,9 +22,10 @@ var (
 )
 
 type FleetManager struct {
-	Kubeconfig    string `usage:"Kubeconfig file"`
-	Namespace     string `usage:"namespace to watch" default:"cattle-fleet-system" env:"NAMESPACE"`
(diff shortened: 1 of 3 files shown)


manno:
Since we don't intend to keep this as an argument, what do you think about just using an env var? We could avoid passing the argument around.

In wrangler/lasso there are already:

* CATTLE_PROMETHEUS_METRICS
* CATTLE_RESYNC_DEFAULT
* CATTLE_DEV_MODE

moio:
> Since we don't intend to keep this as an argument, what do you think about just using an env var?

I think it's a good idea, and also something that crossed my mind when writing this. Besides, I already got feedback as a user just applied the patch individually - so I am closing this for now, and will reopen (as an environment variable) if there is a need going forward.

Thanks for reviewing.

PR #1053, opened 2022-10-24
Cluster specific registry replaces global registry

### LINKED ISSUE #948: Feature Request: Cluster specific registry that the agent image is pulled from

manno:
If both systemDefaultRegistry and privateRepoURL are present, remove systemDefaultRegistry from the image arg before prefixing with privateRepoURL.

To be able to remove the systemDefaultRegistry, we need to store it separately in the fleet config.

Fix #948

## Test

```shell
go test ./pkg/agent/...
```

### DIFF
--- a/pkg/agent/manifest_test.go
+++ b/pkg/agent/manifest_test.go
@@ -1,0 +1,23 @@
+package agent
+
+import "testing"
+
+func TestImageResolve(t *testing.T) {
+	tests := []struct {
+		systemDefaultRegistry string
+		privateRepoURL        string
+		image                 string
+		expected              string
+	}{
+		{"", "", "rancher/fleet:dev", "rancher/fleet:dev"},
+		{"mirror.example/", "", "mirror.example/rancher/fleet:dev", "mirror.example/rancher/fleet:dev"},
+		{"mirror.example/", "local.example", "mirror.example/rancher/fleet:dev", "local.example/rancher/fleet:dev"},
+	}
+
+	for _, d := range tests {
+		image := resolve(d.systemDefaultRegistry, d.privateRepoURL, d.image)
+		if image != d.expected {
+			t.Errorf("expected %s, got %s", d.expected, image)
+		}
+	}
+}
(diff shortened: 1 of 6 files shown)


pkg/agent/manifest_test.go
```diff
@@ -0,0 +1,23 @@
+package agent
+
+import "testing"
+
+func TestImageResolve(t *testing.T) {
+	tests := []struct {
+		systemDefaultRegistry string
+		image                 string
```
thardeck, line 8: I think it would be easier to read if the struct follows the succession of the `resolve` function parameters. So the second should be `privateRepoURL` and the third `image`.

thardeck (changes requested):
lgtm besides the nitpick

pkg/agent/manifest_test.go
manno: Yes, I was just too lazy to re-order the strings :)

thardeck (approved):

PR #1058, opened 2022-10-26
Bugfix: do not ignore agentCheckinInterval

### LINKED ISSUE #1057: Helm agentCheckinInterval parameter is ignored

### Current Behavior

By default, agents check in every 15 minutes, but the amount of time is supposed to be configurable via the `agentCheckinInterval` Helm value:

https://github.com/rancher/fleet/blob/73e9fa5b43265c947bca353d3ac480fdb8df8a62/charts/fleet/values.yaml#L19-L20

In reality that parameter has no effect, indeed running:

```
helm -n ... upgrade ... --set agentCheckinInterval=30m fleet
```

does not change the default behavior.

### Expected Behavior

`agentCheckinInterval` is honored.

### Steps To Reproduce

1. install fleet with debugging logs enabled (`debug=true` Helm value) and custom checkin interval (`agentCheckinInterval=10s` Helm value)
2. observe fleetcontroller logs. The following line will appear (twice) every 15 minutes:

```
Cluster 'fleet-local/local' changed, enqueue matching cluster groups
```

The line should appear every 10s instead.

### Environment

```markdown
- Architecture: amd64
- Fleet Version: current `master` (73e9fa5b43265c947bca353d3ac480fdb8df8a62)
- Cluster:
  - Provider: K3D
  - Options: 1 node, no customizations
  - Kubernetes Version: v1.23.6+k3s1
```

### Logs

_No response_

### Anything else?

Related to internal issue [SURE-4967](https://jira.suse.com/browse/SURE-4967)

moio:
Fixes #1057 

A typo in a configuration parameter name prevented it from being correctly applied. rancher/rancher seems not to be affected (a grep search only surfaces the correct parameter name), so this might only affect standalone installations.

## Test

See reproduction steps in https://github.com/rancher/fleet/issues/1057

I tested this manually.

### Tradeoff

Keeping the old name around and use it if defined. From my understanding nobody really used it so far.

### Potential improvement

None.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -38,7 +38,7 @@ type Config struct {
 	AgentImage                      string            `json:"agentImage,omitempty"`
 	AgentImagePullPolicy            string            `json:"agentImagePullPolicy,omitempty"`
 	SystemDefaultRegistry           string            `json:"systemDefaultRegistry,omitempty"`
-	AgentCheckinInternal            metav1.Duration   `json:"agentCheckinInternal,omitempty"`
+	AgentCheckinInterval            metav1.Duration   `json:"agentCheckinInterval,omitempty"`
 	ManageAgent                     *bool             `json:"manageAgent,omitempty"`
 	Labels                          map[string]string `json:"labels,omitempty"`
 	ClientID                        string            `json:"clientID,omitempty"`
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -92,7 +92,7 @@ func AgentWithConfig(ctx context.Context, agentNamespace, controllerNamespace, a
 	mo.AgentImage = cfg.AgentImage
 	mo.SystemDefaultRegistry = cfg.SystemDefaultRegistry
 	mo.AgentImagePullPolicy = cfg.AgentImagePullPolicy
-	mo.CheckinInterval = cfg.AgentCheckinInternal.Duration.String()
+	mo.CheckinInterval = cfg.AgentCheckinInterval.Duration.String()
 
 	objs = append(objs, Manifest(agentNamespace, agentScope, mo)...)
 
--- a/pkg/controllers/cluster/import.go
+++ b/pkg/controllers/cluster/import.go
@@ -264,7 +264,7 @@ func (i *importHandler) importCluster(cluster *fleet.Cluster, status fleet.Clust
 			},
 			ManifestOptions: agent.ManifestOptions{
 				AgentEnvVars:    cluster.Spec.AgentEnvVars,
-				CheckinInterval: cfg.AgentCheckinInternal.Duration.String(),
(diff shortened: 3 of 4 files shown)


moio:
@manno, @thardeck: what are your feelings about testing this patch?

I tested this manually in my setup, and expect to soon have at least a user independently testing it too. Do we want any sort of automated test?

 - `go test`: I do not see a good way to test this
 - e2e tests: it's a possibility to add this parameter as an extra `--set agentCheckinInterval=21s` here:

https://github.com/rancher/fleet/blob/e2e4b564690287708870084781177dfcd91a78dc/.github/scripts/deploy-fleet.sh#L19-L25

and then check that the value is correct in the agent's deployment, something conceptually similar to:

```sh
~ kubectl get --namespace=$NAMESPACE deployment fleet-agent --output=yaml | grep -A1 CHECKIN_INTERVAL
        - name: CHECKIN_INTERVAL
          value: 21s
```

(I chose the value of 21s for illustration purposes only)

It's probably not a test with a terribly good future breakage predicting power - I'd leave it up to you decide whether it's worth the code or not. Or maybe you have better ideas 😉

pkg/config/config.go
```diff
@@ -38,7 +38,7 @@ type Config struct {
 	AgentImage                      string            `json:"agentImage,omitempty"`
 	AgentImagePullPolicy            string            `json:"agentImagePullPolicy,omitempty"`
 	SystemDefaultRegistry           string            `json:"systemDefaultRegistry,omitempty"`
-	AgentCheckinInternal            metav1.Duration   `json:"agentCheckinInternal,omitempty"`
+	AgentCheckinInterval            metav1.Duration   `json:"AgentCheckinInterval,omitempty"`
```
manno, line 41:
```suggestion
	AgentCheckinInterval            metav1.Duration   `json:"agentCheckinInterval,omitempty"`
```

manno (changes requested):

manno:
> @manno, @thardeck: what are your feelings about testing this patch?

> I tested this manually in my setup, and expect to soon have at least a user independently testing it too. Do we want any sort of automated test?

Not needed right now.  Long term we could look into https://github.com/helm/chart-testing (example: https://github.com/epinio/helm-charts/blob/main/.github/workflows/main-ci.yml#L46-L52).

In general it depends if you want to test the chart, or the CLI, or both in an acceptance/e2e test. However, the helm arguments are not documented and we don't really support standalone/helm use. (well, we use it for testing, so it's very supported in a way)

moio:
Thanks @manno, I tend to agree with your perspective. I would proceed merging this as soon as we have user positive feedback, unless you have any other objection.

moio:
@manno Just rebased, positive feedback from user, I'd merge unless you have any reasons not to.

manno (approved):

Issue #606, opened 2021-11-09
Fleet controller consuming high cpu


StrongMonkey:
SURE-3613

**Issue description**:
fleet controller is consuming a large amount of CPU, possibly querying the Kubernetes API server at a high rate, which is impacting other k8s services that consume the API.

**Business impact**:
This might be impacting the customers ability to deploy a certain chart or affect the K8S api servers availability.

**Troubleshooting steps**:
deploy a hurt chart? I believe in this case it could be due to a chart trying to interpret `deletionGracePeriodSeconds`

**Repro steps**:
deploy a hurt chart? I believe in this case it could be due to a chart trying to interpret `deletionGracePeriodSeconds`

**Additional notes**:
Seeing a lot of this in the fleet controller
```
gitjobs(map[objectset.rio.cattle.io/hash:88bd6abcf757f7fb2377854899cb72a44cc5482a]) 1.004423631s"
time="2021-11-05T00:07:12Z" level=info msg="rate limited gitjobs(map[objectset.rio.cattle.io/hash:7734899c5a839e11a1f4e2361e91d27084de0146]) 1.005166593s"
time="2021-11-05T00:07:12Z" level=info msg="rate limited gitjobs(map[objectset.rio.cattle.io/hash:3aa974cf2fbc6cee42ac2569146e5c253243e1b4]) 1.003825323s"
time="2021-11-05T00:07:12Z" level=info msg="rate limited gitjobs(map[objectset.rio.cattle.io/hash:96a01aeda7a8eaf214e354b37837e0c3fa77073e]) 1.00407412s"
time="2021-11-05T00:07:12Z" level=info msg="rate limited
```

From `top`:
```
top - 00:13:54 up 30 days, 23:51, 0 users, load average: 6.87, 6.41, 5.18
Tasks: 203 total, 2 running, 201 sleeping, 0 stopped, 0 zombie
%Cpu(s): 57.6 us, 15.3 sy, 0.2 ni, 26.3 id, 0.0 wa, 0.0 hi, 0.7 si, 0.0 st
MiB Mem : 7705.7 total, 163.6 free, 4436.1 used, 3105.9 buff/cache
MiB Swap: 0.0 total, 0.0 free, 0.0 used. 4137.6 avail Mem

PID USER PR NI VIRT RES SHR S %CPU %MEM TIME+ COMMAND
3735043 ubuntu 20 0 913240 232788 42364 S 66.1 3.0 37:16.76 fleetcontroller
[...]

PR #1062, opened 2022-10-27
Scaling: avoid needlessly looking up agent bundle deployments

moio:
- #606

## TL;DR

Fix for a handler triggered every 15' resulted and taking `O(n²)` CPU time, where `n` is the number of clusters.

In my test setup with 2000 clusters it results in a **~16x speedup** (22 seconds down from 6 minutes).

## Performance analysis

A user experienced constant CPU load on fleetcontroller, and shared the CPU pprof profile. Flame graph looked like the following:

Screen Shot 2022-10-27 at 12 28 40

Observations:
 - the first small flame is GC work
 - the second big flame has two smaller flames:
    - one is about [bundle controller's OnBundleChange](https://github.com/rancher/fleet/blob/e2e4b564690287708870084781177dfcd91a78dc/pkg/controllers/bundle/controller.go#L171) (this can deduced because it's registered via `RegisterBundleGeneratingHandler`, which adds the specific `FromBundleHandlerToHandler`/`sync`/`Handler` wrappers and results in a Wrangler `Apply` operation)
    - one is about [bundle controller's OnClusterChange](https://github.com/rancher/fleet/blob/e2e4b564690287708870084781177dfcd91a78dc/pkg/controllers/bundle/controller.go#L97)

I am focusing on the last one to start, as it takes the majority of CPU time (>55%).

Further observations:
 - `OnClusterChange` spends most of its time in `GetBundleDeploymentsForBundleInCluster`
 - `GetBundleDeploymentsForBundleInCluster` is called for all cluster/bundle combinations*:

https://github.com/rancher/fleet/blob/e2e4b564690287708870084781177dfcd91a78dc/pkg/controllers/bundle/controller.go#L104-L109

*to be more precise, only bundles which are classified as "to clean up" by `BundlesForCluster` are considered. But actually `BundlesForCluster` selects the vast majority of bundles, so that is a safe approximation according to my tests.

Further observations:
 - there is always at least one bundle per cluster (the cluster agent's)
 - thus we have **an ~n² number of calls to `GetBundleDeploymentsForBundleInCluster`** where n is the cluster count

Theoretical bottom line: even if `GetBundleDeploymentsForBundleInCluster` were quite fast (eg. 100 microseconds), with 2000 clusters we can expect 2000x2000 = 4 million calls so it will still take long (eg. several minutes at 100% CPU). This is compatible with reported symptoms.

## Proposed Fix

Avoid `BundlesForCluster` to return other clusters' agent bundles as "bundles to clean up". IOW, every time `OnClusterChange` fires, [which is by default at least every 15 minutes for each cluster](https://github.com/rancher/fleet/blob/e2e4b564690287708870084781177dfcd91a78dc/charts/fleet/values.yaml#L19-L20)), only clean up that cluster's agent bundle, and any applicable user-defined bundle.

## Results

In my setup, with 2000 fake clusters this patch yields a **~16x speedup** (6 minutes unpatched, 22 seconds patched). Details about the setup follow.

I am waiting on results from the affected user and will update this once I have them.

According to my tests fleetcontroller's ability to update fleet-agents on downstream clusters on version changes remains intact.

## Test

1. install fleet, register one downstream cluster
2. generate 2000 fake cluster resources:
```bash
NAMESPACE=fleet-local

FIRST_CLUSTER_NAME=$(kubectl get --namespace=$NAMESPACE cluster --output=custom-columns=":metadata.name" | grep cluster- | head -1)
kubectl get --namespace=$NAMESPACE cluster $FIRST_CLUSTER_NAME --output=yaml > cluster.yaml

for i in {1..2000}
do
    CLUSTER_ID=`echo $i | md5sum | head --bytes 12`
    cat cluster.yaml | sed "s/$FIRST_CLUSTER_NAME/cluster-$CLUSTER_ID/g" | kubectl apply --filename=-
done
```

3. stop fleetcontroller:
```sh
kubectl scale --namespace=fleet-system deployments/fleet-controller --replicas=0
```

4. "touch" all clusters:
```sh
kubectl get --all-namespaces cluster --output yaml | kubectl apply --filename=- --server-side
```

5. bring fleetcontroller up again, CPU is expected to go 100% for some time:
```sh
kubectl scale --namespace=fleet-system deployments/fleet-controller --replicas=1
vmstat --unit=M --wide --one-header 1 | nl --starting-line-number=-2
```

6. measure how long it takes before it returns back to normal, by paying attention to the CPU `us` column. The first column (output of `nl`) indicates the number of seconds since the start
7. apply the patch, replacing fleetcontroller, then repeat steps 3.-6. and calculate the speedup

To clean up:
```sh
for i in {1..2000}
do
    CLUSTER_ID=`echo $i | md5sum | head --bytes 12`
    kubectl delete --namespace=$NAMESPACE cluster cluster-$CLUSTER_ID --cascade=foreground
    kubectl delete --namespace=$NAMESPACE bundle fleet-agent-cluster-$CLUSTER_ID --cascade=foreground
done
```

### Tradeoff

Bundle selection could be made more precise by adding specific labels to agent bundles instead of relying on bundle names.

Even better, potentially, use one bundle for all agents - and many BundleDeployments (one per cluster). There might be good reasons why the setup is of one Bundle per Cluster, I just do know them.

### Potential improvement

Similar work could be done for the second flame in the flame graph above.

pprof cpu profiling should become a standard fleet option.

### DIFF
--- a/pkg/target/mapping.go
+++ b/pkg/target/mapping.go
@@ -110,9 +110,13 @@ func (b *bundleSet) insert(bundles []*fleet.Bundle, err error) error {
 		return err
 	}
 	for _, bundle := range bundles {
-		key := bundle.Namespace + "/" + bundle.Name
-		b.bundleMap[key] = bundle
-		b.bundleKeys.Insert(key)
+		b.insertSingle(bundle)
 	}
 	return nil
+}
+
+func (b *bundleSet) insertSingle(bundle *fleet.Bundle) {
+	key := bundle.Namespace + "/" + bundle.Name
+	b.bundleMap[key] = bundle
+	b.bundleKeys.Insert(key)
 }
--- a/pkg/target/target.go
+++ b/pkg/target/target.go
@@ -105,8 +105,20 @@ func (m *Manager) clusterGroupsForCluster(cluster *fleet.Cluster) (result []*fle
 func (m *Manager) getBundlesInScopeForCluster(cluster *fleet.Cluster) ([]*fleet.Bundle, error) {
 	bundleSet := newBundleSet()
 
-	if err := bundleSet.insert(m.bundleCache.List(cluster.Namespace, labels.Everything())); err != nil {
+	// all bundles in the cluster namespace are in scope
+	// except for agent bundles of other clusters
+	bundles, err := m.bundleCache.List(cluster.Namespace, labels.Everything())
+	if err != nil {
 		return nil, err
+	}
+	for _, b := range bundles {
+		if b.Annotations["objectset.rio.cattle.io/id"] == "fleet-manage-agent" {
+			if b.Name == "fleet-agent-"+cluster.Name {
+				bundleSet.insertSingle(b)
+			}
+		} else {
+			bundleSet.insertSingle(b)
+		}
 	}
 
 	mappings, err := m.bundleNamespaceMappingCache.List("", labels.Everything())


moio:
@manno what would be your thoughts about adding automated tests for this?

moio:
@manno Just rebased, positive feedback from user, I'd merge unless you have any reasons not to.

manno (approved):

PR #1063, opened 2022-10-28
Reading contents&secrets uses specific service account

manno:
Fixes an issue, by moving the two roles onto specific service accounts.

RBAC previously allowed any user to get contents and any secrets in
the system registration namespace.
However, their names are random and listing was not allowed.

The agents on downstream clusters already use the`request-` account to
retrieve bundledeployments from the upstream cluster and report back the
status. We extend that account to be able to retrieve 'contents', too.

The agents use a `import-` account for reqstration, e.g. to create a
ClusterRegistration. That one is extended, so it can read the secrets in
the agent's cluster registration namespace.

I previously observed issues when updating the agent, but can no longer reproduce these, e.g. the agent log would show
`level=error msg="Failed to register agent: looking up secret cattle-fleet-system/fleet-agent-bootstrap: looking up secret cattle-fleet-system/fleet-agent-bootstrap: secrets \"fleet-agent-bootstrap\" not found".`

Since an upgrade wouldn't modify the dynamically created roles, the old permissions should still be in place. Only new fleet installations benefit from the stricter role bindings.

### DIFF
--- a/e2e/installation/verify_test.go
+++ b/e2e/installation/verify_test.go
@@ -1,0 +1,68 @@
+package installation_test
+
+import (
+	"os"
+	"path"
+
+	"github.com/rancher/fleet/e2e/testenv"
+	"github.com/rancher/fleet/e2e/testenv/kubectl"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+// This runs after an upgrade of fleet to verify workloads are still there and
+// new workload can be created
+var _ = Describe("Fleet Installation", func() {
+	var (
+		asset string
+		k     kubectl.Command
+	)
+
+	BeforeEach(func() {
+		k = env.Kubectl.Namespace(env.Namespace)
+	})
+
+	Context("sanity checks", func() {
+		It("finds the original workload", func() {
+			out, _ := k.Namespace("bundle-diffs-example").Get("services")
+			Expect(out).To(ContainSubstring("app-service"))
+		})
+
+	})
+
+	When("Deploying another bundle", func() {
+		var tmpdir string
+		BeforeEach(func() {
+			asset = "simple/gitrepo.yaml"
+		})
+
+		JustBeforeEach(func() {
+			tmpdir, _ = os.MkdirTemp("", "fleet-")
+			gitrepo := path.Join(tmpdir, "gitrepo.yaml")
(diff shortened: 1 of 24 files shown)


thardeck:
> Fixes #853 by moving the two roles onto specific service accounts.

The issue reference seems to be wrong. It refers to #853

manno:
> > Fixes #853 by moving the two roles onto specific service accounts.
> 
> The issue reference seems to be wrong.

Thanks, it was  in https://github.com/rancherlabs/rancher-security

e2e/installation/verify_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+	When("Deploying another bundle", func() {
+		var tmpdir string
+		BeforeEach(func() {
+			asset = "simple/gitrepo.yaml"
+		})
+
+		JustBeforeEach(func() {
+			tmpdir, _ = os.MkdirTemp("", "fleet-")
+			output := path.Join(tmpdir, testenv.RandomFilename("manifests.yaml"))
```
thardeck, line 42: I would use a more fitting variable name like `file` or `filePath`.

thardeck (changes requested):

thardeck (approved):

manno:
@rancher/qa

## Additionnal QA
### Issue: https://github.com/rancherlabs/rancher-security/issues/853

### Problem
After installation Fleet did not pass hardening validation.

### Solution

Two roles were moved to specific service accounts. These service accounts are already created per cluster.
 
### Testing

Install fleet, harden the cluster by [removing default tokens from the service accounts](https://docs.ranchermanager.rancher.io/v2.5/reference-guides/rancher-security/rancher-v2.5-hardening-guides/hardening-guide-with-cis-v1.6-benchmark#set-automountserviceaccounttoken-to-false-for-default-service-accounts).

Run the [script](https://github.com/rancher/security-scan/blob/fd39c51fd16e522815f244069b497e060e8c5f9c/package/helper_scripts/check_for_default_sa.sh) to verify no roles are bound to the default service accounts.

### Engineering Testing
#### Manual Testing

* Fleet can still deploy
* started documenting the registration flow

#### Automated Testing

* no new tests

### QA Testing Considerations

Since an upgrade wouldn't modify the dynamically created roles, the old permissions should still be in place. Only new fleet installations benefit from the stricter role bindings.
 
Fleet creates a lot of rbac resources dynamically, i.e. per cluster, during the registration process, .... This is hard to audit.

PR #1076, opened 2022-11-04
Add support to continuous pprof CPU profiling

moio:
This PR adds Helm chart values to configure continuous [pprof](https://pkg.go.dev/runtime/pprof#hdr-Profiling_a_Go_program) CPU profiling on fleetcontroller.

Since CPU pprof profiling requires a writer to stream data to, this PR creates an optional volume and environment variables to configure it.

## Test

To test this pull request:

- install Fleet with the following extra Helm values:

```yaml
cpuPprof:
  period: "60s"
  volumeConfiguration:
    hostPath:
      path: /tmp/pprof
      type: DirectoryOrCreate
```

Alternatively, use the following Helm commandline arguments:

```
--set cpuPprof.period=60s \
--set cpuPprof.volumeConfiguration.hostPath.path=/tmp/pprof \
--set cpuPprof.volumeConfiguration.hostPath.type=DirectoryOrCreate \
```

- make sure that `/tmp/pprof` is writable
- start Fleet
- after some minutes, check contents of `/tmp/pprof`. It is expected a `2022-11-04_19_47_18.pprof.fleetcontroller.samples.cpu.pb.gz` file is created every minute
- profiles can be inspected with:

```sh
pprof -http=localhost:5000 ./2022-11-04_19_47_18.pprof.fleetcontroller.samples.cpu.pb.gz
```

Output can be formatted as flame graph, eg.:

example flame graph

### Tradeoff

The volume type could be hard-coded to simplify the values file, at the expense of flexibility.

### Potential improvement

Code could be migrated to wrangler and applied to all wrangler-based projects.

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -256,6 +256,41 @@ That being said, downstream `fleet-agent` deployments can perform Kubernetes API
 While network traffic is major point of consideration, we also have to consider whether our performance issues are **compute-based**, **memory-based**, or **network-based**.
 For example: you may encounter a pod with high compute usage, but that could be caused by heightened network traffic received from the _truly_ malfunctioning pod.
 
+### Using pprof
+
+[http pprof](https://pkg.go.dev/net/http/pprof) handlers are enabled by default with all [default profiles](https://pkg.go.dev/runtime/pprof#Profile) under the `/debug/pprof` prefix.
+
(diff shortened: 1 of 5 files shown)


moio:
@manno DEVELOPING.md instructions added as discussed. Good to merge?

manno (approved):

Issue #721, opened 2022-03-01
rate limited gitjobs and bundle


mmclane:
We have fleet managing our clusters and have started using imagescan to pick up new container versions and push those out automatically.  It's taken me a bit to reverse engineer this feature (the documentation is very light if you ask me) but I think I have figured some things out.  

Currently imagescan runs (once a minute) and identifies a new version.  We can see that it has identified the new version in Rancher.  It doesn't however always write that new version to github right away.  Sometimes it can take 5-10 minutes between when imagescan identifies the new version and when the pods get rolled out.  From my understanding this is because something is keeping imagescan from writing to github and then we also need to wait for fleet to run an update (default is 5m).  

What I noticed today is that I am getting a whole bunch of these "errors" in my fleet-controller logs.  I am guessing they are relavent.

```
time="2022-03-01T16:20:55Z" level=info msg="rate limited gitjobs(map[objectset.rio.cattle.io/hash:682559dbf9496c6e24d53b2dc5182519e59c4aaf]) 1.007998787s"
time="2022-03-01T16:20:57Z" level=info msg="rate limited bundle(map[objectset.rio.cattle.io/hash:41112036500a9095791775c74a4b4e19736a63ee]) 1.005525709s"
time="2022-03-01T16:21:03Z" level=info msg="rate limited bundle(map[objectset.rio.cattle.io/hash:41112036500a9095791775c74a4b4e19736a63ee]) 1.019539797s"
time="2022-03-01T16:21:03Z" level=info msg="rate limited gitjobs(map[objectset.rio.cattle.io/hash:682559dbf9496c6e24d53b2dc5182519e59c4aaf]) 1.013903984s"
time="2022-03-01T16:21:05Z" level=info msg="rate limited bundle(map[objectset.rio.cattle.io/hash:41112036500a9095791775c74a4b4e19736a63ee]) 1.001365362s"
time="2022-03-01T16:21:07Z" level=info msg="rate limited bundle(map[objectset.rio.cattle.io/hash:41112036500a9095791775c74a4b4e19736a63ee]) 1.0093039s"
time="2022-03-01T16:21:07Z" level=info msg="rate limited gitjobs(map[objectset.rio.cattle.io/hash:913163cbf43b76cd3323f081009b1e9f27c16339]) 1.000876569s"
[...]

PR #1080, opened 2022-11-07
Scaling: fold bundle display controller handler into main bundle controller

moio:
- #606
- #721

Every time a bundle changes:
- the `bundle/controller.go`'s `OnBundleChange` handler fires
- the handler code changes the bundle's `Status`
- that, in turn, fires `displaycontrollers.go`'s `OnBundleChange` handler

Both of those handler return objects that are applied via wrangler's `apply` facility, which can be heavy on the CPU.

In a user's large setup case, `apply` is fired ~24 times/s and takes up the majority of CPU time (in purple below):

flame graph

This PR folds the two handlers together, thereby reducing the number of `apply` calls by 50%.

## Test

This PR should be seen as an internal-only refactoring, no user-visible behavior should change.

### DIFF
--- a/pkg/controllers/bundle/controller.go
+++ b/pkg/controllers/bundle/controller.go
@@ -3,6 +3,7 @@ package bundle
 
 import (
 	"context"
+	"fmt"
 	"sort"
 	"time"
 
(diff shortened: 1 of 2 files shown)


manno (approved):

manno:
I wonder if this will delay or speed up status updates...

PR #1081, opened 2022-11-08
[EXPERIMENTAL] Avoid wrangler 3-way merge in bundle/controller

moio:
- #606
- #721

Premise: wrangler's apply mechanism uses a three way patch calculated from the original object (latest version that was ever applied), the current object (as it is currently in the cluster) and the new object (passed as argument to apply). The original object is serialized and stored in a special annotation (LabelApplied) every time an apply occurs - to be available for the next apply. A similar mechanism is implemented in kubectl apply, which inspired wrangler apply.

The above mechanism means any object processed by wrangler's apply will be serialized to JSON, gzipped and base64 encoded into an annotation (and vice versa) every time Apply is called. This can be CPU intensive if it happens very frequently, and for some handlers might be unnecessary.

This patch aims at disabling the mechanism for the bundle/controller OnBundleChange handler.

Please do not merge until the [EXPERIMENTAL] title tag is removed!

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.18
 
 replace (
 	github.com/rancher/fleet/pkg/apis => ./pkg/apis
+	github.com/rancher/wrangler => github.com/moio/wrangler v0.0.0-20221108212643-8df82c78249b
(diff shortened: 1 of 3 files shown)


moio:
Closing temporarily - will reopen if and when it's ready for review.

PR #1085, opened 2022-11-11
[EXPERIMENTAL] Bundle controller optimization

moio:
- #606
- #721

Premise: wrangler's apply mechanism uses a `kubectl apply`-inspired three way patch mechanism to update custom resources. Such mechanism requires any applied resource to be serialized/deserialized to JSON, gzipped/ungzipped and base64 encoded/decoded every time. In some circumstances apply may be applied for many objects very frequently, bringing to a CPU bottleneck.

This patch aims at minimizing objects returned by the `OnBundleChange` handler, in hope of reducing CPU load in a high-scale scenario.

Please do not merge until the [EXPERIMENTAL] title tag is removed!

### DIFF
--- a/pkg/controllers/bundle/controller.go
+++ b/pkg/controllers/bundle/controller.go
@@ -402,5 +402,5 @@ func resetDeployment(target *target.Target, status *fleet.BundleStatus) {
 	}
 
 	status.NewlyCreated++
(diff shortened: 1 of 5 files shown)


moio:
Closing temporarily - will reopen if and when it's ready for review.

PR #1103, opened 2022-11-16
Running helm template to set up status ignores errors

### LINKED ISSUE #1101: Calculating "resourceKey" Calls "helm template" With Insufficient Deployment Options
When fleet creates the bundledeployment from the bundle it runs "helm template" to generate a list of resources which are included in the bundle. This list is stored in the bundle status as `resourceKey`. They're also [synced to the gitrepo resource](https://github.com/rancher/fleet/blob/e5b7646d34124fa725a128543a703b715dd24a6c/pkg/display/gitrepo.go#L194) status. 

The resource list is [used in the dashboard](https://github.com/rancher/dashboard/blob/92c81e6d14657cad83e8be6c63e09504e828994c/shell/detail/fleet.cattle.io.bundle.vue).

The problem is, helm is normally executed on a downstream cluster to install a bundledeployment. Here we run helm on the upstream cluster. So, helm charts that use "helm chart capabilities" fail: https://github.com/rancher/fleet/issues/399

Furthermore, the resourceKey generation does not run against actual, matched target cluster, so deployment options are missing. These options could come from target customizations. The cluster labels are missing, too, so replacement of `global.fleet.clusterLabels.` in helm values fails, which breaks charts that verify their values. 

Ideas: 
* don't treat a failing resourceKey calculation as fatal error (https://github.com/rancher/fleet/pull/1103)
* run resourceKey generation on the [actually matched clusters](https://github.com/manno/fleet/commit/769fcd729eee70edce2eada67b38d95788791a08), which have real deployment options
* optionally disable resourceKey generation for fleet-standalone, as it's only used in the UI
* move helm processing completely to the downstream cluster: https://github.com/rancher/fleet/issues/399

manno:
Fix #1101

### DIFF
--- a/.github/scripts/deploy-fleet.sh
+++ b/.github/scripts/deploy-fleet.sh
@@ -28,3 +28,6 @@ helm upgrade --install fleet charts/fleet \
 kubectl -n cattle-fleet-system rollout status deploy/fleet-controller
 { grep -E -q -m 1 "fleet-agent-local.*1/1"; kill $!; } < <(kubectl get bundles -n fleet-local -w)
 kubectl -n cattle-fleet-system rollout status deploy/fleet-agent
+
+# label local cluster
+kubectl patch clusters.fleet.cattle.io -n fleet-local local --type=json -p '[{"op": "add", "path": "/metadata/labels/management.cattle.io~1cluster-display-name", "value": "local" }]'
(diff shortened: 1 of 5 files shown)


thardeck (approved):
The test does not clean up some remains after the test run, so it stays green even when the fix was removed, but it should not matter since the cluster usually is deleted after each run.

PR #1107, opened 2022-11-18
[v0.5] Update GitHub actions in release v0.5

thardeck:


.github/workflows/e2e-multicluster-ci.yml
```diff
@@ -14,6 +14,9 @@ on:
     paths-ignore:
```
manno, line 14: if we remove the exceptions (because the tests are required either way), we might as well remove these here?

manno: Ah I see, multicluster is not a required test.

PR #1124, opened 2022-11-25
Introduce parameter to debug controller but not agents

moio:
This adds a Helm configuration parameter to selectively debug the fleet controller (without fleet agents).

This contrasts with the current behavior which is to always set the same debug level on both.

Especially in large-scale scenarios, it is undesirable to have all agents restart when temporarily debugging the controller.

## Test

Install with the following Helm options:

```
--set debug=true --set propagateDebugSettingsToAgents=false
```

Or the equivalent `values.yaml` variables.

 - Check that debug logging is enabled on the fleet-controller deployment (there should be a `--debug` line)
 - Check that debug logging is *not* enabled on the fleet-agent deployment (there should *not* be a `--debug` line)

### DIFF
--- a/charts/fleet/templates/deployment.yaml
+++ b/charts/fleet/templates/deployment.yaml
@@ -17,6 +17,8 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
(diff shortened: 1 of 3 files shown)


manno (approved):

manno:
We need to mention this in the next release notes.

PR #1122, opened 2022-11-23
Fix Fleet in Rancher  CI

manno:
Fix #1120

PR #1135, opened 2022-11-29
Revert "Refactoring: avoid needless DeepCopy, the object is not mutated"

thardeck:
This change was preventing Fleet triggering on GitJob changes shown by our "updating a git repository" test.

This reverts commit 866a6159256064aca810c7f3e6fa3fd2e073c37a.


manno (approved):

manno:
I commented on the commit by accident:

> I guess this is because bd is an item from bundleDeployments, err := m.bundleDeploymentCache.List. The cache objects are just pointers and by modifying them we can actually modify the cache. This would also explain the problems in https://github.com/rancher/fleet/pull/1122.
The cache suddenly has the desired state, not the current state.

PR #1138, opened 2022-11-30
fleet apply: allow debugging

moio:
This adds mechanisms to attach a debugger (specifically Delve) to fleet apply.

That means:

1. disabling the security context if DEBUG is set
2. passing over one extra environment variable
3. sitting in a loop at the start of fleet apply to get a chance to attach the debugger

Point 3. is implemented in an admittedly hacky way, but I do not really know of better alternatives at this point.

@manno let me know if this patch is acceptable or not.

### DIFF
--- a/modules/cli/cmds/apply.go
+++ b/modules/cli/cmds/apply.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	"time"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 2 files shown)


manno:
You could run `fleet apply` manually outside the cluster in a terminal to debug it. 
The fleet-examples tests do it: https://github.com/rancher/fleet-examples/blob/master/tests/test.sh

Does it really happen often, that you need to debug a job in a full installation?

moio:
> Does it really happen often, that you need to debug a job in a full installation?

TBH I do not know. I needed it once, hacked together a solution and moved along. I think it's the Team's choice to keep this or discard it - and fair either way.

Issue #712, opened 2022-02-08
EPIC: Managed Fleet Agents should support cluster-specific overrides


aiyengar2:
As described in https://github.com/rancher/rancher/issues/35834#issuecomment-1033004363:

> the configuration of the managed Fleet agent is generated solely from the configuration of Fleet in the local cluster, as seen in the following code: https://github.com/rancher/fleet/blob/d3356b2f029a487236ef96864752a03972717ff9/pkg/controllers/manageagent/manageagent.go#L147-L151
> As a result, if you we were to support supplying agent tolerations here, it would be tolerations that are applied on a global level (e.g. you if you tolerated a taint in one cluster, you would also be tolerating that taint in all your clusters).

Fleet should support the ability to configure some options of managed Fleet agents on a cluster-specific level.

Some examples:
- Ability to set tolerations on a fleet agent: https://github.com/rancher/rancher/issues/35834
- Ability to set a cluster-specific registry: https://github.com/rancher/fleet/issues/711

MKlimuszka:
We want this in 2.6.5

aiyengar2:
Moving to Triage to ensure a milestone is added to the issue

mattfarina:
This is related by different to #869. #869 is set by us (SUSE/Rancher) which this issue has to do with configuration set by the end user. They will be accessed with different settings.

kkaempf:
closing for now for rancher/rancher#35834

PR #1154, opened 2022-12-03
feat: fleet agent deployment configures tolerations from cluster CR

rajiteh:
Related:
- https://github.com/rancher/fleet/pull/703
- https://github.com/rancher/fleet/issues/712
- https://github.com/rancher/rancher/issues/35834

This PR introduces the ability for fleet admins to configure tolerations that must be included in the fleet agent deployment by setting them in each cluster CR's spec.

A new configuration field `.agentTolerations` provide a way to define a list of `corev1.Toleration` objects.

This solves the problem of not being able to use Fleet in clusters which all the nodes are tainted (no schedulable  nodes for agent deployment)

## Additional Information

This solution let's the user/system that manages fleet Cluster resources to configure the tolerations for that particular cluster's agent deployment. 

### Tradeoff

In agent initiated registration flow, the initial agent deployment will need to replicate the same set of tolerations.
ie: if using helm then we would have to set `.kubectl.tolerations` and `fleetAgent.tolerations` fields

However, it can be argued that one could simply set a catch all toleration `- operator: Exists` in the agent initiated flow which guarantees that the agent will get scheduled in the downstream cluster. Once it has communicated with the fleet controller, the new bundle pushed out by the controller will contain the proper tolerations as defined by the Cluster CR which schedule the pods in the appropriate nodes.


manno:
Hey thanks. I guess this will not conflict with the existing tolerations in https://github.com/rancher/fleet/blob/master/pkg/agent/manifest.go#L234-L244?

manno (dismissed):

pkg/agent/manifest.go
```diff
@@ -93,6 +94,10 @@ func Manifest(namespace string, agentScope string, opts ManifestOptions) []runti
 	propagateDebug, _ := strconv.ParseBool(os.Getenv("FLEET_PROPAGATE_DEBUG_SETTINGS_TO_AGENTS"))
 	debug := logrus.IsLevelEnabled(logrus.DebugLevel) && propagateDebug
 	dep := agentDeployment(namespace, DefaultName, image, opts.AgentImagePullPolicy, DefaultName, false, debug)
+
+	// additional tolerations
+	dep.Spec.Template.Spec.Tolerations = append(dep.Spec.Template.Spec.Tolerations, opts.AgentTolerations...)
```
rajiteh, line 99: @manno it will preserve the tolerations defined by the `agentDeployment` func and append anything that the user sent. However, it is possible for the user defined tolerations to override the default ones if they have the same key, but I don't see why someone would do that intentionally as the default ones are rancher namespaced.

rajiteh:
@thardeck @manno it was a little too tough for me to grok how the e2e stuff worked in this repo, I can't seem to figure out if there is any code that tests agent deployment behaviour and I'm too lost to come up with my own.  I decided to add some unit tests to verify the functionality instead. Let me know if that's enough for now. Thanks!

manno (approved):

manno:
@rajiteh I don't think we have anything to really check agent deployment. Our E2E tests mostly apply Gitrepos and use kubectl to check the outcome. 
The only test that comes close is our (basic) "installation test: https://github.com/rancher/fleet/blob/master/.github/workflows/fleet-upgrade.yml#L103

PR #1158, opened 2022-12-06
Cluster values full templating

manno:
PR https://github.com/rancher/fleet/pull/671 added templating to the values section in `fleet.yaml`.

Templating in `fleet.yaml` is not the same as in the `templates/` dir of Helm charts. `fleet.yaml` must be valid YAML. Using the default templating delims `{{`, `}}` leads to errors, if they are not quoted, as the YAML parser interprets them as JSON. Therefore this PR switches the delims to `${`, `}`. This will also help to differentiate one kind of templating from the other, as they don't behave the same in practice, e.g. when using `range` or `with`.

To be able to generate lists and to get rid of the recursion and reduce the number of times `template.Execute` is called, this PR treats the whole `values:` section from `fleet.yaml` as one template. With regards to the inputs from the cluster (`TemplateValues`) resource, I see no problem, because being able to change a `Cluster` resource requires much more privileges than changing a `GitRepo` resource.
The `fleet.yaml` however is already under control of the user, so any structure change in the values is likely intentional. 

Updated docs: https://github.com/rancher/fleet-docs/pull/30
Examples branch for e2e test: https://github.com/rancher/fleet-examples/tree/test-cluster-values/single-cluster/helm-cluster-values


pkg/target/target.go
```diff
@@ -595,73 +605,31 @@ func tplFuncMap() template.FuncMap {
...
-		return nil, fmt.Errorf("templated result was expected to be map[string]interface{}, got %T", tplResult)
-	}
-
-	return compiledYaml, nil
-}
 
-func templateSubstitutions(src interface{}, templateContext map[string]interface{}, tplFn *template.Template, recursionDepth int) (interface{}, error) {
-	if recursionDepth > maxTemplateRecursionDepth {
-		return nil, fmt.Errorf("maximum recursion depth of %v exceeded for current templating operation, too many nested values", maxTemplateRecursionDepth)
+	tmpl := template.New("values").Funcs(tplFuncMap()).Option("missingkey=error").Delims("${", "}")
```
mattfarina, line 614: I wonder if it's worth putting a comment ahead of this to let people know that `{}[],` are yaml flow control characters which causes issues with parsing `{{ }}` as valid YAML using the k8s parser we use. Just to explain the reason in a code comment if people wonder why.

pkg/target/target.go
```diff
@@ -595,73 +605,31 @@ func tplFuncMap() template.FuncMap {
...
 
-func processTemplateValues(valuesMap map[string]interface{}, templateContext map[string]interface{}) (map[string]interface{}, error) {
-	tplFn := template.New("values").Funcs(tplFuncMap()).Option("missingkey=error")
-	recursionDepth := 0
-	tplResult, err := templateSubstitutions(valuesMap, templateContext, tplFn, recursionDepth)
+func processTemplateValues(helmValues map[string]interface{}, templateContext map[string]interface{}) (map[string]interface{}, error) {
+	data, err := kyaml.Marshal(helmValues)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to unmarshal helm values template: %w", err)
```
mattfarina, line 611: The error says "unmarshal" but the function the error comes from is `Marshal`. Should the error message say unmarshal?

mattfarina (changes requested):
Just the slightest of nits in an error message.

Issue #1161, opened 2022-12-07
SURE-5496 - read-only file system errors from Fleet handler image-sync in Rancher v2.6.9/Fleet v0.4.0


kkaempf:
https://jira.suse.com/browse/SURE-5496

Issue #797, opened 2022-06-16
Image Scanner known hosts issue


wiikip:
I am trying to use image scan feature.
I configured everything as indicated in the documentation, the CRD ImageScan is successfuly created and updated when I push new images to my Docker Registry ( Private ECR ). 
However it looks like fleet cannot update my Git repository ( AWS Code Commit ). The following error appears in the fleet-controller logs:
`level=error msg="error syncing 'fleet-local/fleet-server': handler image-sync: unable to find any valid known_hosts file, set SSH_KNOWN_HOSTS env variable, requeuing"`

I found that the library used to clone the repo and to commit changes was https://github.com/go-git/go-git, the reason of this error seems to be that the lib is trying to find a known_hosts file but no valid file exists cf line 220 of this file https://github.com/go-git/go-git/blob/master/plumbing/transport/ssh/auth_method.go

I tried to add ´known_hosts´ in the ssh secret used when creating GitRepo crd, but it has no effect.

A workaround would be to mount a known_host file in the fleet-controller pod but it implies editing the helm chart.

Anyone has encoutered this issue ? I feel like its linked with SSH cloning of the repo, I will give it a try using https, if someone succeed to have it working with SSH I would be interested

I am using Fleet v0.3.8

Thanks

manno:
The go-git library used by imagescan released a fix for the known hosts issue: https://github.com/go-git/go-git/pull/548
Once they cut a new release we can update fleet.

Note: In fleet only imagescan uses go-git, but checking and cloning repos is done by shelling out to git. (Well there is also some test code that also uses go-git).

PR #1181, opened 2022-12-15
Fix image scan issues

thardeck:
Documentation of image scan: https://fleet.rancher.io/imagescan

---

Use `known_hosts` provided by ssh secret for go-git  the same way it is used for gitjob. This allows SSH based GIT connections.

The SSH private key and `known_hosts` can be provided as `clientSecretName` like this: 
e. g.: `kubectl create -n fleet-local secret generic <CLIENT_SECRET_NAME> --type kubernetes.io/ssh-auth --from-file=ssh-privatekey="<PRIVATE_KEY_PATH>" --from-file=known_hosts="<KNOWN_HOSTS_PATH>"`

Fixes rancher/fleet#797

---
Allow writing to `/tmp` besides having a read only root file system.

Fixes rancher/fleet#1161

### DIFF
--- a/pkg/controllers/image/image.go
+++ b/pkg/controllers/image/image.go
@@ -214,6 +214,8 @@ func (h handler) onChangeGitRepo(gitrepo *v1alpha1.GitRepo, status v1alpha1.GitR
 
 	logrus.Debugf("onChangeGitRepo: gitrepo %s/%s changed, syncing repo for image scans", gitrepo.Namespace, gitrepo.Name)
 
+	// This lock is required to prevent conflicts while using the environment variable SSH_KNOWN_HOSTS.
+	// It was added before the SSH support so there might be other potential conflicts without it.
 	lock.Lock()
 	defer lock.Unlock()
 	// todo: maybe we should preserve the dir


charts/fleet/templates/deployment.yaml
```diff
@@ -56,13 +56,17 @@ spec:
           allowPrivilegeEscalation: false
           readOnlyRootFilesystem: true
         {{- end }}
-        {{- if .Values.cpuPprof }}
         volumeMounts:
+          - mountPath: /tmp
+            name: tmp
+        {{- if .Values.cpuPprof }}
```
manno, line 62: I think this is obsolete now, /tmp is already a volume. As far as I can tell `/tmp/pprof` is fine. @moio ?

pkg/controllers/image/image.go
```diff
@@ -278,6 +279,32 @@ func (h handler) onChangeGitRepo(gitrepo *v1alpha1.GitRepo, status v1alpha1.GitR
 	return status, err
 }
 
+func (h handler) setupKnownHosts(gitrepo *v1alpha1.GitRepo, data []byte) error {
+	tmpdir, err := os.MkdirTemp("", gitrepo.Namespace+"-"+gitrepo.Name)
```
manno, line 283: I wouldn't use user input here to construct the temp dir, should be safe and a [random suffix](https://pkg.go.dev/os#MkdirTemp) is added, but still.

```suggestion
	tmpdir, err := os.MkdirTemp("", "fleet-")
```

manno, line 283: We should probably remove that dir when the operation exits?

manno (changes requested):

charts/fleet/templates/deployment.yaml
manno: Ah thanks. I didn't expand the diff :) 👍

pkg/controllers/image/image.go
olblak: Is there any chance to have some sort of collusion here? Like two different process fetching the digest for the same tag at a slightly different type which could lead to different digest for the same tag?

pkg/controllers/image/image.go
```diff
@@ -278,6 +287,32 @@ func (h handler) onChangeGitRepo(gitrepo *v1alpha1.GitRepo, status v1alpha1.GitR
...
 }
 
+func (h handler) setupKnownHosts(gitrepo *v1alpha1.GitRepo, data []byte) error {
+	tmpdir, err := os.MkdirTemp("", fmt.Sprintf("ssh-%s-%s-", gitrepo.Namespace, gitrepo.Name))
+	if err != nil {
+		return err
+	}
+
+	known := path.Join(tmpdir, "known_hosts")
+	err = os.Setenv("SSH_KNOWN_HOSTS", known)
```
manno, line 297: The controller is started from a goroutine (I think this code in lasso starts controllers: https://github.com/rancher/lasso/blob/master/pkg/controller/controller.go#L121-L156).

Goroutines share environment variables, as they are the in the same OS process.  I couldn't find any explicit documentation for this, however https://go.dev/play/p/e0L5ixUe4qL illustrates the issue.

By setting the variable `SSH_KNOWN_HOSTS` to a temp file with a random name, we could in theory fail a concurrent second image scan. I think this isn't too unlikely if the interval is low enough and the SSH connection is slow enough.

As a workaround, we could add a lock around this code (which is not good for performance) or pass the known hosts to go-git without the env var somehow?

pkg/controllers/image/image.go
```diff
@@ -278,6 +287,32 @@ func (h handler) onChangeGitRepo(gitrepo *v1alpha1.GitRepo, status v1alpha1.GitR
 	return status, err
 }
 
+func (h handler) setupKnownHosts(gitrepo *v1alpha1.GitRepo, data []byte) error {
+	tmpdir, err := os.MkdirTemp("", fmt.Sprintf("ssh-%s-%s-", gitrepo.Namespace, gitrepo.Name))
```
manno, line 291:
```suggestion
	tmpdir, err := os.MkdirTemp("", fmt.Sprintf("fleet-ssh-%s-%s-", gitrepo.Namespace, gitrepo.Name))
```

So we can easily remove the dirs locally, without interfering with other temp files?

manno (approved):

PR #1190, opened 2022-12-19
Fix CRD download url

olblak:
It appears that automation introduced in https://github.com/rancher/fleet/pull/1182 broken the CRD url


raulcabello (approved):
LGTM

Issue #1115, opened 2022-11-21
Investigate Rancher 2.5->2.6 upgrade


manno:

manno:
I upgraded from 2.5.15 to 2.6.9 and everything looked fine. No duplicate agents.

manno:
I was not able to observe multiple agents. 
However it's impossible to test all combinations. According to the release notes, some releases require manual intervention. Maybe that was the issue with the Rancher updates?


In any case, the workflow https://github.com/rancher/fleet/pull/1198 shows, fleet is backwards compatible and can be updated without problems even in old versions of Rancher.

manno:
Check:

- [x] 2.5.15 -> 2.6.9 - ok
- [ ] 2.5.9 -> 2.6.3 - leftover fleet-system?
- [ ] 2.5.10 -> 2.6.3

manno:
Upgrading from 2.5.9 results in a left over agent deployment in the fleet-system namespace.

Upgrade from 2.5.16 does not, instead the config is not copied resulting in https://github.com/rancher/rancher/issues/40127

kkaempf:
Is this still blocked ?

olblak:
@kkaempf I see that you target this issue to be solved in milestone [2023-Q2-v2.6x](https://github.com/rancher/fleet/milestone/22) but considering the latest pullrequeset mentioning this issue, it appears that it has been validated by QA for 2.6.11.

So my understanding is this issue can be closed

kkaempf:
> @kkaempf I see that you target this issue to be solved in milestone [2023-Q2-v2.6x](https://github.com/rancher/fleet/milestone/22) but considering the latest pullrequeset mentioning this issue, it appears that it has been validated by QA for 2.6.11.
> 
> So my understanding is this issue can be closed

@manno do you agree ?!

manno:
We'll have to make sure the release notes I wrote for 0.6.0 appear on the final release. Since they are already written https://github.com/rancher/fleet/releases/tag/v0.6.0-rc.2 we can close this.

PR #1198, opened 2022-12-19
Fleet in Rancher Upgrade

manno:
refers to #1115

### DIFF
--- a/.github/scripts/label-downstream-cluster.sh
+++ b/.github/scripts/label-downstream-cluster.sh
@@ -2,6 +2,8 @@
 
 set -euxo pipefail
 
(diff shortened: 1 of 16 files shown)


manno:
Installation is still flaky.

For example: https://github.com/manno/fleet/actions/runs/3767060471

* one job needed 15 minutes to setup rancher and register a downstream cluster, normally needs 5 minutes
* v2.6.9 flakes out in a multi-cluster test for target customizations
* flakes out with network issues during multi cluster test (gha performance issue?)

kkaempf (approved):
👍🏻 
You fix it if it breaks 😆

manno:
Indeed, the new workflow is not very stable yet. Still, the updated scripts are useful and we have a follow-up issue.

Issue #1141, opened 2022-11-30
Bundle State and Bundle Diff


manno:
Resources created from controllers show up as orphaned and can't be ignored. Bundle diff don't work for deleted resources.

* SURE-2739
* SURE-2811
* SURE-3257
* SURE-4466
* SURE-4968
* rancher/fleet#313
* rancher/fleet#327
* rancher/fleet#461

Mentioned:
* https://github.com/rancher/fleet/issues/726
* https://github.com/rancher/fleet/issues/572

raulcabello:
There is a problem with new resources created by an operator if the operator copies the `objectset.rio.cattle.io/hash` label.
In this case wrangler plan thinks this new resource belongs to the bundle deployment which is not right. Therefore the resource is marked as `Orphaned` and status will never be `Ready`.

Example

Create a `GitRepo` with https://github.com/raulcabello/fleet-test/tree/main/ingress. It contains a `Certificate`, `Issuer` and an `Ingress`. `cert-manager` will create a new `Certificate` called `myingress-cert` for this ingress because it has this [annotation](https://github.com/raulcabello/fleet-test/blob/main/ingress/ingress.yaml#L6). `myingress-cert` does not belong to the helm release, however wrangler returns this certificate in the [plan.Delete](https://github.com/rancher/fleet/blob/master/modules/agent/pkg/deployer/manager.go#L106). This is happening because cert-manager [copies the labels](https://github.com/cert-manager/cert-manager/blob/master/pkg/controller/certificate-shim/sync.go#L360) from the `Ingress` into the new `Certificate`, therefore this new `Certificate` contains the label  `objectset.rio.cattle.io/hash`

raulcabello:
https://github.com/rancher/fleet/issues/327 is similar to the one I described above. In this case the `prometheus-operator` is creating 2 statefulsets that contains `objectset.rio.cattle.io/hash` label and `objectset.rio.cattle.io/id` annotation. The operator adds all the [labels and annotations](https://github.com/prometheus-operator/prometheus-operator/blob/19ca0371f3185e46c8f886e3ed43d7d8783d375b/pkg/alertmanager/statefulset.go#L94-L118) to the `Statefulsets` from the `AlertManager` resource which is created in the bundle deployment, therefore it contains the `objectset.rio.cattle.io/hash` label. Wrangler thinks this two resources were created by the bundle because they contain the `objectset.rio.cattle.io/hash` label.

raulcabello:
[...]

PR #1201, opened 2022-12-20
Fix orphan resources created by operators

raulcabello:
It is possible that some operators copy the `objectset.rio.cattle.io/hash` label into a dynamically created objects. Wrangler plan will look for all resources with that label, and will think they all belong to the release. However they should not be considered part of the release. We need to skip these resources because they are not part of the release, and they would appear as orphaned.

For that, we look into the previous release to see if the resource we have to delete was already present. If it was not present it means it will not be deleted, and should not be part of the release so we just skip it.

This [test](https://github.com/rancher/fleet/commit/65961b344ef6d72e052eb28a3807b1964acd79a2#diff-5d4640ccdcecacd353830fb273e4be6e72f5c1a645e8bebc35f8117c87eb5fbcR143-R164) simulates the problem. It is failing without the fix in this PR

Fix https://github.com/rancher/fleet/issues/1141

### DIFF
--- a/integrationtests/agent/bundle_deployment_status_test.go
+++ b/integrationtests/agent/bundle_deployment_status_test.go
@@ -11,6 +11,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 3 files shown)


manno (approved):

dvarrazzo:
Hello,

I am currently experiencing this problem on a cluster running fleet agent 0.5.1, in conjunction with akv2k8s, where the CRD akvs creates a Secret including the `objectset.rio.cattle.io/hash` header. However it doesn't happen with all these resources: currently only 1 over 9 pairs of resources.

This Secret has the issue, here with the akvs creating it, details omitted:

```
apiVersion: spv.no/v1
kind: AzureKeyVaultSecret
metadata:
  name: xxx-keycloak
  labels:
    objectset.rio.cattle.io/hash: 7939fba2bab3b3ab2d58b8b8b722be49905eebea
  uid: d5a38aa1-1032-42a8-a969-090dc904c513
spec:
  ...

apiVersion: v1
kind: Secret
metadata:
  labels:
    objectset.rio.cattle.io/hash: 7939fba2bab3b3ab2d58b8b8b722be49905eebea
  name: xxx-keycloak
  ownerReferences:
  - apiVersion: spv.no/v2beta1
    blockOwnerDeletion: true
    controller: false
    kind: AzureKeyVaultSecret
    name: xxx-keycloak
    uid: d5a38aa1-1032-42a8-a969-090dc904c513
  uid: d8d007c9-fa30-416e-a465-4431ea62df09
```

This resource doesn't have the problem:

```
apiVersion: spv.no/v1
kind: AzureKeyVaultSecret
metadata:
  labels:
    objectset.rio.cattle.io/hash: 0cb9cf069d9ad582bd0259c9f7e83031e77fe53a
  name: keycloak-admin
  uid: e14505b6-8770-4cf3-be4a-861220b70a73
spec:
  ...

apiVersion: v1
kind: Secret
metadata:
  name: keycloak-admin
  annotations:
  labels:
    objectset.rio.cattle.io/hash: 0cb9cf069d9ad582bd0259c9f7e83031e77fe53a
  ownerReferences:
  - apiVersion: spv.no/v2beta1
    blockOwnerDeletion: true
    controller: false
    kind: AzureKeyVaultSecret
    name: keycloak-admin
    uid: e14505b6-8770-4cf3-be4a-861220b70a73
  uid: ff23d6bc-8f1e-43cd-8668-ef07a90a316b
```

I don't see any relevant difference between the working one and the orphaned one. There seem to be no error in the fleet-agent log, apart from "bundle xxx: secret.v1 xxx/xxx-keycloak extra". The state of the bundle is **Modified(1) [Bundle xxx]; secret.v1 xxx/xxx-keycloak extra**. 

Has the fix in #1141 been released?

PR #1204, opened 2022-12-20
Fix comment about bootstrap gitrepo

manno:
copied the text from the helm chart's values.yaml

### DIFF
--- a/pkg/controllers/bootstrap/bootstrap.go
+++ b/pkg/controllers/bootstrap/bootstrap.go
@@ -102,7 +102,9 @@ func (h *handler) OnConfig(config *config.Config) error {
 		},
 	})
 
-	// in case the agent is to be deployed from git
+	// A repo to add at install time that will deploy to the local cluster. This allows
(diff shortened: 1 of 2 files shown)


pkg/controllers/bootstrap/bootstrap.go
```diff
@@ -102,7 +102,9 @@ func (h *handler) OnConfig(config *config.Config) error {
 		},
 	})
 
-	// in case the agent is to be deployed from git
+	// A repo to add at install time that will deploy to the local cluster. This allows
+	// one to fully bootstrap fleet, it's configuration and all it's downstream clusters
```
kkaempf, line 106:
```suggestion
	// one to fully bootstrap fleet, its configuration and all its downstream clusters
```

kkaempf (changes requested):

pkg/controllers/bootstrap/bootstrap.go
manno: Well...

raulcabello (approved):

PR #1206, opened 2022-12-20
Remove fleet-agent simulators

manno:
Clean up fleet-agent simulators. I'm not aware of anyone who uses this.

@moio you looked into fleet performance, are simulators useful, or could they be useful in the future?


moio:
I did not look into simulators yet, and have no short term plans to do so. I also tend not to like abandoned rotting code, so I am personally fine in removing them now - they are always one `git revert` away in case they turn out to be useful in future.

PR #1207, opened 2022-12-20
Add flag to disable local cluster bootstrap

manno:


raulcabello (commented):
Can we add a condition to not create bootstrap [ `ClusterRole` and `ClusterRoleBinding`](https://github.com/rancher/fleet/blob/master/charts/fleet/templates/rbac.yaml#L88-L112)  and [ `ServiceAccount`](https://github.com/rancher/fleet/blob/master/charts/fleet/templates/serviceaccount.yaml#L7-L10) if bootstrap is disabled? I don't think they'll ever be needed if bootstrap is disabled.

raulcabello (approved):

PR #1213, opened 2022-12-21
Fix gke e2e tests

raulcabello:
[e2e are failing](https://github.com/rancher/fleet/actions/runs/3736842647/jobs/6341532628) because of a warning returned by the gke github action. This is because we  are generating credentials via gcloud container clusters get-credentials. Install `gke-gcloud-auth-plugin` to fix this.

More info https://github.com/google-github-actions/setup-gcloud/issues/561#issuecomment-1137363453


raulcabello:
e2e tests are green https://github.com/rancher/fleet/actions/runs/3749057172

PR #1228, opened 2023-01-02
Fix regression in imagescan test

thardeck:
by making knownHostsPath definition setting the global variable.


raulcabello (approved):
LGTM

Issue #978, opened 2022-09-15
Feature Request: Better control of when helmSecretName is used


uivraeus:
### Is your feature request related to a problem?

Yes.

We specify `GitRepo` definitions with `helmSecretName` to provide (reference to) credentials for accessing private chart registries. In these `GitRepo` definitions, we refer to several different paths to monitor, each containing its own "fleet.yaml".

This works as long as each "fleet.yaml" refers to the same _intended_ (private) registry source, but it is too easy to make a mistake and include a chart reference to some other registry (public or private). If such a mistake is made, the credentials are automatically sent in the request to that registry, which is really bad from a security perspective.

### Solution you'd like

Make it possible to specify applicability/scope for the credentials referred to by `helmSecretName`, e.g. a  URL or hostname regex.

This could be done by introducing another key next to `helmSecretName`, e.g. `helmSecretScope` or similar.

When this has been defined, the chart fetching in Fleet (`go-getter`?) should only provide credentials if the source matches what has been specified. Otherwise, no credentials must be provided.

### Alternatives you've considered

The scope (i.e. regex) for each set of credentials could also be defined via some custom annotations in the Secrets object, which would imply that no new key has to be introduced in the `GitRepo` definition.

mattfarina:
Can you please share what version of Rancher and Fleet you are experiencing this problem with? [This was fixed in Helm 3.6.1](https://github.com/helm/helm/security/advisories/GHSA-56hp-xqp3-w2jf). Fleet uses the Helm SDK. We need to make sure this issue isn't still present in the latest Fleet.

uivraeus:
@mattfarina I don't think anything has changed wrt. the situation I described above and I have verified that it's still possible to reproduce with Fleet v. 0.5.0.

But maybe I can clarify a bit;

The scenario I try to describe involves fetching of a chart (tgz-file) via the general "gogetter" support for the `chart`  key in `fleet.yaml`. 

Let's imagine a `GitRepo` definition like this:

```yaml
spec:
  # ...
  helmSecretName: helmsecret-dummy
  paths:
    - releases/gogetter-simple
    - releases/app-example
  # ...
```
The `helmSecretName` is needed for accessing `private-registry.com` (see below)

Content of `releases/app-example/fleet.yaml`:

```yaml
helm:
  # ...
  chart: oci://private-registry.com/example/charts/spa-example
  version: 0.7.0
```

Content of `releases/gogetter-simple/fleet.yaml`:

```yaml
helm:
[...]

PR #1234, opened 2023-01-09
Add helmRepoUrlRegex to the GitRepo

raulcabello:
Add `helmRepoUrlRegex` to the `GitRepo` CRD and add a new parameter `helm-repo-url-regex` to the fleet apply cli comand. Helm credentials will be passed to the helm repository just if the helm repo url matches the `helmRepoUrlRegex`. That way we can control which helm repositories should receive credentials. Helm credentials will be passed to all helm repositories if `helmRepoUrlRegex` is not provided, which is the current behaviour.

This PR introduces integration tests for the fleet cli. For the helm tests it creates a helm repository running on localhost:3000. 

Fix https://github.com/rancher/fleet/issues/978


charts/fleet-crd/templates/crds.yaml
```diff
@@ -2103,6 +2103,9 @@ spec:
                 type: string
               forceSyncGeneration:
                 type: integer
+              helmRepoUrlRegex:
```
manno, line 2106: I'd prefer helmRepoURLRegex to be consistent with the other spellings in that file

integrationtests/cli/assets/helmrepository/testrepo/templates/NOTES.txt
```diff
@@ -0,0 +1,22 @@
+1. Get the application URL by running these commands:
```
manno, line 1: This seems to be the default chart template from helm. It doesn't seem like the tests need this. Except for the file check against 'svc.yaml' and 'deployment.yaml'.

Can we use/copy a smaller chart, e.g. the "config-chart" from https://github.com/rancher/fleet-examples/commit/f1be9e1bd0387fb6ec0df35f38b147a7016937e6 ?

integrationtests/cli/helm/repository.go
```diff
@@ -0,0 +1,63 @@
...
+	}()
+}
+
+func getHandler(authEnabled bool) http.Handler {
+	fs := http.FileServer(http.Dir(cli.AssetsPath + "helmrepository"))
+	if !authEnabled {
+		return fs
+	}
+	return &authHandler{fs: fs}
+
```
manno, line 40:
```suggestion
```

integrationtests/cli/helm/helm_test.go
```diff
@@ -0,0 +1,160 @@
...
+	numberOfFilesInHelmTestRepo = 9
+)
+
+var repo = repository{}
+
+var _ = Describe("Fleet apply helm release", Ordered, func() {
+	When("apply a folder with fleet.yaml that contains a helm release", func() {
+		When("no auth is required", func() {
+			BeforeAll(func() {
+				repo.startRepository(false)
```
manno, line 25: I'm not sure why this works 🤷 
Tests are executed in parallel and repo always uses the same port? Shouldn't this run into a port conflict.

pkg/bundlereader/resources.go
```diff
@@ -152,9 +152,17 @@ func mergeGenericMap(first, second *fleet.GenericMap) *fleet.GenericMap {
...
 // For every chart that is not on disk, create a directory struct that contains the charts URL as path.
-func addRemoteCharts(directories []directory, base string, charts []*fleet.HelmOptions, auth Auth) ([]directory, error) {
+func addRemoteCharts(directories []directory, base string, charts []*fleet.HelmOptions, auth Auth, helmRepoUrlRegex string) ([]directory, error) {
 	for _, chart := range charts {
 		if _, err := os.Stat(filepath.Join(base, chart.Chart)); os.IsNotExist(err) || chart.Repo != "" {
+			shouldAddAuthToRequest, err := shouldAddAuthToRequest(helmRepoUrlRegex, chart.Repo)
+			if err != nil {
+				return nil, err
+			}
+			if !shouldAddAuthToRequest {
```
manno, line 162: I wonder if it's easier to read if we return a "removeAuth", but it reading it again I think this is fine.

integrationtests/cli/helm/helm_test.go
raulcabello: We are using [Ordered](https://onsi.github.io/ginkgo/#ordered-containers), which will not run tests in parallel to avoid this problem

integrationtests/cli/helpers.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+func IsResourcePresentInBundle(resourcePath string, resources []v1alpha1.BundleResource) (bool, error) {
+	resourceFile, err := os.ReadFile(resourcePath)
+	if err != nil {
+		return false, err
+	}
+
+	for _, resource := range resources {
+		if strings.ReplaceAll(resource.Content, "\n", "") == strings.ReplaceAll(string(resourceFile), "\n", "") {
```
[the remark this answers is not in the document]
raulcabello: yes, I think so

charts/fleet-crd/templates/crds.yaml
raulcabello: done in https://github.com/rancher/fleet/pull/1234/commits/9434f279501df414114de8b0447bfb899db9c871

manno (approved):
👍

PR #1267, opened 2023-01-23
Fix Rancher upgrade

raulcabello:
`CattleNamespaceMigrated` `Cluster` status field is true in Rancher v2.5.16 and fleet 3.9 even though it is still using the old namespace `fleet-system`. That's a problem when migrating to the new namespace `cattle-fleet-system` as downstream clusters thinks migration already happened, and registration was already done because [agentDeployed](https://github.com/rancher/fleet/blob/v0.5.1/pkg/controllers/cluster/import.go#L170) returns true. Because of that, it's missing the `fleet-agent` secret that contains the `kubeconfig` in the `cattle-fleet-system` namespace in the downstream cluster. Therefore, new downstream `fleet-agents` can't connect to the upstream cluster.

Old agent is not removed in downstream clusters because `cluster.Spec.AgentNamespace` is always empty [here](https://github.com/rancher/fleet/blob/v0.5.1/pkg/controllers/cluster/import.go#L278) for downstream clusters. `cluster.Spec.AgentNamespace` is not empty for the local fleet agent as it is set by Rancher.

This PR sets `CattleNamespaceMigrated` to false if the agent status namespace is `fleet-system`, so the new `fleet-agent` is registered with the upstream cluster. It also removes the `cluster.Spec.AgentNamespace` should be empty condition, so old fleet-agents are deleted. 

Fix https://github.com/rancher/rancher/issues/40127

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -21,6 +21,7 @@ const (
 	AgentBootstrapConfigName = "fleet-agent-bootstrap"
 	Key                      = "config"
 	DefaultNamespace         = "cattle-fleet-system"
+	LegacyDefaultNamespace   = "fleet-system"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


raulcabello:
replaced by https://github.com/rancher/fleet/pull/1269

PR #1269, opened 2023-01-24
Fix Rancher upgrade

raulcabello:
`CattleNamespaceMigrated` `Cluster` status field is true in Rancher v2.5.16 and fleet 3.9 even though it is still using the old namespace `fleet-system`. That's a problem when migrating to the new namespace `cattle-fleet-system` as downstream clusters thinks migration already happened, and registration was already done because [agentDeployed](https://github.com/rancher/fleet/blob/v0.5.1/pkg/controllers/cluster/import.go#L170) returns true. Because of that, it's missing the `fleet-agent` secret that contains the `kubeconfig` in the `cattle-fleet-system` namespace in the downstream cluster. Therefore, new downstream `fleet-agents` can't connect to the upstream cluster.

Old agent is not removed in downstream clusters because `cluster.Spec.AgentNamespace` is always empty [here](https://github.com/rancher/fleet/blob/v0.5.1/pkg/controllers/cluster/import.go#L278) for downstream clusters. `cluster.Spec.AgentNamespace` is not empty for the local fleet agent as it is set by Rancher. This condition was introduced to support multiple fleet-agents in the same cluster for Rancher managing Rancher. I think it is safe to remove it and just leave `cluster.Status.Agent.Namespace != agentNamespace || !cluster.Status.AgentNamespaceMigrated`. 
 

This PR sets `CattleNamespaceMigrated` to false if the agent status namespace is `fleet-system`, so the new `fleet-agent` is registered with the upstream cluster. It also removes the `cluster.Spec.AgentNamespace` should be empty condition, so old fleet-agents are deleted. 

Fix https://github.com/rancher/rancher/issues/40127 and https://github.com/rancher/fleet/issues/1115

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -20,10 +20,10 @@ const (
 	AgentConfigName          = "fleet-agent"
 	AgentBootstrapConfigName = "fleet-agent-bootstrap"
 	Key                      = "config"
-
 	// DefaultNamespace is the default for the system namespace, which
 	// contains the manager and agent
-	DefaultNamespace = "cattle-fleet-system"
+	DefaultNamespace       = "cattle-fleet-system"
+	LegacyDefaultNamespace = "fleet-system"
 )
 
 var (
--- a/pkg/controllers/cluster/import.go
+++ b/pkg/controllers/cluster/import.go
@@ -170,6 +170,10 @@ func (i *importHandler) deleteOldAgent(cluster *fleet.Cluster, kc kubernetes.Int
 
 // importCluster is triggered for manager initiated deployments and the local agent,
 func (i *importHandler) importCluster(cluster *fleet.Cluster, status fleet.ClusterStatus) (_ fleet.ClusterStatus, err error) {
+	if cluster.Status.Agent.Namespace == config.LegacyDefaultNamespace {
+		cluster.Status.CattleNamespaceMigrated = false
+	}
+
 	if cluster.Spec.KubeConfigSecret == "" ||
 		agentDeployed(cluster) ||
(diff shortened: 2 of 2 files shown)


manno (approved):

manno (approved):

raulcabello:
@rancher/qa
## Additionnal QA
### Issue: https://github.com/rancher/rancher/issues/40127 and https://github.com/rancher/fleet/issues/1115
### Problem
New `fleet-agent` is not registered with the upstream cluster when is migrating from the `fleet-system` to the `cattle-fleet-system` namespace. And fleet-agent in the `fleet-system` namespace is not removed.

### Solution
Set `CattleNamespaceMigrated` to false if the agent status namespace is `fleet-system`, so the new `fleet-agent` is registered with the upstream cluster. Remove the `cluster.Spec.AgentNamespace` should be empty condition, so old fleet-agents are deleted. 
 
### Testing
### Engineering Testing
#### Manual Testing
I have installed Rancher 2.5.9, then upgrade to 2.6.9. fleet-agent was deleted from the fleet-system namespace, and fleet-agent didn't show any error in cattle-fleet-system-namespace.
I test the same upgrading from 2.5.16 to 2.6.9
I imported a Rancher upstream cluster as downstream cluster to test Rancher within Rancher where 2 fleet-agents are present.

### QA Testing Considerations
The condition I removed was introduced to support multiple agents in the same clusters. I tested it importing a Rancher upstream cluster as a downstream cluster of another Rancher installation and I could not find any problem. Please keep this in mind and try to test all possible scenarios where there can be multiple agents.

PR #1273, opened 2023-01-25
Long release names, shorten only once

### LINKED ISSUE #1272: Long bundle names will still be cleaned up continuously
SURE-4160

This is a follow up to https://github.com/rancher/fleet/issues/523

Under certain conditions a long bundle name might be shortened again, resulting in a different checksum suffix. The agent will then clean up the helm installation and install again.

manno:
Fix #1272

Long release names are shortened and a md5 hash is appended. The
existing code would shorten a release name twice. Once when creating the
bundle name and again when creating the helm release name.
This happened because name.Limit(str, 53) shortens any string smaller
than 53 chars to 53 chars. It will modify a string of 53 chars, so it's
not stable when calling it multiple times.

When the release name is different from the bundle name, fleet will
clean up such a release as it no longer belongs to a bundle.

### DIFF
--- a/e2e/single-cluster/release_names_test.go
+++ b/e2e/single-cluster/release_names_test.go
@@ -14,6 +14,19 @@ var _ = Describe("ReleaseName", func() {
 	var (
 		asset string
 		k     kubectl.Command
+
+		// namespaces from test data in single-cluster/release-names.yaml
+		namespaces = []string{
+			"workloadns1",
+			"workloadns2",
+			"workloadns4",
+			"workloadns5",
+			"workloadns6",
+			"workloadns7",
+			"workloadns8",
+			"workloadns9",
+			"workloadns10",
+		}
 	)
 
 	BeforeEach(func() {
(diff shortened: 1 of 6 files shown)


raulcabello (commented):
> This happened because name.Limit(str, 53) shortens any string smaller
than 53 chars to 53 chars. It will modify a string of 53 chars, so it's
not stable when calling it multiple times.

[name.Limit](https://github.com/rancher/wrangler/blob/master/pkg/name/name.go#L40-L42) returns the same string if it is smaller than 53 chars, so I don't think that's the reason for the release name to be different from the bundle name.

I'm ok with removing the `name.Limit(str, 53) ` as bundle name should never [be bigger than 53](https://github.com/rancher/fleet/blob/master/modules/cli/apply/apply.go#L218-L226)

I would be nice to add a bundle with a name bigger than 53 in the tests and assert the name is shorten and that it matches the helm release

manno:
> [name.Limit](https://github.com/rancher/wrangler/blob/master/pkg/name/name.go#L40-L42) returns the same string if it is smaller than 53 chars, so I don't think that's the reason for the release name to be different from the bundle name.

The problem is they use a "smaller than", not "smaller or equal than", example: https://go.dev/play/p/NtEoe1jw8dc
 
> I would be nice to add a bundle with a name bigger than 53 in the tests and assert the name is shorten and that it matches the helm release

This is what the e2e test I extended does. It's using this repo which contains strange and long release names: https://github.com/rancher/fleet-examples/tree/test-bundle-names

raulcabello:
> The problem is they use a "smaller than", not "smaller or equal than", example: https://go.dev/play/p/NtEoe1jw8dc

ah I see the problem now, I got confused with sentence `This happened because name.Limit(str, 53) shortens any string smaller
than 53 chars to 53 chars`. 
bundle name gets shorten to 53 chars if bigger, then `name.Limit(str, 53)`  will change the name for the release, then cleanup will happen

> This is what the e2e test I extended does. It's using this repo which contains strange and long release names: https://github.com/rancher/fleet-examples/tree/test-bundle-names

:+1:  I missed that one

raulcabello (approved):
LGTM, thanks!

Issue #1133, opened 2022-11-29
Clean up Labels and Annotations


manno:
- [x] Annotation `fleet.cattle.io/managed` appears unused, label of the same name is in use
- [x] Label fleet.cattle.io/bootstrap-token is unused
- [x] Annotations `fleet.cattle.io/cluster`, `fleet.cattle.io/cluster-namespace`  are never used on a `Cluster` resource

manno:
https://fleet.rancher.io/next/ref-configuration#labels

PR #1310, opened 2023-02-07
Clean up orphaned request service accounts, remove unused managed annotations

thardeck:
Fix rancher/fleet#1133

I have also checked [rancher/rancher](https://github.com/rancher/rancher) and [rancher/ui](https://github.com/rancher/ui) for usage of the supposedly obsolete annotations and labels.

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -105,9 +105,6 @@ func agentToken(ctx context.Context, agentNamespace, controllerNamespace string,
 		&v1.Namespace{
 			ObjectMeta: metav1.ObjectMeta{
 				Name: agentNamespace,
-				Labels: map[string]string{
-					fleet.ManagedLabel: "true",
-				},
 			},
 		},
 		// fleet-agent-bootstrap


pkg/agent/agent.go
```diff
@@ -105,9 +105,6 @@ func agentToken(ctx context.Context, agentNamespace, controllerNamespace string,
 		&v1.Namespace{
 			ObjectMeta: metav1.ObjectMeta{
 				Name: agentNamespace,
-				Annotations: map[string]string{
```
manno, line 108: I wonder if this should be a label instead.

The [clean up controller](https://github.com/rancher/fleet/blob/master/pkg/controllers/cleanup/handler.go#L117) will purge resources and namespaces when the "managed" label is "true".

If this namespace has no managed=true label, it will be left behind. Actually, it should already be left behind, because the label is set as an annotation instead.

pkg/controllers/clusterregistration/controller.go
```diff
@@ -160,9 +160,6 @@ func (h *handler) authorizeCluster(sa *v1.ServiceAccount, cluster *fleet.Cluster
 			Labels: map[string]string{
 				fleet.ClusterAnnotation: cluster.Name,
 			},
-			Annotations: map[string]string{
-				fleet.ManagedLabel: "true",
```
manno, line 164: Same as above. This secret will not be purged, maybe it was put into the annotations list by accident? Are we leaving cluster registration secrets behind?

pkg/controllers/clusterregistration/controller.go
```diff
@@ -235,7 +232,6 @@ func (h *handler) OnChange(request *fleet.ClusterRegistration, status fleet.Clus
 				Name:      saName,
 				Namespace: cluster.Status.Namespace,
 				Annotations: map[string]string{
-					fleet.ManagedLabel:                           "true",
```
manno, line 238: Same for the `request-` service accounts? 🤷

manno (changes requested):

pkg/agent/agent.go
```diff
@@ -105,7 +105,7 @@ func agentToken(ctx context.Context, agentNamespace, controllerNamespace string,
...
+				Labels: map[string]string{
```
manno, line 108: We discussed and it's probably safer to remove this as we don't want to delete the agent namespace ("cattle-fleet-local-system") if has no clusters.

manno (approved):

PR #1340, opened 2023-02-20
Simplify Go Container reference

thardeck:
to just reference the Go version and not the specific container base so always the latest container for the specified go version is used which prevents usless go container bump prs.

For relevant BCI image updates we have to bump Fleet anyway so a new build/release process needs to be triggered in any case.


raulcabello (approved):

olblak (approved):
btw 1.20.1 is the latest one now :)

Issue #680, opened 2021-12-09
Being able to keep resources when gitRepo is deleted


StrongMonkey:
Today when gitRepo is deleted, assciated bundles and underlying resources are being deleted. This can cause cascading failure if user deletes their gitRepo accidently, or removed their bundles. A possible solution is that we can keep the resource by default when gitRepo/bundle is deleted, and only clean up resources when there is `purge: true` set on gitRepo.

erSitzt:
Not sure if it fits here, but my issue here https://github.com/rancher/rancher/issues/35987 causes similar problems... and should not delete anything without a warning/confirmation.
Im not sure im manual removal of a label has the same result, because this happens to cause an error with no further information.

0x4c6565:
This would be massively beneficial for those using 'App of Apps' pattern (`GitRepo` resource creating more `GitRepo` resources). Currently when this `GitRepo` is deleted, all child `GitRepo` resources (and their bundles) are removed.

puffitos:
This would also be beneficial in case you don't want the resources to get deleted. Image if you're managing CRDs with Fleet and then for some reason you have to make a change to the way you're managing the CRDs, or just wanna stop managing them with Fleet. Deleting the GitRepo would delete all CRDs and **all** custom resources based on those CRDs... 

An option to be able to "orphan" the resources and then bind them to another GitRepo would be ideal :)

atsai1220:
I wonder if one could add annotations to resources so Fleet does not delete them (since it is using Helm underneath Tekton).

https://helm.sh/docs/howto/charts_tips_and_tricks/#tell-helm-not-to-uninstall-a-resource

```yaml
kind: Secret
metadata:
  annotations:
    "helm.sh/resource-policy": keep
```

raulcabello:
@atsai1220 yes you can use the `helm.sh/resource-policy: keep` annotation to keep the resources. Keep in mind that the `helm.sh/resource-policy: keep` annotation is currently only used if it is included in the rendered manifests of the last deployed release. That means that executing `kubectl annotate <type> <name> helm.sh/resource-policy=keep` directly will not work.

atsai1220:
@raulcabello ty!

We used that annotation on CRDs that are part of the chart and has worked well.

This may not work well if CRDs are part of a job generated by the chart.

Lastly, we were able to remove the gitrepo and keep the resources by using these manual steps:
1. Scale down fleet-agent on down stream clusters
[...]

PR #1345, opened 2023-02-24
Allow resources to be kept after deleting GitRepo 

raulcabello:
Add new field `keepResources` to `GitRepo` CRD. Resources will be deleted when deleting a `GitRepo` if `keepResources` is not provided or is false, which is the current behaviour. 
If `keepResources` is true, fleet will add the [helm.sh/resource-policy](https://helm.sh/docs/howto/charts_tips_and_tricks/#tell-helm-not-to-uninstall-a-resource) annotation to all resources in the post renderer. This annotation will not remove resources when doing a helm uninstall. When we delete a `GitRepo` fleet does a helm uninstall, therefore resources will not be deleted.

`keepResources` can also be provided in the `fleet.yaml`. In this case it will apply just to the `Bundle` created for this `fleet.yaml`

A new parameter `keep-resources` is added to the fleet cli.

e2e tests and integration test for the fleet cli added

Fix https://github.com/rancher/fleet/issues/680

### DIFF
--- a/.github/workflows/e2e-ci.yml
+++ b/.github/workflows/e2e-ci.yml
@@ -96,6 +96,11 @@ jobs:
           FLEET_E2E_NS: fleet-local
         run: |
           ginkgo e2e/single-cluster
+      - name: E2E Tests for keep-resources
+        env:
+          FLEET_E2E_NS: fleet-local
+        run: |
+          ginkgo e2e/keep-resources
       -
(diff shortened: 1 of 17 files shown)


raulcabello:
Please don't review it yet. Removing resources in an upgrade don't work with this approach, they won't be deleted because of the `helm.sh/resource-policy` annotation. I'm working on a fix for this

PR #1346, opened 2023-02-24
Add Github action to release Fleet against Rancher

### LINKED ISSUE #1311: Automate Fleet RC Release (weekly)
Automically update rancher/charts weekly with the lastest fleet rc (currently 0.6).

As a github workflow, maybe update CLI to create a PR in rancher/charts with the changes.

What about rancher/rancher?

- [-] min version
- [-] build image with chart asset in system bundle (for rancher rc)
- [-] fleet api package in go.mod (manually if needed)

thardeck:
The scripts already work and are tested locally. They work for new versions as well as small version bumps like new rc releases.

The issue though during back testing is that there are sometimes issues in the charts repo [or with the make targets of that repo]. I guess in this case we manually have to bump otherwise the script would get to complex and still can not cover all cases.

My current plan is to keep the scripts in a form that they can also be executed by the developers from their local machine in case there was an issue in the automated release process so there is no need to run the steps manually.

Since the Rancher versioning is quite complex and recently there was a case where the chart version changed in between an rc bump we decided to completely go for a manual GH action run and providing the version manually.

At the moment the GH action uses my forks and machine token but I will create a ticket to get an official one.

Fixes #1311


.github/workflows/release-against-charts.yml
```diff
@@ -0,0 +1,64 @@
...
+        description: "New Rancher Chart version"
+        required: true
+        default: ""
+      should_replace:
+        description: "Should the old Fleet version be replaced/removed? (e.g. true in case of release candidate bumps)"
+        required: true
+        default: "true"
+
+jobs:
+  rancher-integration:
```
manno, line 35:
```suggestion
  create-rancher-charts-pr:
```

.github/workflows/release-against-charts.yml
```diff
@@ -0,0 +1,64 @@
...
+          ref: ${{github.event.inputs.charts_ref}}
+          path: charts
+      - name: Run release script
+        run: ./fleet/.github/scripts/release-against-charts.sh ${{github.event.inputs.prev_fleet}} ${{github.event.inputs.new_fleet}} ${{github.event.inputs.prev_chart}}  ${{github.event.inputs.new_chart}}  ${{github.event.inputs.should_replace}}
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v3
+        with:
+          token: ${{secrets.CHART_PR_PAT}}
+          push-to-fork: thardeck/charts
+          title: ' Update Fleet to v${{github.event.inputs.new_fleet}}'
```
manno, line 57:
```suggestion
          title: 'Update Fleet to v${{github.event.inputs.new_fleet}}'
```
Ah, that whitespace is probably required because they prefix the version?

.github/workflows/release-against-rancher.yml
```diff
@@ -0,0 +1,72 @@
...
+        required: true
+        default: "true"
+
+env:
+  GOARCH: amd64
+  CGO_ENABLED: 0
+  SETUP_GO_VERSION: '1.19.*'
+
+jobs:
+  rancher-integration:
```
manno, line 40:
```suggestion
  create-rancher-pr:
```

manno (changes requested):
All fine besides the job names.


I noticed  you omit quotes in shell scripts where they are optional, but you put curly braces all the time. :)
I find that confusing, since curly braces are used for parameter expansion and I try to read them more carefully than plain variables.

manno (approved):

PR #1353, opened 2023-02-28
Ensure fleet-agent's helm release name has valid length

manno:
Refers to https://github.com/rancher/rancher/issues/40710

The cluster resource name is used when creating the bundle resource, however the helm release name is not shortened to 53 chars.

Enforcement of proper release name lengths was introduced with https://github.com/rancher/fleet/pull/964 to avoid reoccurring deletion of bundles. Later the fix was removed from the helmdeployer https://github.com/rancher/fleet/issues/1272 because `fleet apply` would produce valid bundles. However, the `fleet-agent` code does not use `fleet apply` and was no longer protected.

This PR keeps the bundle name intact and specifies the helm release name only.
We also switch away from wrangler's name package and use a "stable" limiter that [doesn't modify a string if its length is less or equal](https://github.com/rancher/wrangler/pull/279).

### DIFF
--- a/pkg/name/name.go
+++ b/pkg/name/name.go
@@ -5,6 +5,17 @@ import (
 	"crypto/md5" // nolint:gosec // Non-crypto use
 	"encoding/hex"
 	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	disallowedChars = regexp.MustCompile(`[^a-zA-Z0-9-]+`)
+	helmReleaseName = regexp.MustCompile(`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`)
+	dnsLabelSafe    = regexp.MustCompile(`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`)
+	multiDash       = regexp.MustCompile("-+")
 )
 
 // Limit the length of a string to count characters. If the string's length is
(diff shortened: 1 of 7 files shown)


pkg/controllers/manageagent/manageagent.go
```diff
@@ -205,6 +207,7 @@ func (h *handler) newAgentBundle(ns string, cluster *fleet.Cluster) (runtime.Obj
 				DefaultNamespace: agentNamespace,
 				Helm: &fleet.HelmOptions{
 					TakeOwnership: true,
+					ReleaseName:   name2.Limit(name.SafeConcatName(AgentBundleName, cluster.Name), 53),
```
raulcabello, line 210: why are we not using the new `HelmReleaseName` func? Or maybe we should not add `ReleaseName` here. Then `ReleaseName` would be nil and `HelmReleaseName` would be called in [getOpts](https://github.com/rancher/fleet/blob/master/pkg/helmdeployer/deployer.go#L258-L261) when installing the helm release.

pkg/name/name.go
```diff
@@ -0,0 +1,86 @@
...
+	str = strings.ToLower(str)
+	str = disallowedChars.ReplaceAllLiteralString(str, "-")
+	str = multiDash.ReplaceAllString(str, "-")
+	str = strings.TrimLeft(str, "-")
+	str = strings.TrimRight(str, "-")
+
+	// shorten name to 53 characters, the limit for helm release names
+	if helmReleaseName.MatchString(str) && dnsLabelSafe.MatchString(str) {
+		logrus.Debugf("shorten bundle name %v to %v\n", str, Limit(str, 53))
+		return Limit(str, 53)
```
raulcabello, line 79: nit: I saw 53 in several places. It might be a good idea to create a const

pkg/name/name_test.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"fmt"
+
+	"github.com/rancher/fleet/pkg/name"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Name", func() {
```
raulcabello, line 12: nit: It is great that the `HelmReleaseName` is now in the `name` package and that we have tests for it!
However I think in this case it would be better to have normal go unit tests and not use Ginkgo here. I think it doesn't add much value here and add unnecessary complexity. I would leave Ginkgo for integration and e2e tests.

This is not a blocker, just my opinion on this. Feel free merge it as it is

manno: Yes true, especially since I use a test with a table, but I'd normally use ginkgo for all tests. It's just that the few existing unit tests are old.

pkg/name/name.go
manno: I can't find too many places. This one has a comment and one usage is just for logging. The other place is the manageagent, which should use the func from the name package as you said.
Finally, there is one more 53 in crds.go. Well, I do want to move more constants into our API package..

pkg/controllers/manageagent/manageagent.go
manno: Thanks, that's a good point. This was my initial fix, before I noticed we derive the release name differently in apply and deployer. But with the change to deployer, this should not be needed anymore.

raulcabello (approved):
LGTM, thanks!

jameson-mcghee:
This PR is related to [40709](https://github.com/rancher/rancher/issues/40709)

PR #1354, opened 2023-02-28
Add ShellCheck linter

thardeck:


dev/setup-k3d
```diff
@@ -20,4 +20,4 @@ EOF
   args="$args --registry-config $TMP_CONFIG"
 fi
 
-k3d cluster create "$name" --servers 3 --api-port $(( 36443 + i )) -p "$(( 8080 + i )):80@server:0" -p "$(( 8443 + i )):443@server:0" $args
+k3d cluster create "$name" --servers 3 --api-port $(( 36443 + i )) -p "$(( 8080 + i )):80@server:0" -p "$(( 8443 + i )):443@server:0" "$args"
```
manno, line 23: No, we need to expand $args into multiple args.

dev/setup-k3ds
```diff
@@ -7,7 +7,7 @@ docker_mirror=${docker_mirror-}
 
 if [ -n "$docker_mirror" ]; then
   TMP_CONFIG="$(mktemp)"
-  trap "rm -f $TMP_CONFIG" EXIT
+  trap 'rm -f $TMP_CONFIG' EXIT
```
manno, line 10: I don't see how that matters. It's always the TMP_CONFIG from the previous line, we don't want to evaluate it later on exit.

dev/setup-k3ds
```diff
@@ -22,8 +22,8 @@ fi
 # https://hub.docker.com/r/rancher/k3s/tags
 #args="$args -i docker.io/rancher/k3s:v1.22.15-k3s1"
 
-k3d cluster create upstream --servers 3 --api-port 36443 -p '80:80@server:0' -p '443:443@server:0' $args
-k3d cluster create downstream --servers 1 --api-port 36444 -p '5080:80@server:0' -p '3444:443@server:0' $args
+k3d cluster create upstream --servers 3 --api-port 36443 -p '80:80@server:0' -p '443:443@server:0' "$args"
+k3d cluster create downstream --servers 1 --api-port 36444 -p '5080:80@server:0' -p '3444:443@server:0' "$args"
```
manno, line 26: Again, we want this to expand to multiple args

package/Dockerfile
```diff
@@ -5,11 +5,11 @@ USER 1000
 
 FROM base AS copy_dapper
 ONBUILD ARG ARCH
-ONBUILD COPY bin/fleetcontroller-linux-${ARCH} /usr/bin/fleetcontroller
+ONBUILD COPY bin/fleetcontroller-linux-"${ARCH}" /usr/bin/fleetcontroller
```
manno, line 8: This is not a shell script. Quotes are a bit uncommon in Dockerfiles and I always assumed they are not necessary. Well, unless you EXEC something or RUN a shell script. 
See docs: https://docs.docker.com/engine/reference/builder/#environment-replacement

Here is a blog article telling us not to quote, if you want to know more: https://dev.to/tvanantwerp/don-t-quote-environment-variables-in-docker-268h

package/Dockerfile-windows.agent
```diff
@@ -1,12 +1,12 @@
...
 ARG RELEASES
 ARG VERSION
-ENV RELEASES=$RELEASES
-ENV VERSION=$VERSION
+ENV RELEASES="$RELEASES"
+ENV VERSION="$VERSION"
 SHELL ["powershell", "-NoLogo", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]
 RUN pushd c:\; \
-    $URL = 'https://github.com/git-for-windows/git/releases/download/v2.33.1.windows.1/MinGit-2.33.1-64-bit.zip'; \
+    "$URL" = 'https://github.com/git-for-windows/git/releases/download/v2.33.1.windows.1/MinGit-2.33.1-64-bit.zip'; \
```
manno, line 9: Is that how powershell works?

dev/airgap-rancher.sh
```diff
@@ -19,7 +19,7 @@ rancher/thanosio-thanos
 rancher/webhook-receiver
 rancher/rancher-webhook"
 
-grep -f <(echo $essentials) rancher-images.txt | grep -v windows >> running-images.txt
+grep -f <(echo "$essentials") rancher-images.txt | grep -v windows >> running-images.txt
```
manno, line 22: I'm quite sure we want this to expand for the newlines?

dev/delve-debug
```diff
@@ -19,9 +19,9 @@ procname=${app/-}
...
 
 kubectl --namespace "$ns" port-forward "pod/$pod" 4000 &
-trap "kill $!" EXIT
+trap 'kill $!' EXIT
 
-echo `tput bold`
+tput bold
 echo 'Please wait for the line "debug layer=debugger continuing" to appear...'
-echo `tput sgr0`
+tput sgr0
```
manno, line 26: Does that work? I was wondering about "echo tput", but didn't read up on it. I would have expected `echo -e $(tput ...)`

Makefile
```diff
@@ -2,7 +2,7 @@ TARGETS := $(shell ls scripts)
 
 .dapper:
 	@echo Downloading dapper
-	@curl -sL https://releases.rancher.com/dapper/latest/dapper-`uname -s`-`uname -m` > .dapper.tmp
+	@curl -sL https://releases.rancher.com/dapper/latest/dapper-$(uname -s)-$(uname -m) > .dapper.tmp
```
manno, line 5: Yeah, makefiles don't even support backticks.

.github/scripts/check-for-gitjob-changes.sh
```diff
@@ -4,11 +4,11 @@ set -ue
 : "${MAIN_BRANCH:=master}"
 
 if [ "$(git diff --name-only "${MAIN_BRANCH}" -- "charts/fleet/charts/gitjob/" | wc -l)" -gt "0" ]; then
-    if ! $(git diff "${MAIN_BRANCH}" -- "charts/fleet/charts/gitjob/Chart.yaml" | grep -qE 'version: [.0-9]+'); then
-        echo "Gitjob needs to be updated in the 'rancher/gitjob' repo first and"
-        echo "then the new Gitjob release can be added to Fleet.\n"
+    if ! git diff "${MAIN_BRANCH}" -- "charts/fleet/charts/gitjob/Chart.yaml" | grep -qE 'version: [.0-9]+'; then
+        printf "Gitjob needs to be updated in the 'rancher/gitjob' repo first and"
+        printf "then the new Gitjob release can be added to Fleet.\n"
```
manno, line 9: No need to change the echo to printf and not even  use a format string. Just remove the `\n`. echo adds a newline by default anyway. Just add more `echo`s for more newlines..

manno (changes requested):

.github/scripts/check-for-auto-generated-changes.sh
```diff
@@ -7,9 +7,9 @@ go run pkg/codegen/main.go
 go run ./pkg/codegen crds ./charts/fleet-crd/templates/crds.yaml > /dev/null
 
 if [ -n "$(git status --porcelain)" ]; then
-    echo "Generated files have either been changed manually or were not updated.\n"
+    printf "Generated files have either been changed manually or were not updated.\n"
```
manno, line 10: No need for printf. Would prefer 
```suggestion
    echo "Generated files have either been changed manually or were not updated."
    echo
```

.github/scripts/check-for-auto-generated-changes.sh
```diff
@@ -7,9 +7,9 @@ go run pkg/codegen/main.go
...
 
-    echo "The following generated files did differ after regeneration:"
+    printf "The following generated files did differ after regeneration:"
```
manno, line 12:
```suggestion
    echo "The following generated files did differ after regeneration:"
```

PR #1357, opened 2023-03-01
[v0.6] ensure valid release name for agent

manno:
Refers to https://github.com/rancher/rancher/issues/40710


manno:
I have trouble testing this locally. This is not ready yet.

.github/workflows/aks.yml
```diff
@@ -17,7 +17,7 @@ on:
 env:
   GOARCH: amd64
   CGO_ENABLED: 0
-  SETUP_GO_VERSION: '^1.19'
+  SETUP_GO_VERSION: '1.19.*'
```
olblak, line 20: Imho it would be better to use the version retrieve from the go.mod file 
https://github.com/actions/setup-go#getting-go-version-from-the-gomod-file

olblak (approved):
my comment is not a blocker but a potential improvement that would reduce the need to update the current workflow.

.github/workflows/aks.yml
manno: Interesting discussion for master, but not here. I only want to avoid newer versions that we currently see with 1.20 on some github action caches. 

Note: It's totally fine to build a 1.19 go.mod file with 1.20, it's just the version of the file format.
However, the go version in CI should be synced with the version in `Dockerfile.dapper` (which we use for the release build).

kkaempf:
https://github.com/rancher/rancher/issues/40684 got closed

PR #1368, opened 2023-03-03
Refactor agent deployment into one function

manno:
Just a clean up to make the deployment manifest more visible.


raulcabello (approved):
LGTM, thanks!

PR #1382, opened 2023-03-07
Allow resources to be kept when deleting a GitRepo or a Bundle

raulcabello:
This PR adds a new field `keepResources` to the `GitRepo` and `Bundle` CRDs. It also adds the `keep-resources` flag to the fleet cli.  `keepResources`  can also be specified in the `fleet.yaml`

When a `GitRepo` has the `keepResources` field set to true it sets the `keep-resources` flag to true when executing fleet apply in the job. Then `keepResources` is true in the generated `Bundles` and `BundleDeployments`.

The chart annotation `fleet.cattle.io/keep-resources` is set to true if `keepResources` is true when installing the helm chart. This is done because when a `BundleDeployment` is deleted it is nil in the [cleanup](https://github.com/rancher/fleet/blob/master/modules/agent/pkg/controllers/bundledeployment/controller.go#L87), so we can use this annotation to know if resources should be kept.

Resources will not be deleted if `keepResources` is true. Helm release secrets will be removed instead, like it is done for the [fleet-agent](https://github.com/rancher/fleet/blob/master/pkg/helmdeployer/deployer.go#L632-L635)

e2e test added

Fix https://github.com/rancher/fleet/issues/680
Supersedes https://github.com/rancher/fleet/pull/1345

### DIFF
--- a/.github/workflows/e2e-ci.yml
+++ b/.github/workflows/e2e-ci.yml
@@ -98,6 +98,11 @@ jobs:
           FLEET_E2E_NS: fleet-local
         run: |
           ginkgo e2e/single-cluster
+      - name: E2E Tests for keep-resources
+        env:
+          FLEET_E2E_NS: fleet-local
+        run: |
+          ginkgo e2e/keep-resources
       -
         name: Fleet Tests Requiring Github Secrets
         # These tests can't run for PRs, because PRs don't have access to the secrets
--- a/.github/workflows/e2e-ci.yml
+++ b/.github/workflows/e2e-ci.yml
@@ -97,12 +97,7 @@ jobs:
         env:
           FLEET_E2E_NS: fleet-local
         run: |
-          ginkgo e2e/single-cluster
-      - name: E2E Tests for keep-resources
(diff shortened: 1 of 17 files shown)


.github/workflows/e2e-ci.yml
```diff
@@ -98,6 +98,11 @@ jobs:
           FLEET_E2E_NS: fleet-local
         run: |
           ginkgo e2e/single-cluster
+      - name: E2E Tests for keep-resources
+        env:
+          FLEET_E2E_NS: fleet-local
+        run: |
+          ginkgo e2e/keep-resources
```
manno, line 105: Should be possible in one line
```suggestion
          ginkgo e2e/single-cluster e2e/keep-resources
```

raulcabello: Thanks! done in https://github.com/rancher/fleet/pull/1382/commits/e6275c45de59dab3e5aff760cf54ca6cc1d7e685

manno (approved):

PR #1410, opened 2023-03-15
Switch Updatecli Token to allow automatic triggering

thardeck:


olblak (approved):
Let see if that token has enough permission

Issue #637, opened 2021-11-22
fleet-agent in state `ErrApplied` with the following reason: `another operation (install/upgrade/rollback) is in progress`


papanito:
I currently have my `fleet-agent` in state `ErrApplied` with the following reason: 
```
another operation (install/upgrade/rollback) is in progress.
```
I had two fleet-agents under "Installed Apps": One in ns `cattle-fleet-system` and the other in `fleet-system`

Not sure how I came to that state, but I recently upgraded from Rancher 2.5.11 to 2.6.2.

The app in `cattle-fleet-system` was in status `Pending-Upgrade`

![image](https://user-images.githubusercontent.com/15034009/142824402-58cad5fd-43b8-4c91-8e13-9fbdacabee9b.png)

I deleted both apps and made an "upgrade" of fleet app on the local cluster. The fleet-agent seems now properly deployed

![image](https://user-images.githubusercontent.com/15034009/142824060-bbc6b869-b873-47f7-a85e-85a0137c08d3.png)

sw34:
I also am having fleet issues after upgrade from rancher 2.5 to 2.6

#643

krenaut62:
I also am having fleet issues after upgrade from 2.5.9 to 2.6.3.  The issue was likely caused by either pod security policy and or pod scheduling issues caused by taints.  Unclear how to recover from this, I see no easy/obvious way to redeploy fleet.

I think my issue is the result of me editing the fleet deployment while I was running 2.5.9.  I added a nodeSelector and a toleration to force it to run in a special set of nodes that I have reserved for system applications in the hopes of protecting it from the developers who occasionally wreak havoc on nodes due to poor resource mgmt.

sw34:
I did not find a solution and decided it would be easier to completely rebuild my small deployment with a clean install of 2.6.x rather than spend more hours trying to troubleshoot.

krenaut62:
Rebuilding is certainly a possibility, but we have a tremendous amount of custom security that all disappears and has to be manually reconfigured.  Thank you for the information @sw34

atsai1220:
We removed the Helm release secrets that did not have a successful outcome. 
```
kubectl get secret -A -l status=pending-upgrade
```

krenaut62:
[...]

PR #1424, opened 2023-03-20
retry helm upgrade if it was interrupted

raulcabello:
When a helm upgrade is interrupted it returns the error: `another operation (install/upgrade/rollback) is in progress`. Then the release is in a `pending-upgrade` state. This happens if the `fleet-agent` pod is killed while doing an upgrade.

To solve this we do a rollback if this error was found while doing an upgrade, then retry the upgrade.

fix https://github.com/rancher/fleet/issues/637

### DIFF
--- a/pkg/helmdeployer/deployer.go
+++ b/pkg/helmdeployer/deployer.go
@@ -41,6 +41,7 @@ const (
 	ServiceAccountNameAnnotation = "fleet.cattle.io/service-account"
 	DefaultServiceAccount        = "fleet-default"
 	KeepResourcesAnnotation      = "fleet.cattle.io/keep-resources"
+	HelmUpgradeInterruptedError  = "another operation (install/upgrade/rollback) is in progress"
 )
 
 var (
@@ -405,7 +406,20 @@ func (h *Helm) install(bundleID string, manifest *manifest.Manifest, chart *char
 	if !dryRun {
 		logrus.Infof("Helm: Upgrading %s", bundleID)
 	}
-	return u.Run(releaseName, chart, values)
+	rel, err := u.Run(releaseName, chart, values)
+	if err != nil && err.Error() == HelmUpgradeInterruptedError {
+		logrus.Infof("Helm error: %s for %s. Doing a rollback", HelmUpgradeInterruptedError, bundleID)
+		r := action.NewRollback(&cfg)
+		err = r.Run(releaseName)
+		if err != nil {
+			return nil, err
+		}
+		logrus.Debugf("Helm: retrying upgrade for %s after rollback", bundleID)
+
+		return u.Run(releaseName, chart, values)
+	}
+
+	return rel, err
 }
 
 func (h *Helm) getValues(options fleet.BundleDeploymentOptions, defaultNamespace string) (map[string]interface{}, error) {


pkg/helmdeployer/deployer.go
```diff
@@ -405,7 +406,20 @@ func (h *Helm) install(bundleID string, manifest *manifest.Manifest, chart *char
 	if !dryRun {
 		logrus.Infof("Helm: Upgrading %s", bundleID)
 	}
-	return u.Run(releaseName, chart, values)
+	rel, err := u.Run(releaseName, chart, values)
+	if err != nil && err.Error() == HelmUpgradeInterruptedError {
+		logrus.Infof("Helm error: another operation (install/upgrade/rollback) is in progress for %s. Doing a rollback", bundleID)
```
olblak, line 411: Don't you want to reuse HelmUpgradeInterruptError such as 
```suggestion
		logrus.Infof("Helm error: %s for %s. Doing a rollback", HelmUpgradeInterrruptError, bundleID)
```

olblak (dismissed):

manno (dismissed):
👍

pkg/helmdeployer/deployer.go
manno: I feel like it's more readable without the constant, but 🤷 
@raulcabello if you change the commit, can you make the commit message start with a capital letter? :)

raulcabello: @olblak suggestion didn't compile because it is `HelmUpgradeInterrruptedError`, no `HelmUpgradeInterrruptError`. I changed this and the commit message. Please review it again

w13915984028 (changes requested):
When `fleet-agent` finds some bundles are in `pending-upgrade` status, it could be a last time left, and also could be real conflict, say another fleet-agent is upgrading.

The first-thing is to identify this situation.

Ideally, `fleet-agent` needs to lablel? the helm release (e.g. k8s secrets), to have a mark that the current upgrade is from which `fleet-agent`. Or check the time-stamp, assumes a 10-minute or some value is exceeding the possible upgrade timeout.

For rollback, should we try at most one time, or more?

Those robustness parts need to be carefully checked before doing an automatic rollback, to avoid `from a bad status to an even bad status`.


BTW: can `fleet-agent/helm` recover from simply deleting the current pending helm `pending-upgrade` release ?

raulcabello:
> When fleet-agent finds some bundles are in pending-upgrade status, it could be a last time left, and also could be real conflict, say another fleet-agent is upgrading.

That should never happen. There could not be more than one fleet agent running at the same time. 
fleet-agent uses leader election. Just the leader is doing upgrades, the slaves are waiting until they become leaders.

> For rollback, should we try at most one time, or more?

I think if rollback fails we should just return the error. I don't see why rollback would fail, but I wouldn't retry.

> BTW: can fleet-agent/helm recover from simply deleting the current pending helm pending-upgrade release ?

It looks like doing rollback is a good approach for fixing this issue https://github.com/helm/helm/issues/8987#issuecomment-724036145 and https://github.com/helm/helm/issues/4558#issuecomment-648352068

olblak (approved):

w13915984028:
@raulcabello :
> fleet-agent uses leader election

The `peer` may be a personal operation like `helm upgrade`, or another POD running similar function like `fleet-agent`, which also use helm code to upgrade.  can `fleet-agent` avoid clash with them?


>> For rollback, should we try at most one time, or more?

> I think if rollback fails we should just return the error. I don't see why rollback would fail, but I wouldn't retry.

When the controller `OnChange` runs again, will the code try to `rollback` again?  thats my concern, the k8s controller reconciller may run your code again and again.

Indeed, I am afraid the current code is still fragie in case of error happens. It is assuming that it can blindly run rollback, the rollback will success or no bad result when fail.

w13915984028:
@raulcabello 

Considering the long integration process from `fleet-agent` to `Rancher` to `Harvester`, it takes month to have a new release.

It is important to have an escape route.

Could you please add an enhancement, that via `field/label/annotation` in the bundle, `fleet-agent` can decide to skip the automatic rollback. Or configuration option to `fleet-agent`.


@manno @raulcabello I guess the issue was not tested out in fleet nightlies/rc in the past, are some dedicated test cases added for this bug?

As @bk201 worried, the auto rollback may bring the system into some uncontrolled status.

A bunch of test cases are needed to cover all those potential scenarios.

cc @yasker

thanks.

manno:
We know there is very little automatic testing around the helm deployer. However, we assumed we won't have time in Q2 to build that out. It's hard to test error conditions like this. We would need to kill the deployer at the right moment to get into the inconsistent ("rollback") state. We also don't know about the other possible error conditions you are speculating about. Do they exist, how would we reproduce them? 

I'd suggest to use a nightly https://github.com/rancher/fleet/actions/workflows/nightly.yml build, inject that into Rancher and produce separate issues.

Issue #1153, opened 2022-12-02
[SURE-5668] Fleet BundleDeployments does not reflect new labels from Bundles


atsai1220:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

If we update an existing Fleet Bundle with additional labels then we can observe that BundleDeployments do not have the newly added labels unless we either:
- Delete the Bundle
- Manually patch the BundleDeployment in the cluster's Fleet namespace in the local cluster

This is problematic as we iterate on our Fleet deployments and leveraging the dependsOn feature. The dependent BundleDeployment will say this until we patch the BundleDeployment object with the new label.
```
No bundles matching labels x=y in namespace fleet-default
```

### Expected Behavior

We expect BundleDeployments to be a reflection of Bundle objects, including from labels and annotations updates.

### Steps To Reproduce

1. Create `GitRepo A` object with labels `a=b` inside `fleet.yaml`.
2. Observe `Bundle` and `BundleDeployment` objects created with labels `a=b`
3. Update `GitRepo A` with a new label `x=y` by updating `fleet.yaml`
4. Create second `GtiRepo B` object that `dependsOn` the `GtiRepo A` with the label `x=y` inside of `fleet.yaml`.
5. See that the cluster's `BundleDeployment` object from `GitRepo B` display this error:
```
No bundles matching labels x=y in namespace fleet-default
``` 
6. Manually patch `BundleDeployment` object from `GitRepo A` with label `x=y` (or delete `Bundle` objecet from `GitRepo A` but this is not ideal as this will cause the application to be reinstalled).

### Environment

```markdown
- Architecture: amd
- Fleet Version: 0.5.0
- Cluster:
  - Provider: RKE, RKE2, K3S
  - Kubernetes Version: v1.21.x
```

manno:
I don't understand the reproduction steps.

> Create GitRepo A object with labels a=b inside fleet.yaml.

I think labels are only used in fleet.yaml as selectors.

I'm under the impression that the propagation of labels from a gitrepo resource to a bundle was not actually implemented before 0.6.x-alpha. This feature only existed in the docs.

See  https://github.com/rancher/fleet/pull/1007 for the implementation.

manno:
This is the true fleet.yaml struct, which does support labels: https://github.com/rancher/fleet/blob/master/pkg/bundlereader/read.go#L129-L135

manno:
[...]

PR #1428, opened 2023-03-21
Always get labels from Bundle when creating a BundleDeployment

raulcabello:
When a label is modified in the `fleet.yaml` is also modified in the `Bundle`, but no in the `BundleDeployment`

`BundleDeployment` labels are not being updated when changed in the `fleet.yaml`. This is because we just get the labels from the `Bundle` the first time the `BundleDeployment` is created, this is done in the [resetDeployment method](https://github.com/rancher/fleet/blob/c7b02620e82b16afdca91bc466b527fa7c6232aa/pkg/target/target.go#L433). 

This PR always get the labels from the `Bundle` when creating a `BundleDeployment`

BundleDeployment patch was failing in the integration tests because wrangler was using `application/strategic-merge-patch+json` strategy for the patch. Then I was getting the error:
```
time="2023-03-24T17:04:02+01:00" level=error msg="error syncing 'test-0e60e6e/name': handler bundle: failed to update test-0e60e6e/name fleet.cattle.io/v1alpha1, Kind=BundleDeployment for bundle test-0e60e6e/name: the body of the request was in an unknown format - accepted media types include: application/json-patch+json, application/merge-patch+json, application/apply-patch+yaml, requeuing"
```
This was happening because I registered the `Bundle` and `BundleDeployment` in the client, then wrangler was choosing this strategy [here](https://github.com/rancher/wrangler/blob/v1.1.1/pkg/patch/style.go#L64). I fixed this by not registering the `Bundle` and `BundleDeployment` in the client as they are not needed, then I did the same for the `BundleDeployment` integration tests [here](https://github.com/rancher/fleet/pull/1428/commits/db34da7a96eb94492b42de41e9615d1e09fff315) in order to prevent this error in the future.

refers to https://github.com/rancher/fleet/issues/1153

### DIFF
--- a/integrationtests/agent/suite_test.go
+++ b/integrationtests/agent/suite_test.go
@@ -32,12 +32,10 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/discovery/cached/memory"
 	"k8s.io/client-go/dynamic"
-	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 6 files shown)


manno (approved):
👍

PR #1431, opened 2023-03-22
Don't ignore config map reference in valuesFrom list element

manno:
Refers to https://github.com/rancher/fleet/issues/1432

If both references are provided in a `valuesFrom` list element, use both.

## TODO
- [x] add test from manual testing at https://github.com/manno/fleet-experiments/tree/valuesFrom


raulcabello (approved):
LGTM

Issue #1392, opened 2023-03-09
[Epic] Agent Config for Tolerations/Affinity/Resources per Cluster (Q2)


manno:
We want to make affinity and resource limits for the fleet agent configurable per cluster. 

- [x] extend structs, crds
- [x] update cluster controller to redeploy agents if one of the fields, "agentTolerations", "agentAffinity" and "agentResources", changes.

- [x] rancher/fleet#1445
- [x] rancher/fleet#1446
- [x] rancher/fleet#1447

sowmyav27:
Need info/design doc for determining QA LOE

sowmyav27:
@kkaempf Need info/design doc for determining QA LOE

kkaempf:
QA is done for the sub-issues.

PR #1433, opened 2023-03-23
Add agent config for tolerations, affinity, resources on cluster

manno:
Changes are tracked via the, hashed, status fields and will redeploy the agent.

This also adds more code comments to the custom resources.

refers to #1392


pkg/agent/manifest.go
```diff
@@ -229,6 +231,16 @@ func agentDeployment(namespace string, agentScope string, opts ManifestOptions,
...
 	dep.Spec.Template.Spec.Tolerations = append(dep.Spec.Template.Spec.Tolerations, opts.AgentTolerations...)
 
+	// overwrite affinity if present on cluster
+	if opts.AgentAffinity != nil {
+		dep.Spec.Template.Spec.Affinity = opts.AgentAffinity
+	}
+
+	// set resources if present on cluster
+	if opts.AgentResources != nil {
+		dep.Spec.Template.Spec.Containers[0].Resources = *opts.AgentResources
```
raulcabello, line 241: shouldn't we change Resources in all containers? Imagine we have a pod with multiple containers, I think it should be applied to all containers.

pkg/controllers/manageagent/manageagent_test.go
```diff
@@ -0,0 +1,276 @@
+package manageagent
+
+import (
+	"testing"
+
+	// counterfeiter -o internal/fakes github.com/rancher/wrangler/pkg/generated/controllers/core/v1.NamespaceController
```
raulcabello, line 6: can we remove the counterfeiter comment as we are using gomock?

raulcabello (changes requested):
I left a comment regarding applying `Resources` to all containers instead of just the first one.

Unit tests are great!

pkg/agent/manifest.go
manno: The agent deployment is only one container. This is in line with the env vars below, they're also added just to the first container.

pkg/controllers/manageagent/manageagent_test.go
manno: I turned that into a `//go:generate` comment. `go generate ./pkg/controllers/manageagent` will re-create the mock.

pkg/agent/manifest.go
raulcabello: what if someone is using something like Istio that injects a sidecar container?

raulcabello (approved):

PR #1485, opened 2023-04-21
Scaling: improve behavior on failed deployments

moio:
Attempts to address [SURE-6125](https://jira.suse.com/browse/SURE-6125)

# Context

User has a large number of applied Bundles, several of which are failing due to non-recoverable reasons (eg. broken YAML in manifests in Git).

fleet-agent continuously attempts redeploying those Bundles, and that causes very high resource load, which in turn affects the overall stability of the downstream cluster.

# Proposed changes

Fixes:
- catch more unrecoverable conditions and do not retry them, extending #1042
- fleet-controller: when Cluster status needs to be recomputed due to errors, use exponential backoff provided by Lasso instead of a fixed delay. Tune backoff parameters to be less aggressive than defaults (minimum 2 seconds, maximum 10 minutes)

Damage reduction measures:
- fleet-agent: prevent bundle deployments when the associated Bundle, Cluster or GitRepo are disabled. This allows for better management of symptoms in case the above fixes are not sufficient
- fleet-agent: when BundleDeployments are retried because of errors, use tuned exponential backoff parameters that are less aggressive than defaults (minimum 2 seconds, maximum 10 minutes)

Other fixes:
- fleet-agent: when bundle deployments are triggered because of bundle changes, do not rate limit them (unlike errors). This used to be the case before 

## Test

Unit/integration tests to be done if necessary.

To test manually, import a GitRepo with a YAML error, eg.

```
kubectl apply -f - --kubeconfig=$KUBECONFIG --context=$UPSTREAM_CONTEXT <<EOF
kind: GitRepo
metadata:
  name: moio-test
  annotations:
    {}
  labels:
    {}
  namespace: fleet-default
spec:
  branch: master
  insecureSkipTLSVerify: false
  paths:
    - single-cluster/helm
    - ''
  repo: https://github.com/moio/fleet-examples.git
  targets:
    - clusterSelector:
        matchExpressions:
          - key: provider.cattle.io
            operator: NotIn
            values:
              - harvester
EOF
```

and observe fleet-agent logs.

### Tradeoff

Tuning down the Lasso rate limiters has an effect on `Watch` calls. One is used in the Bundle controller and this will result in worse performance in case the same BundleDeployments change quickly (the rate limit is per-BundleDeployment).

### Potential improvement

It may be possible to disable more computation in fleet-controller Controllers when resources are paused. It is most likely possible to catch more nonrecoverable error conditions.


moio:
Currently being tested by some users. Will report results.

moio:
User reported positive outcomes in terms of memory and CPU load - both on fleet-agent pods and etcd/kubeapiserver.

modules/agent/pkg/controllers/bundledeployment/controller.go
```diff
@@ -247,7 +247,7 @@ func (h *handler) Trigger(key string, bd *fleet.BundleDeployment) (*fleet.Bundle
 		logrus.Debugf("Adding OnChange for bundledeployment's '%s' resource list", key)
 		return bd, h.trigger.OnChange(key, resources.DefaultNamespace, func() {
 			// enqueue bundledeployment if any resource changes
-			h.bdController.Enqueue(bd.Namespace, bd.Name)
+			h.bdController.EnqueueAfter(bd.Namespace, bd.Name, 0)
```
raulcabello, line 255: why don't we want to use the rate limit here? what if the `Bundle` is in an error state and this is triggered too many times in a small period of time? 
I think it is safer to use `Enqueue` here, and use the backoff from wrangler

pkg/helmdeployer/deployer.go
```diff
@@ -500,23 +509,63 @@ func getReleaseNameVersionAndNamespace(bundleID, resourcesID string) (string, in
...
+
+		// return the right one if found
+		for _, release := range releases {
+			if release.Name == releaseName && release.Version == version && release.Namespace == namespace {
+				return release, nil
+			}
+		}
+
+		// desired release was not found in Helm's storage. Put a "dummy" in the cache to remember it was not found
+		err = h.releaseCache.Add(&release.Release{
```
raulcabello, line 548: I wouldn't add a "dummy" release if it was not found. What if a release is not found now, but it becomes available one second later? We would get `ErrNoRelease` because is not in the cache until the cache expires in 5 minutes

modules/agent/pkg/controllers/bundledeployment/controller.go
moio: It is my understanding this fires every time a _resource_ that was deployed by bundledeployment change.

The reasoning was: if a resource has changed, why should we wait to react to the change? This is not about the bundle changing or error conditions, AFAIU.

raulcabello: It waits just if there was an error. The first time it shouldn't wait, the first time it retries it should wait 2^1 * time, the second 2^2 * time, third 2^3 * time and so on.

pkg/helmdeployer/deployer.go
raulcabello: Based in your research do you think it returns `ErrNoRelease` too often that might impact the performance? I think that it will not happen very often, but I might be wrong

modules/agent/pkg/controllers/bundledeployment/controller.go
moio: My understanding is that the underlying work queue rate limits depending on its key (the bundle namespace+name in this case) and it has no notion of what a handler `error` is. So it is true all handler `error`s result in Enqueuing, but not all Enqueuings originate from `error`s.

As a consequence, I thought, we should generally use Enqueue in error-like conditions - cases to be treated as if the handler exited with an `error`. OTOH, the reasoning went, we should not use `Enqueue`, but rather `EnqueueAfter` or something else, in case we are triggering something for a condition which is not an `error` - because we do not want it treated like an `error`.

If a bunch of non-error situations happen, I imagined, we do not want to artificially slow down the work queue, but rather work through them as fast as possible. Only if we have errors we want to slow down, because they might be temporary and we do not want to busy wait.

Does the above make any sense to you?

pkg/helmdeployer/deployer.go
moio: I admit I do not have exact numbers but I remember seeing a difference in CPU between builds that had and did not have these lines, at least in the test environment.

modules/agent/pkg/controllers/bundledeployment/controller.go
raulcabello: it makes sense, thanks for the explanation!

pkg/helmdeployer/deployer.go
raulcabello: > I admit I do not have exact numbers but I remember seeing a difference in CPU between builds that had and did not have these lines, at least in the test environment.

then it's ok to leave it if it's affecting performance. Worst case scenario cache will be removed in 5 minutes, and it will work after that

moio:
Is there anything I should do to have this merged?

raulcabello (approved):

Issue #1120, opened 2022-11-22
Rancher Integration CI is stuck in WaitApplied


manno:
Our Rancher CI is broken.

When the workflow updates fleet to a dev version, the bundles for the fleet-agents are stuck in 'WaitApplied':

https://github.com/rancher/fleet/blob/master/.github/workflows/rancher-integration.yml#L182

Only the fleet-controller is updated to the "rancher/fleet:dev" image. The bundle deployments are not  updated and point to the old contents (containing `image: rancher/fleet-agent:v0.5.0`).

It's unclear how long this is happening, the previous check to make sure the update has finished, did not work correctly and was [flaky](https://github.com/rancher/fleet/actions/workflows/rancher-integration.yml?query=event%3Aschedule):
```
+# wait for fleet to be up on downstream cluster
+sleep 30
+{ grep -q -m 1 "cattle-fleet-system"; kill $!; } < <(kubectl get namespace -w)
+{ grep -q -m 1 "fleet-agent"; kill $!; } < <(kubectl get deployment -n cattle-fleet-system -w)
+# FLAKE: rollout fails: object has been deleted
+# https://github.com/manno/fleet/issues/4
+kubectl -n cattle-fleet-system rollout status deploy/fleet-agent
```

manno:
`helm upgrade --reuse-values` helps with the local agent bundle.

PR #1535, opened 2023-05-15
Fix Upgrade CI

thardeck:
Do not use parameter `--reuse-values` during Fleet upgrades with helm to prevent Gitjob from being updated because its image version reference is stored in the values.
 
 `--reuse-values:  when upgrading, reuse the last release's values and merge in any overrides from the command line via --set and -f. If '--reset-values' is specified, this is ignored`
 
 In general we should consider to move the version information from the value store to the templates.

Github Action Upgrade results:
* [Upgrade Fleet in Rancher to Latest Release](https://github.com/rancher/fleet/actions/runs/4979385018)
* [Upgrade Fleet in Rancher To HEAD](https://github.com/rancher/fleet/actions/runs/4979385765/jobs/8910908099)
* [Upgrade Fleet Standalone to HEAD](https://github.com/rancher/fleet/actions/runs/4979387138/jobs/8911115327)

### DIFF
--- a/.github/scripts/upgrade-rancher-fleet-to-dev-fleet.sh
+++ b/.github/scripts/upgrade-rancher-fleet-to-dev-fleet.sh
@@ -35,8 +35,7 @@ helm upgrade fleet charts/fleet \
   --set image.tag="$fleetTag" \
   --set agentImage.repository="$agentRepo" \
   --set agentImage.tag="$agentTag" \
-  --set agentImage.imagePullPolicy=IfNotPresent \
-  --reuse-values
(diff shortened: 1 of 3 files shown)


manno:
We ran into some trouble without the `--reuse-values` flag a while ago: https://github.com/rancher/fleet/issues/1120#issuecomment-1324078642

Doesn't seem to be relevant anymore?

Matt suggested to use a helper that falls back to the app version if no version, like 'dev', was provided via `--set`.

manno (approved):

PR #1562, opened 2023-05-31
Remove obsolete Github Action cron jobs

thardeck:
there should be no gain in the cron runs since they are already executed against each new pr.

### DIFF
--- a/.github/workflows/e2e-multicluster-ci.yml
+++ b/.github/workflows/e2e-multicluster-ci.yml
@@ -1,6 +1,9 @@
 name: E2E Multi-Cluster Fleet
 
 on:
+  schedule:
+    # Run everyday day at 9:00 AM
+    - cron: '0 9 * * *'
   workflow_dispatch:
     inputs:
       enable_tmate:


weyfonk (dismissed):
LGTM :)

olblak:
> there should be no gain in the cron runs since they are already executed against each new pr.

I am curious, what's the motivation? Are you solving a specific problem?

.github/workflows/ci.yml
```diff
@@ -1,8 +1,6 @@
 name: CI
 
 on:
-  schedule:
```
manno, line 4: This workflow runs unit and integration tests.

.github/workflows/golangci-lint.yml
```diff
@@ -4,8 +4,6 @@ env:
   SETUP_GO_VERSION: '1.19.*'
 
 on:
-  schedule:
-    - cron:  '0 5 * * *'
```
manno, line 8: This workflow only shows *all* issues if run on schedule.

```
only-new-issues: "${{ github.event_name == 'pull_request'}}"
```

manno (changes requested):

.github/workflows/ci.yml
manno: But could turn red because of external dependencies, which we wouldn't notice in periods with little or no PRs.

.github/workflows/golangci-lint.yml
manno: Updates to the linters could quickly change this situation.

PR #1563, opened 2023-05-31
Migrate e2e tests to local git server

### LINKED ISSUE #1556: Enable local end-to-end test runs
Fleet's end-to-end test suite requires secrets because they rely on external test repositories for code and Helm charts.
This prevents that test suite from being run locally as well as from each PR.

Let's refactor it to make use of ephemeral, local repositories, so that:
* each contributor could run the test suite on their machine, both to detect regressions and to expand the test suite
* CI workflows running on pull requests could run the test suite without having to disclose any external secrets.

Tasks:
- [x] refactor `gitrepo` tests to make use of a local git server
- [x] refactor `imagescan` tests to make use of a local git server
- [x] refactor OCI tests to make use of a local Helm repository

weyfonk:
Partly fixes #1556 

This migrates existing `GitRepo` and `imagescan` end-to-end tests to using a local git server, built upon an Nginx instance, which lives inside the test cluster and is set up from within the test cases themselves.

## Test

Run `single-cluster` end-to-end tests locally.

### Tradeoff

* HTTP credentials for basic auth to the git server are provided in the repo URL, in the form `<username>:<password>@<server_url>:<port>`. This is not secure, but arguably this should not be a concern for a local, ephemeral test setup.

### Potential improvement

* Move common git server setup code to eg. `e2e/testenv/githelper/git.go`
* Improve git credentials handling to make use of secrets instead of providing them inline

### DIFF
--- a/.github/workflows/e2e-ci.yml
+++ b/.github/workflows/e2e-ci.yml
@@ -95,6 +95,9 @@ jobs:
         name: E2E Tests
         env:
           FLEET_E2E_NS: fleet-local
+          # Git credentials are here used in a local, ephemeral environment. Leaks would be harmless.
+          GIT_HTTP_USER: "fleet-ci"
+          GIT_HTTP_PASSWORD: "foo"
         run: |
           ginkgo e2e/single-cluster e2e/keep-resources
       -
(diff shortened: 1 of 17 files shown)


go.mod
```diff
@@ -150,6 +150,7 @@ require (
 	github.com/jbenet/go-context v0.0.0-20150711004518-d14ea06fba99 // indirect
 	github.com/jmespath/go-jmespath v0.4.0 // indirect
 	github.com/jmoiron/sqlx v1.3.5 // indirect
+	github.com/joho/godotenv v1.5.1 // indirect
```
manno, line 153: How did this end up here?

.github/workflows/e2e-ci.yml
```diff
@@ -62,6 +62,8 @@ jobs:
         run: |
           docker build -f package/Dockerfile -t rancher/fleet:dev --build-arg="ARCH=$GOARCH" .
           docker build -f package/Dockerfile.agent -t rancher/fleet-agent:dev --build-arg="ARCH=$GOARCH" .
+          cd e2e/assets/gitrepo
+          docker build -f Dockerfile.gitserver -t nginx-git:test --build-arg="passwd=$(openssl passwd foo)" .
```
manno, line 66: How long does this take? Should we publish the image in a separate pipeline, so that we don't have to rebuild every time?

dev/import-images-k3d
```diff
@@ -12,3 +12,8 @@ if [ "$upstream_ctx" != "$downstream_ctx" ]; then
 else
   echo "not importing agent to any downstream clusters. Set FLEET_E2E_CLUSTER_DOWNSTREAM"
 fi
+
+# Build and import git server image
+cd e2e/assets/gitrepo
```
manno, line 17: I run this script a lot while debugging. Can we make gitserver build optional? Maybe via an env var?

weyfonk: Good point, I've added support for env var `FORCE_GIT_SERVER_BUILD` to rebuild the image even if it already exists. Otherwise it should only be built once on your machine.

e2e/require-secrets/gitrepo_test.go
```diff
@@ -36,20 +39,20 @@ var _ = Describe("Git Repo", func() {
 
 	BeforeEach(func() {
 		k = env.Kubectl.Namespace(env.Namespace)
-		gh = githelper.New()
+		gh = githelper.New("", true)
```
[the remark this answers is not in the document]
weyfonk: Agreed! Refactored to use `NewSSH()` and `NewHTTP(addr string)`, which is more readable. Thanks :)

go.mod
weyfonk: This was a leftover from helper code I had added to a `suite_test.go` to read my `.envrc` directly and hence be able to debug test cases from within VS Code. Happy to share that at some point if someone is interested, or to discover better ways of using env vars there :)

manno (approved):

Issue #1552, opened 2023-05-23
Add option to skip bundle targets and customize targeting


manno:
Users can modify targeting from fleet.yaml:

- [ ] skip a bundle deployment on a cluster (doNotDeploy in fleet.yaml)
- [ ] override the cluster selection from fleet.yaml (overrideTargets in fleet.yaml)

Links:
* https://github.com/rancher/fleet/issues/462
* https://github.com/rancher/fleet/issues/1358
* replaces https://github.com/rancher/fleet/issues/952

raulcabello:
# QA Template

## Problem
The only way to create targets is when defining a `GitRepo`. Targets cannot be modified in Bundles using the `fleet.yaml`

## Solution
- Skip a bundle deployment on a cluster using `doNotDeploy` inside of `targetCustomization` in `fleet.yaml`. Clusters that match won't deploy any resource.
- Override the cluster selection using `overrideTargets` in `fleet.yaml`. Targets defined in the `GitRepo` will be overridden by the targets defined in  `overrideTargets` for this `Bundle`

## Testing

### doNotDeploy

Make sure that resources are not deployed if `doNotDeploy` is true in Clusters that contains the `env: test` label. Verify resources are created if `doNotDeploy` is false.
```
targetCustomizations:
- name: test
  doNotDeploy: true
  clusterSelector:
    matchLabels:
      env: test
```

### overrideTargets

Create a `GitRepo` with a different target that the `overrideTargets`. Verify resources are deployed in the `overrideTargets` and not in the targets defined in the `GitRepo`. Targets from GitRepo should be used if  `overrideTargets` is not provided.
```
overrideTargets:
- clusterSelector:
     matchLabels:
        env: test
```

sbulage:
**QA TEST PLAN for `doNotDeploy`**

Scenarios
|Scenario|Test Case|
|---|---|
|1| Test bundle did not get deploy when `doNotDeploy` value set to `true` option is used in the fleet.yaml|
|2| Test bundle deployed when `doNotDeploy: false` is present in the fleet.yaml|
|3| Test existing bundle gets removed when fleet.yaml updated with `doNotDeploy` with value 'true'|

sbulage:
**REPRO STEPS**

**Scenario 1**
1. Add `doNotDeploy: true` option to the fleet.yaml as added [here](https://github.com/sbulage/test-fleet/blob/main/donot-deploy/fleet.yaml#L4 )
[...]

PR #1580, opened 2023-06-05
Add DoNotDeploy and OverrideTargets

raulcabello:
### DoNotDeploy

`BundleDeployments` will not be created for targets that matches a `targetCustomization` that has `DoNotDeploy` set to true, which means no resources will be deployed. If a `targetCustomization` match has `DoNotDeploy` set to true we skip this `BundleDeployment` creation [here](https://github.com/raulcabello/fleet/blob/4e36ea8f82b6bf9b2a805a169850bb234f55183f/pkg/target/target.go#L279-L282). `targetCustomization`  are defined in the `fleet.yaml`.

In the following example:
```yaml
targetCustomizations:
- name: test
  doNotDeploy: true
  clusterSelector:
    matchLabels:
      env: test
```
Resources will not be deployed for `Clusters` that contains the `env: test` label.

### OverrideTargets

Targets defined in a `GitRepo` will be overridden if a `Bundle` defines `overrideTargets` in the `fleet.yaml`. 

Fleet controller creates a target file that contains targets defined in the GitRepo. Then passes this file to `fleet apply` that runs in the job. This PR replaces the content of this file when executing the `fleet apply` command if `overrideTargets` is found in the `fleet.yaml`

For example if a `fleet.yaml` contains:

```yaml
overrideTargets:
- clusterSelector:
     matchLabels:
        env: test
```

The `Bundle` created for this directory will have the targets defined in `overrideTargets` , which means that targets from `GItRepo` will not be used. 

relates to https://github.com/rancher/fleet/issues/1552

### DIFF
--- a/integrationtests/controller/bundle/bundle_targets_test.go
+++ b/integrationtests/controller/bundle/bundle_targets_test.go
@@ -52,6 +52,11 @@ var _ = Describe("Bundle targets", Ordered, func() {
 
 	AfterEach(func() {
 		Expect(bundleController.Delete(env.namespace, bundleName, nil)).NotTo(HaveOccurred())
+		bdList, err := bundleDeploymentController.List("", metav1.ListOptions{LabelSelector: labels.SelectorFromSet(bdLabels).String()})
+		Expect(err).NotTo(HaveOccurred())
+		for _, bd := range bdList.Items {
+			Expect(bundleDeploymentController.Delete(bd.Namespace, bd.Name, &metav1.DeleteOptions{})).NotTo(HaveOccurred())
+		}
 	})
 
 	When("a GitRepo targets all clusters without customization", func() {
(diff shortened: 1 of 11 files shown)


integrationtests/controller/bundle/bundle_targets_test.go
```diff
@@ -301,14 +306,98 @@ var _ = Describe("Bundle targets", Ordered, func() {
...
+					ClusterGroup: "one",
+				},
+			}
+			targetRestrictions = make([]v1alpha1.BundleTarget, len(targetsInGitRepo))
+			copy(targetRestrictions, targetsInGitRepo)
+			targets = append(targets, targetsInGitRepo...)
+		})
+
+		It("one BundleDeployment is created", func() {
+			_ = verifyBundlesDeploymentsAreCreated(expectedNumberOfBundleDeployments, bdLabels, bundleName)
```
weyfonk, line 392: Why do we not wait for the bundle to be ready here, as we do in the two test cases above this one?
Also, would we want to wait for the bundle to be ready in already existing test cases further up in this file?

weyfonk (approved):

integrationtests/controller/bundle/bundle_targets_test.go
raulcabello: In this case it will never be ready because we are just testing this controller, so the BundleDeployment won't be ready as other things need to happen outside of the controller. The previous test case will be ready because it has 0 BundleDeployments

Issue #1465, opened 2023-04-05
[SURE-6125] Master node failures due to fleet high resource usage


kkaempf:
Internal reference: [SURE-6125](https://jira.suse.com/browse/SURE-6125)

#### Issue description:
In one of the downstream clusters, all of the master nodes have been failing one by one consuming all CPU, Memory and I/O.
It has OPA Gatekeeper, Fleet and custom operators for creating network policies.
They have one gitRepo which handles 68 bundles and creates 565 Resources. Scaling down the fleet-agent to zero fixes the issue, and enabling it again causes the issues.
Checked the number of requests in the API server, 134584 out of 206481 requests arises from system:serviceaccount:cattle-fleet-system:fleet-agent

#### Business impact:
Not able to use CI/CD as fleet-agent is causing issues when enabled

#### Troubleshooting steps:
1) One of the 3 Downstream Master machines starts consuming all CPU and RAM.
2) The volume of IO increases a lot, up to 1Gb/s
3) When it occurs, the rke2-server service fails and restarts many times.
3) If the machine was the etcd leader, This leads to the leader election process.
4) With no intervention, after a few hours, a second Downstream Master also consumes all CPU and RAM.
5) Then etcd cluster fails because 2 of the 3 machines are down, and the downstream cluster becomes unavailable under the rancher UI.

#### Repro steps:

Scaled down to 0 the deployment of fleet-agent, no issues for a day
Scaled up fleet-agent to 1 replica, and it failed 30 minutes and even faster later

#### Workaround:
Is workararound available and implemented? no

#### Actual behavior:
Downstream cluster fails when fleet agents are enabled

#### Expected behavior:
Cluster should work flawlessly with fleet-agent enabled

#### Files, logs, traces:

Count of users that made the request
```
2568 "system:node:il0prmb00593"
3093 "system:serviceaccount:trident-system:trident-operator"
4424 "system:serviceaccount:calico-apiserver:calico-apiserver"
5372 "system:admin"
5771 "rke2-cloud-controller-manager"
6470 "system:apiserver"
6574 "system:kube-scheduler"
8052 "system:serviceaccount:kube-system:generic-garbage-collector"
8235 "system:kube-controller-manager"
8247 "system:serviceaccount:kube-system:resourcequota-controller"
10215 "system:serviceaccount:cattle-logging-system:rancher-logging"
134496 "system:serviceaccount:cattle-fleet-system:fleet-agent"
```

[...]

PR #1607, opened 2023-06-19
Rationalize Helm history

moio:
Helps with #1465/[SURE-6125](https://jira.suse.com/browse/SURE-6125)

This patch reduces the number of Helm history entries per deployment from 10 or 5 (defaults were not consistent) down to 2.

Rationale:
1. retrieving the Helm history can be a costly operation in terms of downstream control plane load
1. Helm history retrieval load grows (roughly linearly) with the maximum number of entries kept
1. fleet only ever needs the present and immediately previous entries (at least according to my understanding)
1. users should not interact with Helm directly

Note: current users will see their Helm histories trimmed at first round of updates of the respective deployment bundles/releases. For a one-off cleanup, where needed, this script can be used:

https://gist.github.com/moio/2c9a58cba986a5c771a08de0343d8ced

## Test

This has been tested manually and by users in the context of SURE-6125.

Remarks:
 - main risk is that this patch inadvertently breaks some functionality (my understanding in point 3. above is wrong) - current e2e tests should have that covered
 - `helm history RELEASE_NAME` can be used on the downstream cluster to ensure no more than 2 releases are kept. Open to think about how to test this if it's a blocker

### DIFF
--- a/pkg/helmdeployer/deployer.go
+++ b/pkg/helmdeployer/deployer.go
@@ -45,6 +45,7 @@ const (
 	DefaultServiceAccount        = "fleet-default"
 	KeepResourcesAnnotation      = "fleet.cattle.io/keep-resources"
 	HelmUpgradeInterruptedError  = "another operation (install/upgrade/rollback) is in progress"
+	MaxHelmHistory               = 5
 )
 
 var (
@@ -115,7 +116,7 @@ func NewHelm(namespace, defaultNamespace, labelPrefix, labelSuffix string, gette
 	if err := h.globalCfg.Init(getter, "", "secrets", logrus.Infof); err != nil {
 		return nil, err
 	}
-	h.globalCfg.Releases.MaxHistory = 5
+	h.globalCfg.Releases.MaxHistory = MaxHelmHistory
 	return h, nil
 }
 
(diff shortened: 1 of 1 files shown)


manno:
Shouldn't all the code honor `options.Helm.MaxHistory` if provided by the user?

And since we, apparently, have that option, does '4. users should not interact with Helm directly' really hold? Sadly it seems we're exposing helm a lot and don't treat it as an implementation detail.

manno (approved):

moio:
> Shouldn't all the code honor `options.Helm.MaxHistory` if provided by the user?

That seems reasonable - this PR is only about uniforming values and where they are definted. It does not make that any worse at least, and I wouldn't oppose being more consistent in honoring user-defined options if subsequent PRs.

> And since we, apparently, have that option, does '4. users should not interact with Helm directly' really hold? Sadly it seems we're exposing helm a lot and don't treat it as an implementation detail.

Well, I think that Helm resolves a lot of complexities for us - and that quite naturally translates into a complicated software object with many tunable parameters. FMPOV, many of those knobs make sense in the Fleet context as well as in plain Helm use, so I do not think it is bad to expose them, as they were added for reasons our own users will eventually end up rediscovering. So yes, FMPOV we should expose Helm config parameters and that's not necessarily a bad thing.

That said, an interesting question always comes up when publicizing options for a third-party piece of software: does it make sense to expose them "as-is", or should they be exposed or in a different way (eg. with Fleet-specific names and structures, to retain a feeling of independence)? In this case, I do not think the answer matters any longer though, as the installed base out there is significant, and migrating settings to a different scheme will not add a lot of benefit. So yes, FMPOV we should continue exposing them just as Helm does.

The above does not mean we should encourage direct CLI Helm usage though. One thing is to support the full configuration options that Helm does, making Fleet as flexible as Helm is, but another thing is to claim peaceful coexistence of the two. So far I do not see compelling reasons for that, but I'd like to hear if somebody has. So yes, FMPOV point 4. still holds.

PR #1609, opened 2023-06-20
Rate limit retrieving Helm history

moio:
Helps with #1465/[SURE-6125](https://jira.suse.com/browse/SURE-6125)

This introduces an artificial rate limiting mechanism to avoid Helm history to be retrieved too often, as it can be the heaviest operation on the Kubernetes Control Plane.

This PR was made as a reference and a discussion starter more than an actual fix - ideally a different Helm storage backend could be used, or change Helm storage in such a way only relevant information is retrieved.

This will be discussed in person in week 26.

### DIFF
--- a/pkg/helmdeployer/deployer.go
+++ b/pkg/helmdeployer/deployer.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"os"
 	"strconv"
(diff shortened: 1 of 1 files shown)


manno:
We're implementing a cached helm backend in https://github.com/rancher/fleet/issues/1738

Issue #1553, opened 2023-05-23
Add custom labels to bundle namespaces


manno:
## SURE-6393

[PSA](https://kubernetes.io/docs/concepts/security/pod-security-admission/) need labels on namespaces.

Fleet only creates the `defaultNamespace` or `targetNamespace` and it does so by setting helm's `CreateNamespace=true`.

_Note: We assumed any other namespaces mentioned in one of the the bundle's resources will also be created, but this is not the case. Fleet will error with `'ErrApplied(1) [Cluster fleet-local/local: namespaces "example" not found]'`._

Fleet also creates namespace resources which are contained in the bundle and controls their lifecycle.

Acceptance Criteria:

- [x] namespace resources in a bundle are processed correctly, including labels and annotations (works already)
- [x] annotations and labels for `defaultNamespace` (= `targetNamespace`) can be configured
- [x] mention in the docs that namespace resources contained in a bundle will be deleted with the bundle. this might affect non-bundle resources in the namespace.

## Discussion

Including a resource for the `defaultNamespace` in the bundle doesn't work, we would need to set helm's `Force` option and we would change the uninstall behavior.
We probably need to configure the default namespace explicitly from `fleet.yaml`. 

```
namespaceLabels:
    env: test
namespaceAnnotations: 
    pod: deny
```

* To add the labels and annotations we could use a Helm post install hook.
* We could also create the namespace from code and turn off `CreateNamespace`.

## Related

* https://github.com/rancher/fleet/issues/1484 (labels PSA)
* https://github.com/rancher/fleet/issues/1461 (annotations)

raulcabello:
# QA Template

## Problem
There is no way to add annotations or labels to namespaces created by Fleet.

## Solution
Fleet uses the option `--create-namespace` from helm install to create namespaces. Namespaces are created with only one label with `name` as key and no annotations.

This PR adds two new fields `NamespaceLabels` and `NamespaceAnnotations` to the fleet.yaml in order to add labels or annotations to the namespace created by Fleet.

## Testing

- If no `NamespaceLabels` or `NamespaceAnnotations`  is present no labels or annotations should be added (just the name label added by helm)
- Add `NamespaceLabels` and `NamespaceAnnotations`. Verify annotations and label are added.
- Modify existing labels and annotations
[...]

PR #1627, opened 2023-07-03
Add NamespaceLabels and NamespaceAnnotations

raulcabello:
This adds the ability to append labels or annotations to the namespace created by fleet as part of the helm install process.

`NamespaceLabels` and `NamespaceAnnotations` are pointer to maps. Nothing will be done if they are nil, which means they are not provided by the user.

Helm creates the namespace with one label whose key is `name`. This is the only label we preserve, other labels or annotations that are not provided in `NamespaceLabels` or `NamespaceAnnotations` will be removed.

refers to https://github.com/rancher/fleet/issues/1553


internal/cmd/agent/controllers/bundledeployment/controller.go
```diff
@@ -130,11 +136,110 @@ func (h *handler) DeployBundle(bd *fleet.BundleDeployment, status fleet.BundleDe
...
+
+	if ns.Annotations == nil && bd.Spec.Options.NamespaceAnnotations != nil {
+		ns.Annotations = map[string]string{}
+	}
+	if bd.Spec.Options.NamespaceLabels != nil {
+		addLabelsFromOptions(ns.Labels, *bd.Spec.Options.NamespaceLabels)
+	}
+	if bd.Spec.Options.NamespaceAnnotations != nil {
+		addAnnotationsFromOptions(ns.Annotations, *bd.Spec.Options.NamespaceAnnotations)
+	}
```
weyfonk, line 172: I feel that this would be easier to read. How do you see it?
```suggestion
	if bd.Spec.Options.NamespaceLabels != nil {
		addLabelsFromOptions(ns.Labels, *bd.Spec.Options.NamespaceLabels)
	}
	if bd.Spec.Options.NamespaceAnnotations != nil {
	        if ns.Annotations == nil {
	                ns.Annotations = map[string]string{}
	       }
		addAnnotationsFromOptions(ns.Annotations, *bd.Spec.Options.NamespaceAnnotations)
	}
```

internal/cmd/agent/controllers/bundledeployment/controller.go
```diff
@@ -130,11 +136,110 @@ func (h *handler) DeployBundle(bd *fleet.BundleDeployment, status fleet.BundleDe
...
+	err = h.updateNamespace(ns)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (h *handler) updateNamespace(ns *corev1.Namespace) error {
```
weyfonk, line 181: Do you think it would be clearer to mention that this method applies changes in the cluster? eg. something like:
```suggestion
// updateNamespace updates a namespace resource in the cluster.
func (h *handler) updateNamespace(ns *corev1.Namespace) error {
```

internal/cmd/agent/controllers/bundledeployment/controller.go
```diff
@@ -130,11 +136,110 @@ func (h *handler) DeployBundle(bd *fleet.BundleDeployment, status fleet.BundleDe
...
+	// Delete labels not defined in the options.
+	// Keep the name label as it is added by helm when creating the namespace.
+	for k := range nsLabels {
+		if _, ok := optLabels[k]; k != "name" && !ok {
+			delete(nsLabels, k)
+		}
+	}
+}
+
+func addAnnotationsFromOptions(nsAnnotations map[string]string, optAnnotations map[string]string) {
```
weyfonk, line 230:
```suggestion
// addAnnotationsFromOptions updates nsAnnotations so that it only contains all annotations specified in optAnnotations.
func addAnnotationsFromOptions(nsAnnotations map[string]string, optAnnotations map[string]string) {
```

internal/cmd/agent/controllers/bundledeployment/controller_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+			defer ctrl.Finish()
+
+			mockDynamic := mocks.NewMockInterface(ctrl)
+			mockNamespaceableResourceInterface := mocks.NewMockNamespaceableResourceInterface(ctrl)
+			u, _ := runtime.DefaultUnstructuredConverter.ToUnstructured(&test.ns)
+			mockDynamic.EXPECT().Resource(gomock.Any()).Return(mockNamespaceableResourceInterface).Times(2)
```
weyfonk, line 100: Could we perhaps add a comment explaining where this `2` comes from?

internal/cmd/agent/controllers/bundledeployment/controller_test.go
```diff
@@ -0,0 +1,119 @@
...
+			},
+			release: "namespace/foo/bar",
+			expectedNs: corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels:      map[string]string{"bdLabel": "labelUpdated"},
+					Annotations: map[string]string{"bdAnn": "annUpdated"},
+				},
+			},
+			expectedErr: nil,
+		},
```
weyfonk, line 89: Would it make sense to add a test case where the namespace is not found?

internal/cmd/agent/controllers/bundledeployment/controller_test.go
```diff
@@ -0,0 +1,119 @@
...
+			}).Return(&unstructured.UnstructuredList{
+				Items: []unstructured.Unstructured{{Object: u}},
+			}, nil).Times(1)
+			uns, _ := runtime.DefaultUnstructuredConverter.ToUnstructured(&test.expectedNs)
+			mockNamespaceableResourceInterface.EXPECT().Update(gomock.Any(), &unstructured.Unstructured{Object: uns}, gomock.Any()).Times(1)
+
+			h := handler{
+				dynamic: mockDynamic,
+			}
+			err := h.setNamespaceLabelsAndAnnotations(test.bd, test.release)
```
weyfonk, line 112: Maybe something for a wider discussion: I'm wondering about testing unexported methods, not least because I like having tests isolated in `$package_test` packages, but also because if at some point we omit or comment out the call to `setNamespaceLabelsAndAnnotations` in the controller, this test driver will still pass although Fleet will not behave as we want it to.
Should we add a higher-level test case here, of course without repeating all combinations, but just to make sure that labels and annotations are altered when deploying a namespace from a bundle?

weyfonk (commented):
LGTM mostly, with a few comments; most of them are nitpicks, but I think that one testing question prompts further discussion.
Thanks for this :)

internal/cmd/agent/controllers/bundledeployment/controller.go
```diff
@@ -130,11 +136,110 @@ func (h *handler) DeployBundle(bd *fleet.BundleDeployment, status fleet.BundleDe
...
+	if bd.Spec.Options.NamespaceLabels == nil && bd.Spec.Options.NamespaceAnnotations == nil {
+		return nil
+	}
+
+	ns, err := h.fetchNamespace(releaseID)
+	if err != nil {
+		return err
+	}
+
+	if reflect.DeepEqual(bd.Spec.Options.NamespaceLabels, ns.Labels) && reflect.DeepEqual(bd.Spec.Options.NamespaceAnnotations, ns.Annotations) {
```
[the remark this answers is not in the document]
raulcabello: I don't think reflect will cause any performance issue here. The overhead of using reflection in this case is minimal and won't be noticed by the users. For maps it does exactly what we need, from the docs: 
```
Map values are deeply equal when all of the following are true: they are both nil or both non-nil, they have the same length, and either they are the same map object or their corresponding keys (matched using Go equality) map to deeply equal values.
```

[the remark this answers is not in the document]
weyfonk: Good to know, thanks!

internal/cmd/agent/controllers/bundledeployment/controller_test.go
raulcabello: I had to create different mocks for that. I didn't do it because I thought it was not worh it and it was ok with the other test cases. However you might be right, so I added it in https://github.com/rancher/fleet/pull/1627/commits/d06724c7cba0b85ab6a124f3064bd1692dd98572

internal/cmd/agent/controllers/bundledeployment/controller_test.go
raulcabello: I wish we could. However the current code doesn't allow that, a refactor would need to be done in other to inject mocks for all dependencies.

weyfonk: I see; and an end-to-end test case for that could be a bit overkill.
It could be nice to follow the Golang proverb "accept interfaces, return structs" to ease mocking, which of course is somewhat beyond the scope of this PR :)

weyfonk (approved):
LGTM :)

internal/cmd/agent/controllers/bundledeployment/controller_test.go
manno: Indeed, would be great if we could use `_test` for tests and do away with the "java-style" interfaces. But a lot of the existing code is that way and wrangler complicates it further. While we're at it, I'd like a proper `context.Context` in the networked methods. I think wrangler puts one in a struct, but it's not passed into the handlers.

PR #1629, opened 2023-07-04
Use one agent for single-cluster e2e tests in CI

weyfonk:
This decreases the number of k3d agents to 1 to reduce system load and thereby improve stability of the test suite in CI. The downstream cluster used by the multi-cluster CI workflow already relies on a single k3d agent.

### Tradeoff

This approach favours simplicity over a more production-like environment, and may need to be revisited in the future if more services are added to our single-cluster test suite setup.


manno:
closed for https://github.com/rancher/fleet/pull/1680

PR #1655, opened 2023-07-18
Fix linting errors

weyfonk:
This fixes a few (mostly formatting) errors output by the linter, although not all of them. The following are left:

```
internal/cmd/agent/deployer/internal/scheme/scheme.go:7:2: blank-imports: a blank import should be only in a main or test package, or have a comment justifying it (revive)
        _ "k8s.io/kubernetes/pkg/apis/admission/install"
        ^
internal/cmd/controller/controllers/clusterregistration/controller.go:39:2: G101: Potential hardcoded credentials (gosec)
        AgentCredentialSecretType     = "fleet.cattle.io/agent-credential"
        ^
internal/cmd/agent/register/register.go:266:15: G107: Potential HTTP request made with variable url (gosec)
        if _, err := http.Get(apiServerURL); err == nil {
                     ^
internal/cmd/agent/deployer/internal/normalizers/knowntypes_normalizer.go:32:79: unexported-return: exported func NewKnownTypesNormalizer returns unexported type *normalizers.knownTypesNormalizer, which can be annoying to use (revive)
func NewKnownTypesNormalizer(overrides map[string]resource.ResourceOverride) (*knownTypesNormalizer, error) {
```

### DIFF
--- a/internal/cmd/agent/deployer/monitor.go
+++ b/internal/cmd/agent/deployer/monitor.go
@@ -233,7 +233,7 @@ func modified(plan apply.Plan, resourcesPreviousRelease *helmdeployer.Resources)
 	for gvk, keys := range plan.Create {
 		for _, key := range keys {
 			if len(result) >= 10 {
-				return
+				return result
 			}
 
 			apiVersion, kind := gvk.ToAPIVersionAndKind()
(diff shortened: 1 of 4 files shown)


internal/cmd/agent/deployer/monitor.go
```diff
@@ -233,7 +233,7 @@ func modified(plan apply.Plan, resourcesPreviousRelease *helmdeployer.Resources)
 	for gvk, keys := range plan.Create {
 		for _, key := range keys {
 			if len(result) >= 10 {
-				return
+				return result
```
manno, line 236: Since when does it complain about that...

manno (approved):

Issue #1615, opened 2023-06-23
Clean up left-over registration resources


manno:
We added clean up for the `request-` service accounts in  https://github.com/rancher/fleet/pull/1310/files 
There were several annotations, which we did not convert to labels. The clean up code seems to ignore resources.

Yet, with Fleet 0.7 we see left over `request-` service account tokens.

sbulage:
When upgraded rancher from 2.7.3 --> 2.7.4 --> 2.7.5-rc5 I see spike in increase of `request-` service accounts/tokens are remaining.

![Selection_038](https://github.com/rancher/fleet/assets/12134754/0935e704-409e-43e4-b540-3ce6845751e4)

manno:
It's been reported that this is not new behavior.

The "ClusterRegistration" CR is created by the fleet-agent. While it is an owner of other resources, like the `request-` rolebinding, it has no owner. From reading the code, it seems it is [only deleted when the `Cluster` resource is deleted](https://github.com/rancher/fleet/blob/fix-resource-status-timestamp/internal/cmd/controller/controllers/cluster/controller.go#L118-L131). 

Related issues:

* https://github.com/rancher/fleet/issues/1605 (same problem)


Why is the registration process failing, why does the agent create multiple cluster registrations, instead of updating a single registration request? Since they use `GenerateName` this seems to be by design.

manno:
closing as duplicate of https://github.com/rancher/fleet/issues/1656

manno:
re-opening since #1656 is a backport issue

kkaempf:
closing as duplicate of #1651

PR #1658, opened 2023-07-20
Clean up old clusterregistrations and remember cluster's api server

manno:
Refers to https://github.com/rancher/fleet/issues/1615

This adds clean-up for outdated clusterregistration resources.
It also adds status fields to the cluster resource, so fleet can decide better if it needs to re-deploy and re-register an agent when starting up, or detecting a config change. 

It also improves  the error messages in the agent's startup.


internal/cmd/controller/controllers/cluster/import.go
```diff
@@ -85,27 +90,48 @@ func (i *importHandler) onConfig(config *config.Config) error {
...
-			if err != nil {
-				return err
+	b := &backoff.Backoff{
+		Min:    1 * time.Second,
+		Max:    3 * time.Minute,
+		Factor: 2,
+		Jitter: true,
+	}
+
+	go func() {
```
raulcabello, line 100: why do we use a goroutine here? I think this might trigger more simultaneous calls

internal/cmd/controller/controllers/cluster/import.go
```diff
@@ -85,27 +90,48 @@ func (i *importHandler) onConfig(config *config.Config) error {
...
-		if string(secret.Data["apiServerURL"]) == "" || string(secret.Data["apiServerCA"]) == "" {
-			logrus.Debugf("API server fallback-config changed, trigger cluster import for cluster %s/%s", cluster.Namespace, cluster.Name)
-			c := cluster.DeepCopy()
-			c.Status.AgentConfigChanged = true
-			_, err := i.clusters.UpdateStatus(c)
-			if err != nil {
-				return err
+	b := &backoff.Backoff{
+		Min:    1 * time.Second,
+		Max:    3 * time.Minute,
```
raulcabello, line 95: maybe 3 minutes is too much?

raulcabello (approved):

Issue #1651, opened 2023-07-17
Fleet creating large number of resources for each cluster


kkaempf:
### Environment
Rancher Cluster:
Rancher version: 2.7.5
Number of nodes: 37 (30 for ARGOcd)

Downstream Cluster:
Number of Downstream clusters: 1132
RKE/RKE2/K3S version: 2.7.5
Kubernetes version: 1.25.6
CNI: calico + traefik

Other:
Underlying Infrastructure: Rancher running in managed AKS cluster

### Issue description:
ETCD has filled with a large number of fleet resources, or resources in the fleet-default namespace. The number of resources is not proportionate to the number of downstream clusters.
~~~
ROLE:
request-6s7nb                                                  2023-07-13T12:46:51Z
ROLEBINDING:
request-6s7nb                                                     Role/request-6s7nb                                                  25h
CLUSTERREGISTRATION:
request-6s7nb   s0409          {"management.cattle.io/cluster-display-name":"s0409","management.cattle.io/cluster-name":"c-m-zcllc9kx","objectset.rio.cattle.io/hash":"d7128da63a4e07ede9c4d36b1a5fd60b31ce3d45","provider.cattle.io":"k3s"}
~~~
For example, the cluster s0409 has 170 similar requests.
Some clusters have 300+ requests for clusterregistration, none of which are cleaned up.

### Business impact:
Kube-apiserver in AKS is hitting its limits and is causing a large number of timeouts when all clusters are attempting to connect at once on startup. It takes ~30 minutes for clusters to be able to connect. ETCD is getting overloaded at is around 2G after compact/defrag

### Troubleshooting steps:
Worked with AKS team to reduce load and pinpoint what is filling ETCD. Argo takes up a large amount of space, but most API calls are coming from Rancher. See images for API volume to kube-apiserver, as well as ETCD object count and size. You can see 80k cluster registrations which is large for the 1132 clusters they have registered. Why would that be?

### Actual behavior:
large number of fleet resources created for each cluster (roles, rolebindings, clusterregistrations, secrets, serviceAccounts) in the fleet default namespace. Customer sees large amount of timeouts to downstream clusters, possibly causing fleet to treat it as new clusters.

### Expected behavior:
Smaller, appropriate number of resources created for each cluster.

kkaempf:
see also #1615

weyfonk:
[...]

PR #1675, opened 2023-07-25
Clusteregistration less aggressive cleanup

manno:
Refers to #1651

Previous fix would clean up cluster registrations that were still used by the agent, if the agent restarted quickly.

### DIFF
--- a/internal/cmd/controller/controllers/clusterregistration/controller.go
+++ b/internal/cmd/controller/controllers/clusterregistration/controller.go
@@ -128,7 +128,7 @@ func (h *handler) OnCluster(key string, cluster *fleet.Cluster) (*fleet.Cluster,
 	}
 	for _, cr := range crs {
 		if !cr.Status.Granted {
-			logrus.Infof("Namespace assigned to %s/%s triggering %s/%s", cluster.Namespace, cluster.Name,
+			logrus.Infof("Namespace assigned to cluster '%s/%s' enqueues cluster registration '%s/%s'", cluster.Namespace, cluster.Name,
 				cr.Namespace, cr.Name)
 			h.clusterRegistration.Enqueue(cr.Namespace, cr.Name)
 		}
@@ -182,7 +182,7 @@ func (h *handler) OnChange(request *fleet.ClusterRegistration, status fleet.Clus
 	sa, err := h.serviceAccountCache.Get(cluster.Status.Namespace, saName)
 	if err == nil {
 		if secret, err := h.authorizeCluster(sa, cluster, request); err != nil {
-			return nil, status, err
+			return nil, status, fmt.Errorf("failed to authorize cluster: %w", err)
 		} else if secret != nil {
 			status.Granted = true
 			objects = append(objects, secret)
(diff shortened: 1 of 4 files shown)


internal/cmd/controller/controllers/clusterregistration/controller.go
```diff
@@ -191,18 +191,18 @@ func (h *handler) OnChange(request *fleet.ClusterRegistration, status fleet.Clus
...
-			logrus.Infof("Deleting old clusterregistration %s/%s", creg.Namespace, creg.Name)
-			if err := h.clusterRegistration.Delete(creg.Namespace, creg.Name, nil); err != nil && !apierrors.IsNotFound(err) {
-				return nil, status, err
+	if status.Granted {
+		logrus.Debugf("Cluster registration request '%s/%s' granted, creating registration secret", request.Namespace, request.Name)
+
+		crlist, _ := h.clusterRegistration.List(request.Namespace, metav1.ListOptions{})
+		for _, creg := range crlist.Items {
+			// delete any other cluster registration with the same clientID, but different random and older creation timestamp
+			if creg.Spec.ClientID == request.Spec.ClientID && creg.Spec.ClientRandom != request.Spec.ClientRandom && creg.Name != request.Name && creg.CreationTimestamp.Time.Before(request.CreationTimestamp.Time) {
```
weyfonk, line 201: This, or a version with proper indentation for the last line (my editor would take care of this, but Github doesn't) would be a bit easier to read:
```suggestion
			if creg.Spec.ClientID == request.Spec.ClientID &&
			   creg.Spec.ClientRandom != request.Spec.ClientRandom &&
			   creg.Name != request.Name &&
			   creg.CreationTimestamp.Time.Before(request.CreationTimestamp.Time) {
```

internal/cmd/controller/controllers/clusterregistration/controller.go
```diff
@@ -191,18 +191,18 @@ func (h *handler) OnChange(request *fleet.ClusterRegistration, status fleet.Clus
...
 
-	// delete old clusterregistrations
-	crlist, _ := h.clusterRegistration.List(request.Namespace, metav1.ListOptions{})
-
-	for _, creg := range crlist.Items {
-		if creg.Spec.ClientID == request.Spec.ClientID && creg.Spec.ClientRandom != request.Spec.ClientRandom {
-			logrus.Infof("Deleting old clusterregistration %s/%s", creg.Namespace, creg.Name)
-			if err := h.clusterRegistration.Delete(creg.Namespace, creg.Name, nil); err != nil && !apierrors.IsNotFound(err) {
-				return nil, status, err
+	if status.Granted {
```
weyfonk, line 195: What should happen if `!status.Granted`? Should the cluster registration, role bindings, etc, be created anyway?
In other words, should we return here if `!status.Granted`?

weyfonk (commented):
Thanks for this! Looks good, with a nitpick and a doubt on my end.

internal/cmd/controller/controllers/clusterregistration/controller.go
manno: Sure, I don't need to have the statemement it in one line... I thought about extracting it into a nicely named function, but couldn't come up with a name...

Why is there a new line in 204?

weyfonk: > Sure, I don't need to have the statemement it in one line... I thought about extracting it into a nicely named function, but couldn't come up with a name...
How about something like `shouldDeleteClusterRegistration`, or a shorter equivalent, eg. `shouldDelete` since we're already in package `clusterregistration`?

> Why is there a new line in 204?

The new line is there because I messed up formatting, not sure if clumsiness-related or caused by an issue with the web editor. Ideally line 204 would be indented like the lines above it, without additional line break :)

raulcabello (dismissed):

internal/cmd/controller/controllers/clusterregistration/controller.go
manno: Should be fine, but this is an optimization that doesn't necessarily belong into this PR? Or does it...

weyfonk: Right, as discussed, let's handle that via a separate PR, no worries.

weyfonk (approved):

PR #1676, opened 2023-07-26
Return early from clusterregistration handler

manno:
Refers to #1651 

The clusterregistration.OnChange handler would create objects before the secret token was generated. It's not necessary to create the objects for requests that are not granted.

This also removes the deep nesting from the little state machine of if statements and makes the outcome more explicit.


internal/cmd/controller/controllers/clusterregistration/controller.go
```diff
@@ -180,51 +180,47 @@ func (h *handler) OnChange(request *fleet.ClusterRegistration, status fleet.Clus
...
 	saName := name.SafeConcatName(request.Name, string(request.UID))
 	sa, err := h.serviceAccountCache.Get(cluster.Status.Namespace, saName)
-	if err == nil {
-		if secret, err := h.authorizeCluster(sa, cluster, request); err != nil {
-			return nil, status, fmt.Errorf("failed to authorize cluster: %w", err)
-		} else if secret != nil {
-			status.Granted = true
-			objects = append(objects, secret)
-		}
+	if err != nil {
```
raulcabello, line 183: should we check if it is a not found error here before creating the sa?

internal/cmd/controller/controllers/clusterregistration/controller_test.go
```diff
@@ -0,0 +1,254 @@
...
+				Expect(objs).To(BeEmpty())
+				Expect(newStatus.Granted).To(BeFalse())
+				Expect(newStatus.ClusterName).To(Equal("cluster"))
+			})
+		})
+
+		When("service account does not exist", func() {
+			BeforeEach(func() {
+				cluster.Status = fleet.ClusterStatus{Namespace: "fleet-default"}
+				saCache.EXPECT().Get(gomock.Any(), gomock.Any()).AnyTimes().Return(nil, notFound)
```
raulcabello, line 173: I'd be nice to assert the number of times methods should be called instead of always using `AnyTimes`

internal/cmd/controller/controllers/clusterregistration/controller.go
manno: Yes

raulcabello (approved):

PR #1680, opened 2023-07-27
Remove k8s 1.20 from PR CI matrix

manno:
We have a couple of flakes in our CI, however the setup itself seems unstable. It's not just specific tests.

This reduces the k3d docker containers to one.

Also remove the two year old `rancher/k3s:v1.20.15-k3s1` from the matrix. 1.20 is no longer in the support matrix of Rancher manager: https://www.suse.com/suse-rancher/support-matrix/all-supported-versions/rancher-v2-7-1/


kkaempf (dismissed):
👍🏻

weyfonk (approved):

Issue #1679, opened 2023-07-27
Remove agent restart after import


manno:
During the initial deployment of an agent, with manager initiated registration, first a deployment is created. Then the deployment is "adopted" by creating a bundle. The downstream agent will update itself according to the information in the bundle. This leads to multiple restarts of the agent deployment.

Previously: https://github.com/rancher/fleet/issues/1388
Docs: https://github.com/rancher/fleet-docs/pull/86

Implemented by:
* https://github.com/rancher/fleet/pull/1687
* https://github.com/rancher/fleet/pull/1678

### Solution

Only one agent pod runs the cluster registration. When importing a cluster (or re-importing it via `redeployAgentGeneration`) no restart of the deployment is forced by changing an env variable.
 
### Testing

* Set up Rancher
*  See that fleet correctly registered the local cluster
* force a redeployment of the agen, either via UI, or patching `redeployAgentGeneration` on the local cluster resource
* See that the agent is restarted and reregistered
* Deploy a bundle, see that it still works

This probably needs to be done a few times, as the installation is prone to timing problems.
The agent should still restart, when changing the cluster configuration, e.g. tolerations.

#### Manual Testing

Manually tested with manager initiated registration and the local cluster.

#### Automated Testing

We have no advanced installation tests. 

### QA Testing Considerations

Everything around installing and upgrading Rancher/Fleet should work as before. The fleet agent would restart less often.
This change also shouldn't affect existing cluster registrations.

#### Regressions Considerations

This should not affect agent initiated registration.

Issue #1388, opened 2023-03-09
Clean  Up: Agent Deployment


manno:
During the initial deployment of an agent, with manager initiated registration, first a deployment is created. Then the deployment is "adopted" by creating a bundle. The downstream agent will update itself according to the information in the bundle. This leads to multiple restarts of the agent deployment.

Can we deploy agent less often? Why is “generation” switched from clusterid to “bundled”?

Investigate and document here:

- [ ] How does cluster registration behave when we [remove the `GENERATION` env var](https://github.com/rancher/fleet/blob/master/pkg/agent/manifest.go#L174).
- [ ] What does the agent do during startup, is restarting the agent helpful for updates, cluster registrations?
- [ ] We sometimes observe multiple restarts during upgrades. Are the restarts from [import.go](https://github.com/rancher/fleet/blob/master/pkg/controllers/cluster/import.go#L71) or [manageagent.go](https://github.com/rancher/fleet/blob/master/pkg/controllers/manageagent/manageagent.go#L61)?

manno:
I looked into this and I'm not sure we can improve the situation as there are too many edge cases.

## Notes

When adding a cluster, we see that the agent restarts at least once. 

This is triggered by a difference in the agent’s deployment. Namely, the `GENERATION` env variable.
In case of an agent initiated cluster registration, the fleet-agent helm chart does not specify the variable. In case of a manager-initiated registration, the deployment done by “import.go” will set `GENERATION` to `cluster.UID+cluster.Generation`.

When the cluster, or it’s namespace change, the deployment is adopted by creating a bundle. That bundle uses “bundled” as value for `GENERATION`. 

The `GENERATION` env var is not used by any code. It’s was added in:

db171f73 - Fix issues with agent deployment - Darren Shepherd (2 years, 6 months ago)

Probably only to trigger redeployments of the agent.

How would the `GENERATION` env var trigger a redeployment/restart?
* for every change

Which install operations could depend on a restart?
* uninstallation, the downstream agent will remain in the cluster, only the helm history is deleted. if the agent is adopted, a restart might be needed.
* reinstallation via chart, maybe restart to reregister?
* reinstallation via `cluster.Spec.RedeployAgentGeneration` (not the same generation, uses `cluster.Status.AgentDeployedGeneration` instead)
* expired token/registration?
* upgrading fleet will also restart the fleet-controller, that might happen in parallel with the agent restarts

PR #1687, opened 2023-08-01
Restrict registration to leader agent

manno:
Refers to https://github.com/rancher/fleet/issues/1679 (see also https://github.com/rancher/fleet/issues/1388 and https://github.com/rancher/fleet/issues/1651)

# Problem

When upgrading fleet  the agent's deployment is deleted and recreated and maybe updated. 
The k8s deployment controller doesn't ensure that only one pod is running. In fact it tries to keep the old pod running, so that it can finish remaining requests, while it simultaneously starts the new pod for new requests. Even switching to a "replace" update strategy doesn't guarantee that only one pod is running.

However, when the agent pod starts up and a `fleet-bootstrap-secret` is present, it will run the cluster registration. Once the registration is finished the bootstrap secret is deleted. The cluster registration in `register.Register` is a blocking loop, that keeps k8s from terminating the pod. 
Registration also has to wait for the upstream cluster to generate the service account's token secret, so it might exit with an error and restart itself. Only once it has a valid kubeconfig for the upstream cluster and can reach the API server, it will delete the bootstrap secret and register controllers to watch for events.

Since the agent update is simultaneously initiated from fleet-controller's "import.go", as well as the agent itself (due to the bundle update from "manageagent"), the possibility of two pods being started at once is high. It seems the agressive re-registering code in "import.go" makes this worse, by deleting the deployment, then immediately deleting the deployment's pods and creating a new deployment. Deleting a deployment also orphans the pods, so k8s garbage collection tries to delete them at the same time.

See https://fleet.rancher.io/ref-registration#fleet-agent---clusterregistration for more information.

# Solution

This PR moves the registration code inside the leader election. That way, only pods that won the election can register. 

# Notes

- [x] clean up code after moving it inside RunOrDie
- [x] remove debug sleep
- [x] add pod label for debugging
- [x] ~~switch agent to statefulset/just pod?~~

This adds the agent's pod hostname to the clusterregistration's labels for debugging: ` "fleet.cattle.io/created-by-agent-pod"] = os.Getenv("HOSTNAME")`

Not switching the agent deployment to a pod/statefulset at this time. With all the registration code only running on "leaders", we should see less registrations.

Note: There is no way to configure an agent's replica count, so normally there is only one agent running. Multiple agents only run at the same time  during deployment upgrades. Switching to the `recreate` rollout strategy didn't change that, probably because multiple controllers are updating the deployment, even [deleting its pods(s) directly](https://github.com/rancher/fleet/blob/e6946481ce04bc36bc89ded29eeccb6976a83a4b/internal/cmd/controller/controllers/cluster/import.go#L203-L205), at the same time.

### DIFF
--- a/internal/cmd/agent/start.go
+++ b/internal/cmd/agent/start.go
@@ -2,20 +2,22 @@ package agent
 
 import (
 	"context"
-	"sync"
 	"time"
 
 	"github.com/rancher/fleet/internal/cmd/agent/controllers"
 	"github.com/rancher/fleet/internal/cmd/agent/register"
+	"github.com/sirupsen/logrus"
 
 	"github.com/rancher/lasso/pkg/mapper"
 	"github.com/rancher/wrangler/pkg/kubeconfig"
+	"github.com/rancher/wrangler/pkg/leader"
 	"github.com/rancher/wrangler/pkg/ratelimit"
 	"github.com/rancher/wrangler/pkg/ticker"
 
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/discovery/cached/memory"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/restmapper"
 	"k8s.io/client-go/tools/clientcmd"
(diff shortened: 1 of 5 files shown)


internal/cmd/agent/start.go
```diff
@@ -42,59 +44,62 @@ func start(ctx context.Context, kubeConfig, namespace, agentScope string, opts *
...
+	k8s, err := kubernetes.NewForConfig(localConfig)
 	if err != nil {
 		return err
 	}
 
-	fleetNamespace, _, err := agentInfo.ClientConfig.Namespace()
-	if err != nil {
-		return err
-	}
+	leader.RunOrDie(ctx, namespace, "fleet-agent-lock", k8s, func(ctx context.Context) {
```
raulcabello, line 55: why do we want to do the controllers registration inside the leader block? Controllers will not start receiving events until the factories are started in https://github.com/rancher/fleet/blob/master/internal/cmd/agent/controllers/controllers.go#L129 

I'd prefer to leave everything ready for the slave agent, so it takes less time to start handling events when it takes the leadership. And if there is any error when registering we know it before it is leader. Am I missing any advantage of waiting for it to be leader?

manno: I updated the explanation in the PR :)

raulcabello: thanks!

raulcabello (approved):

Issue #1674, opened 2023-07-25
Some resources remain even after downstream cluster was deleted


kkaempf:
### Issue description:
Some resources remain even after downstream cluster was deleted.
Is this a bug?
As a workaround, could customer remove these resources?

### Business impact:
100,000 of resources remain and etcd servers are under pressure.

Troubleshooting steps:
none.

### Repro steps:
I could reproduce it in local aws environment.

 
[Rancher server]
OS: Ubuntu 22.04
Docker: 20.10.23
RKE: 1.4.6
Kubernetes: rancher/hyperkube:v1.26.4-rancher2
Rancher: 2.7.5
Fleet: 0.7.0

[Downstream cluster]
OS: Ubuntu 22.04
Docker: 20.10.23
Kubernetes: v1.26.4-rancher2

[Create custom Downstream cluster]
Create -> RKE1 -> Custom

[Delete the custom Downstream cluster]
From UI.

[Trailing resources remains]
 
~~~
$ kubectl -n fleet-default get clusterregistration.fleet.cattle.io
NAME            CLUSTER-NAME   LABELS
request-rj6t5   c-c88j8        {"management.cattle.io/cluster-display-name":"ds-rke1","management.cattle.io/cluster-name":"c-c88j8","objectset.rio.cattle.io/hash":"7e3568c9948eb7abcae51039a25f718ab2b8e53f","provider.cattle.io":"rke"} 

$ kubectl -n fleet-default get rolebinding.rbac.authorization.k8s.io
NAME            ROLE                 AGE
request-rj6t5   Role/request-rj6t5   5m8s 

$ kubectl -n fleet-default get role.rbac.authorization.k8s.io
NAME            CREATED AT
request-rj6t5   2023-07-03T01:44:37Z
~~~
 
### Workaround:
Is workararound available and implemented? no

### Actual behavior:
Some resouces remain after cluster deletion.

### Expected behavior:
No resource remains after cluster deletion.

manno:
The fleet controller would clean them up when it sees the cluster resource being deleted. However, if the fleet controller misses the event, the resources will remain.

This should be fixed by the new hook implemented in https://github.com/rancher/fleet/issues/1690
Every time fleet is upgraded it will remove orphaned cluster registrations.  This can also be done manually with the fleet cli, by running `fleet cleanup --min 1ms --max 1ms`

kkaempf:
This should be QA validated

sbulage:
[...]

PR #1689, opened 2023-08-01
Add hook on upgrade to clean up old, duplicate clusterregistrations

manno:
Refers to #1690 

- [x] translate into go
- [ ] add deletion of creg without  a cluster for https://github.com/rancher/fleet/issues/1674


charts/fleet/templates/job_cleanup_clusterregistrtions.yaml
```diff
@@ -0,0 +1,87 @@
...
+    metadata:
+      labels:
+        app: fleet-job
+    spec:
+      serviceAccountName: fleet-controller
+      restartPolicy: Never
+      containers:
+      - name: cleanup
+        image: "{{ template "system_default_registry" . }}{{.Values.agentImage.repository}}:{{.Values.agentImage.tag}}"
+        imagePullPolicy: {{ .Values.global.imagePullPolicy }}
```
manno, line 20: This is actually empty by default, but we do the same in the [fleet-agent job](https://github.com/rancher/fleet/blob/master/charts/fleet-agent/templates/patch_default_serviceaccount.yaml).

internal/cmd/cli/cleanup/cleanup.go
```diff
@@ -0,0 +1,113 @@
...
+		}
+
+		// figure out the latest granted registration request per cluster
+		ts := latestGranted[clusterKey]
+		if cr.Status.Granted && ts.Before(&cr.CreationTimestamp) {
+			latestGranted[clusterKey] = cr.CreationTimestamp
+		}
+	}
+
+	// sleep at max 10 seconds to not overload the API server
```
weyfonk, line 67: Where does that 10 come from, apart form being the default max value?
If we're talking about not overloading the server, shouldn't the _min_ sleep duration be the determining factor here?

internal/cmd/cli/cleanup/cleanup.go
```diff
@@ -0,0 +1,113 @@
...
+		for _, cr := range crList.Items {
+			time.Sleep(b.Duration())
+
+			clusterKey := types.NamespacedName{Namespace: cr.Namespace, Name: cr.Status.ClusterName}
+			latest := latestGranted[clusterKey]
+			if cr.Status.Granted && cr.CreationTimestamp.Before(&latest) {
+				if err := clusterRegistration.Delete(cr.Namespace, cr.Name, nil); err != nil && !apierrors.IsNotFound(err) {
+					logrus.Errorf("Failed to delete clusterregistration %s/%s: %v", cr.Namespace, cr.Name, err)
+				}
+				// also try to delete orphans
```
weyfonk, line 88: I feel that this is a bit clearer:
```suggestion
				// also try to delete orphan child resources
```
Btw, given that this is triggered as a post-install/upgrade script, wouldn't the Fleet controller being running at this point, making this additional check unnecessary?
Or is this simply a way of enforcing orphan resources cleanup even in case this ends up being called in other contexts, being a new command in the Fleet CLI?

internal/cmd/cli/cleanup/cleanup.go
```diff
@@ -0,0 +1,113 @@
...
+				_ = roleBinding.Delete(cr.Namespace, cr.Name, nil)
+
+				filter := labels.Set(map[string]string{"fleet.cattle.io/cluster-registration": cr.Name, "fleet.cattle.io/cluster-registration-namespace": cr.Namespace})
+				saList, _ := serviceAccount.List("", metav1.ListOptions{LabelSelector: filter.AsSelector().String()})
+				for _, sa := range saList.Items {
+					_ = serviceAccount.Delete(sa.Namespace, sa.Name, nil)
+				}
+
+				owner := labels.Set(map[string]string{"objectset.rio.cattle.io/owner-name": cr.Name, "objectset.rio.cattle.io/owner-namespace": cr.Namespace}).AsSelector().String()
+				rbList, _ := roleBinding.List("", metav1.ListOptions{LabelSelector: owner})
```
weyfonk, line 99: nit, for style consistency with the above block:
```suggestion
				owner := labels.Set(map[string]string{"objectset.rio.cattle.io/owner-name": cr.Name, "objectset.rio.cattle.io/owner-namespace": cr.Namespace})
				rbList, _ := roleBinding.List("", metav1.ListOptions{LabelSelector: owner.AsSelector().String()})
```

internal/cmd/cli/cleanup/cleanup.go
```diff
@@ -0,0 +1,113 @@
...
+	// sleep at max 10 seconds to not overload the API server
+	b := &backoff.Backoff{
+		Min:    time.Duration(opts.Min) * time.Millisecond,
+		Max:    time.Duration(opts.Max) * time.Second,
+		Factor: opts.Factor,
+		Jitter: true,
+	}
+
+	go func() {
+		// it should be safe to delete up to the latest granted registration
```
weyfonk, line 76: This looks less ambiguous to me ;)
```suggestion
		// it should be safe to delete up to, but not including, the latest granted registration
```

weyfonk (dismissed):
Nice one! With a few nitpicks :)

internal/cmd/cli/cleanup/cleanup.go
manno: I wanted to make sure these are removed. I'm not sure if helm will wait for the controller pod to be ready before starting the hook. I would expect it not to wait, because normally it needs `--wait` to wait for anything.

Issue #1496, opened 2023-04-28
Update dev docs for testing


manno:
Which tests do we have, what are they good for, which setup is needed to run them?
Which tests don't run for PR, which ones make sense to run locally?

* e2e tests that require secret tests
* e2e testenv infra setup
* fleet integration tests

* [x] create table which workflow runs which tag
* [x] https://github.com/rancher/fleet/pull/1764
* [x] https://github.com/rancher/fleet/pull/1765

manno:
Flaky Tests

**bundlenamespacemapping**
https://github.com/rancher/fleet/actions/runs/5739312968/job/15554640659

**Gitrepo**

 • [FAILED] [601.841 seconds]
Monitoring Git repos via HTTP for change when updating a git repository monitored via webhook [It] updates the deployment [infra-setup]
/home/runner/work/fleet/fleet/e2e/single-cluster/gitrepo_test.go:174

  Timeline >>
  STEP: checking the pod exists @ 07/17/23 12:50:11.717
  STEP: updating the git repository @ 07/17/23 12:50:11.856
  STEP: checking for the updated commit hash in gitrepo @ 07/17/23 12:50:12.007
  [FAILED] in [It] - /home/runner/work/fleet/fleet/e2e/single-cluster/gitrepo_test.go:192 @ 07/17/23 13:00:12.03
  << Timeline

  [FAILED] Timed out after 600.022s.
  Expected
      <string>: apiVersion: fleet.cattle.io/v1alpha1
      kind: GitRepo
      metadata:
        annotations:
          kubectl.kubernetes.io/last-applied-configuration: |
            {"apiVersion":"fleet.cattle.io/v1alpha1","kind":"GitRepo","metadata":{"annotations":{},"name":"gitrepo-test","namespace":"fleet-local"},"spec":{"branch":"master","paths":["examples"],"pollingInterval":"24h","repo":"***git-service.fleet-local.svc.cluster.local:8080/webhook-test"}}
        creationTimestamp: "2023-07-17T12:50:11Z"
        generation: 1
        name: gitrepo-test
        namespace: fleet-local
        resourceVersion: "2435"
        uid: ee97a741-ede2-4079-900a-4d8554033284
      spec:
        branch: master
        paths:
        - examples
        pollingInterval: 24h
        repo: ***git-service.fleet-local.svc.cluster.local:8080/webhook-test
      status:
        commit: 28edcf7420a8a445f06e9b468293f24ec76e6ba0
        conditions:
[...]

PR #1704, opened 2023-08-09
E2E: Increase testenv reliability

p-se:
Relates to #1496

Improves the reliability of setting up `e2e/testenv/infra` by introducing a timeout of (by default) 5 minutes on three positions. One of which has not been repeatedly tried, the others have received an increased timeout. All three can be controlled by a newly introduced environment variable `CI_OCI_TIMEOUT_DURATION` which is parsed by `time.ParseDuration` and is set to `5m` if no value has been provided.

### Trade-off

Possibly longer waiting times.

### Potential improvement

Tests should be less flaky.

### DIFF
--- a/.github/workflows/e2e-ci.yml
+++ b/.github/workflows/e2e-ci.yml
@@ -56,16 +56,16 @@ jobs:
 
           go build -o "bin/fleet-linux-$GOARCH" ./cmd/fleetcli
           go build -o "bin/fleetagent-linux-$GOARCH" ./cmd/fleetagent
-          # cd e2e/testenv/infra
-          # go build -o . ./...
+          cd e2e/testenv/infra
+          go build -o . ./...
       -
         name: Build Docker Images
         run: |
           docker build -f package/Dockerfile -t rancher/fleet:dev --build-arg="ARCH=$GOARCH" .
           docker build -f package/Dockerfile.agent -t rancher/fleet-agent:dev --build-arg="ARCH=$GOARCH" .
-          # cd e2e/assets/gitrepo
-          # # Buildkit needed here for proper here-document support
-          # DOCKER_BUILDKIT=1 docker build -f Dockerfile.gitserver -t nginx-git:test --build-arg="passwd=$(openssl passwd foo)" .
+          cd e2e/assets/gitrepo
+          # Buildkit needed here for proper here-document support
+          DOCKER_BUILDKIT=1 docker build -f Dockerfile.gitserver -t nginx-git:test --build-arg="passwd=$(openssl passwd foo)" .
       -
         name: Provision k3d Cluster
         uses: AbsaOSS/k3d-action@v2
@@ -81,7 +81,7 @@ jobs:
         name: Import Images Into k3d
         run: |
           k3d image import rancher/fleet:dev rancher/fleet-agent:dev
-          #k3d image import nginx-git:test nginx-git:test
+          k3d image import nginx-git:test nginx-git:test
       -
         name: Set Up Tmate Debug Session
         if: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.enable_tmate == 'true' }}
@@ -93,13 +93,13 @@ jobs:
         name: Deploy Fleet
         run: |
           ./.github/scripts/deploy-fleet.sh
-      # -
-      #   name: Create Zot certificates for OCI tests
-      #   env:
-      #     FLEET_E2E_NS: fleet-local
-      #   run: |
-      #     # Generate cert and key for TLS
-      #     ./.github/scripts/create-zot-certs.sh "FleetCI-RootCA"
+      -
+        name: Create Zot certificates for OCI tests
+        env:
+          FLEET_E2E_NS: fleet-local
+        run: |
+          # Generate cert and key for TLS
+          ./.github/scripts/create-zot-certs.sh "FleetCI-RootCA"
       -
         name: E2E Tests
         env:
@@ -111,12 +111,12 @@ jobs:
           CI_OCI_PASSWORD: "foo"
           HELM_PATH: /usr/local/bin/helm
         run: |
-          # export CI_OCI_CERTS_DIR="$(git rev-parse --show-toplevel)/FleetCI-RootCA"
-          # e2e/testenv/infra/infra setup
+          export CI_OCI_CERTS_DIR="$(git rev-parse --show-toplevel)/FleetCI-RootCA"
+          e2e/testenv/infra/infra setup
 
-          ginkgo --label-filter="!infra-setup" e2e/single-cluster e2e/keep-resources
+          ginkgo e2e/single-cluster e2e/keep-resources
 
-          # e2e/testenv/infra/infra teardown
+          e2e/testenv/infra/infra teardown
       -
         name: Acceptance Tests for Examples
         if: >
--- a/.github/workflows/e2e-ci.yml
+++ b/.github/workflows/e2e-ci.yml
@@ -21,6 +21,7 @@ jobs:
   e2e-fleet-test:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         k3s_version:
           # k3d version list k3s | sed 's/+/-/' | sort -h
--- a/dev/README.md
+++ b/dev/README.md
@@ -1,9 +1,12 @@
+# Dev Scripts
+
 These scripts are used for running tests locally in k3d. Don't use these on
 production systems.
 
 ## Configuration
 
-You can set these manually or put them in an `.envrc`:
+You can set these manually, but it is advised to put them in an `.envrc` and
+source them before running any of the scripts.
 
     # use fleet-default for fleet in Rancher, fleet-local for standalone
     export FLEET_E2E_NS=fleet-local
@@ -23,7 +26,9 @@ You can set these manually or put them in an `.envrc`:
     # for running tests on darwin/arm64
     #export GOARCH=arm64
 
-    # needed for gitrepo tests
+    # needed for gitrepo tests, which are currently disabled but part of the
+    # single-cluster tests
+
     #export FORCE_GIT_SERVER_BUILD="yes" # set to an empty value to skip rebuilds
     #export GIT_REPO_USER="git"
     #export GIT_REPO_URL="git@github.com:yourprivate/repo.git"
@@ -33,7 +38,9 @@ You can set these manually or put them in an `.envrc`:
     #export GIT_HTTP_USER="fleet-ci"
     #export GIT_HTTP_PASSWORD="foo"
 
-    # needed for OCI tests
+    # needed for OCI tests, which are currently disabled but part of the
+    # single-cluster tests
+
     #export CI_OCI_USERNAME="fleet-ci"
     #export CI_OCI_PASSWORD="foo"
     #export CI_OCI_CERTS_DIR="../../FleetCI-RootCA"
@@ -41,13 +48,17 @@ You can set these manually or put them in an `.envrc`:
 
 ## Running Tests on K3D
 
-This should set up k3d, and the fleet standalone images for single cluster tests
+These commands should set up k3d, and the fleet standalone images for single
+cluster tests.
 
-    export FLEET_E2E_NS=fleet-local FLEET_E2E_CLUSTER=k3d-upstream
+    export FLEET_E2E_NS=fleet-local
+    export FLEET_E2E_CLUSTER=k3d-upstream
     export FLEET_E2E_CLUSTER_DOWNSTREAM=k3d-upstream
+
     dev/setup-k3d
     dev/build-fleet
     dev/import-images-k3d
+
     # needed for gitrepo tests
     dev/import-images-tests-k3d
     dev/create-zot-certs 'FleetCI-RootCA' # for OCI tests
@@ -56,26 +67,29 @@ This should set up k3d, and the fleet standalone images for single cluster tests
     # This should be needed only once
     cd e2e/testenv/infra
     go build -o . ./...
-
-    ./infra setup
     cd -
+
+    ./e2e/testenv/infra setup
 
     ginkgo e2e/single-cluster
 
-    ./infra teardown # optional
+    ./e2e/testenv/infra teardown # optional
 
 Optional flags for reporting on long-running tests: `--poll-progress-after=10s --poll-progress-interval=10s`.
 
-For multi-cluster tests we need to configure two clusters. You also need to
-make the upstream clusters API accessible to the downstream cluster. The
-default `url` in [dev/setup-fleet-downstream] should work with most systems.
+For multi-cluster tests we need to configure two clusters. You also need to make
+the upstream clusters API accessible to the downstream cluster. The default
+`url` in [dev/setup-fleet-downstream] should work with most systems.
 
-    export FLEET_E2E_NS=fleet-local FLEET_E2E_CLUSTER=k3d-upstream
+    export FLEET_E2E_NS=fleet-local
+    export FLEET_E2E_CLUSTER=k3d-upstream
     export FLEET_E2E_CLUSTER_DOWNSTREAM=k3d-downstream
+
     dev/setup-k3ds
     dev/build-fleet
     dev/import-images-k3d
     dev/setup-fleet-multi-cluster
+
     ginkgo e2e/multi-cluster
 
 To test changes incrementally, rebuild just one binary, update the image in k3d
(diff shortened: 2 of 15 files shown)


.github/workflows/e2e-ci.yml
```diff
@@ -21,6 +21,7 @@ jobs:
   e2e-fleet-test:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
```
raulcabello, line 24: I would leave `fail-fast` to true, so we get faster feedback if something fails

e2e/single-cluster/gitrepo_test.go
```diff
@@ -51,13 +53,19 @@ var _ = Describe("Monitoring Git repos via HTTP for change", Label("infra-setup"
...
+		gitrepoName = testenv.RandomFilename("gitjob-test", r)
 	})
 
 	AfterEach(func() {
-		os.RemoveAll(tmpdir)
-		_, _ = k.Delete("gitrepo", "gitrepo-test")
+		report := CurrentSpecReport()
+		if report.Failure.IsZero() {
+			os.RemoveAll(tmpDir)
+			_, _ = k.Delete("gitrepo", gitrepoName)
```
raulcabello, line 66: shouldn't we try to delete the gitrepo even if there was a failure? We need to make sure there is no resources from previous tests

dev/setup-fleet
```diff
@@ -14,11 +14,18 @@ fi
 host=$( docker inspect -f '{{range.NetworkSettings.Networks}}{{.IPAddress}}{{end}}' k3d-$cluster_name-server-0 )
 ca=$( kubectl config view --flatten -o jsonpath='{.clusters[?(@.name == "k3d-$cluster_name")].cluster.certificate-authority-data}' | base64 -d )
 server="https://$host:6443"
-helm -n cattle-fleet-system upgrade --install --create-namespace --wait fleet-crd charts/fleet-crd
-helm -n cattle-fleet-system upgrade --install --create-namespace --wait --reset-values \
+while !  helm -n cattle-fleet-system upgrade --install --atomic --create-namespace --wait fleet-crd charts/fleet-crd
```
aruiz14, line 17: Wouldn't this cause to retry indefinitely regardless of the reason of the failure? If the purpose of the loop is to give it enough time for it to deploy, I think we should instead use `--timeout` with a sensible value (defaults to 5 minutes)

dev/setup-fleet
```diff
@@ -14,11 +14,18 @@ fi
...
+do
+  sleep 1
+done
+
+while ! helm -n cattle-fleet-system upgrade --install --atomic --create-namespace --wait --reset-values \
```
aruiz14, line 22: ditto about using `--timeout` instead

e2e/assets/gitrepo/post-receive.sh
```diff
@@ -11,4 +11,12 @@ after=$(cat $ref)
...
 github_header="X-GitHub-Event: push"
 
-curl gitjob.cattle-fleet-system.svc.cluster.local -H "$github_header" -d "{\"ref\": \"$ref\", \"after\": \"$after\", \"repository\": {\"html_url\": \"{{.RepoURL}}\"}}"
+while ! curl --fail-with-body gitjob.cattle-fleet-system.svc.cluster.local \
+    -H "$github_header" \
+    -d "{\"ref\": \"$ref\", \"after\": \"$after\", \"repository\": {\"html_url\": \"{{.RepoURL}}\"}}"
+do
+    echo "Sending webhook failed, retrying..."
+    sleep 1
+done
```
aruiz14, line 20: Depending on the `curl` version available, we could use some of the retry flags supported:
 - https://curl.se/docs/manpage.html#--retry
 - https://everything.curl.dev/usingcurl/downloads/retry

e2e/testenv/infra/cmd/setup.go
```diff
@@ -67,39 +76,52 @@ Parallelism is used when possible to save time.`,
...
+			if time.Now().After(startTime.Add(timeoutDuration)) {
+				fail(fmt.Errorf("timed out waiting for external IP"))
+			}
+
 			externalIP, err = k.Get("service", "zot-service", "-o", "jsonpath={.status.loadBalancer.ingress[0].ip}")
 			if err != nil {
 				fail(fmt.Errorf("get external Zot service IP: %v", err))
 			}
+			time.Sleep(200 * time.Millisecond)
 		}
```
aruiz14, line 91: Can we extract this to a function? Also, I think using `context` for timeout will make it cleaner as well.

raulcabello (commented):
Thanks for improving the tests! I haven't run them locally, but I like the approach! I think the timeouts will help with the problems we were facing. It would be a good idea to run the tests multiple times in github actions to make sure they are not flaky. 
I left some comments

e2e/testenv/infra/cmd/setup.go
```diff
@@ -13,13 +12,14 @@ import (
...
-	"github.com/spf13/cobra"
-	"helm.sh/helm/v3/pkg/registry"
-
 	"github.com/rancher/fleet/e2e/testenv"
 	"github.com/rancher/fleet/e2e/testenv/kubectl"
+	"github.com/spf13/cobra"
+	"helm.sh/helm/v3/pkg/registry"
 )
 
+var timeoutDuration = 30 * time.Minute // default timeout duration
```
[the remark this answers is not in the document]
p-se: I thought so too, but was curious about other opinions. That being said, I think 5 minutes could make a test fail if it could still recover. Can we agree on 10 minutes?

.github/workflows/e2e-ci.yml
p-se: Oh, fail fast in a Github Action for a test that has a matrix means that it will abort the other test running at the time the first failed. If we'd like to just rerun the failed test, it'd be good not to abort the other run. In the tests I've run, rarely two tests failed, but of course, if there's an actual issue that's not related to timings, both would fail. If you still think it'd be better to have fail-fast, let me know and I'll remove the change that disables it.

e2e/single-cluster/gitrepo_test.go
p-se: That's a good point. I've used this code to keep the resources in case of failures so that I could inspect them. The adaptation of the test makes sure that they names will most likely not be the same, so there shouldn't be any issue with leaving those around for collecting more information about the issue using the `dump-failed-k3ds.sh` script. That being said, I don't have a strong opinion on this one.

dev/setup-fleet
p-se: The purpose is to retry on transient issues, but if it's a permanent one, you're right. Using `--timeout` might not fix that issue, though. So I might need to add a bit more of complexity to it to make it fail, perhaps after three attempts. What do you think?

e2e/assets/gitrepo/post-receive.sh
p-se: I like that!

e2e/testenv/infra/cmd/setup.go
p-se: I'll take care of it.

p-se:
> Thanks for improving the tests! I haven't run them locally, but I like the approach! I think the timeouts will help with the problems we were facing. It would be a good idea to run the tests multiple times in github actions to make sure they are not flaky.

Thanks for the comments! I actually have run them several times on Github Actions from my fork. The results can be seen [here](https://github.com/p-se/fleet/actions/workflows/e2e-ci.yml?query=event%3Aworkflow_dispatch). Without pushing the branch to `rancher/fleet` I cannot trigger them manually, except if I change the code of the PR and then those tests will run that are configured to run for PRs.

e2e/testenv/infra/cmd/setup.go
```diff
@@ -67,39 +76,52 @@ Parallelism is used when possible to save time.`,
...
+		for externalIP == "" {
+			if time.Now().After(startTime.Add(timeoutDuration)) {
+				fail(fmt.Errorf("timed out waiting for external IP"))
+			}
+
 			externalIP, err = k.Get("service", "zot-service", "-o", "jsonpath={.status.loadBalancer.ingress[0].ip}")
 			if err != nil {
 				fail(fmt.Errorf("get external Zot service IP: %v", err))
 			}
+			time.Sleep(200 * time.Millisecond)
```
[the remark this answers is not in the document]
p-se: By extracting this into a function as suggested by @aruiz14, this will only be a single occurrence.

.github/workflows/e2e-ci.yml
raulcabello: ok, that's fine then

dev/setup-fleet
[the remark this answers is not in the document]
raulcabello: would increment the timeout help as suggested by @aruiz14 https://github.com/rancher/fleet/pull/1704#discussion_r1302796344 ?

e2e/single-cluster/gitrepo_test.go
raulcabello: perhaps resources deployed by a previous test case might affect other test cases. I'd prefer to have a clean state when starting a new test case

dev/setup-fleet
manno: I'd consider that to be a part of the test setup. It shouldn't fail. If it fails we can just ctrl-c it locally. It's just a dev script.

e2e/single-cluster/gitrepo_test.go
manno: I prefer the fast cleanup we had before:

* Clean up is not immediate, we can't trust it without extensive waiting.
* I don't want to wait for a proper clean up and fail the test if clean up fails.  
* Tests run in parallel and should not conflict with each other anyhow.
* I would focus and comment the clean up, if I have to inspect a failure.

e2e/testenv/infra/cmd/setup.go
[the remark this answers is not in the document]
manno: Is this like the interval in between checks for the ingress? We could just use 1s, waiting for ingress is usually really slow?

internal/cmd/controller/controllers/bundle/controller.go
```diff
@@ -342,6 +342,7 @@ func (h *handler) updateStatusAndTargets(status *fleet.BundleStatus, allTargets
 	}
 
 	for _, partition := range partitions {
+		partition := partition
```
manno, line 345: Is there some issue here, while partitions is not a pointer array, the next line `partition.Targets` is, but I can't spot a problem on first glance.

internal/cmd/cli/cleanup/cleanup.go
```diff
@@ -54,6 +54,7 @@ func ClusterRegistrations(ctx context.Context, client Getter, opts Options) erro
 	// figure out the latest granted registration request per cluster
 	latestGranted := map[types.NamespacedName]metav1.Time{}
 	for _, cr := range crList.Items {
+		cr := cr
```
manno, line 57: Is this really necessary? I thought Items is `Items []ClusterRegistration`, so not an array of pointers.

internal/cmd/controller/controllers/bundle/controller.go
p-se: To be honest, I do don't know for sure. But apparently due to the move to Go 1.21, or perhaps because I'm running a more recent version of `golangci-lint`, it became one for `golangci-lint`:

```go
internal/cmd/controller/controllers/bundle/controller.go:358:40: G601: Implicit memory aliasing in for loop. (gosec)
                        updateTarget(currentTarget, status, &partition.Status)
                                                            ^
internal/cmd/controller/controllers/bundle/controller.go:361:37: G601: Implicit memory aliasing in for loop. (gosec)
                if target.UpdateStatusUnavailable(&partition.Status, partition.Targets) {
```

Do you think `golangci-lint` is wrong about it?

internal/cmd/cli/cleanup/cleanup.go
p-se:
```go
internal/cmd/cli/cleanup/cleanup.go:63:37: G601: Implicit memory aliasing in for loop. (gosec)
                if cr.Status.Granted && ts.Before(&cr.CreationTimestamp) {
```

dev/setup-fleet
p-se: It appears I've confused it with `deploy-fleet.sh`, which is used in the Github Action. But `dev/setup-fleet` is not. I'll remove it and leave it out for the time being.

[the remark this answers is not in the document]
p-se: It appears I've confused it with `deploy-fleet.sh`, which is used in the Github Action. But `dev/setup-fleet` is not. I'll remove it and leave it out for the time being.

e2e/single-cluster/gitrepo_test.go
p-se: Alright, I'll undo that.

p-se:
> I've switched to the `fix-testenv` branch for testing Github Actions and there are a few transient issues again.

It was an issue with one of my updates. Seems to be more stable now.

internal/cmd/controller/controllers/bundle/controller.go
manno: I researched it and yes, we should hold off on golangci-lint 1.54.2: `https://github.com/golangci/golangci-lint/issues/4037`
It's probably another false positive from Gosec.

Unless someone can show me invalid usage of the pointer, like adding it to an array `allthesameptrs = append(allthesameptrs, &partition)` or using it in a closure. In that case we would have a serious bug in our targeting and would probably need to backport fixes.


We can also discuss if we always want to copy pointers in a for loop to be defensive about it, but 🤷

manno (dismissed):
See https://github.com/rancher/fleet/pull/1704#discussion_r1305580561

internal/cmd/controller/controllers/bundle/controller.go
p-se: I'm happy to discuss, but it does not have to be in this PR. I've not found an issue that would require a fix and therefore removed the two changed from the PR.

dev/README.md
```diff
@@ -56,26 +67,29 @@ This should set up k3d, and the fleet standalone images for single cluster tests
...
-
-    ./infra setup
     cd -
 
+    ./e2e/testenv/infra setup
+
     ginkgo e2e/single-cluster
 
-    ./infra teardown # optional
+    ./e2e/testnenv/infra teardown # optional
```
weyfonk, line 76:
```suggestion
    ./e2e/testenv/infra teardown # optional
```

e2e/testenv/infra/cmd/setup.go
```diff
@@ -13,13 +13,38 @@ import (
...
-
 	"github.com/rancher/fleet/e2e/testenv"
 	"github.com/rancher/fleet/e2e/testenv/kubectl"
+	"github.com/spf13/cobra"
+	"helm.sh/helm/v3/pkg/registry"
 )
 
+var timeoutDuration = 10 * time.Minute // default timeout duration
+
+func eventually(f func() (string, error)) string {
```
weyfonk, line 24: Liking this, wrapping anything that can fail on first attempts into this, as you did, makes more sense than the initial approach. Thanks for this!

dev/README.md
p-se: Fixed, thanks.

e2e/testenv/kubectl/kubectl.go
```diff
@@ -44,7 +44,7 @@ func (c Command) Workdir(dir string) Command {
 }
 
 func (c Command) Apply(args ...string) (string, error) {
-	return c.Run(append([]string{"apply"}, args...)...)
+	return c.Run(append([]string{"apply", "--wait"}, args...)...)
```
manno, line 47: > If true, wait for resources to be gone before returning. This waits for finalizers.

Making this the default for all `Apply` makes sense to me.

e2e/testenv/env.go
```diff
@@ -57,8 +57,10 @@ func (e *Env) getShellEnv() {
 // NewNamespaceName returns a name for a namespace that is unique to the test
 // run. e.g. as a targetNamespace for workloads
 func NewNamespaceName(name string) string {
-	rand.Seed(time.Now().UnixNano())
```
manno, line 60: What do you think about passing in the randomness source, like you do with `RandomFilename`, and using `GinkgoRandomSeed()` when calling from ginkgo tests?

e2e/single-cluster/gitrepo_test.go
```diff
@@ -27,13 +27,15 @@ const (
...
-		tmpdir           string
+		tmpDir           string
 		clonedir         string
 		k                kubectl.Command
 		gh               *githelper.Git
 		clone            *git.Repository
 		repoName         string
 		inClusterRepoURL string
+		gitrepoName      string
+		r                = rand.New(rand.NewSource(1))
```
manno, line 38: We could use `GinkgoRandomSeed()`
```suggestion
		r                = rand.New(rand.NewSource(GinkgoRandomSeed()))
```

e2e/testenv/template.go
```diff
@@ -39,7 +41,7 @@ func ApplyTemplate(k kubectl.Command, asset string, data interface{}) error {
 	tmpdir, _ := os.MkdirTemp("", "fleet-")
 	defer os.RemoveAll(tmpdir)
 
-	output := path.Join(tmpdir, RandomFilename(asset))
+	output := path.Join(tmpdir, RandomFilename(asset, rand.New(rand.NewSource(time.Now().UnixNano())))) // nolint:gosec // test code
```
manno, line 44: We probably don't care about this filename and don't need to control the seed, as it's just an intenral temp file.

PR #1707, opened 2023-08-10
Re-enable testenv/infra

p-se:
Fix #XXX


p-se:
Tests can be run in a fork instead.

PR #1712, opened 2023-08-12
Increase speed of clusterregistration clean up hook

manno:
Improperly placed wait statement. Additionally, the client is already rate limited. Previous PR takes 90minutes for ~500 registrations?

Refers to #1689


weyfonk (approved):
LGTM

Issue #1643, opened 2023-07-09
Traefik - Helm >= 3.9.0 is required


JamesTurland:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Have been using fleet to deploy Traefik for almost a year. Have just rebuilt my cluster and connected my Git account. When I try to deploy Traefik shows the following error:

ErrApplied(1) [Cluster fleet-local/local: execution error at (traefik/templates/deployment.yaml:3:8): ERROR: Helm >= 3.9.0 is required]

### Expected Behavior

Should deploy without error.

`helm version
version.BuildInfo{Version:"v3.11.3-rancher1", GitCommit:"8853d6f9875eb3138d10f4af7a4a67130039e28b", GitTreeState:"clean", GoVersion:"go1.19.4"}`

### Steps To Reproduce

Try to deploy Traefik using Fleet.

### Environment

```markdown
- Architecture: x86
- Fleet Version: 2.7.5
- Cluster:
  - Provider: K3S
  - Options: 8 nodes, Longhorn, Metallb
  - Kubernetes Version: 1.25.9
```

### Logs

```markdown
ErrApplied(1) [Cluster fleet-local/local: execution error at (traefik/templates/deployment.yaml:3:8): ERROR: Helm >= 3.9.0 is required]
```

weiyentan:
am getting the same issue

manno:
The [traefik chart has an explicit check](https://github.com/traefik/traefik-helm-chart/blob/2dd3b1848ff88f053a7cb01fdc997ab47bc3afd3/traefik/templates/deployment.yaml#L2) for the version:

```
{{- if (semverCompare "<v3.9.0" (.Capabilities.HelmVersion.Version | default "v3.0.0")) -}}
    {{- fail "ERROR: Helm >= 3.9.0 is required" -}}
{{- end -}}
```

Fleet supports capabilities now, but doesn't set a helm version: https://github.com/rancher/fleet/blob/master/internal/helmdeployer/capabilities.go#L56-L63

To fix this bug, it's probably sufficient to add this line: https://github.com/helm/helm/blob/main/pkg/chartutil/capabilities.go#L48

manno:
## Additional QA
 
### Problem

Fleet didn't provide all ["Capabilities"](https://helm.sh/docs/chart_template_guide/builtin_objects/) to helm charts it deployed. For the helm version it provided an old default.

### Solution

Update how we use the helm SDK.
 
[...]

PR #1743, opened 2023-08-24
Add helm version to capabilities

manno:
Some charts, like cert-manager check the helm version, since fleet uses the SDK we need to set it.

Refers to https://github.com/rancher/fleet/issues/1643


aruiz14 (approved):
LGTM!

PR #1745, opened 2023-08-24
Bump all modules

manno:
Result of running `go get -u ./...`

* cannot bump go-containerregistry to 0.16.1
* ineffective bumps, modules haver their versions replaced in `replace`
   block
* bumps gitjob, too


manno:
> Controller Runtime was not bumped, was it intentional? #1698

That needs a newer k8s version. Seems like we only use it for the integration tests.

manno:
> Could you do the same for `pkg/apis`? It seems that the k8s api module was not bumped to 0.27.4: https://github.com/rancher/fleet/blob/master/pkg/apis/go.mod#L7

Sure. We are just doing it for the cosmetics, right? Or am I missing something?

PR #1755, opened 2023-09-04
Allow configuring EnableDNS property in Helm through fleet.yaml

### LINKED ISSUE #1332: Expose new EnableDNS property in Helm through fleet.yaml
The [Helm v3.11.1 release](https://github.com/helm/helm/releases/tag/v3.11.1) introduced a new flag/property. When Fleet updates to this version of Helm is should expose this property (which defaults to false) so that it can optionally be enabled. Prior to 3.11.1 the property did not exist and was the equivalent of setting it to true.

aruiz14:
Fix #1332

This PR allows to configure the value of Helm's `--enable-dns` feature, which is set to `true` by Fleet for compatibility reasons. If an user intends to omit this flag, `disableDNS: true` must be set in the corresponding `fleet.yaml` file.

## Test

To test this pull request, you add the following top-level property to a `fleet.yaml`:

```yaml
disableDNS: true
```
Then it must be applied in a cluster running Fleet, where the updated version of the CRDs are installed and an updated version of the `fleet-agent` image is used.
Then, after a GitRepo pointing to the updated `fleet.yaml` file is synced, the resulting `bundle` CRD object must preserve `disableDNS: true`.

### Trade-off

Please note that the naming was inverted (`DisableDNS` vs. `EnableDNS`) to reflect the difference in the default behavior.


internal/helmdeployer/deployer.go
```diff
@@ -428,7 +428,7 @@ func (h *Helm) install(bundleID string, manifest *manifest.Manifest, chart *char
 
 	u := action.NewUpgrade(&cfg)
 	u.Adopt = true
-	u.EnableDNS = true
+	u.EnableDNS = !options.DisableDNS
```
weyfonk, line 431: Could we perhaps automate testing for this? Either through unit tests, checking that Helm is called with the right arguments, or by means of e2e tests maybe to validate actual deployments? How do you see it?

manno (changes requested):
I think this needs another line in https://github.com/rancher/fleet/blob/master/internal/cmd/controller/options/calculate.go#L76 to support usage in target customizations (see list of [supported target customizations](https://fleet.rancher.io/gitrepo-targets#supported-customizations)). 


I'm not sure about the testing, there seems to be no existing unit/integration test that could be extended, `Atomic`, `WaitForJob` or `TakeOwnership` and the others don't seem to have testing to make sure they pass from manifest to helmdeployer. E2E seems to heavy for all of those, especially with the DNS server in this issue...

aruiz14:
> I think this needs another line in [`master`/internal/cmd/controller/options/calculate.go#L76](https://github.com/rancher/fleet/blob/master/internal/cmd/controller/options/calculate.go?rgh-link-date=2023-09-05T08%3A22%3A52Z#L76) to support usage in target customizations (see list of [supported target customizations](https://fleet.rancher.io/gitrepo-targets#supported-customizations)).

Oh, I totally missed that part. Will take a look, thanks.

---
 
> I'm not sure about the testing, there seems to be no existing unit/integration test that could be extended, `Atomic`, `WaitForJob` or `TakeOwnership` and the others don't seem to have testing to make sure they pass from manifest to helmdeployer. E2E seems to heavy for all of those, especially with the DNS server in this issue...

I was thinking on creating a very simple chart in `fleet-test-data` that used `getHostByName()` in one of the templates, then make the test check that it indeed fails when `DisableDNS` is set to `true`. WDYT?

manno:
> Oh, I totally missed that part. Will take a look, thanks.

It's very easy to miss, sadly. Maybe you can add a comment to helmdeployer for the future.

> I was thinking on creating a very simple chart in `fleet-test-data` that used `getHostByName()` in one of the templates, then make the test check that it indeed fails when `DisableDNS` is set to `true`. WDYT?

Yes, that sounds like a good way to e2e test this. We'd be dependent on the DNS resolver, maybe test with `github.com`.

weyfonk (dismissed):
LGTM :)

manno (approved):

PR #1756, opened 2023-09-04
Testenv infra uses local fleet module

manno:
Instead of using a hardcoded version, use the local files. Avoids confusion when changing both at once.


e2e/testenv/infra/go.mod
```diff
@@ -1,11 +1,15 @@
 module github.com/rancher/fleet/e2e/testenv/infra
 
-go 1.20
+go 1.21
+
+toolchain go1.21.0
```
weyfonk, line 5: What does this line bring in this case? Do we want to [prevent this module from loading](https://go.dev/doc/toolchain) if, when Fleet evolves past `go 1.21`, this module's toolchain has not been updated? In other words, is it a way of keeping both modules in sync, Go toolchain-wise?

manno: `go mod tidy` added that automatically, I think.

According to the docs you linked it should be fine (?):

> In the standard configuration, the go command uses its own bundled toolchain when that toolchain is at least as new as the go or toolchain lines in the main module or workspace. For example, when using the go command bundled with Go 1.21.3 in a main module that says go 1.21.0, the go command uses Go 1.21.3.

And if your go is older than toolchain:

>  For example, when using the go command bundled with Go 1.21.3 in a main module that says go 1.21.9, the go command finds and runs Go 1.21.9 instead. It first looks in the PATH for a program named go1.21.9 and otherwise downloads and caches a copy of the Go 1.21.9 toolchain. This automatic toolchain switching can be disabled, but in that case, for more precise forwards compatibility, the go command will refuse to run in a main module or workspace in which the go line requires a newer version of Go.

PR #1759, opened 2023-09-05
Improve release scripts

thardeck:
to handle different release.yamls. Since everyone seems to use no indentation for yaml arrays in the `release.yaml` I am using sed to not change all the other entries all the time.

The updated script is used for this pr: https://github.com/rancher/charts/pull/2937

### DIFF
--- a/.github/scripts/release-against-charts.sh
+++ b/.github/scripts/release-against-charts.sh
@@ -35,20 +35,16 @@ else
     exit 1
 fi
 
-if [ "${REPLACE}" == "true" ]; then
-    for i in fleet fleet-crd fleet-agent; do
+for i in fleet fleet-crd fleet-agent; do
+    if [ "${REPLACE}" == "true" ]; then
         yq --inplace "del( .${i}.[] | select(. == \"${PREV_CHART_VERSION}+up${PREV_FLEET_VERSION}\") )" release.yaml
-        yq --inplace ".${i} += \"${NEW_CHART_VERSION}+up${NEW_FLEET_VERSION}\"" release.yaml
-    done
-else
-    for i in fleet fleet-crd fleet-agent; do
-        yq --inplace ".${i} += \"${NEW_CHART_VERSION}+up${NEW_FLEET_VERSION}\"" release.yaml
-    done
-fi
+    fi
+    yq --inplace ".${i} += \"${NEW_CHART_VERSION}+up${NEW_FLEET_VERSION}\"" release.yaml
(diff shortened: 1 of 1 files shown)


.github/scripts/release-against-charts.sh
```diff
@@ -27,34 +27,36 @@ if [ ! -f bin/charts-build-scripts ]; then
...
-    sed -i -z -e  "s/[[:space:]]*\nfleet:/\nfleet:/g" release.yaml
-    fi
+    for i in fleet fleet-crd fleet-agent; do
+        yq --inplace ".${i} += \"${NEW_CHART_VERSION}+up${NEW_FLEET_VERSION}\"" release.yaml
+    done
 fi
 
+# Follow the currently applied no identation style for arrays in release.yaml,
+# which is not possible with yq v4 atm: https://github.com/mikefarah/yq/issues/1309
+sed -i "s/^  -/-/g" release.yaml
```
weyfonk, line 51: nit: I suppose the `g` may not be necessary in this `sed` expression since it deals with line beginnings.

.github/scripts/release-against-charts.sh
```diff
@@ -27,34 +27,36 @@ if [ ! -f bin/charts-build-scripts ]; then
...
+    find ./packages/fleet/ -type f -exec sed -i -e "s/version: ${PREV_CHART_VERSION}/version: ${NEW_CHART_VERSION}/g" {} \;
+else
+    echo "Previous Fleet version references do not exist in ./packages/fleet/ so replacing it with the new version is not possible. Exiting..."
+    exit 1
+fi
 
 if [ "${REPLACE}" == "true" ]; then
-    sed -i -e "s/${PREV_CHART_VERSION}+up${PREV_FLEET_VERSION}/${NEW_CHART_VERSION}+up${NEW_FLEET_VERSION}/g" release.yaml
+    for i in fleet fleet-crd fleet-agent; do
+        yq --inplace "del( .${i}.[] | select(. == \"${PREV_CHART_VERSION}+up${PREV_FLEET_VERSION}\") )" release.yaml
```
weyfonk, line 40: I wonder how entries with the previous chart and Fleet versions are still part of  `release.yaml` after this, but that would have been the case before these changes as well, as it already looked like those entries were deleted. Are they restored later in the workflow? :thinking:

weyfonk (dismissed):
LGTM :)

weyfonk (approved):

PR #1761, opened 2023-09-07
Improve release scripts

thardeck:


.github/workflows/release-against-rancher.yml
```diff
@@ -53,8 +54,7 @@ jobs:
           path: rancher
       - uses: actions/setup-go@v4
         with:
-          go-version-file: 'go.mod'
-          check-latest: true
+          go-version: ${{ env.SETUP_GO_VERSION }}
```
olblak, line 57: By curiousity
What's the motivation to explicitly specify the golang version here? 
Instead of reusing the one specified in the go.mod file? 

From my perspective, it's one additional dependency that we need to monitor

aruiz14 (approved):
Changes look good to me, but maybe you can get approval from someone else who is more familiar with these workflows.

weyfonk (approved):

PR #1762, opened 2023-09-07
Fixes for running multi cluster E2E tests locally

### LINKED ISSUE #1496: Update dev docs for testing

p-se:
Part of #1496

* Makes a change to be able to run the multi-clusters tests locally using the dev scripts and documents the necessary configuration.
* Makes a change to be able to run the multi-clusters tests locally with [nektos/act](https://github.com/nektos/act).

### DIFF
--- a/.github/scripts/label-downstream-cluster.sh
+++ b/.github/scripts/label-downstream-cluster.sh
@@ -2,7 +2,7 @@
 
 set -euxo pipefail
 
-ns=${FLEET_E2E_NS-fleet-default}
+ns=${FLEET_E2E_NS_DOWNSTREAM-fleet-default}
 
 { grep -q -m 1 -e "k3d-downstream"; kill $!; } < <(kubectl get clusters.fleet.cattle.io -n "$ns" -w)
 name=$(kubectl get clusters.fleet.cattle.io -o=jsonpath='{.items[0].metadata.name}' -n "$ns")
--- a/.github/workflows/e2e-multicluster-ci.yml
+++ b/.github/workflows/e2e-multicluster-ci.yml
@@ -116,14 +116,30 @@ jobs:
 
           token=$(kubectl get secret -n fleet-default second-token -o go-template='{{index .data "values" | base64decode}}' | yq eval .token -)
           ca=$(kubectl get secret -n cattle-fleet-system fleet-controller-bootstrap-token -o go-template='{{index .data "ca.crt" | base64decode}}')
+          apiServerIP=$(kubectl get node k3d-upstream-server-0 -o jsonpath='{.status.addresses[?(@.type=="InternalIP")].address}')
 
-          kubectl config use-context k3d-downstream
-          helm -n cattle-fleet-system upgrade --install --create-namespace --wait fleet-agent charts/fleet-agent \
-            --set-string labels.env=test \
-            --set apiServerCA="$ca" \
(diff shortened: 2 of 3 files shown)


.github/workflows/e2e-multicluster-ci.yml
```diff
@@ -116,14 +116,25 @@ jobs:
 
           token=$(kubectl get secret -n fleet-default second-token -o go-template='{{index .data "values" | base64decode}}' | yq eval .token -)
           ca=$(kubectl get secret -n cattle-fleet-system fleet-controller-bootstrap-token -o go-template='{{index .data "ca.crt" | base64decode}}')
+          apiServerURL=$(kubectl get node k3d-upstream-server-0 -o jsonpath='{.status.addresses[?(@.type=="InternalIP")].address}')
```
p-se, line 119: This is a change that was necessary for me as I, if I remember correctly, have extended the range of default networks available to docker, which results in a different network being used at the start. It seems to work well this way.

.github/scripts/label-downstream-cluster.sh
```diff
@@ -2,7 +2,7 @@
 
 set -euxo pipefail
 
-ns=${FLEET_E2E_NS-fleet-default}
+ns=${FLEET_E2E_NS_DOWNSTREAM-fleet-default}
```
p-se, line 5: I have had issues starting the test environment from the dev scripts _and_ running them with the same values of the those two environment variables:

* FLEET_E2E_NS
* FLEET_E2E_NS_DOWNSTREAM

With this change I've been able to start and run the tests with the same values of those env vars.

.github/workflows/e2e-multicluster-ci.yml
```diff
@@ -116,14 +116,25 @@ jobs:
...
+                --wait \
+              fleet-agent charts/fleet-agent \
+                --set-string labels.env=test \
+                --set apiServerCA="$ca" \
+                --set apiServerURL="https://$apiServerURL:6443" \
+                --set clusterNamespace="fleet-default" \
+                --set token="$token"
+
+          echo "waiting for downstream cluster to be registered..."
+          { grep -q -m 1 "k3d-downstream"; kill $!; } < <(kubectl get cluster -n fleet-default -w)
```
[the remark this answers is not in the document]
manno: Yes. `kubectl -w` exits too early? Maybe we need another check to see if fleet cluster resource is ready, too?

.github/workflows/e2e-multicluster-ci.yml
p-se: I wonder what the difficulties on Mac would be with this line, perhaps I can address them as well?

manno: Not sure, maybe it works. Act runs everything in containers? So the address is probably reachable.

p-se: Yes, it runs everything in containers.

manno (approved):

PR #1764, opened 2023-09-08
Add dev scripts to simplify using E2E tests

### LINKED ISSUE #1496: Update dev docs for testing

p-se:
Part of #1496

Adding scripts to not have to copy and paste instructions from `dev/README.md` with sane defaults to spin up the environment and run the tests out-of-the-box. Contains appropriate adaptations of the `dev/README.md`.

### DIFF
--- a/dev/README.md
+++ b/dev/README.md
@@ -3,13 +3,79 @@
 These scripts are used for running tests locally in k3d. Don't use these on
 production systems.
 
+## Requirements
+
+- docker
+- git
+- go
+- helm
+- jq
+- k3d
+- kubectl
+- ...
+
+## Running Tests on K3D
+
+These commands should set up k3d and the fleet standalone images for single
+cluster tests and run those.
+
+    source dev/setup-single-cluster
+    ginkgo e2e/single-cluster
+
+Optional flags to `ginkgo` for reporting on long-running tests:
+`--poll-progress-after=10s --poll-progress-interval=10s`.
+
+For multi-cluster tests we need to configure two clusters. You also need to make
+the upstream clusters API accessible to the downstream cluster. The default URL
+in `dev/setup-fleet-downstream` should work with most systems.
+
+    source dev/setup-multi-cluster
+    ginkgo e2e/multi-cluster
+
+### Testing changes incrementally
+
+To test changes incrementally, rebuild just one binary, update the image in k3d
+and restart the controller. Make sure you have sourced the right configuration
+for the current setup.
+
+    dev/update-agent-k3d
+    dev/update-controller-k3d
+
 ## Configuration
 
-You can set these manually, but it is advised to put them in an `.envrc` and
-source them before running any of the scripts.
+### Running scripts manually
+
+You can set these environment variables for configuration manually, but it is
+advised to put them in `.envrc` and source them before running any of the
+scripts (except for `dev/setup-{single,multi}-cluster`), if the scripts are run
+manually. You can rely on the environment variables being set correctly if you
+source the `dev/setup-{single,multi}-cluster` scripts.
+
+```bash
+source .envrc
+```
+
+### Running setup scripts
+
+If you use `dev/setup-single-cluster` or `dev/setup-multi-cluster` you can
+simply put your custom configuration in the root of the repository as
+`env.single-cluster` and `env.multi-cluster`. Those files will then be used
+instead of the defaults in `dev/env.single-cluster-defaults` or
+`dev/env.multi-cluster-defaults`, respectively.
+
+If you occasionally want to specify a different file, you can set the
+`FLEET_TEST_CONFIG` environment variable to point to your custom configuration
+(like `.envrc`) file. This will make those scripts use the file specified in the
+`FLEET_TEST_CONFIG` environment variable instead of the defaults in
+`dev/env.single-cluster-defaults` and `dev/env.multi-cluster-defaults` and also
+instead of the custom configuration in `env.single-cluster` and
+`env.multi-cluster`.
+
+### A list of all environment variables
 
     # use fleet-default for fleet in Rancher, fleet-local for standalone
     export FLEET_E2E_NS=fleet-local
+    export FLEET_E2E_NS_DOWNSTREAM=fleet-default
 
     # running single-cluster tests in Rancher Desktop
     #export FLEET_E2E_CLUSTER=rancher-desktop
@@ -38,66 +104,14 @@ source them before running any of the scripts.
     #export GIT_HTTP_USER="fleet-ci"
     #export GIT_HTTP_PASSWORD="foo"
 
-    # needed for OCI tests, which are currently disabled but part of the
-    # single-cluster tests
+    # needed for OCI tests, which are part of the single-cluster tests
 
     #export CI_OCI_USERNAME="fleet-ci"
     #export CI_OCI_PASSWORD="foo"
     #export CI_OCI_CERTS_DIR="../../FleetCI-RootCA"
-    #export HELM_PATH="/usr/bin/helm" # optional, for selecting Helm versions (see [Troubleshooting](#troubleshooting))
 
-## Running Tests on K3D
-
-These commands should set up k3d, and the fleet standalone images for single
-cluster tests.
-
-    export FLEET_E2E_NS=fleet-local
-    export FLEET_E2E_NS_DOWNSTREAM=fleet-default
-    export FLEET_E2E_CLUSTER=k3d-upstream
-    export FLEET_E2E_CLUSTER_DOWNSTREAM=k3d-upstream
-
-    dev/setup-k3d
-    dev/build-fleet
-    dev/import-images-k3d
-
-    # Needed for gitrepo tests
-    dev/import-images-tests-k3d
-    dev/create-zot-certs 'FleetCI-RootCA' # for OCI tests
-    dev/setup-fleet
-
-    # This should be needed only once
-    cd e2e/testenv/infra
-    go build -o . ./...
-    cd -
-
-    ./e2e/testenv/infra/infra setup
-
-    ginkgo e2e/single-cluster
-
-    ./e2e/testenv/infra teardown # optional
-
-Optional flags for reporting on long-running tests: `--poll-progress-after=10s --poll-progress-interval=10s`.
-
-For multi-cluster tests we need to configure two clusters. You also need to make
-the upstream clusters API accessible to the downstream cluster. The default
-`url` in [dev/setup-fleet-downstream] should work with most systems.
-
-    export FLEET_E2E_NS=fleet-local
-    export FLEET_E2E_CLUSTER=k3d-upstream
-    export FLEET_E2E_CLUSTER_DOWNSTREAM=k3d-downstream
-
-    dev/setup-k3ds
-    dev/build-fleet
-    dev/import-images-k3d
-    dev/setup-fleet-multi-cluster
-
-    ginkgo e2e/multi-cluster
-
-To test changes incrementally, rebuild just one binary, update the image in k3d
-and restart the controller:
-
-    dev/update-agent-k3d
-    dev/update-controller-k3d
+    # optional, for selecting Helm versions (see [Troubleshooting](#troubleshooting))
+    #export HELM_PATH="/usr/bin/helm"
 
 ### Troubleshooting
 
@@ -115,19 +129,9 @@ does not reuse dev scripts, however dev scripts may use CI scripts. We want to
 keep CI scripts short, targeted and readable. Dev scripts may change in an
 incompatible way at any day.
 
-## Requirements
-
-* docker
-* git
-* go
-* helm
-* jq
-* k3d
-* kubectl
-* ...
-
 ## Run integration tests
 
     ./dev/run-integration-tests.sh
 
-This will download and prepare setup-envtest, then it will execute all the integration tests.
+This will download and prepare setup-envtest, then it will execute all the
+integration tests.
--- a/dev/README.md
+++ b/dev/README.md
@@ -85,40 +85,38 @@ export FLEET_E2E_NS=fleet-local
 export FLEET_E2E_NS_DOWNSTREAM=fleet-default
 
 # running single-cluster tests in Rancher Desktop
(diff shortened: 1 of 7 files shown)


dev/k3d-clean
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+k3d cluster list -o json | \
+    jq -r '.[].name' | \
+    xargs -I{} k3d cluster delete {}
```
manno, line 5: Is this different from ` k3d cluster delete --all`?

dev/k3d-clean
```diff
@@ -0,0 +1,17 @@
...
+
+ids=$(docker ps -a --filter name=k3d --format "{{.ID}}")
+for id in $ids; do
+    docker stop "$id"
+    docker rm "$id"
+done
+
+ids=$(docker ps -a --filter name=act --format "{{.ID}}")
```
manno, line 13: Maybe we can put `act` in the script's name, I was surprised that it deletes something else beside k3d.

dev/k3d-clean
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+k3d cluster list -o json | \
```
weyfonk, line 3: This looks like an aggressive deletion method, and I wonder if it may be a bit heavy-handed, ie deleting clusters that the user may still want to keep, not necessarily related to Fleet tests. Is this really necessary (just curious)?

How about checking env vars `FLEET_E2E_CLUSTER` and `FLEET_E2E_CLUSTER_DOWNSTREAM` and `k3d cluster delete` them (ie deleting at most 2 clusters)?

dev/setup-single-cluster
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
```
weyfonk, line 1: I like the idea of these `setup-{single,multi}-cluster`, which simplifies the dev `README` and offers flexibility for users to override the `envrc` with their own variables. It is nicely explained in the `README`, too :)

weyfonk (commented):
Keeping an approval for later as this is still a draft, but I like where this is going. Curious to know what others think.

dev/setup-multi-cluster
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+
+if [ ! -f dev/envrc-mc ]; then
+    echo "Run this from the root of the repo"
+    exit 1
+fi
+
+./dev/k3d-clean
+
+if [ -n "$ENVRC" ]; then
```
manno, line 10: I see the convenience, but I'm not sure if the name is right. If we treat this like config, maybe we should call it config. 

What do you think about?
* .env.multicluster  (or something) 
* .env.singlecluster
* dev/env.multicluster-defaults
* dev/env.singlecluster-defaults

What trips me up is that `.envrc` is managed by https://direnv.net/. It's used to provide default env vars for different directories. 
I think `.env` files are more commonly loaded by the script/app itself (dotenv), not by the shell. That's also why we'd want to avoid `.env` without a suffix, it's loaded by Docker for example.

dev/k3d-clean
p-se: No, and that's a good point. But now I'm considering the suggestion of @weyfonk to make it only delete clusters that are set in environment variables, which will make it look similar to as it currently does.

dev/k3d-clean
p-se: Yes, it deletes something else as well. If `act` is used to create an environment, it creates a container to use which executes all those steps defined in the workflow files. 

If there is an issue while setting the environment up (means, an issue in the workflow files), the container is not being removed automatically. It may even be reused, which can cause issues. Those are shortly explained in the [pull requests that extends the `dev/README.md` with information about how to use `act`](https://github.com/rancher/fleet/pull/1765/files#diff-e14025c0fa40d4857e4b40fc96ea5ee995afe300014626b68cb55a479fa5b8fcR277-R288).

I'll change the name to `k3d-act-clean`.

dev/k3d-clean
p-se: No, it is not necessary. It was the brute force approach I've taken on my machine without ever having to wonder if anything would still remain. But it does make sense to think more critically about this now.

Since those new scripts don't require having any environment variables set, and because I'd like to keep it that way, I'm even thinking about taking the defaults of single- and multi-cluster environment variables and use those to selectively but always delete **both** kind of clusters, unless environment variables are set. In which case it would depend on the values of those environment variables which clusters would be removed (but it'd still be only up to two, because it would only consider `FLEET_E2E_CLUSTER` and `FLEET_E2E_CLUSTER_DOWNSTREAM`.

What do you think about that?

dev/setup-single-cluster
p-se: Thank you very much!

dev/setup-multi-cluster
p-se: As shortly discussed, I'll change the environment variable to `FLEET_TEST_CONFIG`. I also very much appreciate the other suggestions, of which I like a slight adaptation the most:

- dev/env.multi-cluster-defaults
- dev/env.single-cluster-defaults

p-se: Due to the greater proximity to the names of the scripts, `dev/setup-single-cluster` and `dev/setup-multi-cluster`.

dev/k3d-clean
weyfonk: Yes, that makes more sense to me; we'd keep the flexibility of being able to set cluster names while only impacting those when cleaning up test environments.

dev/env.multi-cluster-defaults
```diff
@@ -0,0 +1,14 @@
...
+
+export FLEET_E2E_CLUSTER=k3d-upstream
+export FLEET_E2E_CLUSTER_DOWNSTREAM=k3d-downstream
+
+export GIT_HTTP_USER=fleet-ci
+export GIT_HTTP_PASSWORD=foo
+
+export CI_OCI_USERNAME=fleet-ci
+export CI_OCI_PASSWORD=foo
+export CI_OCI_CERTS_DIR=/home/user/src/fleet/FleetCI-RootCA/
```
weyfonk, line 12: This line means that on most machines, the defaults will not work; should we mention this in the `README`, encouraging users to set up their own `env.{single,multi}-cluster` files?

p-se: Nice find! I'll look into it, perhaps it's possible to use a relative path to make it work out-of-the-box. If not, I think, we'll need to update the docs as you suggested.

dev/setup-cluster-config
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+
+set -ex
+
+if [[ -n "$1" && "$1" = "teardown" ]]; then
```
weyfonk, line 5: nit: wouldn't this have the same effect?
```suggestion
if [[ "$1" = "teardown" ]]; then
```

dev/setup-cluster-config
```diff
@@ -0,0 +1,28 @@
...
+    fi
+    # shellcheck source=/dev/null
+    source "$FLEET_TEST_CONFIG"
+elif [ -f "$DEFAULT_CONFIG" ]; then
+    # shellcheck source=/dev/null
+    source "$DEFAULT_CONFIG"
+else
+    echo >&2 "File not found: \$DEFAULT_CONFIG: $DEFAULT_CONFIG"
+    exit 1
+fi
```
weyfonk, line 28: The `else` block is unreachable, since we've already checked existence of `$DEFAULT_CONFIG` in L10-12, exiting when it is not found as a regular file.
```suggestion
else
    # shellcheck source=/dev/null
    source "$DEFAULT_CONFIG"
fi
```

weyfonk (commented):
Looking good, with a couple of notes. Thanks!

dev/env.multi-cluster-defaults
```diff
@@ -0,0 +1,14 @@
...
+export FLEET_E2E_CLUSTER_DOWNSTREAM=k3d-downstream
+
+export GIT_HTTP_USER=fleet-ci
+export GIT_HTTP_PASSWORD=foo
+
+export CI_OCI_USERNAME=fleet-ci
+export CI_OCI_PASSWORD=foo
+export CI_OCI_CERTS_DIR=FleetCI-RootCA/
+
+export HELM_PATH=/usr/bin/helm
```
aruiz14, line 14: I would set those variables only if they are not set already, which allows customizing their values (e.g. `HELM_PATH` on Mac is under the Homebrew installation directory):

```
export HELM_PATH="${HELM_PATH:-/usr/bin/helm}"
```

EDIT: I've just seen that this was probably the purpose of the `CUSTOM_CONFIG_FILE`, so maybe we can get rid of that and simplify the scripts a little bit (`setup-cluster-config` no longer needed?)

dev/env.single-cluster-defaults
```diff
@@ -0,0 +1,13 @@
...
+export FLEET_E2E_CLUSTER_DOWNSTREAM=k3d-upstream
+
+export GIT_HTTP_USER=fleet-ci
+export GIT_HTTP_PASSWORD=foo
+
+export CI_OCI_USERNAME=fleet-ci
+export CI_OCI_PASSWORD=foo
+export CI_OCI_CERTS_DIR=FleetCI-RootCA/
+
+export HELM_PATH=/usr/bin/helm
```
aruiz14, line 13: Same about conditionally set these variables

dev/README.md
```diff
@@ -3,101 +3,121 @@
...
-    export FLEET_E2E_CLUSTER_DOWNSTREAM=k3d-downstream
+the upstream clusters API accessible to the downstream cluster. The default URL
+in `dev/setup-fleet-downstream` should work with most systems.
 
-    dev/setup-k3ds
-    dev/build-fleet
-    dev/import-images-k3d
-    dev/setup-fleet-multi-cluster
+```bash
+source dev/setup-multi-cluster
```
aruiz14, line 35: Is `source` really needed? Please note that this step differs from the `single-cluster` case above

p-se: That single and multi cluster weren't the same was a mistake. The answer is no, it is not needed for the `dev/setup-{single,multi}-cluster` scripts. They will run successfully without it. But for `ginkgo` in the following lines, it will set the environment variables correctly and `ginkgo` requires them. So, it's either `source script && ginkgo ...` or `./script && source <the-same-env-vars> && ginkgo ...`. That's why I think sourcing them should be preferred.

dev/env.multi-cluster-defaults
p-se: Yes, that is the purpose of `setup-cluster-config`. Considering that the `setup-{single,multi}-cluster` scripts are recommended to be sourced ([see this explanation](https://github.com/rancher/fleet/pull/1764#discussion_r1325508492)) to set the environment variables correctly, changing them to not overwrite previously set environment variables would not change them, e.g.

```bash
source dev/setup-single-cluster && ginkgo e2e/single-cluster
```

would set the environment variables for "single-cluster". A successive run of

```bash
./dev/k3d-act-clean && source dev-setup-multi-cluster && ginkgo e2e/multi-cluster
```

would, provide your suggestions was implemented, not update the environment variables from single to multi cluster. But I understand how convenient it would be to just have to `export HELM_PATH=...` globally once. Perhaps `HELM_PATH` is an exception to those environment variables, since it's not supposed to change. But the same could be said for `CI_OCI_` and `GIT_HTTP_` environment variables. Adding your suggestion to all of them may complicate those scripts, since we'd be having two types of environment variables. At this point I think it may be simpler to create a `env.single-cluster` and `env.multi-cluster` file to customize the setup.

But generally, we can

- leave it as it is,
- change it just for the `HELM_PATH` variable or
- change it for most of the variables, but not all of them.

What do you think?

dev/README.md
aruiz14: Thanks for explaining!

dev/env.multi-cluster-defaults
aruiz14: Agree on the 2 types of variables:
 - `FLEET_E2E_*` ones are probably not meant to be modified. If that's the case then it means we are under a different test scenario and they'll probably deserve its own script
 -  Git and OCI user/password, certificates directory or Helm path may be dependent on your local setup and are independent from the test scenario, so I think it makes sense to respect if values are already set (e.g. using `direnv`).

dev/setup-multi-cluster
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+export DEFAULT_CONFIG="dev/env.multi-cluster-defaults"
+export CUSTOM_CONFIG_FILE="env.multi-cluster"
```
manno, line 4: Where is this used?

manno (changes requested):

dev/setup-cluster-config
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+set -ex
```
manno, line 3: Should this warn for uninitialized variables? 
```suggestion
set -exu
```

dev/k3d-act-clean
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bash
+
+set -e
```
manno, line 3:
```suggestion
set -eu
```

dev/setup-multi-cluster
p-se: I've removed it accidentally.

dev/k3d-act-clean
p-se: Should not be added. There are optional environment variables in it as well.

dev/setup-cluster-config
p-se: Should not. "$1" is optional and there are other optional environment variables in it as well.

manno (approved):

Issue #1740, opened 2023-08-23
Controllers Support Sharding via Labels


manno:
Sharding will allow Fleet to scale horizontally, by spawning multiple fleet controllers. This will allow Fleet to handle larger fleets of clusters, more git repos, etc.

Each controller will be responsible for one shard. The controllers can process resources in parallel. Each shard has an ID, only resources labeled with the shard ID are processed and cached by the controller.
Unlabeled resources are handled by the main controller.

As a first iteration, shard controllers are created when Fleet is installed. They are clones of the [main fleet controller](https://github.com/rancher/fleet/blob/master/internal/cmd/controller/operator.go). Each one is configured with their shard ID, e.g. via an env var. All resources created by the shard controllers must be labeled with the ID, to ensure they're only processed by the shard controller.

## Implementation

* leader election in current controllers makes sure, that only one controller is processing events. Even when increasing the replica count, the replicas are just in standby, they don't process events. The ID must be part of the leader election id.
* deployment names must differ.
* agentmanagement and clean up are not sharded. The fleet-agent is not sharded.
* users will want to control where the shards are started, e.g. on which node.
* controllers can share the configmap config file.
* controllers don't conflict on shared resources.
* controllers only create gitjobs for their labeled gitrepos.
* controllers only create bundledeployments for their clusters/gitrepos.
* shard id label needs to be present on all watched resources, but since labels are inherited, they should not trigger re-deployments? 
* label git resources with `fleet.cattle.io/shard=abc123` or `shard.fleet.cattle.io/id=abc123` or what's appropriate for the implementation.

Since we're using helm to configure the shards, this is a rather static solution:
```
helm install fleet --set shards=abc123,abc234
-> one main deployment (fleetcontroller, cleanup, agentmgt)
-> 2 fleetcontroller deployments  (ENV shard=abc123, shard=abc234, ...)
```

Configuration would be done through [Rancher settings](https://github.com/rancher/fleet/issues/1717).
Changing the shard setting needs to trigger a redeployment of Fleet in Rancher.

## Notes

Here's how Flux did sharding:
* https://github.com/weaveworks/flux-shard-controller/tree/main/docs
* and cache config: https://github.com/fluxcd/source-controller/blob/main/main.go#L349-L357

## Future possibilities

To allow dynamic changes to the shards, we could add a controller, e.g. to react to config map changes:

```
"fleet-controller" configmap
  shards: [abc123, abc234]
[...]

Issue #1734, opened 2023-08-23
Convert fleet-agent to controller-runtime


manno:
We discussed wrangler vs controller-runtime multiple times.  How feasible is it to switch frameworks?

Fleet-agent is relatively limited. When migrating to controller-runtime:

* how easy is it to port the different "timers", `time.After`,`EnqueueAfter`
* are we missing something when we drop `apply`, e.g. cleanup, updates
* agent sets up dynamic watches on deployed resources, ?
* should we extract existing "skip" logic into [predicates/event filters](https://sdk.operatorframework.io/docs/building-operators/golang/references/event-filtering/) first or are wrangler handlers too different?
* will probably need a separate go.mod

If it's not possible to move the "monitor" part of the agent, because of timers and `internal/cmd/agent/trigger/watcher.go`, does it help to [split the agent](https://github.com/rancher/fleet/issues/1733), so we can at least port the deployer?

---

The [fleet-agent controller was split into three processes](https://github.com/rancher/fleet/pull/1905):

* registration
* cluster status updater
* fleet-agent controller, watches bundledeployments

Only the last part is an actual controller. Cluster status updates work on a timer and registration is done when the agent starts up. A new function `registration.Get` was added, to just return an existing registration to the other processes.

The agent to a statefulset, to avoid leader election. The code of the handlers was spread between the "handler", the deploy "manager" and the helm deployer. It has been consolidated into the `deployer` package.

"Trigger" was renamed to `driftdetect` as it uses the helm history to detect drift for deployments. Which is similar to the`monitor`, which updates the bundle deployments status.

After splitting the agent, the [fleet-agent controller was converted to controller-runtime](https://github.com/rancher/fleet/pull/1772):
* new bundledeployment controller + reconciler
* agent reconciler only sets status when exiting reconcile
* when deployed resources change, re-trigger bundledeployment via status update
* simplify bundlestatus condition handling
* switch to zap logging

aruiz14:
After chatting about this, and how we are using leader election, I was wondering if it is really necessary.
At the moment, it's used to prevent multiple instances of the `fleet-agent` (although it's deployment specifies only 1 replica), to act on the same `BundleDeployment` resource, ending up with repeated work or conflicts in the worst case.

I believe we can use an alternative approach based on optimistic locking, hence eliminating the restriction of having one single active instance:
 1. `BundleDeployment` status reflects that an instance is already processing it (e.g. a new in progress condition).
[...]

Issue #1491, opened 2023-04-26
[SURE-3805] Make leaderelection poll period configurable (was: Fleet lock api calls account for 33% of all api calls)


kkaempf:
Internal reference: [SURE-3805](https://jira.suse.com/browse/SURE-3805)

### Issue description:
For the top 10 calls to the kube-apiserver, Fleet locks account for 33% of these calls. Here's the top 10:
```
   9399 "/api/v1/namespaces/cattle-fleet-local-system/configmaps/fleet-agent-lock"
   9374 "/api/v1/namespaces/kube-system/configmaps/cattle-controllers?timeout=15m0s"
   8227 "/api/v1/namespaces/cattle-fleet-system/configmaps/fleet-agent-lock"
   5387 "/api/v1/namespaces/cattle-fleet-system/configmaps/fleet-controller-lock"
   4933 "/api/v1/namespaces/kube-system/configmaps/k3s"
   4736 "/api/v1/namespaces/cattle-fleet-system/configmaps/gitjob"
   4220 "/api/v1/namespaces/ingress-nginx/configmaps/ingress-controller-leader-nginx"
   1588 "/api/v1/namespaces/kube-system/configmaps/coredns-autoscaler"
   1588 "/api/v1/namespaces/default/services/kubernetes"
   1588 "/api/v1/namespaces/default/endpoints/kubernetes"
```

The first, third, and forth URIs in the list are for locks related to Fleet. The log file used for this top 10 had 69,128 requests over a 4+ hour period - "2021-12-13T09:30:07.389612Z" to "2021-12-13T13:54:50.179191Z". That amounts to a rate of 1.45 requests per second. Example JSON request:

```
{
  "kind": "Event",
  "apiVersion": "audit.k8s.io/v1",
  "level": "RequestResponse",
  "auditID": "6d513d96-104c-47ad-be40-9c55b4261981",
  "stage": "ResponseComplete",
  "requestURI": "/api/v1/namespaces/cattle-fleet-local-system/configmaps/fleet-agent-lock",
  "verb": "get",
  "user": {
    "username": "system:serviceaccount:cattle-fleet-local-system:fleet-agent",
    "uid": "9d6c91c0-8e7f-4208-b449-b3ffe59781da",
    "groups": [
      "system:serviceaccounts",
      "system:serviceaccounts:cattle-fleet-local-system",
      "system:authenticated"
    ]
  },
  "sourceIPs": [
    "172.16.0.17"
  ],
[...]

PR #1772, opened 2023-09-11
Fleet agent controller runtime

manno:
Refers to https://github.com/rancher/fleet/issues/1734

Requires:
- [x] https://github.com/rancher/fleet/pull/1798
- [x] https://github.com/rancher/fleet/pull/1905

Follow up:

* find better mechanism to "enqueue" bundledeployment from drift correction ("trigger" mini-controller)?
  * maybe disable resync in agent to avoid triggering for all bd?  
* when [converting fleetcontroller](https://github.com/rancher/fleet/issues/1972), filter bd status events
* cleanup should use finalizers instead of iterating all the releases: 
  * https://github.com/rancher/fleet/blob/master/internal/cmd/agent/deployer/cleanup/cleanup.go
* bring back leader election, configurable with args, should fit sharding strategy, see:
  * https://github.com/rancher/fleet/issues/1491
  * https://github.com/rancher/fleet/issues/1740
  * logging, field names (fieldNames), log levels (1,5?)

- [ ] get rid of nodestatus reporting in clusterstatus.go, https://github.com/rancher/fleet/issues/1973
- [ ] is it possible to remove [`requeueAfter` with `bd.Status.AppliedDeploymentID = ""`](https://github.com/rancher/fleet/blob/7dbfa6d7efaa50b16d4afb542c8271ca8816182a/internal/cmd/agent/controller/bundledeployment_controller.go#L140-L151), or is it still needed? - https://github.com/rancher/fleet/pull/1977/commits/b7810f0d5b9815e650a4273caa502349bd4fdaf5
- [ ] clean up agent resources needed? https://github.com/rancher/fleet/pull/1977/commits/b7810f0d5b9815e650a4273caa502349bd4fdaf5
- [ ] clean up calls listdeployment too often, tries to delete k3s charts - https://github.com/rancher/fleet/pull/1976
- [ ] variables instead constants in types - https://github.com/rancher/fleet/pull/1977/commits/14382e4213b07c24289aa35ded1d2d98b68d313f

Manual testing:
- [ ] do agent restarts, agent updates work?
- [ ] keeps resources when upstream is offline
- [ ] cleans up orphaned resources
- [x] `e2e/single-cluster/bundle_diffs_test.go:145` is flaky  -> https://github.com/rancher/fleet/pull/1975


internal/cmd/agent/deployer/monitor/updatestatus.go
```diff
@@ -1,56 +1,150 @@
...
+		}
+	}
+
+	return errors.New(msg)
+}
+
+// updateFromResources updates the status with information from the
+// helm release history and an apply dry run.
+func (m *Monitor) updateFromResources(ctx context.Context, bd *fleet.BundleDeployment, resources *helmdeployer.Resources) error {
+	logger := log.FromContext(ctx).WithName("updateBundleDeploymentStatus")
```
aruiz14, line 146: Are the logger names okay in this file? `updateFromResources()` uses `updateBundleDeploymentStatus`, while `UpdateStatus()` uses `MonitorBundle`.

If the idea is to use something meaningful to the overall process that is happening, maybe we would use `logr.NewContext` to set the logger name at a higher level, so all deeper calls to `log.FromContext(ctx)` will have it already configured?

internal/cmd/agent/deployer/driftdetect/driftdetect.go
```diff
@@ -1,57 +1,147 @@
...
+
+	// This mechanism of triggering requeues for changes is not ideal.
+	// It's a workaround since we can't enqueue directly from the trigger
+	// mini controller. Triggering via a status update is expensive.
+	// It's hard to compute a stable hash to make this idempotent, because
+	// the hash would need to be computed over the whole change. We can't
+	// just use the resource version of the bundle deployment. We would
+	// need to look at the deployed resources and compute a hash over them.
+	// However this status update happens for every changed resource, maybe
+	// multiple times per resource. It will also trigger on a resync.
```
aruiz14, line 119: Happy to talk or brainstorm about this, but could it be an option to order all resources by `kind` + `key`, hash their resource versions and compare that against `SyncGeneration`? This should avoid unnecessary re-deployments, for example on fleet-agent restarts?

Also, we should consider using `Generation` instead of the `ResourceVersion`. So status updates are omitted and don't requeue bundle deployments?

internal/cmd/agent/apply.go
```diff
@@ -0,0 +1,47 @@
...
+		for range ticker.Context(ctx, 30*time.Second) {
+			disc.Invalidate()
+			mapper.Reset()
+		}
+	}()
+
+	dyn, err := dynamic.NewForConfig(localconfig)
+	if err != nil {
+		return nil, nil, nil, err
+	}
```
aruiz14, line 38: Isn't this already in `newMappers()`?

internal/cmd/agent/controller.go
```diff
@@ -2,109 +2,246 @@ package agent
...
-		setupLog.Error(err, "failed to register controllers")
+	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
+		setupLog.Error(err, "unable to set up ready check")
 		return err
 	}
 
-	if err := appCtx.Start(ctx); err != nil {
-		setupLog.Error(err, "failed to start app context")
+	setupLog.Info("starting manager")
+	//if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
```
aruiz14, line 108: Is signals handling already configured somewhere else?

dev/setup-k3ds
```diff
@@ -22,7 +22,7 @@ fi
 # https://hub.docker.com/r/rancher/k3s/tags
 #args="$args -i docker.io/rancher/k3s:v1.22.15-k3s1"
 
-k3d cluster create upstream --servers 3 --api-port 36443 -p '80:80@server:0' -p '443:443@server:0' $args
+k3d cluster create upstream --servers 1 --api-port 36443 -p '80:80@server:0' -p '443:443@server:0' $args
```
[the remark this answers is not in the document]
moio: It is possible to set up a 1-node k3s/k3d cluster backed by etcd by adding the `--cluster-init@server:0` argument if desired.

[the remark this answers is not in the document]
manno: Interesting info about embedded etc vs sqlite.

It's already set to `1` in `dev/setup-k3ds`, but it doesn't have to be the same.
I put it to 1, because images have to be uploaded into each node. Especially for scripts like https://github.com/rancher/fleet/blob/master/dev/update-agent-k3d one node is noticeable faster.

integrationtests/agent/bundle_deployment_status_test.go
```diff
@@ -1,38 +1,38 @@
...
-			},
-		}, "v2": {
-			{
-				Name:     "deployment-v2.yaml",
-				Content:  string(v2),
-				Encoding: "",
-			},
+func init() {
+	v1, _ := os.ReadFile(assetsPath + "/deployment-v1.yaml")
+	v2, _ := os.ReadFile(assetsPath + "/deployment-v2.yaml")
```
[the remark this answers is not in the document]
manno: No, the data needs to be there before the `BeforeSuite` in `suite_test.go`. 
I can move the preparation of the lookup mock into the `suite_test.go`, but test data wouldn't be in the same file as the test anymore. Or I can add a panic if there was an error loading the test data from the assets folder.

I don't care much about the error as it only occurs if a commit removes the test assets. However, then the test would fail a little later, so technically that commit shouldn't get merged...

internal/helmdeployer/helmcache/secret.go
```diff
@@ -47,7 +47,9 @@ func (s *SecretClient) DeleteCollection(ctx context.Context, opts metav1.DeleteO
 
 // Get gets a secret from the wrangler cache.
 func (s *SecretClient) Get(ctx context.Context, name string, opts metav1.GetOptions) (*v1.Secret, error) {
```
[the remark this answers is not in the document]
manno: I'll add a `var _ corev1.SecretInterface = &SecretClient{}` to make it more obvious that this client implements a specific interface, remove wrangler from code comments and make opts an `_` arg.

internal/helmdeployer/helmcache/secret.go
```diff
@@ -3,25 +3,25 @@ package helmcache
...
 	applycorev1 "k8s.io/client-go/applyconfigurations/core/v1"
 	"k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // SecretClient implements methods to handle secrets. Get and list will be retrieved from the wrangler cache, the other calls
 // will be make to the Kubernetes API server.
 type SecretClient struct {
-	cache     corecontrollers.SecretCache
+	cache     client.Client
```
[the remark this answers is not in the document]
manno: Well, as opposed to the client, this is the cached client.

internal/helmdeployer/helmcache/secret.go
```diff
@@ -56,19 +58,16 @@ func (s *SecretClient) List(ctx context.Context, opts metav1.ListOptions) (*v1.S
 	if err != nil {
 		return nil, err
 	}
-	secrets, err := s.cache.List(s.namespace, labels)
+	secrets := v1.SecretList{}
+	err = s.cache.List(ctx, &secrets, &client.ListOptions{
+		Namespace:     s.namespace,
+		LabelSelector: labels,
```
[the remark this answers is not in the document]
manno: No, that  would break the interface.

pkg/apis/fleet.cattle.io/v1alpha1/bundle_types.go
```diff
@@ -288,7 +288,8 @@ var (
 	// BundleDeploymentConditionDeployed is used by the bundledeployment
 	// controller. It is true if the handler returns no error and false if
 	// an error is returned.
-	BundleDeploymentConditionDeployed = "Deployed"
+	BundleDeploymentConditionDeployed  = "Deployed"
+	BundleDeploymentConditionMonitored = "Monitored"
```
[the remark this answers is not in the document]
manno: Interesting, there's a couple more string variables at the beginning of the file. Not sure I feel comfortable to change types in the public api pkg in this PR. Maybe in a follow-up? It does still compile with the change, though.

internal/cmd/controller/controllers/bundle/controller.go
```diff
@@ -154,7 +154,7 @@ func (h *handler) OnBundleChange(bundle *fleet.Bundle, status fleet.BundleStatus
 	}
 
 	if status.ObservedGeneration != bundle.Generation {
-		if err := setResourceKey(&status, bundle, manifest, h.isNamespaced); err != nil {
+		if err := setResourceKey(context.Background(), &status, bundle, manifest, h.isNamespaced); err != nil {
```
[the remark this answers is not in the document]
manno: Hm, because we don't have a good context to pass in, yet? I'm fine with that change, but it's only used for "helm template".

internal/cmd/agent/deployer/monitor/updatestatus.go
manno: Ah thanks. I renamed the funcs. So far I only used `log.IntoContext` in higher level files. I'll just pass in the logger directly.

internal/cmd/agent/deployer/driftdetect/driftdetect.go
manno: Yes let's talk. I'm currently assuming it's cheaper to trigger than to compute a proper hash.

internal/cmd/agent/apply.go
manno: In this one https://github.com/rancher/fleet/blob/fleet-agent-controller-runtime/internal/cmd/agent/clusterstatus.go#L172?
That's only used for the clusterstatus sub-command.

internal/helmdeployer/deployer.go
```diff
@@ -697,27 +705,28 @@ func (h *Helm) deleteByRelease(bundleID, releaseName string, keepResources bool)
...
 	}
 
 	u := action.NewUninstall(&cfg)
 	_, err = u.Run(releaseName)
 	return err
 }
 
-func (h *Helm) delete(bundleID string, options fleet.BundleDeploymentOptions, dryRun bool) error {
+func (h *Helm) delete(ctx context.Context, bundleID string, options fleet.BundleDeploymentOptions, dryRun bool) error {
+	logger := log.FromContext(ctx).WithName("HelmDeployer").WithName("delete").WithValues("dry run", dryRun)
```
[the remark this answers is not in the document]
manno: I think `bundleID` is just the `bundledeployment.Name`

internal/cmd/agent/controller.go
manno: Took me a while to find it again: https://github.com/rancher/fleet/blob/fleet-agent-controller-runtime/cmd/fleetagent/main.go#L14

internal/cmd/agent/controller/bundledeployment_controller.go
```diff
@@ -0,0 +1,241 @@
...
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.15.0/pkg/reconcile
+func (r *BundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx).WithName("bundledeployment")
+	ctx = log.IntoContext(ctx, logger)
+	key := req.String()
+
+	// get latest BundleDeployment from cluster
+	bd := &fleetv1.BundleDeployment{}
+	err := r.Get(ctx, req.NamespacedName, bd)
+	if apierrors.IsNotFound(err) {
```
raulcabello, line 79: This will not work if the `BundleDeployment` is deleted when the agent is not available. In that case helm releases will be left orphaned forever.

internal/cmd/agent/controller.go
```diff
@@ -2,109 +2,245 @@ package agent
...
+	setupLog.Info("listening for changes on upstream cluster", "cluster", clusterName, "namespace", fleetNamespace)
+
+	metricsAddr := ":8080"
+	probeAddr := ":8081"
+
+	mgr, err := ctrl.NewManager(upstreamConfig, ctrl.Options{
+		Scheme:                 scheme,
+		Metrics:                metricsserver.Options{BindAddress: metricsAddr},
+		HealthProbeBindAddress: probeAddr,
+		LeaderElection:         false,
```
raulcabello, line 71: Are we sure we want to remove leader election? If that is the case, do we want to do it in gitjob and fleet controller as well?

We would be losing the ability to be highly available, and we would see some incosistent behaviour if someone manually scale the statefulset

internal/cmd/agent/controller/bundledeployment_controller.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	// update our mini controller, which watches deployed resources for drift
+	_, err = r.DriftDetect.Refresh(logger, req.String(), bd, resources)
+	if err != nil {
+		logger.V(1).Error(err, "Failed to refresh drift detection", "step", "drift")
+		merr = append(merr, fmt.Errorf("failed refreshing drift detection: %w", err))
+	}
+
+	err = r.Cleanup.CleanupReleases(ctx, key, bd)
```
raulcabello, line 168: We iterate through all BundleDeployments each time any of them changes to see if they need to be cleaned up. That's not very efficient. We can use finalizers as I mentioned in the previous comment

internal/cmd/agent/deployer/driftdetect/driftdetect.go
```diff
@@ -1,57 +1,147 @@
...
-func (m *DriftDetect) AllResources(bd *fleet.BundleDeployment) (*helmdeployer.Resources, error) {
-	resources, err := m.deployer.Resources(bd.Name, bd.Status.Release)
+func (d *DriftDetect) Clear(bdKey string) error {
+	return d.trigger.Clear(bdKey)
+}
+
+func (d *DriftDetect) Refresh(logger logr.Logger, bdKey string, bd *fleet.BundleDeployment, resources *helmdeployer.Resources) (*fleet.BundleDeployment, error) {
+	logger = logger.WithName("DriftDetect")
+	if bd == nil {
+		return bd, d.trigger.Clear(bdKey)
```
raulcabello, line 62: This will never be reached as bd will never be nil when this is called. Again finalizers can help here. Happy to discuss this

internal/cmd/agent/deployer/driftdetect/driftdetect.go
raulcabello: yes, let's talk about this

internal/cmd/agent/controller/bundledeployment_controller.go
```diff
@@ -0,0 +1,241 @@
...
+		if len(status.ModifiedStatus) > 0 && bd.Spec.CorrectDrift.Enabled {
+			err = r.Deployer.RemoveExternalChanges(ctx, bd)
+			if err != nil {
+				merr = append(merr, fmt.Errorf("failed reconciling drift: %w", err))
+			}
+		}
+
+		// TODO is this needed with drift correction?
+		if len(bd.Status.ModifiedStatus) > 0 && monitor.ShouldRedeploy(bd) {
+			result.RequeueAfter = durations.MonitorBundleDelay
```
raulcabello, line 149: why do we want to requeue here?

internal/manifest/lookup.go
```diff
@@ -1,27 +1,25 @@
...
 }
 
-type lookup struct {
-	content fleetcontrollers.ContentClient
+type Lookup struct {
 }
 
-func (l *lookup) Get(id string) (*Manifest, error) {
-	c, err := l.content.Get(id, metav1.GetOptions{})
+func (l *Lookup) Get(ctx context.Context, client client.Reader, id string) (*Manifest, error) {
```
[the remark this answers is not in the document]
aruiz14: Oh I see, didn't think of that, thanks

internal/helmdeployer/helmcache/secret.go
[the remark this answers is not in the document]
aruiz14: Right, sorry for the noise 😅

internal/helmdeployer/helmcache/secret.go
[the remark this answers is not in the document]
aruiz14: I didn't notice this was implementing an interface from Helm, my bad

internal/cmd/agent/controller.go
aruiz14: So this commented line can go away?

internal/cmd/agent/deployer/driftdetect/driftdetect.go
aruiz14: > Yes let's talk. I'm currently assuming it's cheaper to trigger than to compute a proper hash.

I'm not sure it's cheaper, especially since BundleDeployments updates may trigger as well updates on other resources on the upstream cluster (e.g. updating status of Bundles, GitRepos or Clusters resources). Also, beware of the multiplying factor in the case of many downstream clusters.

internal/cmd/agent/controller/bundledeployment_controller.go
manno: It's old behavior. I'm not sure we can remove it.

internal/helmdeployer/helmcache/secret.go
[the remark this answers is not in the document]
manno: The `client.Client` is provided by controller-runtime. The package is still a caching storage driver for helm, even if we inject dependencies. If you have a better name, please, but I think `helmcache` is to the point and short enough. This PR didn't change anything about that, the `corecontrollers.SecretCache` did all the caching before and it's also injected.

.golangci.json
```diff
@@ -64,6 +64,14 @@
           "revive"
         ]
       },
+      {
+        "path": "internal/cmd/cli/cleanup/cleanup.go",
+        "text": "G601: Implicit memory aliasing in for loop."
+      },
+      {
+        "path": "internal/cmd/controller/controllers/bundle/controller.go",
+        "text": "G601: Implicit memory aliasing in for loop."
```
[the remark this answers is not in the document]
manno: I have a separate PR for that: https://github.com/rancher/fleet/pull/1958

I'll remove it here, but than the PR will show linter errors.

internal/cmd/agent/controller/bundledeployment_controller.go
manno: Yes, that's old logic again (https://github.com/rancher/fleet/blob/master/internal/cmd/agent/deployer/cleanup/cleanup.go). I'd not fix it within this PR, but I'll add an item to the follow up items.

internal/helmdeployer/helmcache/secret.go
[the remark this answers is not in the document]
aruiz14: But in that case, the type was `SecretCache` and not `SecretClient`.
I've tried to follow where this `client.Client` is coming from (its part of the `Helm` object, which takes it from the `localClient`, received from this code https://github.com/rancher/fleet/blob/a138a30d1268d0cdb30343f54ed170bd0dbe534d/internal/cmd/agent/controller.go#L228), and there is no explicit mention to this being a cached content. It completely relies on the default settings, which could inadvertently change.
I'm not against changing the method name, but to ensure that the client used by `SecretClient` really caches secrets.

internal/cmd/agent/controller.go
manno: We already removed leader election in  https://github.com/rancher/fleet/pull/1905/files
We can add a follow up item to bring it back. We'll need to discuss how that works together with the upcoming [sharding story](https://github.com/rancher/fleet/issues/1740).

internal/cmd/agent/controller/bundledeployment_controller.go
manno: Fleet agent could never deal with that situation? I think it always checked all bd the next time it deployed one, as you state below.

internal/helmdeployer/helmcache/secret.go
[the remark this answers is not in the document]
manno: This is the main client we're going to use from now on. Construction is a bit special, because we need to build it for upstream by ourselves. Otherwise the controller-runtime manager does that for us.

> It is a common pattern in Kubernetes to read from a cache and write to the API server. This pattern is covered by the creating the Client with a Cache.  (https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client
)

You already linked `newCluster` where I [set up the cache](https://github.com/rancher/fleet/blob/a138a30d1268d0cdb30343f54ed170bd0dbe534d/internal/cmd/agent/controller.go#L236-L242), similar to the code in the manager.


Again, this isn't much different than before, instead of `client.Client`, we relied on `"github.com/rancher/wrangler/v2/pkg/generated/controllers/core/v1".SecretCache` to stay cached.

internal/cmd/agent/deployer/driftdetect/driftdetect.go
raulcabello: We should call trigger.Clear, currently we are not doing it.  We are leaving the watches forever

internal/cmd/agent/controller/bundledeployment_controller.go
raulcabello: maybe we can fix this and [this](https://github.com/rancher/fleet/pull/1772#discussion_r1407382343) using finalizers in a separate PR?

internal/cmd/agent/controller/bundledeployment_controller.go
raulcabello: yes, but checking all `BundleDeployments` in each reconcilation loop, which will happens many times, is not very efficient

internal/cmd/agent/controller.go
```diff
@@ -2,109 +2,245 @@ package agent
...
+
+var (
+	scheme      = runtime.NewScheme()
+	localScheme = runtime.NewScheme()
 )
 
 // defaultNamespace is the namespace to use for resources that don't specify a namespace, e.g. "default"
 const defaultNamespace = "default"
 
+func init() {
```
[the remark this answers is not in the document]
aruiz14: `init` is a special function name in Go (like `main`). I will be called once when the package is loaded:
https://go.dev/ref/spec#Package_initialization

internal/cmd/agent/root.go
```diff
@@ -22,26 +23,32 @@ type UpstreamOptions struct {
...
 	return nil
 }
 
 func (a *FleetAgent) Run(cmd *cobra.Command, args []string) error {
-	zopts := zap.Options{
-		Development: true,
-	}
+	ctx := clog.IntoContext(cmd.Context(), ctrl.Log)
+
+	// for compatibility, override zap opts with legacy debug opts. remove once manifests are updated.
```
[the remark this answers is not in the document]
manno: The helm chart and the https://github.com/rancher/fleet/blob/master/internal/cmd/controller/agent/manifest.go#L295-L297 only know about the old flags `--debug` and `--debug-level`.

internal/helmdeployer/deployer.go
```diff
@@ -242,11 +244,11 @@ func (h *Helm) Deploy(bundleID string, manifest *manifest.Manifest, options flee
 
 // RemoveExternalChanges does a helm rollback to remove changes made outside of fleet.
 // It removes the helm history entry if the rollback fails.
-func (h *Helm) RemoveExternalChanges(bd *fleet.BundleDeployment) error {
-	logrus.Infof("Drift correction: rollback BundleDeployment %s", bd.Name)
+func (h *Helm) RemoveExternalChanges(ctx context.Context, bd *fleet.BundleDeployment) error {
+	log.FromContext(ctx).WithName("RemoveExternalChanges").Info("Drift correction: rollback")
```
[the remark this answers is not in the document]
manno: Yes, we're not only using controller-runtime, we're using the `builder`, so this happens: https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/builder/controller.go#L390
Then the `reconcileID` is added https://github.com/kubernetes-sigs/controller-runtime/blob/v0.16.3/pkg/internal/controller/controller.go#L309 and finally our reconciler adds a name to the names: https://github.com/rancher/fleet/blob/fleet-agent-controller-runtime/internal/cmd/agent/controller/bundledeployment_controller.go#L72-L73

internal/cmd/agent/controller/bundledeployment_controller.go
manno: https://github.com/rancher/fleet/pull/1977/commits/b7810f0d5b9815e650a4273caa502349bd4fdaf5

internal/helmdeployer/deployer.go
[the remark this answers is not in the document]
weyfonk: Ack, thanks for clarifying!

internal/helmdeployer/deployer.go
```diff
@@ -669,10 +676,11 @@ func (h *Helm) Delete(bundleID, releaseName string) error {
 		// Never found anything to delete
 		return nil
 	}
-	return h.deleteByRelease(bundleID, releaseName, keepResources)
+	return h.deleteByRelease(ctx, bundleID, releaseName, keepResources)
 }
 
-func (h *Helm) deleteByRelease(bundleID, releaseName string, keepResources bool) error {
+func (h *Helm) deleteByRelease(ctx context.Context, bundleID, releaseName string, keepResources bool) error {
+	logger := log.FromContext(ctx).WithName("deleteByRelease").WithValues("release name", releaseName, "keep resources", keepResources)
```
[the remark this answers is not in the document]
manno: I switched the fields to camelCase, there is conflicting advice in https://github.com/kubernetes-sigs/controller-runtime/blob/main/TMP-LOGGING.md, but I looked through the controller-runtime code and they don't use spaces: https://github.com/kubernetes-sigs/controller-runtime/blob/2154ffbc22e26ffd8c3b713927f0df2fa40841f2/pkg/internal/controller/controller.go#L309

internal/cmd/agent/apply.go
```diff
@@ -0,0 +1,47 @@
...
+
+	disc := memory.NewMemCacheClient(d)
+	mapper := restmapper.NewDeferredDiscoveryRESTMapper(disc)
+
+	go func() {
+		for range ticker.Context(ctx, 30*time.Second) {
+			disc.Invalidate()
+			mapper.Reset()
```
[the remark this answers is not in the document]
manno: To detect changes to the schema.

Issue #1285, opened 2023-01-27
Take Ownership for manifests/kustomize bundles


manno:
Fleet supports "takeOwnership" for helm based bundles. The feature is missing for manifest/kustomize based bundles.

SURE-3976

PR #1775, opened 2023-09-12
Add integration test to verify fleet.yaml helm options for non-helm type bundles

thardeck:
even when fleet.yaml does not reference a helm chart.

Also make sure that the values are actually evaluated and that no potential default value is interfering with the test.

The test proves that there is no need for an additional `takeOwnership` like requested in  rancher/fleet#1285 because helm options are evaluated despite `fleet.yaml` referencing a non-helm type bundle.
Which makes sense because helm is used underneath and otherwise users could not provide a  `releaseName`.

### DIFF
--- a/integrationtests/cli/apply/apply_test.go
+++ b/integrationtests/cli/apply/apply_test.go
@@ -222,5 +222,22 @@ var _ = Describe("Fleet apply", Ordered, func() {
 				}).Should(BeTrue())
 			})
 		})
+
+		When("passes along helm options with a kustomize bundle", func() {
+			BeforeEach(func() {
+				name = "helm_options_kustomize"
+				options = apply.Options{Output: gbytes.NewBuffer()}
+				dirs = []string{cli.AssetsPath + name}
+			})
+
+			It("publishes the flag in the bundle options", func() {
+				Eventually(func() bool {
+					bundle, err := cli.GetBundleFromOutput(options.Output)
+					Expect(err).NotTo(HaveOccurred())
+					return bundle.Spec.Helm.TakeOwnership &&
+						bundle.Spec.Helm.ReleaseName == "kustomize"
+				}).Should(BeTrue())
+			})
+		})
 	})
 })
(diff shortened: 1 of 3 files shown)


integrationtests/cli/apply/apply_test.go
```diff
@@ -179,4 +179,38 @@ var _ = Describe("Fleet apply", Ordered, func() {
...
+				name = "helm_options_enabled"
+				options = apply.Options{Output: gbytes.NewBuffer()}
+				dirs = []string{cli.AssetsPath + name}
+			})
+
+			It("publishes the flag in the bundle options", func() {
+				Eventually(func() bool {
+					bundle, err := cli.GetBundleFromOutput(options.Output)
+					Expect(err).NotTo(HaveOccurred())
+					return bundle.Spec.Helm.TakeOwnership && bundle.Spec.Helm.Atomic && bundle.Spec.Helm.Force && bundle.Spec.Helm.WaitForJobs && bundle.Spec.Helm.DisablePreProcess && bundle.Spec.Helm.ReleaseName == "enabled"
```
weyfonk, line 195: This looks somewhat easier to ready to me (as would be future diffs on this):
```suggestion
					return bundle.Spec.Helm.TakeOwnership &&
					        bundle.Spec.Helm.Atomic &&
					        bundle.Spec.Helm.Force &&
					        bundle.Spec.Helm.WaitForJobs &&
					        bundle.Spec.Helm.DisablePreProcess &&
					        bundle.Spec.Helm.ReleaseName == "enabled"
```

Same remark about L211.

integrationtests/cli/assets/helm_options_disabled/fleet.yaml
```diff
@@ -0,0 +1,7 @@
+helm:
```
weyfonk, line 1: While not referencing a Helm chart, this `fleet.yaml` still specifies a `helm` tag. Do we know if, say, specifying a `kustomize` tag with a directory would result in the same behaviour? My understanding is that it should, but that might be something we would like to validate as well.

weyfonk (approved):

PR #1776, opened 2023-09-12
Incorrect agent's DebugLevel when settings propagation is set

### LINKED ISSUE #1694: DebugLevel Is Never Updated

### Current Behavior

In [manifest.go](https://github.com/rancher/fleet/blob/e6946481ce04bc36bc89ded29eeccb6976a83a4b/internal/cmd/controller/agent/manifest.go#L94) the agent DebugLevel is initialized as 0 but never updated to match what the controller debug level is set as when propagateDebug is set to true. 

### Expected Behavior

when propagateDebug is set to true the agent manifest should inherit the debug level of the controller 

### Steps To Reproduce

1. in the fleet controller values file set the following values
  - debug=true
  - debugLevel=5
  - propagateDebugSettingsToAgents=true
2.  update the fleet helm chart using those values
3. check an agent managed by the controller and the debugLevel will be set to 0 rather than the value in the fleet controller

### Environment

```markdown
- Architecture: x86
- Fleet Version: 0.7.0
- Cluster:
  - Provider: tkg
  - Options:
  - Kubernetes Version: 1.22
```

### Logs

```markdown
fleet-controller yaml
containers:                                                                                                                                                          │
│       - command:                                                                                                                                                           │
│         - fleetcontroller                                                                                                                                                  │
│         - --disable-bootstrap                                                                                                                                              │
│         - --debug                                                                                                                                                          │
│         - --debug-level                                                                                                                                                    │
│         - "4"       
fleet-agent yaml
containers:                                                                                                                                                          │
│       - command:                                                                                                                                                           │
│         - fleetagent                                                                                                                                                       │
│         - --debug                                                                                                                                                          │
│         - --debug-level                                                                                                                                                    │
│         - "0"                                                                                                                                                              │
│         env:
```

### Anything else?

_No response_

aruiz14:
Fixes #1694

The variable used to populate the agent's `--debug-level` flag was being set after starting the controller, which is causing the value not to be ready by the time it's accessed.

I've also taken the chance to move the handling of the environment variable that controls this propagation to the root command.

# Testing

I confirmed the issue by installing Fleet controller with debug enabled, then checking the `fleet-agent` deployment that gets automatically installed:

```
❯ kubectl -n cattle-fleet-system get deploy/fleet-agent -o=jsonpath='{.spec.template.spec.containers[0].command}'
["fleetagent","--debug","--debug-level","0"]
```

After my changes, we can see that the debug level flag matches the value used in the controller:
```
❯ kubectl -n cattle-fleet-system get deploy/fleet-agent -o=jsonpath='{.spec.template.spec.containers[0].command}'
["fleetagent","--debug","--debug-level","99"]
```

### DIFF
--- a/internal/cmd/controller/agent/manifest.go
+++ b/internal/cmd/controller/agent/manifest.go
@@ -1,7 +1,6 @@
 package agent
 
 import (
-	"os"
 	"path"
 	"strconv"
 	"strings"
@@ -20,7 +19,8 @@ import (
 )
 
 var (
-	DebugLevel = 0
+	DebugEnabled bool
+	DebugLevel   = 0
 )
 
 const (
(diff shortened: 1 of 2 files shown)


weyfonk (approved):
LGTM

Issue #1777, opened 2023-09-12
Bump to wrangler/v2


manno:
Mocks break, see https://github.com/rancher/rancher/pull/41817

aruiz14:
related: https://github.com/rancher/gitjob/issues/315

aruiz14:
@manno I'm afraid that we are still blocked here. I tried upgrading the import paths to `wrangler/v2` but there is still no version of `lasso` using `v2`. It's not possible to keep both versions of `wrangler` because `steve`'s `/pkg/aggregation` expects a `v1` handler.

aruiz14:
Not blocked anymore, since we tackled: rancher/gitjob#339

PR #1785, opened 2023-09-13
Update to wrangler/v2

aruiz14:
Part of #1777

Updates `rancher/wrangler` after it was refactored to use generics (see https://github.com/rancher/rancher/pull/41817).
Mocks interfaces within the wrangler module are no longer needed and have been replaced by `generic/fake` mocks.

There is an upcoming module rename in `wrangler` (`github.com/rancher/wrangler -> github.com/rancher/wrangler/v2`), but there is no tagged version yet. I would wait until that happens before marking this draft as ready as well.

Meanwhile, it should be good for testing the new library version on CI.


internal/cmd/agent/register/register.go
```diff
@@ -105,6 +105,13 @@ func tryRegister(ctx context.Context, namespace, clusterID string, cfg *rest.Con
 	}, nil
 }
 
+// coreInterface is a subset of corecontrollers.Interface
+type coreInterface interface {
```
aruiz14, line 109: I changed this to avoid using a full `Interface`, only specifying the APIs that are used. This makes testing easier, as seen below (avoiding to generate the `Interface` mock)

manno (approved):

PR #1786, opened 2023-09-13
Add support for spec field identifying KUBECONFIG namespace for imported clusters

aiyengar2:
Fix https://github.com/rancher/fleet/issues/1787

Introduce the `.spec.kubeConfigSecretNamespace` annotation to facilitate specifying an alternate namespace for grabbing the KUBECONFIG of an imported cluster. This should only affect manager-based registration since it's only affecting the imported cluster workflow.

### Trade-off

Listed in issue

### Potential improvement

No regression possibilities. This is an opt-in feature.

### DIFF
--- a/internal/cmd/controller/controllers/cluster/import.go
+++ b/internal/cmd/controller/controllers/cluster/import.go
@@ -229,12 +229,19 @@ func (i *importHandler) importCluster(cluster *fleet.Cluster, status fleet.Clust
 		cluster.Spec.ClientID == "" {
 		return status, nil
 	}
-	secret, err := i.secrets.Get(cluster.Namespace, cluster.Spec.KubeConfigSecret)
+
+	kubeConfigSecretNamespace := cluster.Namespace
+	if cluster.Spec.KubeConfigSecretNamespace != "" {
+		kubeConfigSecretNamespace = cluster.Spec.KubeConfigSecretNamespace
+	}
+	logrus.Debugf("Cluster import for '%s/%s'. Getting kubeconfig from secret in namespace %s", cluster.Namespace, cluster.Name, kubeConfigSecretNamespace)
+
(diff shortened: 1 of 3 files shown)


internal/cmd/controller/controllers/cluster/import.go
```diff
@@ -229,12 +231,19 @@ func (i *importHandler) importCluster(cluster *fleet.Cluster, status fleet.Clust
 		cluster.Spec.ClientID == "" {
 		return status, nil
 	}
-	secret, err := i.secrets.Get(cluster.Namespace, cluster.Spec.KubeConfigSecret)
+
+	kubeConfigSecretNamespace := cluster.Namespace
+	if val, ok := cluster.Annotations[KubeConfigNamespaceAnnotation]; ok {
```
aruiz14, line 236: I'm not sure if using annotations is appropriate here. It probably deserves a new field in the spec, so it can be treated as any other configuration option (documentation, tests, etc.).

manno: I agree, `KubeConfigSecret` is part of the spec, it's strange to have the secret's namespace in an annotation.

aiyengar2: I'll modify it to be a spec field, but FWIW the logic for making it an annotation was to avoid making a breaking change.

The standard practice (if I understand correctly) is to introduce breaking changes to current CRD version (i.e. `v1alpha1`) as annotations, which will be converted to new spec fields in a newer CRD version (i.e. `v1alpha2` or `v1beta1`) as a spec field. That way, conversion webhooks can neatly handle the conversion of the CRD to a new view.

However, in this case this annotation is not a breaking change since, if unset, it will not be used, so it's fine to make this a spec field. Will make the corresponding change on the PR, but feel free to resolve this comment once you see it.

aruiz14 (approved):

PR #1793, opened 2023-09-18
Retry k3d image import in workflows

manno:
Refers to https://github.com/rancher/fleet/issues/1496

This is a workaround until [k3d import is fixed upstream](https://github.com/k3d-io/k3d/issues/1072).

### DIFF
--- a/.github/scripts/k3d-import-retry.sh
+++ b/.github/scripts/k3d-import-retry.sh
@@ -2,21 +2,13 @@
 
 set -euxo pipefail
 
-container=${1:-k3d-upstream}
-name=${2:-fleet}
-shift
-shift
-
-echo "Run k3d import with \"$*\" and retry if '$name' is not found in 'critcl images' output"
-
 i=0
-while ! ( docker exec "$container"-server-0 /bin/crictl images | grep -q "$name" ); do
+while k3d image import "$@" -m direct | grep -q "failed to import"; do
   i=$((i + 1))
   if (( i > 3 )); then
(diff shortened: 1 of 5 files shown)


.github/scripts/k3d-import-retry.sh
```diff
@@ -0,0 +1,20 @@
...
+
+container=${1:-k3d-upstream}
+name=${2:-fleet}
+shift
+shift
+
+echo "Run k3d import with \"$*\" and retry if '$name' is not found in 'critcl images' output"
+
+i=0
+while ! ( docker exec "$container"-server-0 /bin/crictl images | grep -q "$name" ); do
```
aruiz14, line 13: Why retry based on name match absence instead of checking every image individually?
`/bin/crictl inspecti "${image}" >/dev/null` may do the trick

.github/scripts/k3d-import-retry.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+set -euxo pipefail
```
aruiz14, line 3: Is `-x` intentional?

manno: Yes, script is not working in CI..

.github/scripts/k3d-import-retry.sh
manno: Interesting, but "crictl" output wasn't reliable.

aruiz14 (approved):

Issue #1799, opened 2023-09-22
Feature Request: Include nonResourceURLs Fleet-Agent's RBAC 


KevinJoiner:
### Is your feature request related to a problem?

When applying a Rancher GlobalRole resource that specifies nonResourceUrl set, the fleet-agent SA is denied since the SA’s RBAC only identifies `{"apiGroups": ["*"],"resources": [ "*"],"verbs": ["*"] }`
``` yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  annotations: (...)
  labels: (...)
  name: cattle-fleet-local-system-fleet-agent-role
rules:
- apiGroups:
  - '*'
  resources:
  - '*'
  verbs:
  - '*'
```

### Solution you'd like

This could be solved by updating the fleet-agent rules to include nonResourceURLs
``` yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  annotations: (...)
  labels: (...)
  name: cattle-fleet-local-system-fleet-agent-role
rules:
- apiGroups:
  - '*'
  resources:
  - '*'
  verbs:
  - '*'
- nonResourceURLs:
  - "*"
  verbs:
  - "*"
```

### Alternatives you've considered

The alternative is to do nothing and have Rancher bind the local fleet-agent to a new Role.

## Reproduction
- Start a new Rancher instance
- Create a git repo with the following GlobalRole
``` yaml
apiVersion: management.cattle.io/v3
kind: GlobalRole
metadata:
  name: my-admin-gr-non-resource-url
rules:
  - apiGroups:
      - "*"
    resources:
      - "*"
    verbs:
      - "*"
  - nonResourceURLs:
      - "*"
    verbs:
      - "*"
```
- Add a new GitRepo resource for your repo in the local cluster.
- Verify that the fleet-agent fails to deploy the CRD because it is blocked by the webhook for having insufficient permissions.

kkaempf:
Fixed in fleet-0.9.0-rc.3

PR #1806, opened 2023-09-25
Include nonResourceURLs Fleet-Agent's RBAC

manno:
Refers to https://github.com/rancher/fleet/issues/1799

### DIFF
--- a/charts/fleet-agent/templates/rbac.yaml
+++ b/charts/fleet-agent/templates/rbac.yaml
@@ -9,7 +9,10 @@ rules:
   - '*'
   verbs:
   - '*'
-
+- nonResourceURLs:
+  - "*"
+  verbs:
+  - "*"
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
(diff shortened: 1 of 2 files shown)


aruiz14 (approved):

charts/fleet-agent/templates/rbac.yaml
```diff
@@ -9,7 +9,10 @@ rules:
   - '*'
   verbs:
   - '*'
-
+- nonResourceURLs:
+  - "*"
+  verbs:
```
kkaempf, line 14: That's a duplicate from lines 10/11 ?!

kkaempf (changes requested):
I miss a testcase

charts/fleet-agent/templates/rbac.yaml
manno: The diff is pretty unreadable, but this is a new array element for `rules:`, so it comes out at

```
rules: [ 
  {"apiGroups": ['*'], "resources": ['*'], "verbs": ['*']}, 
  {"nonResourceURLs": ["*"], "verbs": ['*']}
 ],
```

kkaempf: 🙈 - github diffs again ...

kkaempf (approved):

Issue #1738, opened 2023-08-23
Helm Storage Backend With Caching


manno:
The fleet agent stores the history in the default helm storage, which is secrets. However the backend does not cache,  limiting the number of history items helps with performance, but as discussed we need caching.

This should fix #1465 and replace https://github.com/rancher/fleet/pull/1609

mattfarina:
For some history on history... the current lookup mechanism was designed for use with the Helm CLI which is not long running. Caching did not make sense for that case.

I would be curious how we keep the cache up to date. A watch of Helm secrets? Helm secrets are typed (i.e., `helm.sh/release.v1`) so they can be filtered from other secrets. Something else?

moio:
> For some history on history... the current lookup mechanism was designed for use with the Helm CLI which is not long running. Caching did not make sense for that case.

👍 

> I would be curious how we keep the cache up to date. A watch of Helm secrets? Helm secrets are typed (i.e., helm.sh/release.v1) so they can be filtered from other secrets. Something else?

Fleet (specifically `fleet-agent`) originally had no caching whatsoever - all reads just go through Helm, and Helm fetches all data from the Kubernetes API client - no Informers or similar caching mechanisms were in place. In #1485 I added a cache to the whole `Storage.History()` results, which is expected to not always be coherent (cache results expire after 5 minutes) - that can arguably be described as a suboptimal workaround. Details here: https://github.com/rancher/fleet/pull/1485/commits/8b0d9b7031570b7f6c0b8b5e969e34542339d181

A better strategy could indeed be Informer-based. Or not saving history in Kubernetes/etcd altogether - which opens up the interesting question what other DB could be used.

BTW: one additional problem we have is that Helm's Storage API fetches more than we actually need, and then we filter results in-memory. It'd help if the new backend, however it looks like, could deal with this natively. Details:

https://github.com/rancher/fleet/blob/26c247a999decdda7e639db9a0cef6e36f8ec5a2/internal/helmdeployer/deployer.go#L538-L555

PR #1809, opened 2023-09-26
Helm Storage Backend With Caching

raulcabello:
Helm is internally listing all secrets each time we retrieve information about the release. We were using the helm storage secrets driver, which is meant to be used as a cli. It uses a k8s client to get/list/create.. secrets, and it doesn't use any cache (see implementation [here](https://github.com/helm/helm/blob/v3.12.3/pkg/action/lazyclient.go#L60-L125))

This PR adds a new helm storage driver that uses the `SecretCache` from wrangler for getting and listing secrets. All other operations uses a k8s client like the helm secrets driver does.  `SecretCache` from wrangler is internally using a `ShareInformer` for creating the cache.

This PR undoes the cache implemented in https://github.com/rancher/fleet/commit/8b0d9b7031570b7f6c0b8b5e969e34542339d181

The helm globalCfg is created with an [empty namespace](https://github.com/rancher/fleet/blob/v0.8.0/internal/helmdeployer/deployer.go#L116), which means that it will list secrets from all namespaces. We are using it in some [places](https://github.com/rancher/fleet/blob/v0.8.0/internal/helmdeployer/deployer.go#L598) for retrieving releases information. That means we are unnecessarily listing secrets from all namespaces, and that could cause issues if two releases have the same name in two different namespaces. This will be fixed in a separate PR. 

refers to https://github.com/rancher/fleet/issues/1738

### DIFF
--- a/integrationtests/agent/suite_test.go
+++ b/integrationtests/agent/suite_test.go
@@ -162,6 +162,9 @@ func registerBundleDeploymentController(cfg *rest.Config, namespace string, look
 	err = factory.Start(ctx, 50)
 	Expect(err).ToNot(HaveOccurred())
 
+	err = coreFactory.Start(ctx, 50)
+	Expect(err).ToNot(HaveOccurred())
+
 	return factory.Fleet().V1alpha1().BundleDeployment()
 }
 
--- a/internal/helmdeployer/helmcache/secret.go
+++ b/internal/helmdeployer/helmcache/secret.go
@@ -23,7 +23,7 @@ func NewSecretClient(cache corecontrollers.SecretCache, client kubernetes.Interf
 	return &SecretClient{cache, client, namespace}
 }
 
-func (s *SecretClient) Create(ctx context.Context, secret *v1.Secret, opts metav1.CreateOptions) (result *v1.Secret, err error) {
+func (s *SecretClient) Create(ctx context.Context, secret *v1.Secret, opts metav1.CreateOptions) (*v1.Secret, error) {
 	return s.client.CoreV1().Secrets(s.namespace).Create(ctx, secret, opts)
 }
 
--- a/internal/helmdeployer/helmcache/secret.go
+++ b/internal/helmdeployer/helmcache/secret.go
@@ -13,6 +13,8 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+// SecretClient implements methods to handle secrets. Get and list will be retrieved from the wrangler cache, the other calls
+// will be make to the Kubernetes API server.
 type SecretClient struct {
 	cache     corecontrollers.SecretCache
 	client    kubernetes.Interface
@@ -23,26 +25,32 @@ func NewSecretClient(cache corecontrollers.SecretCache, client kubernetes.Interf
 	return &SecretClient{cache, client, namespace}
 }
 
+// Create creates a secret using a k8s client that calls the Kubernetes API server
 func (s *SecretClient) Create(ctx context.Context, secret *v1.Secret, opts metav1.CreateOptions) (*v1.Secret, error) {
 	return s.client.CoreV1().Secrets(s.namespace).Create(ctx, secret, opts)
 }
 
+// Update updates a secret using a k8s client that calls the Kubernetes API server
 func (s *SecretClient) Update(ctx context.Context, secret *v1.Secret, opts metav1.UpdateOptions) (*v1.Secret, error) {
 	return s.client.CoreV1().Secrets(s.namespace).Update(ctx, secret, opts)
 }
 
+// Delete deletes a secret using a k8s client that calls the Kubernetes API server
 func (s *SecretClient) Delete(ctx context.Context, name string, opts metav1.DeleteOptions) error {
 	return s.client.CoreV1().Secrets(s.namespace).Delete(ctx, name, opts)
 }
 
+// DeleteCollection deletes a secret collection using a k8s client that calls the Kubernetes API server
 func (s *SecretClient) DeleteCollection(ctx context.Context, opts metav1.DeleteOptions, listOpts metav1.ListOptions) error {
 	return s.client.CoreV1().Secrets(s.namespace).DeleteCollection(ctx, opts, listOpts)
 }
 
+// Get gets a secret from the wrangler cache.
 func (s *SecretClient) Get(ctx context.Context, name string, opts metav1.GetOptions) (*v1.Secret, error) {
 	return s.cache.Get(s.namespace, name)
 }
 
+// List lists secrets from the wrangler cache.
 func (s *SecretClient) List(ctx context.Context, opts metav1.ListOptions) (*v1.SecretList, error) {
 	labels, err := labels.Parse(opts.LabelSelector)
 	if err != nil {
@@ -63,14 +71,17 @@ func (s *SecretClient) List(ctx context.Context, opts metav1.ListOptions) (*v1.S
 	}, nil
 }
 
+// Watch watches a secret using a k8s client that calls the Kubernetes API server
 func (s *SecretClient) Watch(ctx context.Context, opts metav1.ListOptions) (watch.Interface, error) {
 	return s.client.CoreV1().Secrets(s.namespace).Watch(ctx, opts)
 }
 
+// Patch patches a secret using a k8s client that calls the Kubernetes API server
 func (s *SecretClient) Patch(ctx context.Context, name string, pt types.PatchType, data []byte, opts metav1.PatchOptions, subresources ...string) (*v1.Secret, error) {
 	return s.client.CoreV1().Secrets(s.namespace).Patch(ctx, name, pt, data, opts, subresources...)
 }
 
+// Apply applies a secret using a k8s client that calls the Kubernetes API server
 func (s *SecretClient) Apply(ctx context.Context, secretConfiguration *applycorev1.SecretApplyConfiguration, opts metav1.ApplyOptions) (*v1.Secret, error) {
 	return s.client.CoreV1().Secrets(s.namespace).Apply(ctx, secretConfiguration, opts)
 }
--- a/internal/helmdeployer/helmcache/secret.go
+++ b/internal/helmdeployer/helmcache/secret.go
@@ -1,0 +1,76 @@
+package helmcache
+
+import (
+	"context"
+
+	corecontrollers "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
+	v1 "k8s.io/api/core/v1"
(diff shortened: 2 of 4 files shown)


internal/helmdeployer/helmcache/secret.go
```diff
@@ -0,0 +1,76 @@
...
+	client    kubernetes.Interface
+	namespace string
+}
+
+func NewSecretClient(cache corecontrollers.SecretCache, client kubernetes.Interface, namespace string) *SecretClient {
+	return &SecretClient{cache, client, namespace}
+}
+
+func (s *SecretClient) Create(ctx context.Context, secret *v1.Secret, opts metav1.CreateOptions) (result *v1.Secret, err error) {
+	return s.client.CoreV1().Secrets(s.namespace).Create(ctx, secret, opts)
```
weyfonk, line 27: Are we sure that write operations (`Create`, `Update`, `Delete`, `DeleteCollection`, `Patch`, `Apply`?) write to the cache too? Is it worth testing here, or is this handled in Wrangler internals?
How the K8s client updates the cache is not clear to me here.

For instance:
* unit tests for the `List` operation already assume that the requested entries are in the cache
* those for `Create` don't expect any call to the cache
* `Create`/`Delete`, etc. test cases could check the state of the cache before/after the operation

Are these concerns relevant here, or should such behaviour be tested further up (eg. integration tests)?

internal/helmdeployer/helmcache/secret_test.go
```diff
@@ -0,0 +1,178 @@
...
+	secretPatch := corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      secretName,
+			Namespace: secretNamespace,
+		},
+		Data: map[string][]byte{"test": []byte("content")},
+	}
+	client := fake.NewSimpleClientset(&secret)
+	secretClient := NewSecretClient(nil, client, secretNamespace)
+	patch := []byte(`{"data":{"test":"Y29udGVudA=="}}`)
```
weyfonk, line 143: nit:
```suggestion
	patch := []byte(`{"data":{"test":"Y29udGVudA=="}}`) // "content", base64-encoded
```

internal/helmdeployer/helmcache/secret.go
```diff
@@ -0,0 +1,76 @@
...
+	cache     corecontrollers.SecretCache
+	client    kubernetes.Interface
+	namespace string
+}
+
+func NewSecretClient(cache corecontrollers.SecretCache, client kubernetes.Interface, namespace string) *SecretClient {
+	return &SecretClient{cache, client, namespace}
+}
+
+func (s *SecretClient) Create(ctx context.Context, secret *v1.Secret, opts metav1.CreateOptions) (result *v1.Secret, err error) {
```
aruiz14, line 26: Suggestion to @weyfonk's suggestion :) 
```suggestion
func (s *SecretClient) Create(ctx context.Context, secret *v1.Secret, opts metav1.CreateOptions) (*v1.Secret, error) {
```

weyfonk: Oops! Yes, that's what it should have looked like, for some reason I missed `err` :upside_down_face:

internal/helmdeployer/helmcache/secret.go
```diff
@@ -0,0 +1,76 @@
...
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/watch"
+	applycorev1 "k8s.io/client-go/applyconfigurations/core/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+type SecretClient struct {
```
aruiz14, line 16: If `SecretClient` is meant to implement [Kubernetes' SecretsInterface](https://pkg.go.dev/k8s.io/client-go/kubernetes/typed/core/v1#SecretInterface), maybe it's worth adding a compile-time assertion:
```
var _ v1.SecretInterface = (*SecretClient)(nil)
```

internal/helmdeployer/helmcache/secret.go
```diff
@@ -0,0 +1,76 @@
...
+	cache     corecontrollers.SecretCache
```
aruiz14, line 17: Is there any chance to use [`client-go`'s cache](https://pkg.go.dev/k8s.io/client-go/tools/cache) instead of this one? (in order to not mix clients).
On the other hand, I do agree this implementation is pretty clean.

internal/helmdeployer/helmcache/secret.go
```diff
@@ -0,0 +1,76 @@
...
+	client    kubernetes.Interface
+	namespace string
+}
+
```
manno, line 21: This is implementing https://pkg.go.dev/helm.sh/helm/v3@v3.12.3/pkg/storage/driver#Driver right? Can we add something make it obvious and get early errors if we break the interface?

```
var _ storage.Driver = &SecretClient{}
```

manno: Oops, it implements https://github.com/helm/helm/blob/v3.12.3/pkg/storage/driver/secrets.go#L50

internal/helmdeployer/helmcache/secret.go
manno: From my memory (and the PR description) we only want to cache Get/List. Maybe add function level comments to make it clear that's intentional.

internal/helmdeployer/helmcache/secret.go
raulcabello: I don't see the benefit of adding this. It would fail [here](https://github.com/raulcabello/fleet/blob/helm-cache/internal/helmdeployer/deployer.go#L774) as it is expecting a SecretInterface

internal/helmdeployer/helmcache/secret.go
raulcabello: Wrangler and lasso are using client-go `ShareInformer` for the cache. I didn't want to create another client-go informer as the wrangler one is already created, and it is used in more places.

internal/helmdeployer/helmcache/secret.go
raulcabello: oh true. I copied and pasted it from the helm's code https://github.com/helm/helm/blob/v3.12.3/pkg/action/lazyclient.go#L64 . Not sure why it is there in helm, but I'll remove it.

internal/helmdeployer/helmcache/secret.go
aruiz14: That's my point, to add this assertion with whatever interface is expected to be implemented from the users.

internal/helmdeployer/helmcache/secret.go
raulcabello: it would break [here](https://github.com/raulcabello/fleet/blob/helm-cache/internal/helmdeployer/deployer.go#L774) if we break the interface. I don't think we need this extra check or https://github.com/rancher/fleet/pull/1809#discussion_r1338480259 . In my opinion that would be against the implicit interfaces in go

integrationtests/agent/suite_test.go
```diff
@@ -162,6 +162,9 @@ func registerBundleDeploymentController(cfg *rest.Config, namespace string, look
 	err = factory.Start(ctx, 50)
 	Expect(err).ToNot(HaveOccurred())
 
+	err = coreFactory.Start(ctx, 50)
```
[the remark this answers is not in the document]
weyfonk: Interesting that we hadn't detected this before then :thinking: 
Thanks for clarifying :)

internal/helmdeployer/helmcache/secret.go
manno: Yeah true, since the struct is only created to implement the one interface I thought it might be useful. But as you point out, it will report errors when we use it.  I just read https://go.dev/doc/effective_go#interfaces and it says:

>  The appearance of the blank identifier in this construct indicates that the declaration exists only for the type checking, not to create a variable. Don't do this for every type that satisfies an interface, though. By convention, such declarations are only used when there are no static conversions already present in the code, which is a rare event.

internal/helmdeployer/helmcache/secret.go
raulcabello: Added comments in https://github.com/rancher/fleet/pull/1809/commits/d9161ac38876582549e13a759ca8406691d7efae 

> How the K8s client updates the cache is not clear to me here.

This is handled by the wrangler `corev1controller.SecretCache`. You can read [wrangler](https://github.com/rancher/wrangler) and [lasso](https://github.com/rancher/lasso) docs to understand how they work. Also you can read about client-go [SharedInformers](https://aly.arriqaaq.com/kubernetes-informers/), which is what wrangler is using.

I don't think this is the right place to explain how cache works in wrangler, as we are using it extensively throughout the fleet code. In order to find where we are using wrangler caches just do a global search for `SecretCache`, `BundleCache`, `BundleDeploymentCache`, `ServiceAccountCache`, ...

internal/helmdeployer/helmcache/secret.go
aruiz14: Resolving after reading: https://github.com/rancher/fleet/pull/1809/files#r1338677031

moio:
> The helm globalCfg is created with an [empty namespace](https://github.com/rancher/fleet/blob/v0.8.0/internal/helmdeployer/deployer.go#L116), which means that it will list secrets from all namespaces. We are using it in some [places](https://github.com/rancher/fleet/blob/v0.8.0/internal/helmdeployer/deployer.go#L598) for retrieving releases information. That means we are unnecessarily listing secrets from all namespaces, and that could cause issues if two releases have the same name in two different namespaces. This will be fixed in a separate PR.

@raulcabello: is work for this separate PR already ongoing? I would like the main affected user to try work you've done here, but I'd be hesitant delivering a version with a known potential issue.

raulcabello:
@moio not yet. It will be done in https://github.com/rancher/fleet/issues/1812 
I wouldn't wait for it, and I would tell the user to try this now. We are now always listing secrets from the cache. So even if we list secrets from all namespaces (when we shouldn't) they will be retrieved from the cache.

shfy70:
Gentlemen,

    Kindly let us know which Rancher version (2.7.8? or maybe 2.8.0? I notice the matrix is always 2-3 weeks behind the released Rancher version here on github) will fix this fleet-agent issue highlighted in my earlier post, many thanks.

moio:
2.8.0 will have this fix.

A backport to the 2.7.x line has not been decided yet - that might depend on customer feedback.

PR #1810, opened 2023-09-27
Fix patch_crd_descriptions.sh when yq is not installed

raulcabello:
Docker build output was appended to the `crds.yaml` when `yq` is not installed. 

This PR discards the output of the docker build command, so it doesn't get appended to the `crds.yaml`

### DIFF
--- a/cmd/codegen/hack/patch_crd_descriptions.sh
+++ b/cmd/codegen/hack/patch_crd_descriptions.sh
@@ -18,7 +18,7 @@ run_yq() {
     image="fleet-codegen-hack:yq"
     log "yq (from https://github.com/kislyuk/yq) is missing, building a helper docker image ($image)..."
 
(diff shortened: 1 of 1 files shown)


cmd/codegen/hack/patch_crd_descriptions.sh
```diff
@@ -18,7 +18,7 @@ run_yq() {
     image="fleet-codegen-hack:yq"
     log "yq (from https://github.com/kislyuk/yq) is missing, building a helper docker image ($image)..."
 
-    docker build -t $image - << EOF
+    docker build -t $image - << EOF >/dev/null
```
aruiz14, line 21: May I suggest redirecting to stderr instead?
```suggestion
    docker build -t $image - >&2 << EOF
```

aruiz14 (approved):

PR #1840, opened 2023-10-06
Reduce replica count in rancher integration tests

manno:


weyfonk (approved):
Worth trying :thinking:

PR #1847, opened 2023-10-09
Add tmate debugging info to Nightly E2E workflow

weyfonk:
This should help troubleshoot failures happening in tests requiring secrets, which can therefore not be run locally nor against Fleet forks.


aruiz14 (approved):
LGTM!

Issue #1842, opened 2023-10-06
[SURE-6990] fleet-controller consuming high cpu


kkaempf:
### Issue description:
fleet-controller is consuming 5GHz plus of CPU. customer is using AzureDevOps with 300+ git repos

### Business impact:
fleet gets unstable, or even the node for rancher management

### Troubleshooting steps:
scaling nodes 4x

### Repro steps:
no repro steps yet

### Workaround:
Is workararound available and implemented? yes
What is the workaround: scaling vertically the nodes

### Actual behavior:
fleet-controller consuming 5GHz+ of CPU

skaven81:
We are observing this same problem.  It seems to occur when there are any bundles out of sync.  Once all the Bundles are back in sync, the CPU utilization calms back down. But if anything goes out of sync, fleet-controller will start using all the CPU it can get its hands on.

skaven81:
I was able to capture a core file from the fleet-controller process using Delve.  If somebody can tell me what to do with such a file (or if there are other Delve commands I can run to extract the useful information for debugging this) please let  me know.  I cannot attach the core file itself.

skaven81:
If anybody else sees this and wants to do the same, here's how I attached to it:

First get the container ID of fleet-controller (`8021a57ab5de` in this example)
```
$ docker ps | grep fleet-controller
8021a57ab5de   fdcac788e741                                                       "fleetcontroller --d…"   3 minutes ago   Up 3 minutes             k8s_fleet-controller_fleet-controller-5649d6545-fql99_cattle-fleet-system_31ae6828-7972-4893-b278-d7cf4113b245_0
663317e93319   docker-registry.qualcomm.com/rancher/mirrored-pause:3.6            "/pause"                 3 minutes ago   Up 3 minutes             k8s_POD_fleet-controller-5649d6545-fql99_cattle-fleet-system_31ae6828-7972-4893-b278-d7cf4113b245_0
```

Then attach to it with a container that has the `dlv` binary in it:
```
$ docker run -it --rm --privileged --pid=container:8021a57ab5de your-registry.example.com/org/delve:latest
```

Once inside, the fleet-controller is running as PID 1 and you can attach to it.  You may need to temporarily set this to get the attach to work:
```
$ echo 1 | sudo tee /proc/sys/kernel/yama/ptrace_scope
```

PR #1857, opened 2023-10-13
Use index when listing BundleDeployments by Bundle

aruiz14:
Part of #1842

These changes add an indexer for BundleDeployments' Informer, which will build an index of BundleDeployment objects based on the Bundle that generated them (uniquely identified by their namespace + name). This provides an optimization over constantly using `List` with multiple label selectors.

This will be also back-ported to 0.7 and 0.8 to provide a fix for the original reporters.

## Test

Unit tests have been adapted to match the new code.

```shell
go test ./internal/cmd/controller/target
```

### Trade-off

No expected side effects despite a marginal memory usage increase due to keeping the new index in memory.

### Potential improvement

CPU usage drop proportional to the number of bundles and downstream clusters.

### DIFF
--- a/internal/cmd/controller/target/target.go
+++ b/internal/cmd/controller/target/target.go
@@ -8,6 +8,7 @@ package target
 import (
 	"bytes"
 	"fmt"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
@@ -45,6 +46,7 @@ var (
 const (
 	maxTemplateRecursionDepth = 50
 	clusterLabelPrefix        = "global.fleet.clusterLabels."
+	byBundleIndexerName       = "fleet.byBundle"
 )
 
 type Manager struct {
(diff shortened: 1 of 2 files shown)


internal/cmd/controller/target/target.go
```diff
@@ -193,15 +205,13 @@ func (m *Manager) BundlesForCluster(cluster *fleet.Cluster) (bundlesToRefresh, b
 }
 
 func (m *Manager) GetBundleDeploymentsForBundleInCluster(bundle *fleet.Bundle, cluster *fleet.Cluster) (result []*fleet.BundleDeployment, err error) {
```
weyfonk, line 207: In case these changes were not enough, I wonder if applying similar changes eg. [here](https://github.com/rancher/fleet/blob/master/internal/cmd/controller/controllers/git/git.go#L525) could help reduce CPU load.

aruiz14: Nice point Tan. That code surely benefit from using a similar approach, although it would need a different index, based on `GitRepo` instead of `Bundles`. I can create a new issue for that, and add it separately. Do you agree?

weyfonk: Of course, this can be handled via a separate PR. I think it is relevant to the same issue as this one (#1842), but no strong opinions here.

raulcabello (approved):
lgtm

Issue #1845, opened 2023-10-05
[BUG] gitRepo resources aren't applying to clusters when changing fleet workspace


slickwarren:
**Rancher Server Setup**
- Rancher version: v2.8-head (fc12e7d)
- Installation option (Docker install/Helm Chart): rke2
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): v1.27.6+rke2r1
- Proxy/Cert Details: valid certs

**Information about the Cluster**
- Kubernetes version: v1.27.6 (rke2, k3s, rke1)
- Cluster Type (Local/Downstream): downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): aws

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom)
  - If custom, define the set of permissions: tested with admin

**Describe the bug**

when changing a cluster's workspace, the git-repo resources for fleet aren't applying to the clusters.

**To Reproduce**

* create a cluster (by default, deploys to fleet-default workspace)
* create a new workspace
* in the new workspace, create a git-repo
* transfer the cluster to the new workspace

**Result**
resources from fleet's new workspace aren't deploying. The cluster count remains at 0

**Expected Result**

resources from the gitRepo in the user added workspace should apply to any clusters that are matched. 

**Screenshots**

**Additional context**

adding a git-repo and switching the cluster back to fleet-default results in the expected behavior
tested by using the 'all clusters in workspace' setting and manually selecting a cluster in the workspace, but neither is working

possibly affected by https://github.com/rancher/rancher/issues/43078 as this was found on the same setup

slickwarren:
tested on a released version of rancher, 2.7.8, with a hardened local cluster and am experiencing the same error there:
```
[...]

PR #1875, opened 2023-10-17
Add SecurityContext to job container

raulcabello:
Make sure the `gitjob` pod and the job init container runs in an unprivileged `securityContext`. The following `securityContext` has been added:
```
securityContext:
  allowPrivilegeEscalation: false
  readOnlyRootFilesystem: true
  privileged: false
  runAsNonRoot: true
  seccompProfile:
    type: RuntimeDefault
  capabilities:
    drop:
      - ALL
```

`emptyDir` volume is mounted in the ` /tmp` dir in order to prevent issues when creating temporary files as `readOnlyRootFilesystem` is set to true.

`emptyDir` volume is mounted in the ` /home/fleet-ci` dir in order to prevent issue when creating helm config in the home directory.

refers to https://github.com/rancher/fleet/issues/1845

### DIFF
--- a/internal/cmd/controller/controllers/git/git.go
+++ b/internal/cmd/controller/controllers/git/git.go
@@ -38,6 +38,8 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+const fleetHomeDir = "/fleet-home"
+
 var (
 	two = int32(2)
 )
@@ -652,7 +654,7 @@ func volumes(
 		},
 		{
 			Name:      emptyDirHomeVolumeName,
-			MountPath: "/home/fleet-apply",
+			MountPath: fleetHomeDir,
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


internal/cmd/controller/controllers/git/git.go
```diff
@@ -611,13 +627,33 @@ func volumes(
...
+			Name:      configVolumeName,
 			MountPath: "/run/config",
 		},
+		{
+			Name:      emptyDirTmpVolumeName,
+			MountPath: "/tmp",
+		},
+		{
+			Name:      emptyDirHomeVolumeName,
+			MountPath: "/home/fleet-apply",
```
aruiz14, line 655: Discussed offline: I believe we should not assume the user running the container will configure `/home/fleet-aply` as the home. Better to explicitly set the`HOME` env. variable as well

weyfonk (dismissed):

internal/cmd/controller/controllers/git/git.go
raulcabello: added in https://github.com/rancher/fleet/pull/1875/commits/73b3e1ae0f97c2a6fcb869009ffa9535739dda87

aruiz14 (approved):

manno (approved):

PR #1885, opened 2023-10-20
Move cleanup to a different container

raulcabello:
Move the cleanup controller and the code that's cleaning orphan `contents`,  `bundles` and `imageScans` to a different binary.

Run the fleet-cleanup in a different container

### DIFF
--- a/internal/cmd/controller/cleanup/root.go
+++ b/internal/cmd/controller/cleanup/root.go
@@ -1,0 +1,38 @@
+package cleanup
+
+import (
+	"fmt"
+
+	command "github.com/rancher/fleet/internal/cmd"
+	"github.com/rancher/fleet/pkg/version"
+	"github.com/spf13/cobra"
+)
+
+type CleanUp struct {
+	command.DebugConfig
+	Kubeconfig string `usage:"kubeconfig file"`
+	Namespace  string `usage:"namespace to watch" env:"NAMESPACE"`
+}
+
+func (c *CleanUp) Run(cmd *cobra.Command, args []string) error {
+	if err := c.SetupDebug(); err != nil {
+		return fmt.Errorf("failed to setup debug logging: %w", err)
+	}
+
+	if c.Namespace == "" {
+		return fmt.Errorf("--namespace or env NAMESPACE is required to be set")
+	}
+	if err := start(cmd.Context(), c.Kubeconfig, c.Namespace); err != nil {
+		return err
+	}
+	<-cmd.Context().Done()
+
+	return nil
+}
+
+func App() *cobra.Command {
+	return command.Command(&CleanUp{}, cobra.Command{
+		Version: version.FriendlyVersion(),
+		Use:     "cleanup",
+	})
+}
--- a/internal/cmd/controller/cleanup/root.go
+++ b/internal/cmd/controller/cleanup/root.go
@@ -17,12 +17,7 @@ func (c *CleanUp) Run(cmd *cobra.Command, args []string) error {
 	if c.Namespace == "" {
 		return fmt.Errorf("--namespace or env NAMESPACE is required to be set")
 	}
-	if err := start(cmd.Context(), c.Kubeconfig, c.Namespace); err != nil {
-		return err
-	}
-	<-cmd.Context().Done()
-
-	return nil
+	return start(cmd.Context(), c.Kubeconfig, c.Namespace)
 }
 
 func App() *cobra.Command {
--- a/internal/cmd/controller/cleanup/root.go
+++ b/internal/cmd/controller/cleanup/root.go
@@ -9,16 +9,11 @@ import (
 )
 
 type CleanUp struct {
-	command.DebugConfig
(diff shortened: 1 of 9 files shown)


manno (approved):

internal/cmd/controller/cleanup/root.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"fmt"
+
+	command "github.com/rancher/fleet/internal/cmd"
+	"github.com/rancher/fleet/pkg/version"
+	"github.com/spf13/cobra"
+)
+
+var (
+	debugConfig command.DebugConfig
```
aruiz14, line 12: This should not be needed, since the root command already configures it

go.sum
```diff
@@ -820,6 +821,10 @@ github.com/rancher/helm/v3 v3.12.3-fleet1 h1:CCvZttHJChIG097BZIMYHX9rcuPFhgjmmnc
 github.com/rancher/helm/v3 v3.12.3-fleet1/go.mod h1:KPKQiX9IP5HX7o5YnnhViMnNuKiL/lJBVQ47GHe1R0k=
 github.com/rancher/lasso v0.0.0-20230830164424-d684fdeb6f29 h1:+kige/h8/LnzWgPjB5NUIHz/pWiW/lFpqcTUkN5uulY=
 github.com/rancher/lasso v0.0.0-20230830164424-d684fdeb6f29/go.mod h1:kgk9kJVMj9FIrrXU0iyM6u/9Je4bEjPImqswkTVaKsQ=
+github.com/rancher/wrangler v1.1.1 h1:wmqUwqc2M7ADfXnBCJTFkTB5ZREWpD78rnZMzmxwMvM=
+github.com/rancher/wrangler v1.1.1/go.mod h1:ioVbKupzcBOdzsl55MvEDN0R1wdGggj8iNCYGFI5JvM=
+github.com/rancher/wrangler-cli v0.0.0-20220624114648-479c5692ba22 h1:ADMwgJyVwmLXJBSm/nNobB1XGSmFCTA+TY/otxgIPu4=
+github.com/rancher/wrangler-cli v0.0.0-20220624114648-479c5692ba22/go.mod h1:vyO9SU60oplNFa5ZqoEAFWmYKgj2F6remdy8p6H0SgI=
```
aruiz14, line 827: There must be something wrong. Please beware that we have removed `wrangler-cli` and `wrangler/v1` recently

aruiz14 (changes requested):
Please review the comments before merging

internal/cmd/controller/cleanup/root.go
manno: That's because we merged something else first, this PR is in conflict.

raulcabello: @manno is right. It has conflicts with https://github.com/rancher/fleet/pull/1899 which is the PR that changed the debug approach. I'm fixing the conflicts

go.sum
raulcabello: this was gone after fixing the conflicts with https://github.com/rancher/fleet/pull/1899

internal/cmd/controller/cleanup/root.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"fmt"
+
+	command "github.com/rancher/fleet/internal/cmd"
+	"github.com/rancher/fleet/pkg/version"
+	"github.com/spf13/cobra"
+)
+
+type CleanUp struct {
+	command.DebugConfig
```
aruiz14, line 12: Doesn't `FleetManager` already includes a `DebugConfig`? Since it uses `PersistentPreRun`, I believe this is not needed.

raulcabello: you're right! It's not needed to add it again in the subcomand. Removed in https://github.com/rancher/fleet/pull/1885/commits/8b1d535b72856e7501dd25512d867efcc9a5fe50

internal/cmd/controller/cleanup/root.go
```diff
@@ -0,0 +1,33 @@
...
+func (c *CleanUp) Run(cmd *cobra.Command, args []string) error {
+	if c.Namespace == "" {
+		return fmt.Errorf("--namespace or env NAMESPACE is required to be set")
+	}
+	if err := start(cmd.Context(), c.Kubeconfig, c.Namespace); err != nil {
+		return err
+	}
+	<-cmd.Context().Done()
+
+	return nil
```
aruiz14, line 25: Why is this needed? `RunOrDie` used in `start` is blocking, isn't it?
https://github.com/rancher/wrangler/blob/v2.1.2/pkg/leader/leader.go#L29

```suggestion
	return start(cmd.Context(), c.Kubeconfig, c.Namespace)
```

raulcabello: I must have added it because I copied pasted from the [agent](https://github.com/rancher/fleet/blob/master/internal/cmd/agent/root.go#L48)

aruiz14 (approved):

PR #1900, opened 2023-10-25
Test ability to use template functions in fleet.yaml

aruiz14:
Adapt test expectations to rancher/fleet-test-data#10


aruiz14:
The latest commit here must be removed before merging, after https://github.com/rancher/fleet-test-data/pull/10 is merged.

Issue #1815, opened 2023-09-28
Replace Wrangler's CRD generator


aruiz14:
Switch from wrangler's crd generator to controller-tools.

This was originally part of #1495, but this migration is not trivial and the ultimate goal is to enable "kubectl describe" to show our docs.

Controller-runtime only allows customizing the produced schema by using [markers syntax](https://pkg.go.dev/sigs.k8s.io/controller-tools/pkg/markers). The biggest drawback is losing the ability to modify the `JSONSchema` properties during the generation, which is currently used to inject certain validations in the inner spec fields (see [crds.go](https://github.com/rancher/fleet/blob/master/pkg/crd/crds.go#L66). Many of those validations apply to fields that use core Kubernetes types, so it's not possible to add markers on them.
The alternatives would be to replace those types with our own, or maybe use the new [Validation Rules feature](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-rules) to redesign those validations, in case we don't want to just drop them.

There are other differences like properties marked as nullable and the ability to specify the allowed values for enums.

In the scope of this, we could revisit if the logic to install CRDs at controller boot is really used/needed, and perform any necessary cleanup.

PR #1904, opened 2023-10-26
Replace wrangler for kubebuilder for generating CRDs

raulcabello:
This PR replaces wrangler CRD generation with `controller-gen` (kubebuilder). It also removes the dynamic creation of CRDs from the `fleet-controller`. 
`code-gen` from wrangler is still used to generate code for wrangler controllers (e.g. `BundleController`). This can't be removed until wrangler controllers are not used anymore.

refers to https://github.com/rancher/fleet/issues/1815


cmd/codegen/hack/generate_and_sort_crds.sh
```diff
@@ -24,11 +24,7 @@ FROM bitnami/python:3.10
 RUN install_packages jq
 RUN python -m pip install yq
 EOF
-    container_id=$(docker create --rm -i -v ${PWD}:${PWD}:ro -w ${PWD} -v ${tmpdir}:${tmpdir} ${image} yq $@ )
-    if [ -n "${COPY_FILE}" ] ; then
-      # When running on CI, docker-in-docker may be used, so the generated input file, which is outside the working directory, is not mounted and not available
-      docker cp "${COPY_FILE}" "${container_id}:${COPY_FILE}"
-    fi
+    container_id=$(docker create --rm -i -v ${PWD}:${PWD} -w ${PWD} -v ${tmpdir}:${tmpdir} ${image} yq $@ )
     docker start -ai "${container_id}"
```
aruiz14, line 28: I think this can now be simplified to:

```suggestion
    docker run --rm -ai -v ${PWD}:${PWD} -w ${PWD} -v ${tmpdir}:${tmpdir} ${image} yq $@
```

cmd/codegen/hack/generate_and_sort_crds.sh
```diff
@@ -45,13 +41,8 @@ if ! $CONTROLLERGEN --version | grep -q "${CONTROLLERGEN_VERSION}" ; then
...
-COPY_FILE="${CONTROLLERGEN_CRDS_YAML}" run_yq -f cmd/codegen/hack/patch_crd_descriptions.jq "${CRDS_YAML}" "${CONTROLLERGEN_CRDS_YAML}" \
-  --slurp --sort-keys --explicit-start --yaml-output > "${PATCHED_CRDS_YAML}"
-
-# Override previous CRDs file
-cp "${PATCHED_CRDS_YAML}" "${CRDS_YAML}"
+${CONTROLLERGEN} crd webhook paths="./pkg/apis/..." output:stdout > $CRDS_YAML
+# Sort
+run_yq  --slurp --sort-keys --explicit-start --yaml-output -i 'sort_by(.metadata.name)[]' $CRDS_YAML
+# Set spec.preserveUnknownFields=false since it can't be done with controller-gen https://github.com/kubernetes-sigs/controller-tools/issues/476
+run_yq -yi '.spec.preserveUnknownFields=false' $CRDS_YAML
```
aruiz14, line 48: According to [these docs](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#field-pruning), this is the default for `apiextensions.k8s.io/v1`? Maybe it's not needed anymore and safe to just remove it?

Also, this could be done in a single call to `yq`

cmd/codegen/hack/generate_and_sort_crds.sh
```diff
@@ -45,13 +41,8 @@ if ! $CONTROLLERGEN --version | grep -q "${CONTROLLERGEN_VERSION}" ; then
...
-# Patch existing CRDs with the descriptions from controller-gen
-PATCHED_CRDS_YAML="${tmpdir}/crds-patched.yaml"
-COPY_FILE="${CONTROLLERGEN_CRDS_YAML}" run_yq -f cmd/codegen/hack/patch_crd_descriptions.jq "${CRDS_YAML}" "${CONTROLLERGEN_CRDS_YAML}" \
-  --slurp --sort-keys --explicit-start --yaml-output > "${PATCHED_CRDS_YAML}"
-
-# Override previous CRDs file
-cp "${PATCHED_CRDS_YAML}" "${CRDS_YAML}"
+${CONTROLLERGEN} crd webhook paths="./pkg/apis/..." output:stdout > $CRDS_YAML
+# Sort
+run_yq  --slurp --sort-keys --explicit-start --yaml-output -i 'sort_by(.metadata.name)[]' $CRDS_YAML
```
aruiz14, line 46: 🦅 👁️ 
```suggestion
run_yq --slurp --sort-keys --explicit-start --yaml-output -i 'sort_by(.metadata.name)[]' $CRDS_YAML
```

aruiz14 (commented):
It's great to see so much red lines!

cmd/codegen/hack/generate_and_sort_crds.sh
raulcabello: that fails with
```
invalid argument "i" for "-a, --attach" flag: valid streams are STDIN, STDOUT and STDERR
See 'docker run --help'.

```

cmd/codegen/hack/generate_and_sort_crds.sh
raulcabello: you're right. Removed in https://github.com/rancher/fleet/pull/1904/commits/6a8c0e141caf98e8a9d344141c4c29d07b413033

cmd/codegen/hack/generate_and_sort_crds.sh
aruiz14: Oh right, it expects an argument. We can try just dropping it? Also, I noticed `tmpdir` is not really used here anymore, so what about removing that volume as well?:

```suggestion
    docker run --rm -i -v ${PWD}:${PWD} -w ${PWD} ${image} yq $@
```

aruiz14 (approved):

PR #1905, opened 2023-10-26
Split agent

manno:
Refers to #1772 

To prepare the agent for controller-runtime, split the code into three processes:
* registration
* cluster status updater
* agent controller, watches bundledeployments

Only the last part is an actual controller. Cluster status updates work on a timer and registration is done when the agent starts up. A new function "registration.Get" was added, to just return an existing registration to the other processes.

This PR switches the agent to a statefulset, to avoid leader election. I don't expect the agent to scale horizontally, as a controller. 
Updating the cluster status with information about the cluster nodes seems unrelated to Fleet's purpose and we might discuss this in the future.
The code of the handlers in the bundledeployment controller is currently spread around the "handler", a deploy "manager" and the helm deployer. The PR takes first steps to consolidate the functionality into the deployer package, making it more obvious which client is used when during reconciliation.

"Trigger" was renamed to "driftdetect" as it uses the helm history to detect drift for deployments. Which is similar to the "monitor", which updates the bundle deployments status.


internal/cmd/controller/agent/manifest_test.go
```diff
@@ -32,10 +32,10 @@ func TestImageResolve(t *testing.T) {
 	}
 }
 
-func getDeploymentFromManifests(namespace string, scope string, opts ManifestOptions) *appsv1.Deployment {
+func getDeploymentFromManifests(namespace string, scope string, opts ManifestOptions) *appsv1.StatefulSet {
```
aruiz14, line 35: nit-pick: outdated function name

internal/cmd/controller/controllers/cluster/import.go
```diff
@@ -188,33 +188,11 @@ func (i *importHandler) deleteOldAgent(cluster *fleet.Cluster, kc kubernetes.Int
 		return err
 	}
 
-	deployment, err := kc.AppsV1().Deployments(namespace).Get(i.ctx, config.AgentConfigName, metav1.GetOptions{})
-	if apierrors.IsNotFound(err) {
-		return nil
-	} else if err != nil {
-		return err
-	}
```
aruiz14, line 196: Should we keep this or some other sort of backwards compatibility to support older agent versions that used `Deployment`?

internal/cmd/agent/root.go
```diff
@@ -4,53 +4,61 @@ import (
...
 }
 
+var setupLog = ctrl.Log.WithName("setup")
+
 func (a *FleetAgent) Run(cmd *cobra.Command, args []string) error {
+	zopts := zap.Options{
+		Development: true,
+	}
+	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&zopts)))
+	ctx := clog.IntoContext(cmd.Context(), ctrl.Log)
```
aruiz14, line 36: dumb question: should this use `setupLog`?

internal/cmd/agent/register.go
```diff
@@ -0,0 +1,87 @@
...
+	localConfig.RateLimiter = ratelimit.None
+	k8s, err := kubernetes.NewForConfig(localConfig)
+	if err != nil {
+		return err
+	}
+
+	setupLog.Info("starting registration on upstream cluster", "namespace", r.Namespace)
+
+	ctx, cancel := context.WithCancel(ctx)
+	leader.RunOrDie(ctx, r.Namespace, "fleet-agent-register-lock", k8s, func(ctx context.Context) {
```
aruiz14, line 64: Does `leaderelection` make sense anymore?

internal/cmd/agent/register.go
```diff
@@ -0,0 +1,87 @@
...
+func NewRegister() *cobra.Command {
+	cmd := command.Command(&Register{}, cobra.Command{
+		Use:   "register [flags]",
+		Short: "Register agent with an upstream cluster",
+	})
+	return cmd
+}
+
+type Register struct {
+	command.DebugConfig
```
aruiz14, line 31: This is already in the root command. Since it uses `PersistentPreRun`, it's always executed hence not needed here nor in the other subcommands.

internal/cmd/agent/register.go
```diff
@@ -0,0 +1,87 @@
...
+	// provide a logger in the context to be compatible with controller-runtime
+	zopts := zap.Options{
+		Development: true,
+	}
+	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&zopts)))
+	ctx := log.IntoContext(cmd.Context(), ctrl.Log)
+
+	if err := r.SetupDebug(); err != nil {
+		return fmt.Errorf("failed to setup debug logging: %w", err)
+	}
```
aruiz14, line 45: Not needed, see above comment.

internal/cmd/controller/controllers/cluster/import.go
manno: Not sure. I find this part problematic. The agent will delete it's own deployment (via Helm) when the bundle is updated. So this is more like a fail-safe?

internal/cmd/agent/root.go
manno: No, this is the logger in the context. I want one there in every process, in case share code uses it in some package.

internal/cmd/agent/register.go
manno: I wonder if we get anything from it. The code runs in an init container. 
Do we agree to keep the statefulset? Then I could remove all leader election from the agent, I think.

internal/cmd/agent/register.go
manno: Nice

weyfonk (commented):
Looking good, thanks! Leaving a few nitpicks.

dev/setup-fleet
```diff
@@ -21,9 +21,9 @@ helm -n cattle-fleet-system upgrade --install --create-namespace --wait --reset-
   --set debug=true --set debugLevel=1 fleet charts/fleet
 
 # wait for controller and agent rollout
-kubectl -n cattle-fleet-system rollout status deploy/fleet-controller
+kubectl -n cattle-fleet-system rollout status deployment/fleet-controller
 { grep -E -q -m 1 "fleet-agent-local.*1/1"; kill $!; } < <(kubectl get bundles -n fleet-local -w)
-kubectl -n cattle-fleet-system rollout status deploy/fleet-agent
+kubectl -n cattle-fleet-system rollout status statefulset/fleet-agent
```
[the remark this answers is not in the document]
manno: Not sure if I get this,`dev/setup-rancher-clusters` doesn't wait for agent deployments directly?

.github/scripts/deploy-fleet.sh
```diff
@@ -39,7 +39,7 @@ eventually helm upgrade --install fleet charts/fleet \
 # wait for controller and agent rollout
 kubectl -n cattle-fleet-system rollout status deploy/fleet-controller
 { grep -E -q -m 1 "fleet-agent-local.*1/1"; kill $!; } < <(kubectl get bundles -n fleet-local -w)
-kubectl -n cattle-fleet-system rollout status deploy/fleet-agent
+kubectl -n cattle-fleet-system rollout status statefulset/fleet-agent
```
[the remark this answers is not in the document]
manno: * .github/scripts/deploy-fleet-latest-release.sh - not possible until release exists that uses a statefulset, because that script deploys the latest released fleet

* .github/scripts/setup-latest-rancher.sh - are you saying this should also wait for agents?
    .github/scripts/setup-rancher.sh - add another wait here?
    .github/workflows/rancher-upgrade-fleet.sh - I think this waits for agent bundles, which is a stronger check than waiting for the statefulset

internal/cmd/agent/clusterstatus.go
```diff
@@ -0,0 +1,153 @@
...
+
+type ClusterStatus struct {
+	UpstreamOptions
+	CheckinInterval string `usage:"How often to post cluster status" env:"CHECKIN_INTERVAL"`
+}
+
+func (cs *ClusterStatus) Run(cmd *cobra.Command, args []string) error {
+	// provide a logger in the context to be compatible with controller-runtime
+	zopts := zap.Options{
+		Development: true,
```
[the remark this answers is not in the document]
manno: Yes, I would say we will have to come up with logging configuration later. I was hoping we could switch to https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/log/zap#Options.BindFlags

internal/cmd/agent/deployer/plan/plan.go
```diff
@@ -0,0 +1,144 @@
...
+		return plan, err
+	}
+
+	var errs []error
+	for gvk, objs := range plan.Update {
+		for key := range objs {
+			desiredObj := desired[gvk][key]
+			if desiredObj == nil {
+				desiredKey := key
+				// if different namespace options to guess if resource is namespaced or not
```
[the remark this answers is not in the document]
manno: Not at all.

dev/setup-fleet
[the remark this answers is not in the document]
manno: I added https://github.com/rancher/fleet/pull/1905/commits/b49809cfde25bdfb3b5c72b61ae292e91a338b84

.github/scripts/deploy-fleet.sh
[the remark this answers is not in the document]
weyfonk: Oh of course, this is an agent-only change, so my comment does not make much sense (except perhaps for `deploy-fleet-latest-release.sh`, which as you've noticed needs to wait until Fleet is released with stateful sets).
Sorry for the noise.

dev/setup-fleet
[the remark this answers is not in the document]
weyfonk: You're right, feel free to disregard my comment, I had overlooked the fact that this is only about the agent.

aruiz14 (approved):
LGTM, just a minor suggestion

internal/cmd/agent/register.go
```diff
@@ -0,0 +1,67 @@
...
+		logrus.Fatal(err)
+	}
+
+	_, err = agentInfo.ClientConfig.ClientConfig()
+	if err != nil {
+		logrus.Fatal(err)
+	}
+
+	setupLog.Info("successfully registered with upstream cluster", "namespace", ns)
+	cancel()
```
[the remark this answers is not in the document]
manno: I'll add it in the next PR

internal/cmd/agent/controller.go
```diff
@@ -0,0 +1,105 @@
...
+func start(ctx context.Context, kubeConfig, namespace, agentScope string) error {
+	clientConfig := kubeconfig.GetNonInteractiveClientConfig(kubeConfig)
+	kc, err := clientConfig.ClientConfig()
+	if err != nil {
+		return err
+	}
+
+	agentInfo, err := register.Get(ctx, namespace, kc)
+	if err != nil {
+		logrus.Fatal(err)
```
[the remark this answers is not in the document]
manno: Yes, all these logrus calls should be removed in the future https://github.com/rancher/fleet/pull/1772/files#diff-a08a0ecbbab37eaea426b6b8bc55a0d5107c75cac3db92153ef7d3329dc47650R79. But the second point you make is interesting. Why do we exit instead of return...
I'll change it in the next PR.

PR #1915, opened 2023-10-30
Replace json-based implementation of DeepCopy on GenericMap

aruiz14:
Relates to #1842

The current `GenericMap`'s `DeepCopy` method is calling `wrangler/pkg/convert.ToObj` in order to be a copy of the map. This function uses `encoding/json` to serialize and deserialize in order to get a new independent copy of the entire map. However, since `encoding/json` uses reflection, this approach is more CPU-intensive than other alternatives, like using native recursion.

<p align="center">
  
</p>
This shows how up to a 9% of the time from a 30s profile is spent on `convert.ToObj`.
Also, I ran a simple benchmark to get an estimation on how both implementations behave:

```
BenchmarkDeepCopy/json-small-12         	  160172	      7147 ns/op
BenchmarkDeepCopy/native-small-12       	 1204741	       994.6 ns/op
BenchmarkDeepCopy/json-big-12           	     781	   1510766 ns/op
BenchmarkDeepCopy/native-big-12         	    4351	    254641 ns/op
BenchmarkDeepCopy/json-huge-12          	       4	 299542114 ns/op
BenchmarkDeepCopy/native-huge-12        	      32	  36687546 ns/op
```

where:
- `small` is the contents of a sample `values.yaml` with just a handful of keys.
- `big` is an arbitrary ~280KB JSON file
- `huge` is a ~25MB JSON file I found on GitHub with random data

This means that using this implementation could be one order of magnitude faster than using `convert.ToObj` , hence causing an speedup proportional to the amount and size of `GenericMap`s used.

## Test

```shell
cd pkg/apis/fleet.cattle.io/v1alpha1/
go test
```

### Trade-off

Possible none, just the chance of introducing a bug for a non-tested corner case.

### Potential improvement

Review other possible usages of `convert.ToObj` and migrate to this approach if applicable.

### DIFF
--- a/pkg/apis/fleet.cattle.io/v1alpha1/generic_map.go
+++ b/pkg/apis/fleet.cattle.io/v1alpha1/generic_map.go
@@ -29,8 +29,29 @@ func deepCopyMap(src, dest map[string]interface{}) {
 			destValue := make(map[string]interface{}, len(value))
 			deepCopyMap(value, destValue)
 			dest[key] = destValue
+		case []any:
+			destValue := make([]any, len(value))
+			deepCopySlice(value, destValue)
+			dest[key] = destValue
 		default:
 			dest[key] = value
+		}
+	}
+}
+
+func deepCopySlice(src, dest []any) {
+	for i := range src {
+		switch value := src[i].(type) {
+		case map[string]interface{}:
+			destValue := make(map[string]interface{}, len(value))
+			deepCopyMap(value, destValue)
+			dest[i] = destValue
+		case []any:
+			destValue := make([]any, len(value))
+			deepCopySlice(value, destValue)
+			dest[i] = destValue
+		default:
+			dest[i] = value
 		}
 	}
 }
--- a/pkg/apis/fleet.cattle.io/v1alpha1/generic_map.go
+++ b/pkg/apis/fleet.cattle.io/v1alpha1/generic_map.go
@@ -33,8 +33,8 @@ func deepCopyMap(src, dest map[string]interface{}) {
 			destValue := make(map[string]interface{}, len(value))
 			deepCopyMap(value, destValue)
 			dest[key] = destValue
-		case []any:
-			destValue := make([]any, len(value))
+		case []interface{}:
+			destValue := make([]interface{}, len(value))
 			deepCopySlice(value, destValue)
 			dest[key] = destValue
 		default:
@@ -45,15 +45,15 @@ func deepCopyMap(src, dest map[string]interface{}) {
 
 // deepCopyMap will recursively copy each entry from src into dest, ensuring that nested maps and slices are copies as well
 // It assumes that both slices have the same capacity.
-func deepCopySlice(src, dest []any) {
+func deepCopySlice(src, dest []interface{}) {
 	for i := range src {
 		switch value := src[i].(type) {
 		case map[string]interface{}:
 			destValue := make(map[string]interface{}, len(value))
 			deepCopyMap(value, destValue)
 			dest[i] = destValue
-		case []any:
-			destValue := make([]any, len(value))
+		case []interface{}:
+			destValue := make([]interface{}, len(value))
 			deepCopySlice(value, destValue)
 			dest[i] = destValue
 		default:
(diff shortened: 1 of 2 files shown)


pkg/apis/fleet.cattle.io/v1alpha1/values.go
```diff
@@ -1,27 +0,0 @@
-package v1alpha1
-
-import (
-	"encoding/json"
-
-	"github.com/rancher/wrangler/v2/pkg/data/convert"
-)
-
-type GenericMap struct {
```
aruiz14, line 9: I renamed the file to match the type name, but only `DeepCopyInto` was modified.

pkg/apis/fleet.cattle.io/v1alpha1/generic_map.go
```diff
@@ -0,0 +1,57 @@
...
+		switch value := src[key].(type) {
+		case map[string]interface{}:
+			destValue := make(map[string]interface{}, len(value))
+			deepCopyMap(value, destValue)
+			dest[key] = destValue
+		case []any:
+			destValue := make([]any, len(value))
+			deepCopySlice(value, destValue)
+			dest[key] = destValue
+		default:
```
moio, line 36: Can pointer types ever come as input?


Eg. pointers to slices or maps.

moio (commented):
If you have a reproducible case where you can see improvements in a quantitative way (eg. via flame graphs or benchmarks) it would be good to add the speedup value in the PR description.

Thanks for fixing this!

pkg/apis/fleet.cattle.io/v1alpha1/generic_map.go
aruiz14: These changes rely on `GenericMap` always being instantiated from calling Unmarshal over a JSON or YAML payload, which should not produce such pointers. Note that this code also won't work for for slices and maps with a specific type (e.g. `map[string]string` vs. `map[string]interface{}`). That's the main reason I kept this function here instead of a common place where it could be reused.

In fact, I should probably add this as a comment so that we can prevent blindly copy-pasting it somewhere that may not fit that restriction. Thanks for your comment!

aruiz14: Comments added at b9ddf4a437fae3895494b5e68c4be5d8ea671b87

pkg/apis/fleet.cattle.io/v1alpha1/generic_map.go
```diff
@@ -0,0 +1,63 @@
...
+	in.Data = map[string]interface{}{}
+	return json.Unmarshal(data, &in.Data)
+}
+
+func (in *GenericMap) DeepCopyInto(out *GenericMap) {
+	out.Data = make(map[string]interface{}, len(in.Data))
+	deepCopyMap(in.Data, out.Data)
+}
+
+// deepCopyMap will recursively copy each entry from src into dest, ensuring that nested maps and slices are copies as well
```
weyfonk, line 25: Did you mean this?
```suggestion
// deepCopyMap will recursively copy each entry from src into dest, ensuring that nested maps and slices are copied as well
```

pkg/apis/fleet.cattle.io/v1alpha1/generic_map.go
```diff
@@ -0,0 +1,63 @@
...
+			destValue := make([]interface{}, len(value))
+			deepCopySlice(value, destValue)
+			dest[key] = destValue
+		default:
+			dest[key] = value
+		}
+	}
+}
+
+// deepCopyMap will recursively copy each entry from src into dest, ensuring that nested maps and slices are copies as well
```
weyfonk, line 46:
```suggestion
// deepCopySlice will recursively copy each entry from src into dest, ensuring that nested maps and slices are copied as well
```

weyfonk (approved):
Thanks for taking care of this!

raulcabello (dismissed):
LGTM, thanks!

> This shows how up to a 9% of the time from a 30s profile is spent on `convert.ToObj`

Do we know how many times this method was called? It looks to me like if this method is impacting the performance it's because it's being called more times than it should

aruiz14:
> LGTM, thanks!
> 
> > This shows how up to a 9% of the time from a 30s profile is spent on `convert.ToObj`
> 
> Do we know how many times this method was called? It looks to me like if this method is impacting the performance it's because it's being called more times than it should

Sadly, CPU profiles do not contain that information, only a measure of time spent in each method. Nonetheless, I could observe many parts of the code (including generated controllers' code) perform calls to `DeepCopy` on top-level entities, which ultimately end up executing this code if any nested field contain a `GenericMap`.

weyfonk (approved):

PR #1920, opened 2023-10-31
Remove leftovers from controller-gen migration

aruiz14:
Refers to #1815

There were leftover generated files from wrangler's controller-gen that were not deleted. A similar version of them is used in controller-gen but it's `kubebuilder` which generates it as part of the `_types.go` files. In addition to removing those files, this PR is also adding those missing bits.


cmd/codegen/hack/generate_and_sort_crds.sh
```diff
@@ -40,6 +40,7 @@ if ! $CONTROLLERGEN --version | grep -q "${CONTROLLERGEN_VERSION}" ; then
 fi
 
 # Run controller-gen
+${CONTROLLERGEN} object:headerFile=scripts/boilerplate.go.txt,year="$(date +%Y)" paths="./pkg/apis/..."
```
aruiz14, line 43: This step creates `zz_generated` files. It's part of `kubebuilder`'s Makefile boilerplate.

internal/cmd/controller/controllers/clusterregistration/controller.go
```diff
@@ -16,7 +16,6 @@ import (
 	"github.com/rancher/fleet/internal/config"
 	fname "github.com/rancher/fleet/internal/name"
 	"github.com/rancher/fleet/internal/registration"
-	fleetgroup "github.com/rancher/fleet/pkg/apis/fleet.cattle.io"
```
aruiz14, line 19: This package (see `pkg/apis/fleet.cattle.io/zz_generated_register.go` below) only contained one constant. Hence I directly used the group name defined in the versioned package, as it's also available there.

raulcabello (approved):

PR #1940, opened 2023-11-15
Switch container images to scratch and suse/git

manno:
closes #1877

### DIFF
--- a/package/Dockerfile.agent
+++ b/package/Dockerfile.agent
@@ -1,7 +1,7 @@
 ARG BUILD_ENV=dapper
 
-FROM registry.suse.com/bci/bci-base:15.5.36.5.52 AS base
-RUN zypper in --no-recommends -y git bash openssh && groupadd -g 1000 fleet-apply && useradd -u 1000 -g 1000 -m fleet-apply; rm -fr /var/cache/* /var/log/*log
+FROM registry.suse.com/suse/git:2.35 AS base
+USER 1000
 COPY package/log.sh /usr/bin/
 
 FROM base AS copy_dapper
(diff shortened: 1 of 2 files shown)


package/Dockerfile.agent
```diff
@@ -1,7 +1,7 @@
 ARG BUILD_ENV=dapper
 
-FROM registry.suse.com/bci/bci-base:15.5.36.5.52 AS base
-RUN zypper in --no-recommends -y git bash openssh && groupadd -g 1000 fleet-apply && useradd -u 1000 -g 1000 -m fleet-apply; rm -fr /var/cache/* /var/log/*log
+FROM registry.suse.com/suse/git:2.35 AS base
```
mattfarina, line 3: Dropping to a major.minor version for the tag could be done everywhere. When rebuilt (e.g., for the final) you would have the fixed packages automatically be there.

mattfarina (commented):
If scratch doesn't work as an image you can use a different one of the BCI base images. For example, the busybox one. There are fewer packages installed so there's a lower opportunity for a CVE to show up. You can find a table of differences at https://documentation.suse.com/container/all/html/Container-guide/index.html#container-bci-general

package/Dockerfile.agent
manno: I think automation will set a specific version when it runs 🤷

manno:
> If scratch doesn't work as an image you can use a different one of the BCI base images. For example, the busybox one. 
There are fewer packages installed so there's a lower opportunity for a CVE to show up. You can find a table of differences at https://documentation.suse.com/container/all/html/Container-guide/index.html#container-bci-general

Scratch didn't work for some reason, the imagescan test failed. `bci-busybox` passes tests.

weyfonk (approved):
LGTM

PR #1941, opened 2023-11-16
Improve pull request template

thardeck:
to make contribution easier.

Fixes rancher/fleet#1238

### DIFF
--- a/docs/qa_template.md
+++ b/docs/qa_template.md
@@ -1,0 +1,27 @@
+Pull request content for Rancher QA verification.
+
+## Additional QA
+### Issue: <link the issue or issues this PR resolves here>
+< If your PR depends on changes from another pr link them here and describe why they are needed on your solution section. >
+ 
+### Problem
+< Describe the root cause of the issue you are resolving. This may include what behavior is observed and why it is not desirable. If this is a new feature describe why we need this feature and how it will be used. >
+ 
+### Solution
+< Describe what you changed to fix the issue. Relate your changes back to the original issue / feature and explain why this addresses the issue.>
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -1,62 +1,9 @@
 <!-- Specify the issue ID that this pull request is solving -->
-Fix #XXX
+Fixes #XXX
```
weyfonk, line 2: This could prevent issues from being automatically closed by a single PR, as we often close our issues through multiple PRs.
```suggestion
Refers to #XXX
```

.github/pull_request_template.md
```diff
@@ -1,62 +1,9 @@
...
-
-<!-- Please describe, if any, the trade-offs that you found acceptable in this pull request -->
-
-### Potential improvement
-
-<!-- Please describe, if any, potential improvement that you are envisioning -->
-
-<!-- Please Uncomment following block if looking for QA validation
-
-## Additional QA
```
weyfonk, line 36: Shouldn't we _move_ the QA template somewhere else, instead of deleting it? Could we have it eg. in `.github/ISSUE_TEMPLATE`, or somewhere we could use it as reference when needed?

weyfonk: > could prevent users from creating issues because of the huge amount of content required
Agreed; I think we should have that template _somewhere_ (TBD) for reference, without making it part of issue templates per se. It should enable the dev team to communicate expectations in a standardised way with QA, without getting in the way of issue submission.

weyfonk (approved):

PR #1942, opened 2023-11-16
Make sure that all go.mods are tidy for new prs

thardeck:
to prevent failing ci runs.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -15,6 +15,13 @@ updates:
     labels:
       - "dependencies"
   -
+    package-ecosystem: "gomod"
+    directory: "/e2e/testenv/infra"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "dependencies"
+  -
     package-ecosystem: "docker"
     directory: "/"
     schedule:


.github/scripts/check-for-go-mod-changes.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+set -ue
+
+for DIRECTORY in . ./pkg/apis ./e2e/testenv/infra; do
+    cd "$DIRECTORY"
+    go mod tidy
+    go mod verify
+    cd "$OLDPWD"
```
raulcabello, line 8: what's the value of `$OLDPWD` ?

.github/scripts/check-for-go-mod-changes.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+set -ue
+
+for DIRECTORY in . ./pkg/apis ./e2e/testenv/infra; do
```
aruiz14, line 4: What about `find`-ing `go.mod` files in the repository and run this on every location?

.github/scripts/check-for-go-mod-changes.sh
raulcabello: ah ok

raulcabello (approved):

.github/scripts/check-for-go-mod-changes.sh
aruiz14: Hmm not worth it then, thanks!

aruiz14 (approved):

.github/scripts/check-for-go-mod-changes.sh
weyfonk: Maybe something like this?
`find . -name 'go.mod' -execdir go mod tidy \;`

weyfonk (approved):

PR #1952, opened 2023-11-22
Reduce the number of times that Bundle resources is serialized

aruiz14:
Refers to #1842

`Content` cluster-wide objects are stored in Kubernetes, and their names is based on the SHA256 sum of the resources' JSON serialization. The current approach requires the controller to compute this serialization on every handler execution in order to ensure the `Content` objects exist in Kubernetes. Since these handlers are executed quite frequently, the time spent doing this work is proportional to the number of bundles and clusters, turning into a non negligible CPU time usage.

The proposed solution calculates makes the controller store the calculated checksum in the status field, pre-loading in the `Manifest` wrapper so for cases that don't need access to the payload directly the serialization steps are avoided. Only if the observed generation differs from the generation field in the `Bundle` metadata, we will then intentionally compute the hash and save the new value (if it did change, as `generation` changes for any change in the `spec`, not only the `resources` field).

These changes should be transparent from the users, besides the need to accordingly update the CRDs.


internal/manifest/store.go
```diff
@@ -23,35 +23,49 @@ func NewStore(content fleetcontrollers.ContentController) Store {
...
 	compressed, err := content.Gzip(data)
 	if err != nil {
-		return id, err
+		return err
 	}
 
 	_, err = c.content.Create(&fleet.Content{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: id,
+			Annotations: map[string]string{
```
raulcabello, line 64: why do we use an annotation here? would it make sense to use a new status field as you added for the `Bundle` ?

aruiz14: It's used by Lookup for verifying the integrity of the payload.
Since Lookup only has access to `Content` objects, not `Bundles`, you mean to add a new field to the `Content` object, right? (in fact, I would add it to the `Spec`, not in the `Status`.
That's a fair point, I don't mind adding that as well. The more aggressive approach would be to completely get rid of the integrity check, which I honestly think it does not make much sense.

raulcabello: yes I meant the `Content` resource. The integrity check does not make much sense to me either, I think it can be removed

aruiz14: Replaced annotation with a dedicated spec field in `Content`, thanks!
If you feel strongly about the integrity check, please reopen this conversation.

raulcabello (approved):

Issue #1953, opened 2023-11-22
[SURE-7269] Update v0.8.0 images


kkaempf:
We need to do an oob update of

    rancher/fleet-agent:v0.8.0
    rancher/fleet:v0.8.0
    rancher/gitjob:v0.1.76

PR #1960, opened 2023-11-23
Backport OCI test changes

thardeck:
Refers to rancher/fleet#1953


aruiz14 (approved):
Multi-cluster test failed but it's probably flaky

PR #1964, opened 2023-11-23
Use UniqueApplyForResourceVersion in Bundle and GitRepo GeneratingHandlers

aruiz14:
Refers to #1842

_WIP_

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -251,3 +251,5 @@ require (
 )
 
 replace github.com/json-iterator/go v1.1.10 => github.com/json-iterator/go v1.1.12
+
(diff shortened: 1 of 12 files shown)


go.mod
```diff
@@ -251,3 +251,5 @@ require (
 )
 
 replace github.com/json-iterator/go v1.1.10 => github.com/json-iterator/go v1.1.12
+
+replace github.com/rancher/wrangler/v2 => github.com/aruiz14/wrangler/v2 v2.1.1-0.20231123160640-9dd87242b681
```
aruiz14, line 255: This must be replaced with a new wrangler release when rancher/wrangler#341 is merged and published

aruiz14:
`master` no longer uses wrangler so this is not needed.

PR #1975, opened 2023-11-28
Switch bundle diff test to use random target namespaces

manno:
The test was flaky, probably since both specs used the same namespace. The specs tried to work around that by waiting, but that's probably not effective.

The bundle-diffs.yaml was used in another test and was cloned, so that test doesn't have to change.

### DIFF
--- a/e2e/single-cluster/bundle_diffs_test.go
+++ b/e2e/single-cluster/bundle_diffs_test.go
@@ -78,7 +78,7 @@ var _ = Describe("BundleDiffs", func() {
 
 			os.RemoveAll(tmpdir)
 
-			k.Delete("namespace", targetNamespace)
+			_, _ = k.Delete("namespace", targetNamespace)
 
 			// test cases use the same namespace, so we have to
 			// make sure resources are cleaned up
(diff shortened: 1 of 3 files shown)


raulcabello (approved):
make sure golangci-lint is green before merging

e2e/single-cluster/bundle_diffs_test.go
```diff
@@ -43,20 +50,40 @@ var _ = Describe("BundleDiffs", func() {
...
 		Expect(err).ToNot(HaveOccurred(), out)
 
 		DeferCleanup(func() {
-			out, err := k.Delete("-f", testenv.AssetPath(asset))
+			out, err := k.Delete("-f", gitrepo)
 			Expect(err).ToNot(HaveOccurred(), out)
 
+			os.RemoveAll(tmpdir)
+
+			k.Delete("namespace", targetNamespace)
```
aruiz14, line 81:
```suggestion
			_, _ = k.Delete("namespace", targetNamespace)
```

manno: Not sure what to do here, exiting from clean up because one step failed just leaves you with a mess... Maybe we should in the future log the error and continue.

PR #1976, opened 2023-11-28
Agent Bundle/Release Cleanup Ignores Non-Fleet Releases

manno:
This is a follow-up to https://github.com/rancher/fleet/pull/1772

There are a lot of `Deleting orphan bundle ID k3s, release kube-system/traefik` messages in the logs. The clean up only works on helm charts with the `fleet.cattle.io/bundle-id` annotation.

Turns out k3s patches this annotation into all its charts: https://github.com/k3s-io/rke2-v2/commit/ba9e82c11bcb8f4c9360d3a251dff30889dbaefc#diff-ae6fb519767e6661e6cfd6bbebbf3b8c432aee7a0166fa70d35851cc6d55ca38R27

The only reason fleet-agent doesn’t delete, for example, traefik, is that at a later time it also checks for the namespace annotation.
The logic seems weird, calling ListDeployments multiple times when deleting a release. Finally another call to `globalCfg.Releases.List`, which filters for the namespace annotation `fleet.cattle.io/agent-namespace`. At this point the k3s charts are skipped and not deleted.

This PR checks for the second annotation much earlier and avoids the confusing log message.


weyfonk (dismissed):
LGTM; any tests showing that the confusing log messages no longer appear?

manno:
> LGTM; any tests showing that the confusing log messages no longer appear?

I didn't keep any logs from manual testing. However, this is a good opportunity for some refactoring and adding a basic unit test.

weyfonk (approved):
Thanks for adding test cases for listing deployments :100: !

Issue #1717, opened 2023-08-10
[SURE-6723] Allow to specify additional arbitrary Fleet chart values


kkaempf:
## SURE-6723

**Is your feature request related to a problem? Please describe.**

### Request description:
Goal: allow Rancher users to specify additional arbitrary values for the Fleet chart, so that their Fleet installation can be customized flexibly.

This is particularly useful as a catch-all measure when assisting customers to introduce Fleet-specific experimental tuning settings, debug settings, etc. even if they are not yet covered by the Rancher UI.

### Actual behavior:
Currently Rancher installs the Fleet with fixed values, making the solution less flexible than Fleet standalone cases.

### Expected behavior:
Parameters can be customized - feature parity with Fleet standalone.

### Workaround:
Manually editing Fleet deployments, ConfigMaps, etc.

### Additional notes:
Currently Rancher installs the Fleet with fixed values. Code reference:

https://github.com/rancher/rancher/blob/9c5a25675676b680898f933681abde88f53dba95/pkg/controllers/dashboard/fleetcharts/controller.go#L89-L100

A similar change was introduced for the webhook component: https://github.com/rancher/rancher/pull/42060

**Describe the solution you'd like**

Parameters can be customized - feature parity with Fleet standalone.

**Describe alternatives you've considered**

**Additional context**

moio:
> Upon discussion, transferring to backend since no UI is required.

To elaborate this statement a bit: having a configurable ConfigMap that can achieve the same goal is sufficient as a starting point and should cover most cases, even without a UI (which can always be added later if necessary).

PR #1981, opened 2023-11-30
Configurable leader election via chart values

aruiz14:
Part of #1491

It makes leader election parameters configurable via Helm chart values, and adapts the default values by applying the following reasoning:
* The purpose of leader election on Fleet is to ensure only instance is running at a time.
  * `fleet-controller` deployment is configured to use only 1 replica, but users could still try to scale it up.
  * Rolling updates could cause to temporarily have 2 living instances of fleet.  
  * This could be mitigated by using `StatefulSet` instead of `Deployment`, although this would still not prevent users from manually scaling up the `StatefulSet` too
* The definition of `LeaseDuration` says that any candidate should wait at lease this duration before attempting to become leader, so long duration could slow down the rollout of new images.
  * For this reason, I've decreased **`LeaseDuration`** from 45 to **30 seconds**.
    * `core` Kubernetes clients use 15 seconds, but they also keep a shorter retry period (2s), but 30s is enough for Fleet's use case.
* `RetryPeriod` defines the wait period between actions, including renewing the leader lease. This mean that every `$retryPeriod`, it will acquire the lease for up to `$leaseDuration`. It has a default period of 2 seconds, which causes too much pressure on the Kubernetes API
  * For this reason, and taking into account the value used for `LeaseDuration`, I'm changing **`RetryPeriod`** from 2 to **10 seconds**.
* `RenewalDeadline` is the period during which an active master will keep trying to renew the lock before giving up, which in our implementation means exiting the program. Failures renewing the lease could happen due to network instability in the node running the controller.
  * Considering the `LeaseDuration` and `RetryPeriod`, in order to allow at least 2 attempts to renew the lease before it expires, I'm configuring **`RenewalDeadline`** to **25 seconds**.

Reference: https://pkg.go.dev/k8s.io/client-go/tools/leaderelection#LeaderElectionConfig

### DIFF
--- a/charts/fleet/templates/deployment.yaml
+++ b/charts/fleet/templates/deployment.yaml
@@ -39,6 +39,18 @@ spec:
         - name: FLEET_CPU_PPROF_PERIOD
           value: {{ quote .Values.cpuPprof.period }}
         {{- end }}
+        {{- if .Values.leaderElection.leaseDuration }}
+        - name: CATTLE_ELECTION_LEASE_DURATION
+          value: {{.Values.leaderElection.leaseDuration}}
+        {{- end }}
+        {{- if .Values.leaderElection.retryPeriod }}
+        - name: CATTLE_ELECTION_RETRY_PERIOD
+          value: {{.Values.leaderElection.retryPeriod}}
+        {{- end }}
+        {{- if .Values.leaderElection.renewDeadline }}
+        - name: CATTLE_ELECTION_RENEW_DEADLINE
+          value: {{.Values.leaderElection.renewDeadline}}
+        {{- end }}
         {{- if .Values.debug }}
         - name: CATTLE_DEV_MODE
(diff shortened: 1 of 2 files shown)


moio:
Reasoning sounds solid to me. The only part I believe not being accurate is:

> Closes https://github.com/rancher/fleet/issues/1491

That issue is not only about making these parameters configurable, but crucially _to make them configurable in Rancher for all downstream clusters at once_.

This PR is a step in that direction, but crucially we will still need enablement in Rancher code to make use of these new Helm values, namely https://github.com/rancher/fleet/issues/1717.

Please do close https://github.com/rancher/fleet/issues/1491 when this PR is merged and backported and a fix for https://github.com/rancher/fleet/issues/1717 is also in place.

Thanks!

manno (approved):

PR #1985, opened 2023-12-01
Remove bd requeue, rely on drift correction

manno:
This is a follow-up to https://github.com/rancher/fleet/pull/1772

We discussed if that re-queue is still needed. Also, why are agent resources deleted?

### DIFF
--- a/internal/cmd/agent/controller/bundledeployment_controller.go
+++ b/internal/cmd/agent/controller/bundledeployment_controller.go
@@ -142,7 +142,7 @@ func (r *BundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 			}
 		}
 
-		if len(bd.Status.ModifiedStatus) > 0 && monitor.IsAgent(bd) && monitor.ShouldRedeploy(bd) {
+		if len(bd.Status.ModifiedStatus) > 0 && monitor.ShouldRedeployAgent(bd) {
 			bd.Status.AppliedDeploymentID = ""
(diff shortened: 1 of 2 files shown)


internal/cmd/agent/controller/bundledeployment_controller.go
```diff
@@ -143,16 +142,10 @@ func (r *BundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 			}
 		}
 
-		// TODO is this needed with drift correction?
-		if len(bd.Status.ModifiedStatus) > 0 && monitor.ShouldRedeploy(bd) {
-			result.RequeueAfter = durations.MonitorBundleDelay
-			logger.Info("Redeploying, by resetting the appliedDeploymentID")
+		if len(bd.Status.ModifiedStatus) > 0 && monitor.IsAgent(bd) && monitor.ShouldRedeploy(bd) {
```
raulcabello, line 145: we don't need `montitor.IsAgent` as it is checked inside [ShouldRedeploy](https://github.com/rancher/fleet/blob/30ca349a84bcb80f6b2bc5eb26f86c098aaf6b58/internal/cmd/agent/deployer/monitor/updatestatus.go#L49)

raulcabello:
> Also, why are agent resources deleted?

Maybe because we wanted to remove drift in the agent? Resources are deleted if it is the agent and `len(bd.Status.ModifiedStatus) > 0`, then maybe are recreated after?  

Can we remove this code and just enable drift correction in the agent `Bundle`?

raulcabello (approved):

PR #1995, opened 2023-12-04
Remove s390x as runners are not available

raulcabello:


aruiz14 (approved):

aruiz14:
Will this need to be backported to the release branches as well?

PR #2011, opened 2023-12-14
Pre-process ClusterGroups' Selector on Spec changes

aruiz14:
Refers to _ToBeCreated_

We have profiling data that shows how calculating if a Fleet Cluster belongs to a ClusterGroup can cause a considerable use of CPU under certain scenarios. In particular, most of this usage is dedicated to processing the field `.Spec.Selector` from each ClusterGroup, as shown in this screenshot:

Screenshot 2023-12-14 at 13 18 52

The changes in this PR add some sort of caching mechanism for this processing, making the `clustergroup` controller to store the relevant object every time a change in the `ClusterGroup` spec is observed. Then the targeting mechanism can reuse this pre-processed selector in a more efficient way.

TODO
------
- [ ] Add unit tests for `Store`

### DIFF
--- a/internal/cmd/controller/target/clustergroups.go
+++ b/internal/cmd/controller/target/clustergroups.go
@@ -72,11 +72,9 @@ func (s *ClusterGroupStore) Delete(key string) {
 
 func (s *ClusterGroupStore) GetSelector(cg *fleet.ClusterGroup) (labels.Selector, error) {
 	key := cg.Namespace + "/" + cg.Name
+
 	entry, found := s.getEntry(key)
-	if !found {
-		return nil, nil
-	}
-	if !entry.sameAs(cg) {
+	if !found || !entry.sameAs(cg) {
 		entry = newEntry(cg)
 		s.setEntry(key, entry)
 	}
--- a/internal/cmd/controller/target/clustergroups.go
+++ b/internal/cmd/controller/target/clustergroups.go
@@ -1,11 +1,12 @@
 package target
 
 import (
-	fleet "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sync"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+
+	fleet "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
 )
 
 type clusterGroupEntry struct {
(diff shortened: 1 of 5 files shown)


internal/cmd/controller/target/clustergroups.go
```diff
@@ -0,0 +1,85 @@
...
+	defer s.mu.Unlock()
+
+	delete(s.store, key)
+}
+
+func (s *ClusterGroupStore) GetSelector(cg *fleet.ClusterGroup) (labels.Selector, error) {
+	key := cg.Namespace + "/" + cg.Name
+	entry, found := s.getEntry(key)
+	if !found {
+		return nil, nil
```
raulcabello, line 77: This should never happens as all selectors should be in the cache. Should we return an error here? or at least a log message?

manno: Relying on the update from the clustergroup handler should be safe.. still maybe we can just add the missing element to the store here?

manno:
I quickly added a benchmark test for this here: https://github.com/rancher/fleet/blob/clustegroup-store-0.8-bench/internal/cmd/controller/target/selector_test.go#L19

Amazing that this selector shows up in the profiling, from my quick measurement:

Before:

```
% go test -bench=. ./internal/cmd/controller/target                                                                                                                                                                                                    
goos: darwin
goarch: arm64
pkg: github.com/rancher/fleet/internal/cmd/controller/target
BenchmarkStore-10    	   10000	   1801391 ns/op
PASS
ok  	github.com/rancher/fleet/internal/cmd/controller/target	18.311s
go test -bench=. ./internal/cmd/controller/target  132,37s user 10,70s system 687% cpu 20,806 total
```

After:
```
% go test -bench=. ./internal/cmd/controller/target
goos: darwin
goarch: arm64
pkg: github.com/rancher/fleet/internal/cmd/controller/target
BenchmarkStore-10    	   10000	   1379797 ns/op
PASS
ok  	github.com/rancher/fleet/internal/cmd/controller/target	14.095s
go test -bench=. ./internal/cmd/controller/target  118,32s user 1,60s system 724% cpu 16,563 total
```

internal/cmd/controller/target/clustergroups.go
aruiz14: Oh, right. I should merge this with the `sameAs` check in order to be totally safe.

raulcabello:
Looks good to me. Should we target v0.9 instead of v0.8? just for avoiding the risk of introducing a bug in the stable release.

aruiz14:
Abandoning until receiving further evidence that this is a problem

Issue #1794, opened 2023-09-20
Override drift correction via targetCustomization


manno:

raulcabello:
# QA Template

## Solution
drift correction options added to `targetCustomization` in the `fleet.yaml`, so it can be overridden per clusters 

## Testing
Create a `GitRepo` with `correctDrift` configured, and override the configuration in the `targetCustomization` in the `fleet.yaml`.

## Additional info
Needs a new fleet RC

thehejik:
### Test report

Works as expected on v2.8.3-rc2

#### Environment
Tested on `rancher:2.8.3-rc3` with 2 downstream clusters added. Used images:
```
rancher/fleet:v0.9.1-rc.5
rancher/gitjob:v0.9.3
rancher/fleet-agent:v0.9.1-rc.5
```
#### Testing
Added labels to clusters under CD/Clusters in `fleet-default` namespace:
* First cluster has label `environment: prod`
* second cluster has label `environment: dev`

Added gitRepo with simple nginx deployment.yaml, and with fleet.yaml containing:
```yaml
defaultNamespace: fleet-nginx-github
# This is new feature from 2.8.3
targetCustomizations:
- name: production
  # first cluster
  clusterSelector:
    matchLabels:
      environment: prod
  correctDrift:
    enabled: true
    force: true
    keepFailHistory: true
- name: development
  # second cluster
  clusterSelector:
    matchLabels:
      environment: dev
  correctDrift:
    enabled: false
    force: false
    keepFailHistory: false
```

After its deploying I tried to modify amount of replicas on both clusters:
* on first (production) cluster it was always reverted back to amount of replicas in the last commit of gitrepo
* on second (development) cluster it was using amount of replicas defined by user and it did not reconciled to a value defined in last commit
* After pressing `Force Update` the amount of replicas on second devel server was reverted to a value defined in last commit

PR #2014, opened 2023-12-14
add correctDrift to targetCustomization

raulcabello:
This PR overrides `correctDrift` if it appears in `targetCustomizations`. This allows user to override this setting per cluster.

same as https://github.com/rancher/fleet/pull/1982
refers to https://github.com/rancher/fleet/issues/1794


weyfonk (commented):
Why do we not update `charts/fleet-crd/templates/crds.yaml` here as done in #1982?

raulcabello:
> Why do we not update `charts/fleet-crd/templates/crds.yaml` here as done in #1982?

We changed the way we genereate crds. In fleet 0.8 we were using `wrangler`, and we use `controller-gen` (from kubebuilder) now. 
Wrangler sets `nullable` for pointers, and controller-gen don't. The difference is that with controller-gen the field will be pruned if not specified, but the struct will be nil in the code so I don't think that makes a difference. [More info](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#defaulting-and-nullable)

weyfonk:
> > Why do we not update `charts/fleet-crd/templates/crds.yaml` here as done in #1982?
> 
> We changed the way we genereate crds. In fleet 0.8 we were using `wrangler`, and we use `controller-gen` (from kubebuilder) now. Wrangler sets `nullable` for pointers, and controller-gen don't. The difference is that with controller-gen the field will be pruned if not specified, but the struct will be nil in the code so I don't think that makes a difference. [More info](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#defaulting-and-nullable)

Nice, thanks for clarifying :+1: !

weyfonk (approved):

PR #2015, opened 2023-12-15
Adapt release script to build.yaml

thardeck:
Refers to https://github.com/rancher/rancher/pull/42000

### DIFF
--- a/.github/scripts/release-against-rancher.sh
+++ b/.github/scripts/release-against-rancher.sh
@@ -30,7 +30,8 @@ fi
 
 if [ -e build.yaml ]; then
     sed -i -e "s/fleetVersion: .*$/fleetVersion: ${NEW_CHART_VERSION}+up${NEW_FLEET_VERSION}/" build.yaml
-    git add build.yaml
+    go generate
+    git add build.yaml pkg/buildconfig/constants.go
 else
     sed -i -e "s/ENV CATTLE_FLEET_VERSION=.*$/ENV CATTLE_FLEET_VERSION=${NEW_CHART_VERSION}+up${NEW_FLEET_VERSION}/" package/Dockerfile
     sed -i -e "s/ENV CATTLE_FLEET_MIN_VERSION=.*$/ENV CATTLE_FLEET_MIN_VERSION=${NEW_CHART_VERSION}+up${NEW_FLEET_VERSION}/" package/Dockerfile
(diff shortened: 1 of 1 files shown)


.github/scripts/release-against-rancher.sh
```diff
@@ -28,9 +28,14 @@ if [ ! -e ~/.gitconfig ]; then
...
 fi
 
-sed -i -e "s/ENV CATTLE_FLEET_VERSION=.*$/ENV CATTLE_FLEET_VERSION=${NEW_CHART_VERSION}+up${NEW_FLEET_VERSION}/g" package/Dockerfile
-
-git add package/Dockerfile
+if [ -e build.yaml ]; then
+    sed -i -e "s/fleetVersion: .*$/fleetVersion: ${NEW_CHART_VERSION}+up${NEW_FLEET_VERSION}/" build.yaml
+    git add build.yaml
+else
+    sed -i -e "s/ENV CATTLE_FLEET_VERSION=.*$/ENV CATTLE_FLEET_VERSION=${NEW_CHART_VERSION}+up${NEW_FLEET_VERSION}/" package/Dockerfile
```
weyfonk, line 35: Nice touch, ensuring backward compatibility is a good idea :)

.github/scripts/release-against-rancher.sh
```diff
@@ -28,9 +28,14 @@ if [ ! -e ~/.gitconfig ]; then
     git config --global user.email fleet@suse.de
 fi
 
-sed -i -e "s/ENV CATTLE_FLEET_VERSION=.*$/ENV CATTLE_FLEET_VERSION=${NEW_CHART_VERSION}+up${NEW_FLEET_VERSION}/g" package/Dockerfile
-
-git add package/Dockerfile
+if [ -e build.yaml ]; then
+    sed -i -e "s/fleetVersion: .*$/fleetVersion: ${NEW_CHART_VERSION}+up${NEW_FLEET_VERSION}/" build.yaml
+    git add build.yaml
```
weyfonk, line 33: As per https://github.com/rancher/rancher/blob/release/v2.8/docs/build.md#update-an-existing-value, editing `build.yaml` is not enough. We also need to regenerate the constants file with versions:
```suggestion
    go generate
    git add build.yaml pkg/buildconfig/constants.go
```

weyfonk (changes requested):
Almost there, thanks for taking care of this :)

weyfonk (approved):

PR #2031, opened 2023-12-20
[v0.9] Reduce BundleDeployment triggering on deployed resources updates

aruiz14:
Refers to #2058

These changes aim for reducing the number of time that the BundleDeployment reconciler is executed. This happens due to the Fleet Agent configuring a watcher on every deployed resources, which will trigger a requeue of the BundleDeployment handler in the agent for every watch event. Since this handler refreshes the conditions timestamp in the BundleDeployment status, this has a direct effect on the Fleet controller as well, which will react to the update in the BundleDeployment object.

This patch consists of the following changes:
 - The `TriggerSleep` delay has been removed, since it could cause modification events happening in that time window to be missed (this was the case for the integration tests). This is connected to the next change:
 - `ADDED` watch events are now ignored:
   - The informer creates an initial `ADDED` event for every object, but this does not make sense for objects that were just created. 
   - The only interesting events are `MODIFIED` and `DELETED`, which are the ultimate interest of the trigger (detecting modified or deleted resources). If an object is deleted and later created by the user, the controller may still need to redeploy in order to achieve a stable situation.
- Filters modified events based on `Generation` field, for those kinds that make use of it.
  - Kinds that do not implement this metadata field (e.g. ConfigMaps) will keep the existing behavior.
  - Resources that set the `Generation` field will be checked: based on their UID, if `Generation` didn't change from a previous event, it will be interpreted as the spec didn't change, so will skip the trigger, effectively omitting status updates.

### DIFF
--- a/integrationtests/agent/watcher_trigger_test.go
+++ b/integrationtests/agent/watcher_trigger_test.go
@@ -178,7 +178,7 @@ var _ = Describe("Watches for deployed resources", Ordered, func() {
 				return *triggerCount
 			}).Should(Equal(0))
 		})
-		It("is triggered Spec is changed", func() {
+		It("is triggered on Spec changes", func() {
 			deploy.Spec.Paused = true
 			for i := 1; i <= 5; i++ {
 				deploy.Spec.Replicas = pointer.Int32(int32(i))
--- a/integrationtests/agent/watcher_trigger_test.go
+++ b/integrationtests/agent/watcher_trigger_test.go
@@ -1,0 +1,194 @@
+package agent
+
+import (
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/utils/pointer"
+
+	"github.com/rancher/fleet/internal/cmd/agent/trigger"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ = Describe("Watches for deployed resources", Ordered, func() {
+
+	var (
+		env    *specEnv
+		triggr *trigger.Trigger
+	)
+
+	BeforeAll(func() {
+		env = specEnvs["watchertrigger"]
+		triggr = trigger.New(ctx, env.k8sClient.RESTMapper(), dynamic.NewForConfigOrDie(cfg))
+		DeferCleanup(func() {
+			Expect(k8sClient.Delete(ctx, &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: env.namespace}})).ToNot(HaveOccurred())
+		})
+	})
+
+	registerResource := func(key string, objs ...runtime.Object) *int {
+		_ = triggr.Clear(key)
+		var count int
+		Expect(triggr.OnChange(key, env.namespace, func() {
+			count++
+		}, objs...)).ToNot(HaveOccurred())
+		return &count
+	}
+
+	When("watching a deployed configmap", Ordered, func() {
+		createConfigMap := func() *corev1.ConfigMap {
+			cm := corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					GenerateName: "test-cm-",
+					Namespace:    env.namespace,
+				},
+			}
+
+			err := env.k8sClient.Create(ctx, &cm)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(cm.UID).ToNot(BeEmpty())
+			return &cm
+		}
+
+		var triggerCount *int
+		var cm *corev1.ConfigMap
+		BeforeEach(func() {
+			cm = createConfigMap()
+			triggerCount = registerResource(env.namespace+"/test-configmap", cm)
+			DeferCleanup(func() {
+				Expect(
+					client.IgnoreNotFound(env.k8sClient.Delete(ctx, cm))).
+					NotTo(HaveOccurred())
+			})
+		})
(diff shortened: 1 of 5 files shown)


raulcabello:
I like the approach. This introduces two improvements:
- Removing `ADDED` -> removes an initial enqueue that happens for all resources being watched as the informer creates an initial ADDED event for each resource.
- Filters modified events based on Generation field -> removes a lot of enqueues that happens when status fields are changed. We ignore status changes when detecting drift in [wrangler](https://github.com/rancher/wrangler/blob/master/pkg/apply/desiredset_compare.go#L135-L138)

I don't see any issue with this approach. However, there is a possibility that these extra enqueues are needed in some edge cases we don't see at the moment. So, I wouldn't add this to the stable release (0.8)

internal/cmd/agent/trigger/watcher.go
```diff
@@ -125,10 +131,36 @@ func (t *Trigger) OnChange(key string, defaultNamespace string, trigger func(),
 	return nil
 }
 
-func (t *Trigger) call(gvk schema.GroupVersionKind, key objectset.ObjectKey) {
+func (t *Trigger) call(gvk schema.GroupVersionKind, obj metav1.Object, deleted bool) {
```
weyfonk, line 134: How feasible would it be to cover this with tests? I for one would then feel more comfortable that this reduces triggering without introducing any important, undesired side-effects.
eg.:
* deleted object → triggers called
* added object → triggers not called
* modified object with no generation field → triggers called
* modified object with generation field set to 0 → triggers not called
* modified object with generation field updated → triggers called (possibly repeated a couple of times to validate that subsequent generation updates have the desired effect)

aruiz14: Tests added, thanks for the suggestion. It made me find that there was a corner case in which a trigger would happen but is not desired. Please take another look!

aruiz14: One of the tests is flaky, I'm looking into it

integrationtests/agent/watcher_trigger_test.go
```diff
@@ -0,0 +1,194 @@
...
+		It("is not initially triggered", func() {
+			Consistently(func() int {
+				return *triggerCount
+			}).Should(Equal(0))
+		})
+		It("is triggered on deletion", func() {
+			Expect(env.k8sClient.Delete(ctx, cm)).ToNot(HaveOccurred())
+			Eventually(func() int {
+				return *triggerCount
+			}).WithPolling(100 * time.Millisecond).MustPassRepeatedly(3).
```
weyfonk, line 79: Out of curiosity, where does this `3` come from?

integrationtests/agent/watcher_trigger_test.go
```diff
@@ -0,0 +1,194 @@
...
+					Message:            "tests",
+					Reason:             "tests",
+				},
+			}
+			Expect(env.k8sClient.Status().Update(ctx, deploy)).ToNot(HaveOccurred())
+			Consistently(func() int {
+				return *triggerCount
+			}).Should(Equal(0))
+		})
+		It("is triggered Spec is changed", func() {
```
weyfonk, line 182: nit:
```suggestion
		It("is triggered when Spec is changed", func() {
```
or (more consistent with the previous test description):
```suggestion
		It("is triggered on Spec changes", func() {
```

weyfonk (approved):
LGTM, thanks for taking the time to add integration tests for this :relaxed:

integrationtests/agent/watcher_trigger_test.go
aruiz14: It's a way to express that the trigger count should eventually increase from 0 -> 1, and not increase any longer. In other words, just executed once.
The actual 3 and the polling interval are totally made up values, I didn't want to slow down the tests too much either.
Happy to apply any suggestions you may have!

internal/cmd/agent/trigger/watcher.go
```diff
@@ -125,10 +132,42 @@ func (t *Trigger) OnChange(key string, defaultNamespace string, trigger func(),
...
+			t.seenGenerations.Delete(uid)
+		} else {
+			// keep a map of UID -> generation, using sync.Map and atomic.Int64 for safe concurrent usage
+			// - sync.Map entries are never modified after created, a pointer is used as value
+			// - using atomic.Int64 as values allows safely comparing and updating the current Generation value
+			var previous *atomic.Int64
+			if value, ok := t.seenGenerations.Load(uid); ok {
+				previous = value.(*atomic.Int64)
+			} else {
+				previous = t.storeObjectGeneration(uid, currentGeneration)
```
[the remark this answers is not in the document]
aruiz14: Hmm it makes sense, let me rework it.

internal/cmd/agent/trigger/watcher.go
aruiz14: Fixed in ad49a1638a0aa564300d3cd290e0dbf97da64e3c

integrationtests/agent/watcher_trigger_test.go
weyfonk: No worries, I was just wondering how you came to `3` instead of, say, 5 or 10 (for instance, also made up values).
3 works for me :)

internal/cmd/agent/trigger/watcher.go
```diff
@@ -125,10 +132,40 @@ func (t *Trigger) OnChange(key string, defaultNamespace string, trigger func(),
 	return nil
 }
 
-func (t *Trigger) call(gvk schema.GroupVersionKind, key objectset.ObjectKey) {
+func (t *Trigger) storeObjectGeneration(uid types.UID, generation int64) *atomic.Int64 {
+	value := new(atomic.Int64)
+	value.Store(generation)
+	t.seenGenerations.Store(uid, value)
+	return value
+}
```
weyfonk, line 140: I think this would make the linter happy:
```suggestion
func (t *Trigger) storeObjectGeneration(uid types.UID, generation int64) {
	value := new(atomic.Int64)
	value.Store(generation)
	t.seenGenerations.Store(uid, value)
}
```

PR #2033, opened 2023-12-21
Limit configmap cache&watch

manno:
Refers to #2018 

TODO
* this seems to limit `Get` for configmaps, too. Fails when gitrepo reconciler tries to `CreateOrUpdate` the `target.yaml` configmap
* `DefaultNamespaces` doesn't seem to work for : `"unable to list: test-7af124d because of unknown namespace for the cache"`

Opened: https://github.com/kubernetes-sigs/controller-runtime/issues/2628


manno:
The upstream issue has been closed. `metav1.NamespaceAll` use in `CacheOptions` seems to be intended to define different cache options for namespaces only.

However, using  a simpler cache config seems to limit the cache, which can be proven with some benchmarking.

PR #2045, opened 2024-01-05
Add missing serviceName in fleet-agent statefulset

p-se:
Refers to #2048

### DIFF
--- a/e2e/single-cluster/sharding_test.go
+++ b/e2e/single-cluster/sharding_test.go
@@ -100,36 +100,38 @@ var _ = Describe("Filtering events by shard", Label("infra-setup"), Ordered, fun
 				}).Should(ContainSubstring("sleeper-"))
 
 				for _, s := range shards {
-					logs, err := k.Namespace("cattle-fleet-system").Logs(
-						"-l",
-						"app=fleet-controller",
-						"-l",
-						fmt.Sprintf("shard=%s", s),
-					)
-					Expect(err).ToNot(HaveOccurred())
-					regexMatcher := matchers.MatchRegexpMatcher{
-						Regexp: fmt.Sprintf(`Reconciling GitRepo.*"name":"%s"`, gitrepoName),
-					}
-					hasReconciledGitRepo, err := regexMatcher.Match(logs)
-					Expect(err).ToNot(HaveOccurred())
-					if s == shard {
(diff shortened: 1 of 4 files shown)


internal/cmd/controller/agentmanagement/agent/manifest.go
```diff
@@ -115,9 +115,23 @@ func Manifest(namespace string, agentScope string, opts ManifestOptions) []runti
 		},
 	}
```
manno, line 117:
```suggestion

// added for compatibility with old helm versions, the service is not used and not necessary
```

manno:
Discussed, we don't need to support (now) very old versions of kubectl.

manno:
This adds a headless service to the fleet-agent statefulset.

    StatefulSets currently require a Headless Service to be responsible for the network identity of the Pods. You are responsible for creating this Service.

https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/

If the service is missing, Rancher shows a warning in the UI.

internal/cmd/controller/agentmanagement/agent/manifest.go
```diff
@@ -115,9 +115,25 @@ func Manifest(namespace string, agentScope string, opts ManifestOptions) []runti
...
 
+	// Added for compatibility with old helm versions, the service is not used
+	// and not necessary.
```
manno, line 119:
```suggestion
	// Statefulsets require a headless service. The service is not used.
```

e2e/single-cluster/sharding_test.go
```diff
@@ -100,36 +100,38 @@ var _ = Describe("Filtering events by shard", Label("infra-setup"), Ordered, fun
...
-					} else {
-						Expect(hasReconciledGitRepo).To(BeFalseBecause(
-							"GitRepo %q labeled with shard %q should not have been"+
-								" deployed by controller for shard %q",
-							gitrepoName,
-							shard,
-							s,
-						))
-					}
+					Eventually(func(g Gomega) {
```
weyfonk, line 103: Nice one, thanks for looking into this :muscle:

p-se: Glad you like it :slightly_smiling_face:

manno (approved):

Issue #2041, opened 2024-01-03
CVE Fixes for Rancher 2.8.3


kkaempf:
# SURE-7404

kkaempf:
done

PR #2067, opened 2024-01-12
Bump to gitjob v0.8.0

manno:
Refers to https://github.com/rancher/fleet/issues/2041


weyfonk (commented):
Why get rid of some CI workflows?

manno:
> Why get rid of some CI workflows?

They're broken and do not run automatically in release branches.

weyfonk (approved):
LGTM
As discussed offline, "broken" workflows here means that those CI workflows do not work in their current state with the `toolchain` directive in `go.mod`.
We can backport those workflows if/when needed, which for now seems unclear, hence we decided to spare ourselves the effort to adapt them.

PR #2074, opened 2024-01-15
Bump fleet-agent chart's kubectl to 1.29

manno:
This kubectl is only used by the standalone chart.

Maybe this should be same image as the one used by Rancher to limit downloads?


weyfonk (approved):

weyfonk:
> Maybe this should be same image as the one used by Rancher to limit downloads?

That sounds like a good idea; I wonder how fast that target moves, and whether a single version is used by Rancher itself for a given release (one would hope so).

PR #2081, opened 2024-01-16
Release scripts chart dir var

manno:
Ideally the `.github/scripts` don't contain much configuration complexity and tailored to the CI.
However, we do execute this locally and duping it into dev seems unnecessary, when I can just remove a variable and an if condition instead.

This also allows me to use any charts dir, like `charts-2.7`.


weyfonk (approved):
LGTM

PR #2086, opened 2024-01-17
Remove k8s replace from go.mod

manno:


go.mod
```diff
@@ -3,36 +3,9 @@ module github.com/rancher/fleet
 go 1.21
 
 replace (
+	github.com/json-iterator/go v1.1.10 => github.com/json-iterator/go v1.1.12
```
aruiz14, line 6: Why is this needed? Shouldn't `go get github.com/json-iterator/go@v1.1.12` use that version already?

manno: We want to replace the vulnerable version in all transient dependencies. There were a few left in the `go mod graph`. I just moved the directive to the top of the file.

aruiz14: Oh, I didn't know about the vulnerability.

aruiz14 (approved):

go.mod
manno: Removing it like we discussed.

aruiz14 (approved):

Issue #2008, opened 2023-12-11
Merge Gitjob Controller into Fleet Controller / Chart


manno:
[rancher/gitjob](https://github.com/rancher/gitjob) is only used by Fleet, and adds to its maintenance/release overhead, both for Go code and for charts.
Instead, we could integrate `rancher/gitjob` contents into `rancher/fleet` and manage all Fleet-related logic from there.

Note:
Merging gitjob into the same code base adds some interesting possibilities in the future. We might be able to remove the gitjob CRD, the gitjob container image, so we have less complexity and a smaller download size.
This will become clearer when be build OCI repo support, but gitjob/gitrepo are as much their own component as oci repo is. So the split between gitjob vs gitrepo doesn’t make sense. Eventually, we might have “bundle” producing components (gitrepo, ocirepo) and a component which distributes bundles to clusters (fleetcontroller/targetting).
So, in short, we’re moving gitjob now, so maintenance and releases get easier, but it will also enable a clearer design in the future.

This would require us to:
- [x] move tests
- [x] remove the `gitjob` helm chart, placing it in the `fleet` chart (no more sub chart)
- [x] move the gitjob controller into fleet, keep it in a separate pod (it polls an external resource)
- [x] adapt drone workflow to publish image
- [ ] remove "subchart config" from rancher/charts a
- [x] optional, move into same directory struct as fleet (`internal/cmd/gitjob`)
- [x] can we get rid of the `GitJob` resource in the future and use GitRepo instead?
- [x] keep the webhook in the same pod as gitjob

## More housekeeping (in no particular order)
- [x] remove updatecli job
- [x] check `.gitignore`
- [x] update references to `rancher/gitjob` in docs throughout `rancher/fleet`
- [x] remove script `.github/scripts/check-for-gitjob-changes.sh`

weyfonk:
Once #2098 is merged:
* Cherry-pick [gitjob#421](https://github.com/rancher/gitjob/pull/421)
* Cherry-pick [gitjob#422](https://github.com/rancher/gitjob/pull/422)

weyfonk:
When next releasing Fleet from a branch including #2098:
- [ ] Remove [packages/fleet/gitjob](https://github.com/rancher/charts/tree/dev-v2.9/packages/fleet/gitjob) from `rancher/charts`' target branch (eg. by editing the PR generated by [release-against-charts](https://github.com/rancher/fleet/actions/workflows/release-against-charts.yml))
[...]

PR #2098, opened 2024-01-23
Move `rancher/gitjob` logic to this repository

weyfonk:
This PR moves `rancher/gitjob` logic to the Fleet controller chart.
More specifically:
* `gitjob` now lives in its own pod, via new Docker image `fleet-gitjob`. Drone now publishes that new image beside Fleet's.
* all manifests from `rancher/gitjob` now live in the `fleet` chart (controller chart)
* Fleet no longer depends on a separate `gitjob` chart.  CRDs coming from gitjob are kept in `gitjob.cattle.io` though, to preserve compatibility with existing Fleet deployments.
* the Updatecli job and its dependencies, monitoring `rancher/gitjob` as a dependency is removed.
* Fleet release docs are updated to reflect `rancher/gitjob` no longer being a dependency for this repository.

Also, gitjob integration tests are more lightweight than they were in `rancher/gitjob`, starting fewer `gogs` containers. This makes those tests less likely to fail in CI.

_Note_: Importing `rancher/gitjob` history comes with the following caveats:
* Commit SHAs are not preserved
* Accessing pre-import history requires using the `--follow` flag, eg. `git log --follow -- pkg/apis/gitjob.cattle.io/v1/doc.go` will show the file's full history, whereas `git log pkg/apis/gitjob.cattle.io/v1/doc.go` will only display history from the import commit onwards.

Option `disable-gitops`, which would determine whether the `gitjob` chart would be installed, and still determines whether to start the `GitRepo` reconciler, is kept in place.

Refers to #2008


.github/scripts/build-fleet-binaries.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+# Description: build fleet binary and image with debug flags
+
+set -euxo pipefail
+
+if [ ! -d ./cmd/fleetcontroller ]; then
+  echo "please change the current directory to the fleet repo checkout"
+  exit 1
+fi
```
manno, line 9:
```suggestion
```

Not needed as CI is always in the right directory. I'd remove it so we don't have to maintain the path to `cmd/fleetcontroller`.

.github/scripts/build-fleet-binaries.sh
```diff
@@ -0,0 +1,29 @@
...
+
+export GOARCH="${GOARCH:-amd64}"
+export CGO_ENABLED=0
+
+# re-generate code
+if ! git diff --quiet HEAD origin/master --  pkg/apis/fleet.cattle.io/v1alpha1; then
+  go generate
+fi
```
manno, line 17:
```suggestion
```

I'd also remove this. We have a check in another action, to protect against a missing generate, I think. In any case, the CI shouldn't generate code, but use the already generated code.

weyfonk: Correct, [this workflow](https://github.com/rancher/fleet/blob/master/.github/workflows/check-changes.yml) is called against pull requests and already checks for this. Removed, thanks.

manno (approved):

Issue #2110, opened 2024-01-31
CLI to create bundledeployments and deploy charts


manno:
These commands help with debugging bundles. The deploy command can be used to bring bundles to air-gapped clusters.   More information on how to create bundles with `fleet apply` from folders can be found in the [blog post on external-secrets](https://www.suse.com/c/rancher_blog/fleet-multi-cluster-deployment-with-the-help-of-external-secrets/) and the [fleet docs on bundles](https://fleet.rancher.io/bundle-add).

- [x] https://github.com/rancher/fleet/pull/2111
- [x] https://github.com/rancher/fleet/pull/2102
- [x] https://github.com/rancher/fleet/pull/2113 (optional)
- [x] https://github.com/rancher/fleet/pull/2112 (optional)

# Documentation

Target reads a bundle from a file and works with a live cluster to print out the bundledeployment
/content resource, which fleetcontroller would create. It takes a namespace as an argument, so it can look in that namespace for e.g. cluster resources. It can also dump the data structure which is used during "targeting", so decisions taken regarding labels and cluster names can be checked.

```
% fleet target --help
Print available targets for a bundle

Usage:
  fleet target [flags]

Flags:
  -b, --bundle-file string                Location of the Bundle resource yaml
      --kubeconfig string                 Paths to a kubeconfig. Only required if out-of-cluster.
  -l, --list-inputs                       List the live resources, which are used in targeting, like clusters, as YAML
  -n, --namespace string                  Override the namespace of the bundle. Targeting searches this namespace for clusters.
```

The second command "deploy" takes the output of `fleet target`, or a dumped bundledeployment/content resource and deploys it to a cluster, just like fleet-agent would. It supports a dry run mode, to print out the resources which would be created, instead of installing them with helm. Since the command doesn't create the input resources, a running fleet-agent would likely garbage collect the deployment. I'm still investigating how to deal with manually created deployments. I think adopt or ignore are the most useful options.

```
% fleet deploy --help                                                                                                                                                                                                                                         
Deploy a bundledeployment/content resource to a cluster

Usage:
  fleet deploy [flags]

Flags:
  -a, --agent-namespace string            Set the agent namespace, normally cattle-fleet-system. If set, fleet agent will garbage collect the helm release, i.e. delete it if the bundledeployment is missing.
  -d, --dry-run                           Print the resources that would be deployed, but do not actually deploy them
  -i, --input-file string                 Location of the YAML file containing the content and the bundledeployment resource
[...]

PR #2112, opened 2024-01-31
Fix helm deployer ignoring kubeconfig CLI arg

manno:
Noticed during https://github.com/rancher/fleet/issues/2110

The helmdeployer does not consider `--kubeconfig`.  We don't seem to use the flag, except in upcoming CLI tests to target the testenv cluster.

### DIFF
--- a/internal/cmd/agent/operator.go
+++ b/internal/cmd/agent/operator.go
@@ -152,10 +152,9 @@ func newReconciler(ctx, localCtx context.Context, mgr manager.Manager, localConf
 	}
 	localClient := localCluster.GetClient()
 
-	kubeconfig := flag.Lookup("kubeconfig").Value.String()
-	if kubeconfig != "" {
+	if kubeconfig := flag.Lookup("kubeconfig").Value.String(); kubeconfig != "" {
 		// set KUBECONFIG env var so helm can find it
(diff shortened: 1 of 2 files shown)


internal/cmd/agent/operator.go
```diff
@@ -151,6 +152,12 @@ func newReconciler(ctx, localCtx context.Context, mgr manager.Manager, localConf
 	}
 	localClient := localCluster.GetClient()
 
+	kubeconfig := flag.Lookup("kubeconfig").Value.String()
+	if kubeconfig != "" {
```
weyfonk, line 156: nit:
```suggestion
	if kubeconfig := flag.Lookup("kubeconfig").Value.String(); kubeconfig != "" {
```

internal/cmd/agent/operator.go
```diff
@@ -151,6 +152,12 @@ func newReconciler(ctx, localCtx context.Context, mgr manager.Manager, localConf
...
+		// set KUBECONFIG env var so helm can find it
+		os.Setenv("KUBECONFIG", flag.Lookup("kubeconfig").Value.String())
```
weyfonk, line 158: Why do the lookup again?
```suggestion
		os.Setenv("KUBECONFIG", kubeconfig)
```

weyfonk (approved):

PR #2129, opened 2024-02-07
Remove `GitJob` resource

weyfonk:
Refers to #2008

Resource `GitJob` is only ever used when calling `fleet apply` from the `GitRepo` reconciler, as a proxy to store values either inherited or computed from `GitRepo`.
Instead, the idea implemented here is that the `GitRepo` reconciler can leave k8s job creation to a separate reconciler (previously the `GitJob` reconciler), which now also uses `GitRepo` as input but deals with reconciling (creating, updating, deleting) those jobs directly from `GitRepo`s.

A couple of doubts remain here:
* are status updates done properly? Tests pass, but observed statuses and their conditions might have changed in uncovered ways
* instantiating a k8s job running `fleet apply` requires a Fleet agent image and its pull policy as input. When job specs were created from the `GitRepo` reconciler, those values were available from configuration, reconciled within the Fleet controller pod. However, with this PR, job creation logic now lives in the `gitjob` pod, where configuration is not reconciled (unless we decide to instantiate the configuration controller there as well). In the meantime, the default Fleet agent image is used (`rancher/fleet-agent:$version`) with pull policy `IfNotPresent`. Do we still need those values to be configurable? I do not see value for our users in this.


internal/cmd/controller/gitjob/gitjob.go
```diff
@@ -305,19 +319,300 @@ func (r *GitJobReconciler) newJob(ctx context.Context, obj *v1.GitJob) (*batchv1
...
-func (r *GitJobReconciler) generateInitContainer(ctx context.Context, obj *v1.GitJob) (corev1.Container, error) {
-	args := []string{obj.Spec.Git.Repo, "/workspace"}
+func (r *GitJobReconciler) computeJobSpec(ctx context.Context, gitrepo *v1alpha1.GitRepo) (*batchv1.JobSpec, error) {
+	paths := gitrepo.Spec.Paths
+	if len(paths) == 0 {
+		paths = []string{"."}
+	}
+
+	// XXX: how much sense does it make to recompute the whole configmap here just to extract its name? Can we do
+	// better/simpler?
```
manno, line 336:
```suggestion
	// calculate configmap content, because the name contains a hash
```

Yes, this seems weird. The config map is created by the other controller and the name contains a hash over the data.
Maybe we can fix to something fixed, like `gitrepo.name + gitrepo.commit`, but the current approach seems more fault tolerant, i.e. target configmaps from older/newer gitrepos would still work if the data didn't change?

Maybe we just document the current situation and move on..

internal/cmd/controller/gitrepo/targetsyaml.go
```diff
@@ -12,6 +12,9 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// XXX: move this to internal/cmd/controller/target?
+// Or reconsider the idea of rebuilding job specs altogether?
+
```
manno, line 17:
```suggestion
```

If we want to be super correct, since this is now used by two different binaries, it should live in `internal/`.
But I already ignored the structure when I created the CLI for `target` and didn't move `internal/cmd/controller/target`

pkg/webhook/webhook_test.go
```diff
@@ -47,23 +47,21 @@ func TestGetBranchTagFromRef(t *testing.T) {
 func TestAzureDevopsWebhook(t *testing.T) {
 	const commit = "f00c3a181697bb3829a6462e931c7456bbed557b"
 	const repoURL = "https://dev.azure.com/fleet/git-test/_git/git-test"
-	gitjob := &v1.GitJob{
+	gitRepo := &v1alpha1.GitRepo{ // TODO update this to make tests pass
```
manno, line 50: Can this comment be removed?

manno:
PR failed due to a flaky test. 

This seems to fix it, ran 63 times without failure.

```
diff --git a/integrationtests/controller/gitrepo/gitrepo_test.go b/integrationtests/controller/gitrepo/gitrepo_test.go
index 0b8ddc04..acdb3a32 100644
--- a/integrationtests/controller/gitrepo/gitrepo_test.go
+++ b/integrationtests/controller/gitrepo/gitrepo_test.go
@@ -70,11 +70,11 @@ var _ = Describe("GitRepo", Ordered, func() {
                        }).Should(BeTrue())
                })

-               It("updates the gitrepo status", func() {
+               FIt("updates the gitrepo status", func() {
                        org := gitrepo.ResourceVersion
                        Eventually(func() bool {
                                _ = k8sClient.Get(ctx, types.NamespacedName{Name: "test-gitrepo", Namespace: namespace}, gitrepo)
-                               return gitrepo.ResourceVersion > org
+                               return gitrepo.ResourceVersion > org && gitrepo.Status.Display.State == "GitUpdating"
                        }).Should(BeTrue())

                        Expect(gitrepo.Status.Display.ReadyBundleDeployments).To(Equal("0/0"))
```

pkg/webhook/webhook_test.go
weyfonk: Oops, thanks for spotting this. Of course it could :o)

internal/cmd/controller/gitjob/gitjob.go
weyfonk: Agreed, documentation is a good approach for now. Thanks!

internal/cmd/controller/gitrepo/targetsyaml.go
weyfonk: Yes, I wrote this comment when I thought that we might be building job specs from 2 different places, which is of course not necessary. Comment removed.

weyfonk:
> PR failed due to a flaky test.
It wasn't flaky, just plain failing :D
Thanks for the heads-up though, that was a heavy-handed commit, now reverted and improved.

manno (approved):

PR #2137, opened 2024-02-09
Infra e2e command uses a published fleet module

manno:
We have trouble updating our `go.mod`. Often, we need to update `e2e/testenv/infra/go.mod` when the main `go.mod` updates.

That's because the infra module imports the local fleet module via `replace github.com/rancher/fleet => ../../../`.

This PR imports a published go.mod instead, since the infra command only uses the testenv helpers from e2e, the dependency does not have to be updated often.

When making changes to the infra command, one might want to edit and use the helpers directly. Without the need to tag a version for every change. Go workspaces can help:

```
cd e2e/testenv/infra
go work init
go work use ../../../
go work use .
```

This is similar to a replace statement. Never commit the `go.work` file, though.
It might still be necessary split the PR into two, one for the helpers and a second to update the infra command and switch infra's go.mod to the commit of the first PR.

### DIFF
--- a/e2e/testenv/infra/go.mod
+++ b/e2e/testenv/infra/go.mod
@@ -4,14 +4,10 @@ go 1.21
 
 toolchain go1.21.0
 
-replace github.com/rancher/fleet => ../../../
-
-replace github.com/rancher/fleet/pkg/apis => ../../../pkg/apis
-
 replace k8s.io/client-go => k8s.io/client-go v0.29.0
 
 require (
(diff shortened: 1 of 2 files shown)


weyfonk (approved):
As an alternative to splitting this PR into 2, I'd be happy with some doc (either in `dev/README.md` for instance, or in a commit message), explaining where the new `rancher/fleet` tag in `e2e/testenv/infra/go.mod` comes from and how to use Go workspaces as in the PR description.

PR #2146, opened 2024-02-13
Update helm fork, rename ForceAdopt/Adopt to TakeOwnership

manno:
In tag `v3.14.0-fleet2`, `rancher/helm` replaces `ForceAdopt` with `TakeOwnership` to clarify the purpose of that feature.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.21
 
 replace (
 	github.com/rancher/fleet/pkg/apis => ./pkg/apis
-	helm.sh/helm/v3 => github.com/rancher/helm/v3 v3.14.0-fleet1
+	helm.sh/helm/v3 => github.com/rancher/helm/v3 v3.14.0-fleet2
 )
 
 require (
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -4,7 +4,7 @@ go 1.21
 
 replace (
 	github.com/rancher/fleet/pkg/apis => ./pkg/apis
-	helm.sh/helm/v3 => github.com/rancher/helm/v3 v3.14.0-fleet1
+	helm.sh/helm/v3 => github.com/rancher/helm/v3 v3.14.0-fleet2
```
weyfonk, line 7: Is there a specific reason for updating this tag? Just curious.

weyfonk (approved):

go.mod
manno: fleet2 has one more commit than fleet1: https://github.com/rancher/helm/commits/v3.14.0-fleet2/

manno: Yes, because I explain it wrong every time ;)

Issue #2121, opened 2024-02-05
Replace Drone Pipeline With Github Workflow


manno:
We can start removing drone/dapper from Fleet. Looking at .drone.yaml we need a Github workflow to do the following when a release tag is pushed:
* run FOSSA check
* create Github release
* run some basic tests on the release
* build the binaries
* build the images, tag them and push them

The basic tests are currently unit tests and a check against ‘go generate’, ‘go mod verify’.
In the future we could do linting, unit, and integration.

We’d need to backport the final workflow to supported release branches, like release/v0.9

We already have an experiment with [goreleaser](https://github.com/goreleaser/goreleaser), which was slow and ran into trouble with windows container images: https://github.com/rancher/fleet/pull/1019/files

The nightly workflow builds and pushes images to [ghcr.io](http://ghcr.io/) using buildx: https://github.com/rancher/fleet/blob/04492e3f505aa6b30371623f383b6bb8c7a3701d/.github/workflows/nightly.yml

While we want to push to dockerhub, it might be easier to test with ghcr.io. Supporting both could be easy.

manno:
- [ ] Do we need to build windows images for the agent? Fleet-agent is configured not to start on Windows nodes.
- [ ] Can goreleaser use buildx to build windows images?
- [ ] Need a separate issue/PR to remove drone/dapper.

weyfonk:
Remaining work on images, including getting rid of the Windows-based agent Dockerfile, to be done as part of #2184.

PR #2168, opened 2024-02-20
Replace Drone pipeline with Github actions

weyfonk:
This provides a new `release-fleet.yaml` CI workflow to replace our `.drone.yml` pipelines.
[Here](https://github.com/weyfonk/fleet/releases/tag/v42.0.26-test) is an example of a release generated by this new workflow.

Refers to #2121

## Additions
* Linting
* Integration tests
* `gitjob` and `gitcloner` binaries

## Limitations/Possible improvements
* Single checksum file, as per GoReleaser limitation (this may change [soon](https://github.com/goreleaser/goreleaser/issues/4618)); could be mitigated by using a separate workflow step to compute checksums by target arch
* [No Windows-based images](https://github.com/orgs/goreleaser/discussions/2440) for now, although there may be a way (see [experimental branch](https://github.com/weyfonk/fleet/tree/2121-build-windows-images)) involving a patch to GoReleaser
    * Further discussion suggested that we may not actually _need_ Windows images, because we require at least one Linux node in each cluster, and the agent is not scheduled to run on Windows nodes anyway.
* Fleet binaries are built _twice_ for the Github runner platform: once to run the tests, then again by GoReleaser to prepare a release.

## Doubts
* How much value is there in releasing `gitjob` and `gitcloner` binaries on their own? Would having them in `rancher/gitjob` images be enough?

## Left to do
* Remove `.drone.yml` and Dapper `./scripts/`

### DIFF
--- a/.github/workflows/release-fleet.yml
+++ b/.github/workflows/release-fleet.yml
@@ -1,0 +1,110 @@
+# Fleet release workflow
+name: Fleet release
+
+on:
+  push:
+    tags:
+      - v**
+
+permissions:
+  contents: write
+  packages: write
+
+env:
+  GOARCH: amd64
+  CGO_ENABLED: 0
+  SETUP_K3D_VERSION: 'v5.5.1'
+
+jobs:
+  build-fleet:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out Fleet
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
+          check-latest: true
+
+      - name: Run golangci-lint
+        uses: golangci/golangci-lint-action@v4.0.0
+        with:
+          # Required: the version of golangci-lint is required and must be specified without patch version: we always
+          # use the latest patch version.
+          version: v1.55
+          args: --timeout=10m --config=.golangci.json
+
+      - name: Run FOSSA scan
+        uses: fossas/fossa-action@v1.3.1
+        with:
+          api-key: ${{ secrets.FOSSA_API_KEY }}
+
+      - name: Run FOSSA tests
+        uses: fossas/fossa-action@v1.3.1
+        with:
+          api-key: ${{ secrets.FOSSA_API_KEY }}
+          run-tests: true
+
+      - name: Check for code changes
+        run: |
+          ./.github/scripts/check-for-auto-generated-changes.sh
+          go mod verify
+
+      - name: Run unit tests
+        run: go test -cover -tags=test $(go list ./... | grep -v -e /e2e -e /integrationtests)
+
+      - name: Run integration tests
+        env:
+          SETUP_ENVTEST_VER: v0.0.0-20240115093953-9e6e3b144a69
+          ENVTEST_K8S_VERSION: 1.28
+        run: ./.github/scripts/run-integration-tests.sh
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Log into Container registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v5
+        with:
+          distribution: goreleaser
+          version: v1.24.0
+          args: release --clean --verbose
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Upload charts to release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          repo: "ghcr.io/rancher/fleet"
+          tag: "${GITHUB_REF_NAME}"
+        run: |
+          sed -i \
+              -e "s@repository:.*@repository: $repo@" \
+              -e "s/tag:.*/tag: ${VERSION}/" \
+              charts/fleet/values.yaml
+
+          sed -i \
+              -e "s@repository:.*@repository: $repo@" \
+              -e "s/tag: dev/tag: ${VERSION}/" \
+              charts/fleet-agent/values.yaml
+
+          helm package --version="$VERSION" --app-version="$VERSION" -d ./dist ./charts/fleet
+          helm package --version="$VERSION" --app-version="$VERSION" -d ./dist ./charts/fleet-crd
+          helm package --version="$VERSION" --app-version="$VERSION" -d ./dist ./charts/fleet-agent
+
+          for f in $(find dist/artifacts/ -name '*.tgz'); do
+            gh release upload $tag $f
+          done
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,0 +1,268 @@
+---
+project_name: fleet
+
+before:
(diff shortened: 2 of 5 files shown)


.github/workflows/release-fleet.yml
```diff
@@ -0,0 +1,107 @@
...
+          args: release --clean --verbose
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Upload charts to release
+        if: ${{ startsWith(github.ref, 'refs/tags/') }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          ./scripts/package-helm
```
manno, line 103: This script is a bit dated. I prefer the approach taken in [nightly](https://github.com/rancher/fleet/blob/release/v0.9/.github/workflows/nightly.yml#L111-L126). Less `sed` and don't regenerate go files, crds.

.github/workflows/release-fleet.yml
```diff
@@ -0,0 +1,107 @@
...
+          distribution: goreleaser
+          version: v1.24.0
+          args: release --clean --verbose
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Upload charts to release
+        if: ${{ startsWith(github.ref, 'refs/tags/') }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
manno, line 101:
```suggestion
```

Seems unnecessary

.github/workflows/release-fleet.yml
```diff
@@ -0,0 +1,107 @@
...
+        with:
+          distribution: goreleaser
+          version: v1.24.0
+          args: release --clean --verbose
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Upload charts to release
+        if: ${{ startsWith(github.ref, 'refs/tags/') }}
+        env:
```
manno, line 100:
```suggestion
        env:
            repo="ghcr.io/rancher/fleet"
            tag: "${{github.ref_name}}"
```

.github/workflows/release-fleet.yml
```diff
@@ -0,0 +1,107 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          ./scripts/package-helm
+          tag=${GITHUB_REF#refs/tags/}
```
manno, line 104:
```suggestion
```

Can be set from `env:` block.

.github/workflows/release-fleet.yml
```diff
@@ -0,0 +1,107 @@
...
+        uses: goreleaser/goreleaser-action@v5
+        with:
+          distribution: goreleaser
+          version: v1.24.0
+          args: release --clean --verbose
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Upload charts to release
+        if: ${{ startsWith(github.ref, 'refs/tags/') }}
```
manno, line 99: This should probably be true for the whole workflow / the goreleaser step?  Could also omit this as the trigger is on tags only now.

.goreleaser.yml
```diff
@@ -0,0 +1,271 @@
...
+    binary: gitjob-{{ .Os }}-{{ .Arch }}{{ if .Arm }}64{{ end }}
+    no_unique_dist_dir: true
+    ldflags:
+      - -w -s
+      - -X github.com/rancher/fleet/pkg/version.GitCommit={{ .Commit }}
+      - -X github.com/rancher/fleet/pkg/version.Version={{ .Tag }}
+    targets:
+    - linux_amd64_v1
+    - linux_arm64
+    - windows_amd64
```
manno, line 93:
```suggestion
```

Let's only keep the agent and cli.

.goreleaser.yml
```diff
@@ -0,0 +1,271 @@
...
+    binary: gitcloner-{{ .Os }}-{{ .Arch }}{{ if .Arm }}64{{ end }}
+    no_unique_dist_dir: true
+    ldflags:
+      - -w -s
+      - -X github.com/rancher/fleet/pkg/version.GitCommit={{ .Commit }}
+      - -X github.com/rancher/fleet/pkg/version.Version={{ .Tag }}
+    targets:
+    - linux_amd64_v1
+    - linux_arm64
+    - windows_amd64
```
manno, line 80:
```suggestion
```

Let's only keep the agent and cli.****

.goreleaser.yml
```diff
@@ -0,0 +1,271 @@
...
+    # gcflags:
+    #   - all="-N -l"
+    ldflags:
+      - -w -s
+      - -X github.com/rancher/fleet/pkg/version.GitCommit={{ .Commit }}
+      - -X github.com/rancher/fleet/pkg/version.Version={{ .Tag }}
+    targets:
+    - linux_amd64_v1
+    - linux_arm64
+    - windows_amd64
```
manno, line 54:
```suggestion
```

Let's only keep the agent and cli.

package/Dockerfile
```diff
@@ -10,6 +11,10 @@ FROM base AS copy_buildx
 ONBUILD ARG TARGETARCH
 ONBUILD COPY bin/fleetcontroller-linux-${TARGETARCH} /usr/bin/fleetcontroller
 
+FROM base AS copy_goreleaser
+ONBUILD ARG ARCH
+ONBUILD COPY fleetcontroller-linux-${ARCH} /usr/bin/fleetcontroller
```
manno, line 16: I guess this is needed because of `no_unique_dist_dirs`?

Previously: https://github.com/rancher/fleet/blob/fb18a5e507c5c75de761bc8d73b38be45e8104da/package/Dockerfile#L11-L12

.github/workflows/release-fleet.yml
weyfonk: The [secret](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret) is available, but doesn't seem to be included in [environment variables](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables) by default :thinking:

.github/workflows/release-fleet.yml
weyfonk: That's simpler indeed, and removes a dependency on a Dapper script. Thanks.

.github/workflows/release-fleet.yml
weyfonk: True! Removed.

.github/workflows/release-fleet.yml
```diff
@@ -0,0 +1,107 @@
...
+#          api-key: ${{ secrets.FOSSA_API_KEY }}
+#
+#      - name: Run FOSSA tests
+#        uses: fossas/fossa-action@v1.3.1
+#        with:
+#          api-key: ${{ secrets.FOSSA_API_KEY }}
+#          run-tests: true
+
+      - name: Build Fleet binaries
+        run: ./.github/scripts/build-fleet-binaries.sh
```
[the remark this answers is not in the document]
weyfonk: Yes, we're building the binaries here to run tests against them. The output of this build is not used by GoReleaser.

package/Dockerfile
weyfonk: Correct.

.github/workflows/release-fleet.yml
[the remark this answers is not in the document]
weyfonk: As discussed offline, we don't actually need to build the binaries for unit nor integration testing, and won't be doing e2e testing here, which is too time-consuming.

manno (approved):

Issue #1408, opened 2023-03-15
[SURE-4340] Prometheus Metrics missing 


ulikl:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I would like to monitor the fleet functionality via Prometheus metrics.
I found this was already implemented and merge via PR "Add Prometheus metrics to Fleet #769" at May 13, 2022.

The fleet controller has no prometheus metrics code included in release 0.5.1 (released Jan 23) and master branch.

But the code is no longer there. Just the modules are still in go.mod and go.sum.
What happend to this feature?

### Expected Behavior

I would expect metrics at the controller endpoint /metrics on port 6060 

### Environment

```markdown
- Architecture: amd64
- Fleet Version: 0.5.1
```

### Issues
- [x] #2172
- [x] #2344
- [x] #2355
- [x] #2315
- [x] #2460
- [x] #2295
- [x] #2450

kkaempf:
See SURE-4340

manno:
https://github.com/rancher/fleet/pull/769 was superseded by #770. That was partially reviewed and closed.

At first glance I see a few problems with #770, which make this complex:
* exposing port 6060 of the fleet-controller for `/metrics` also allows access to `/debug/pprof`, which is insecure. We probably want to enable pprof only with `--debug`.
* as mentioned in the previous review we want to enable metrics only with `--enable-metrics`.
* adding `.global` to the bundledeployment, to store the cluster name from rancher is a smart workaround, but I think we should modify fleet to store a fleet cluster identifier (=cleaned up labels from the live cluster) in the bundledeployments status instead of `bundleDep.Spec.StagedOptions.Helm.Values.Global.Fleet.ClusterLabels[clusterNameLabel]`? I'm afraid of side effects with the helm values related code.

manno:
Build this for the new controller-runtime controllers only. This is a requirement for #1850

PR #2172, opened 2024-02-22
[SURE-4340] Add Prometheus Metrics

p-se:
Refers to #1408

Expose Prometheus metrics of the fleet-controller for the following
controllers with corresponding E2E tests:

- GitRepo
- Bundle
- BundleDeployment
- Cluster
- ClusterGroup

### DIFF
--- a/e2e/metrics/bundle_test.go
+++ b/e2e/metrics/bundle_test.go
@@ -1,0 +1,181 @@
+package metrics_test
+
+import (
+	"fmt"
+	"maps"
+	"math/rand"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/rancher/fleet/e2e/metrics"
+	"github.com/rancher/fleet/e2e/testenv"
+	"github.com/rancher/fleet/e2e/testenv/kubectl"
+	fleet "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
+)
+
+var _ = Describe("Bundle Metrics", Label("bundle"), func() {
+	const (
+		gitRepoName = "metrics"
+		branch      = "master"
+	)
+
+	var (
+		// kw is the kubectl command for namespace the workload is deployed to
+		kw        kubectl.Command
+		namespace string
+	)
+
+	BeforeEach(func() {
+		k = env.Kubectl.Namespace(env.Namespace)
+		namespace = testenv.NewNamespaceName(
+			gitRepoName,
+			rand.New(rand.NewSource(time.Now().UnixNano())),
+		)
+		kw = k.Namespace(namespace)
+
+		out, err := k.Create("ns", namespace)
+		Expect(err).ToNot(HaveOccurred(), out)
+
+		err = testenv.CreateGitRepo(
+			kw,
+			namespace,
+			gitRepoName,
+			branch,
+			"simple-manifest",
+		)
+		Expect(err).ToNot(HaveOccurred())
+
+		DeferCleanup(func() {
+			out, err = k.Delete("ns", namespace)
+			Expect(err).ToNot(HaveOccurred(), out)
+		})
+	})
+
+	When("collecting Bundle metrics", func() {
+		bundleMetricNames := map[string]map[string][]string{
+			"fleet_bundle_desired_ready": {},
+			"fleet_bundle_err_applied":   {},
+			"fleet_bundle_modified":      {},
+			"fleet_bundle_not_ready":     {},
+			"fleet_bundle_out_of_sync":   {},
+			"fleet_bundle_pending":       {},
+			"fleet_bundle_ready":         {},
+			"fleet_bundle_wait_applied":  {},
+			"fleet_bundle_state": {
+				"state": []string{
+					string(fleet.Ready),
+					string(fleet.NotReady),
+					string(fleet.WaitApplied),
+					string(fleet.ErrApplied),
+					string(fleet.OutOfSync),
+					string(fleet.Pending),
+					string(fleet.Modified),
+				},
+			},
+		}
+
+		// checkMetrics checks that the metrics exist or not exist. Custom
+		// checks can be added by passing a function to check. This can be used
+		// to check for the value of the metrics.
+		checkMetrics := func(
+			gitRepoName string,
+			expectExists bool,
+			check func(metric *metrics.Metric) error,
+		) func() error {
+			return func() error {
+				et := metrics.NewExporterTest(metricsURL)
+				expectOne := func(metricName string, labels map[string]string) error {
+					metric, err := et.FindOneMetric(metricName, labels)
+					if expectExists && err != nil {
+						return err
+					} else if !expectExists && err == nil {
+						return fmt.Errorf("metric %s found but not expected", metricName)
+					}
+
+					if check != nil {
+						err = check(metric)
+						if err != nil {
+							return err
+						}
+					}
+					return nil
+				}
+
+				for metricName, matchLabels := range bundleMetricNames {
+					identityLabels := map[string]string{
+						"name":      gitRepoName,
+						"namespace": namespace,
+					}
+					labels := map[string]string{}
+					maps.Copy(labels, identityLabels)
+
+					if len(matchLabels) > 0 {
+						for labelName, labelValues := range matchLabels {
+							for _, labelValue := range labelValues {
+								labels[labelName] = labelValue
+								err := expectOne(metricName, labels)
+								if err != nil {
+									return err
+								}
+							}
+						}
+					} else {
+						err := expectOne(metricName, labels)
+						if err != nil {
+							return err
+						}
+					}
+				}
+				return nil
+			}
+		}
+
+		It("should have one metric for each specified metric and label value", func() {
+			Eventually(checkMetrics(gitRepoName+"-simple-manifest", true, func(metric *metrics.Metric) error {
+				// No cluster exists in the namespace where our GitRepo has been deployed, hence
+				// we expect the values of the metrics to be 0.
+				if value := metric.Gauge.GetValue(); value != float64(0) {
+					return fmt.Errorf("unexpected metric value: expected 0, found %f", value)
+				}
+				return nil
+			})).ShouldNot(HaveOccurred())
+		})
+
+		When("the GitRepo (and therefore Bundle) is changed", Label("bundle-modified"), func() {
+			It("should not duplicate metrics if Bundle is updated", Label("bundle-update"), func() {
+				// et := metrics.NewExporterTest(metricsURL)
+				out, err := kw.Patch(
+					"gitrepo", gitRepoName,
+					"--type=json",
+					"-p", `[{"op": "replace", "path": "/spec/paths", "value": ["simple-chart"]}]`,
+				)
+				Expect(err).ToNot(HaveOccurred(), out)
+				Expect(out).To(ContainSubstring("gitrepo.fleet.cattle.io/metrics patched"))
+
+				Eventually(checkMetrics(gitRepoName+"-simple-chart", true, func(metric *metrics.Metric) error {
+					if metric.LabelValue("paths") == "simple-manifest" {
+						return fmt.Errorf("path for metric %s unchanged", metric.Metric.String())
+					}
+					return nil
+				})).ShouldNot(HaveOccurred())
+			})
+
+			It("should not keep metrics if Bundle is deleted", Label("bundle-delete"), func() {
+				gitRepoName := gitRepoName + "-simple-manifest"
+
+				var (
+					out string
+					err error
+				)
+				Eventually(func() error {
+					out, err = kw.Delete("bundle", gitRepoName)
+					return err
+				}).ShouldNot(HaveOccurred(), out)
+
+				Eventually(checkMetrics(gitRepoName, false, nil)).ShouldNot(HaveOccurred())
+			})
+		})
+	})
+})
--- a/e2e/metrics/bundledeployment_test.go
+++ b/e2e/metrics/bundledeployment_test.go
@@ -1,0 +1,168 @@
+package metrics_test
+
+import (
+	"fmt"
+	"math/rand"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/rancher/fleet/e2e/metrics"
+	"github.com/rancher/fleet/e2e/testenv"
+	"github.com/rancher/fleet/e2e/testenv/kubectl"
+)
+
+var _ = Describe("BundleDeployment Metrics", Label("bundledeployment"), func() {
+	const (
+		branch    = "master"
+		namespace = "fleet-local" // required for this test to create BundleDeployments
+	)
+
+	var (
+		// kw is the kubectl command for namespace the workload is deployed to
+		kw kubectl.Command
+		// objName is going to be "randomized" instead of using a dedicated and
+		// random namespace, like it is the case for the other tests.
+		objName string
+	)
+
+	BeforeEach(func() {
+		k = env.Kubectl.Namespace(env.Namespace)
+		kw = k.Namespace(namespace)
+		objName = testenv.AddRandomSuffix(
+			"metrics",
+			rand.New(rand.NewSource(time.Now().UnixNano())),
+		)
+
+		err := testenv.CreateGitRepo(
+			kw,
+			namespace,
+			objName,
+			branch,
+			"simple-manifest",
+		)
+		Expect(err).ToNot(HaveOccurred())
+
+		DeferCleanup(func() {
+			out, err := k.Delete("gitrepo", objName)
+			Expect(err).ToNot(HaveOccurred(), out)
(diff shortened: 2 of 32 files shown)


mig4ng:
Thank you for this work @p-se !
Last week I had some workloads down, which I did not get alerted on due to a failed deployment of a PrometheusRule of in one of my GitRepos.

Looking forward to see this in Fleet and subsequentially in Rancher. 🚀

p-se:
It might be worth looking at when exactly the data is being collected in the reconciliation loops of the controllers, because I'm not sure if it is perfect as it currently is.

internal/metrics/gitrepo_metrics.go
```diff
@@ -0,0 +1,158 @@
...
+			Help:      "The amount of clusters desired to be in a ready state.",
+		},
+		gitRepoLabels,
+	)
+	gitrepoReadyClusters = promauto.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: namespace,
+			Subsystem: gitRepoSubsystem,
+			Name:      "ready_clusters",
+			Help:      "The count of cluster in a Ready state.",
```
weyfonk, line 103:
```suggestion
			Help:      "The count of clusters in a Ready state.",
```

internal/metrics/gitrepo_metrics.go
```diff
@@ -0,0 +1,158 @@
...
+		},
+		gitRepoLabels,
+	)
+	gitrepoObserved = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: namespace,
+			Subsystem: gitRepoSubsystem,
+			Name:      "observations_total",
+			Help:      "The total times that this GitRepo has been observed",
```
weyfonk, line 112: I wonder if this phrasing might be a bit ambiguous to users. WDYT? How about something like this?
```suggestion
			Help:      "The total times that metrics have been computed for this GitRepo",
```

internal/cmd/controller/reconciler/gitrepo_controller.go
```diff
@@ -68,6 +69,8 @@ func (r *GitRepoReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, nil
 	}
 
+	metrics.CollectGitRepoMetrics(gitrepo)
```
weyfonk, line 72: Newbie question: what happens when a resource, in this case a GitRepo, is deleted? Will its metrics live forever with their last computed values, or will they be garbage-collected at some point?

weyfonk: Also, I wonder if we might want to bring metrics collection closer to (concretely right after) resource creation (L139-181), since most gitrepo metrics deal with resource counts.
A downside of this would be that, if resource creation fails for some reason, then metrics would not be refreshed. Would it make sense to compute metrics _more than once_ per reconciliation loop, eg. once here and once after resource creation, just to be on the safe side? Do we want the _observations_total_ metric to reflect a number of reconciliation loops, in which case that would be an issue?

internal/cmd/controller/reconciler/clustergroup_controller.go
```diff
@@ -71,6 +72,8 @@ func (r *ClusterGroupReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	})
 	if err != nil {
 		logger.V(1).Error(err, "Reconcile failed final update to cluster group status", "status", group.Status)
+	} else {
+		metrics.CollectClusterGroupMetrics(group)
 	}
```
weyfonk, line 77: nit: I think this is simpler:
```suggestion
	}
		
	metrics.CollectClusterGroupMetrics(group)
```

internal/cmd/controller/reconciler/bundledeployment_controller.go
```diff
@@ -68,7 +69,13 @@ func (r *BundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 			return err
 		}
 		t.Status = bd.Status
-		return r.Status().Update(ctx, t)
+		err = r.Status().Update(ctx, t)
+		if err != nil {
+			return err
+		}
+
+		metrics.CollectBundleDeploymentMetrics(t)
```
weyfonk, line 77: Wouldn't it be simpler to leave metrics collection out of the retry loop, and to run it right before the final `return` (~L83), which would have the same effect?

weyfonk (commented):
Thanks for this great effort! Leaving mostly nitpicks and some questions, as I am not very familiar with Prometheus.
Happy to discuss :)

internal/metrics/bundle_metrics.go
```diff
@@ -0,0 +1,157 @@
...
+	bundleModifiedDeployments.With(labels).Set(float64(bundle.Status.Summary.Modified))
+	bundleReadyDeployments.With(labels).Set(float64(bundle.Status.Summary.Ready))
+	bundlePendingDeployments.With(labels).Set(float64(bundle.Status.Summary.Pending))
+	bundleDesiredReadyDeployments.With(labels).Set(float64(bundle.Status.Summary.DesiredReady))
+	bundleObserved.With(labels).Inc()
+
+	currentState := summary.GetSummaryState(bundle.Status.Summary)
+
+	for _, state := range bundleStates {
+		labels["state"] = string(state)
```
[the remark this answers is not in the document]
weyfonk: As discussed offline, I had misunderstood this: the idea here is simply to work around the fact that the value for a metric cannot be text (in this case the value of the state for the bundle), hence we report each possible state, where all of them but one (the current state, with 1) will have value 0.

internal/cmd/controller/reconciler/clustergroup_controller.go
weyfonk: Oops, this is not functionally equivalent; we may not actually want to collect metrics when we failed to update the status. Please feel free to dismiss my previous comment.

internal/cmd/controller/reconciler/cluster_controller.go
```diff
@@ -167,6 +168,8 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		logger.V(1).Error(err, "Reconcile failed final update to cluster status", "status", cluster.Status)
 	}
 
+	metrics.CollectClusterMetrics(cluster)
```
bigkevmcd, line 171: I wonder if in addition to these metrics, we want to record the duration of the `Reconcile` call

Record the start time, and record the duration at the end of the function (`defer`) this could be useful to indicate how frequently the cluster can be reconciled, or indicate issues reconciling beyond simple errors (overload cases?).

bigkevmcd:
I think it'd be useful to also gauge the number of Paused resources, this is useful to highlight issues where folks have things that are not being applied.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,31 @@
+package metrics
+
+import fleet "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
+
+var (
+	// The namespace for the metrics, not the Kubernetes namespace of the
```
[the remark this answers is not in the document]
p-se: Would renaming the variable to `prefix` remove the need to provide that link?

dev/setup-fleet-downstream
```diff
@@ -60,7 +60,7 @@ kubectl config use-context "$downstream_ctx"
 helm -n cattle-fleet-system upgrade --install --create-namespace --wait fleet-agent charts/fleet-agent \
   --set-string labels.env=test \
   --set apiServerCA="$ca" \
-  --set apiServerURL="https://$public_hostname:36443" \
+  --set apiServerURL="https://$public_hostname:6443" \
```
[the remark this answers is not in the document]
p-se: That's not meant to be here. It is unfortunately required for my setup.

dev/setup-tests
```diff
@@ -0,0 +1,15 @@
...
+# shellcheck source=dev/setup-cluster-config
+source dev/setup-cluster-config
+
+# needed for gitrepo tests
+./dev/import-images-tests-k3d
+./dev/create-zot-certs 'FleetCI-RootCA' # for OCI tests
+cd e2e/testenv/infra || exit 1
+go build
+cd -
+./e2e/testenv/infra/infra setup
```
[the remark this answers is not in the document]
p-se: Yes!

internal/metrics/gitrepo_metrics.go
p-se: Because I could not come up with a clear use case and because collecting/updating metrics several times inside a single reconciliation loop is possible, I think keeping it is more confusing than helpful. Therefore, I've removed this metric.

internal/cmd/controller/reconciler/cluster_controller.go
p-se: controller-runtime does this out-of-the-box for us, e.g.:

```
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="0.005"} 81
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="0.01"} 113
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="0.025"} 165
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="0.05"} 319
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="0.1"} 413
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="0.15"} 453
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="0.2"} 470
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="0.25"} 490
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="0.3"} 493
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="0.35"} 496
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="0.4"} 499
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="0.45"} 501
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="0.5"} 502
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="0.6"} 502
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="0.7"} 502
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="0.8"} 502
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="0.9"} 502
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="1"} 503
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="1.25"} 503
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="1.5"} 503
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="1.75"} 503
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="2"} 503
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="2.5"} 503
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="3"} 503
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="3.5"} 503
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="4"} 503
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="4.5"} 503
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="5"} 503
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="6"} 503
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="7"} 503
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="8"} 503
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="9"} 503
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="10"} 503
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="15"} 503
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="20"} 503
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="25"} 503
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="30"} 503
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="40"} 503
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="50"} 503
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="60"} 503
controller_runtime_reconcile_time_seconds_bucket{controller="bundle",le="+Inf"} 503
```

p-se:
> I think it'd be useful to also gauge the number of Paused resources, this is useful to highlight issues where folks have things that are not being applied.

https://github.com/rancher/fleet/issues/2314

e2e/metrics/util.go
```diff
@@ -0,0 +1,181 @@
+package metrics
```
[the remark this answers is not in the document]
p-se: `exporter.go` fine?

e2e/metrics/cluster_test.go
```diff
@@ -0,0 +1,101 @@
...
+			for _, cluster := range existingClusters {
+				for metricName, expectedExist := range expectedMetricsExist {
+					metric, err := et.FindOneMetric(
+						metricName,
+						cluster.Name,
+						cluster.Namespace,
+						nil,
+					)
+					if expectedExist {
+						Expect(err).ToNot(HaveOccurred())
```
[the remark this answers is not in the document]
p-se: I haven't fully figured it out how to test cases in which a cluster is modified (if that is even possible and supposed to be possible) or deleted.

On the other topics, the test is "weakened" in the sense that it does not care about the values of the metrics and only considers existing cluster resources, how ever many there might be. It works well with running it in parallel, as do the other tests.

Adding one or two other tests will be done in a separate issue: https://github.com/rancher/fleet/issues/2315

e2e/metrics/bundle_test.go
```diff
@@ -0,0 +1,156 @@
...
+					if err != nil {
+						return err
+					}
+					Expect(metric.Gauge.GetValue()).To(Equal(float64(0)))
+				}
+				return nil
+			}).ShouldNot(HaveOccurred())
+		})
+
+		Context("when the GitRepo (and therefore Bundle) is changed", Label("bundle-altered"), func() {
```
manno, line 86: nit: could us syntactic sugar:

```suggestion
		When("the GitRepo (and therefore Bundle) is changed", Label("bundle-altered"), func() {
```

e2e/metrics/bundle_test.go
```diff
@@ -0,0 +1,156 @@
...
+			It("it should not duplicate metrics if Bundle is updated", Label("bundle-update"), func() {
```
manno, line 87:
```suggestion
			It("should not duplicate metrics if Bundle is updated", Label("bundle-update"), func() {
```

e2e/metrics/bundledeployment_test.go
```diff
@@ -0,0 +1,168 @@
...
+					}
+				}
+				return nil
+			}).ShouldNot(HaveOccurred())
+		})
+
+		Context(
+			"when the GitRepo (and therefore Bundle) is changed",
+			Label("bundle-altered"),
+			func() {
```
manno, line 92: nit: newlines look strange to me
```suggestion
		When("the GitRepo (and therefore Bundle) is changed", Label("bundle-altered"), func() {
```

e2e/metrics/exporter.go
```diff
@@ -0,0 +1,144 @@
...
+
+	return &Metric{Metric: metrics[0]}, nil
+}
+
+type promLabels map[string]string
+
+func (l promLabels) String() string {
+	r := ""
+	for k, v := range l {
+		r += fmt.Sprintf("%s=%q, ", k, v)
```
weyfonk, line 99: nit: not very important here since we're dealing with test code, but this could be done with a [strings.StringBuilder](https://pkg.go.dev/strings#Builder.WriteString).

e2e/metrics/exporter.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	return r
+}
+
+type Metric struct {
```
weyfonk, line 104: Does this type implement a Prometheus-defined interface that we may want to reference/name here, for clarity?

weyfonk (commented):
Looking good, thanks for this effort :rocket: 
Leaving a few comments and doubts.

internal/metrics/bundledeployment_metrics.go
```diff
@@ -0,0 +1,74 @@
...
+			),
+		},
+		collector: collectBundleDeploymentMetrics,
+	}
+)
+
+func collectBundleDeploymentMetrics(obj any, metrics map[string]prometheus.Collector) {
+	bundleDep, ok := obj.(*fleet.BundleDeployment)
+	if !ok {
+		panic("unexpected object type")
```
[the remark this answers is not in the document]
bigkevmcd: I would generally agree that panic is bad, but...this would imply a code bug (connecting the wrong things) which means that it could be improved by making `CollectorCollection` generic with `[T runtime.Object]` ?

manno (approved):

mig4ng:
Amazing! 🎉

PR #2173, opened 2024-02-23
Add git/ssh binaries to images

manno:
Partially reverts https://github.com/rancher/fleet/pull/2125

The go-getter library used in the fleet apply command [uses the git CLI](https://github.com/hashicorp/go-getter/issues/99). Since git uses the ssh binary, both the git-core and openssh packages are needed.

The go-git library, used in the fleetcontroller (imagescan) needs git and probably ssh. Though the [file:// protocol](https://github.com/go-git/go-git/blob/master/COMPATIBILITY.md#transport-schemes) is not used, they give  [misleading error messages](https://github.com/rancher/gitjob/pull/362) if git is missing.

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -1,6 +1,7 @@
 ARG BUILD_ENV=dapper
 
-FROM registry.suse.com/bci/bci-busybox:15.5 AS base
+FROM registry.suse.com/bci/bci-base:15.5 AS base
+RUN zypper in --no-recommends -y git-core openssh; rm -fr /var/cache/* /var/log/*log
 
 FROM base AS copy_dapper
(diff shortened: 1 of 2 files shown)


manno:
According to 

https://github.com/go-git/go-git/blob/master/COMPATIBILITY.md?plain=1#L212  go-git only needs the binary for the `file://` protocol. However, we  observed that the [error message for a missing repo changes](https://github.com/rancher/gitjob/pull/362), if git is not installed.
Even though we [don't allow `file://` for gitrepos](https://github.com/rancher/fleet/blob/main/pkg/git-urls/urls.go#L41-L47) we need the binary to get the correct error message?

Issue #2096, opened 2024-01-22
[SURE-7469] Imagescan breaks fleet controller 


kkaempf:
### Issue description:
Fleet controller in Rancher upstream clusters crashes if the imagen tag does not include the semver range expected when using [ImageScan](https://fleet.rancher.io/imagescan) in the fleet.yaml file.

~~~
="2024-01-17T18:44:44Z" level=debug msg="DesiredSet - No change(2) gitjob.cattle.io/v1, Kind=GitJob fleet-default/project-test2 for gitjobs fleet-default/project-test2"
E0117 18:44:44.351563       1 runtime.go:79] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 1491 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic({0x22b04a0?, 0x3e240f0})
    /go/pkg/mod/k8s.io/apimachinery@v0.25.12/pkg/util/runtime/runtime.go:75 +0x99
k8s.io/apimachinery/pkg/util/runtime.HandleCrash({0x0, 0x0, 0xfffffffe?})
    /go/pkg/mod/k8s.io/apimachinery@v0.25.12/pkg/util/runtime/runtime.go:49 +0x75
panic({0x22b04a0, 0x3e240f0})
    /usr/lib64/go/1.20/src/runtime/panic.go:884 +0x213
github.com/Masterminds/semver/v3.(*Version).Original(...)
    /go/pkg/mod/github.com/!masterminds/semver/v3@v3.2.1/version.go:259
github.com/rancher/fleet/internal/cmd/controller/controllers/image.semverLatest({0x3d70a50?, 0xc0026ff260?}, {0xc0058ff200, 0x2f, 0x0?})
    /go/src/github.com/rancher/fleet/internal/cmd/controller/controllers/image/image.go:501 +0x105
github.com/rancher/fleet/internal/cmd/controller/controllers/image.latestTag({0xc0075afc90?, 0xc0075afca0?}, {0xc0058ff200?, 0xc009d1a01d?, 0x44?}) 
~~~

### Business impact:
Fleet goes into a crash and the GitRepos are not updated. 

### Troubleshooting steps:

From the customer: 
In artifact registry our latest image was built 20 hours ago and has the tag 0.0.0-52. The one deployed with the initial fleet.yaml file is 0.0.0-44. This fails to update.

### Actual behavior:
[...]

PR #2174, opened 2024-02-23
Fixes panic when imagescan uses prereleases and *

0xavi0:
When using pre-releases in imagescan combined with
```yaml
semver:
  range: *
```
it failed because `semver` ignores pre-releases unless they're explicitly described in the `semver.range`.

For example:
```yaml
semver:
  range: ">= 0.0.0-40"
```
works fine because it is explicitly using a pre-release.

This fixes the `*` case by checking if there's any version found (lastVersion != nil)

It also adds `e2e` tests for the following imagescans cases:

* Deploy gitrepo with imagescan using `semver.range = ">= 0.0.0-40"` deploying image `0.0.0-40` and after that pushing image `0.0.0-50`
* Same as above but for the case that whis branch fixed (using `semver.range = "*"` with pre-releases.

Refers to: https://github.com/rancher/fleet/issues/2096

### DIFF
--- a/e2e/assets/imagescan/pre-releases-ignored/fleet.yaml
+++ b/e2e/assets/imagescan/pre-releases-ignored/fleet.yaml
@@ -1,0 +1,21 @@
+imageScans:
+# specify the policy to retrieve images, can be semver or alphabetical order
+- policy:
+    # if range is specified, it will take the latest image according to semver order in the range
+    # for more details on how to use semver, see https://github.com/Masterminds/semver
+    # in this test case we're going to use prerelease versions.
+    # as we're specifying * semver will ignore new tags but should not crash the fleet controller
+    semver:
+      range: "*"
+    # can use ascending or descending order
+    alphabetical:
+      order: asc
+
+  # specify images to scan
+  image: {{.Image}}
+
+  # Specify the tag name, it has to be unique in the same bundle
+  tagName: test-scan
+
+  # Specify the scan interval
+  interval: 5s
--- a/e2e/single-cluster/imagescan_test.go
+++ b/e2e/single-cluster/imagescan_test.go
@@ -1,12 +1,20 @@
 package singlecluster_test
 
 import (
+	"fmt"
 	"os"
+	"os/exec"
 	"path"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"time"
 
+	"github.com/go-git/go-git/v5"
 	"github.com/rancher/fleet/e2e/testenv"
 	"github.com/rancher/fleet/e2e/testenv/githelper"
 	"github.com/rancher/fleet/e2e/testenv/kubectl"
+	"k8s.io/apimachinery/pkg/util/uuid"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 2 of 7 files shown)


e2e/single-cluster/imagescan_test.go
```diff
@@ -92,3 +68,211 @@ var _ = Describe("Image Scan", Label("infra-setup"), func() {
...
+		tmpdir := GinkgoT().TempDir()
+		clonedir = path.Join(tmpdir, "clone")
+		repository = setupRepo(k, tmpdir, clonedir, tmpRepoDir)
+	})
+
+	AfterEach(func() {
+		_, _ = k.Delete("gitrepo", "imagescan")
+		_, _ = k.Delete("secret", "git-auth")
+	})
+	When("deploy imagescan setup with pre-release images", func() {
```
weyfonk, line 95: nit:
```suggestion
	When("deploying imagescan setup with pre-release images", func() {
```

e2e/single-cluster/imagescan_test.go
```diff
@@ -92,3 +68,211 @@ var _ = Describe("Image Scan", Label("infra-setup"), func() {
...
+}
+
+func tagAndPushImage(baseImage, image, tag string) string {
+	imageTag := fmt.Sprintf("%s:%s", image, tag)
+	// tag the image and push it to ttl.sh
+	cmd := exec.Command("docker", "tag", baseImage, imageTag)
+	err := cmd.Run()
+	Expect(err).ToNot(HaveOccurred())
+	// push the image to ttl.sh
+	cmd = exec.Command("docker", "push", imageTag)
```
weyfonk, line 242: Out of curiosity, do we know the default expiration time of `ttl.sh` images? I've tried to reach ttl.sh to check that, but am getting a blank page, which did not use to be the case :thinking: .

e2e/single-cluster/imagescan_test.go
```diff
@@ -92,3 +68,211 @@ var _ = Describe("Image Scan", Label("infra-setup"), func() {
...
+			By("pushing a new tag to the registry and checking the new image tag is found in the bundle")
+			newTag := "0.0.0-50"
+			imageTag = tagAndPushImage("k8s.gcr.io/pause", image, newTag)
+			Eventually(func() string {
+				out, _ := k.Get("bundles", "imagescan-examples", "-o", "yaml")
+				return out
+			}).Should(
+				MatchRegexp(fmt.Sprintf(`image: %s # {"\$imagescan": "test-scan"}`, imageTag)))
+
+			By("checking that the new tag is pushed in the git repository")
```
weyfonk, line 125: This is a thorough check, but arguably if the new image tag has made it into the bundle, then that would mean that it had first been propagated to the git repository.

e2e/single-cluster/imagescan_test.go
```diff
@@ -92,3 +68,211 @@ var _ = Describe("Image Scan", Label("infra-setup"), func() {
...
+				return out
+			}).Should(
+				MatchRegexp(fmt.Sprintf(`image: %s # {"\$imagescan": "test-scan"}`, imageTag)))
+
+			By("pushing a new tag to the registry and checking the fleet controller does not crash")
+			// store number of fleet controller restarts to compare later
+			fleetControllerRestarts := getFleetControllerRestarts(k)
+			newTag := "0.0.0-50"
+			previousImageTag := imageTag
+			imageTag = tagAndPushImage("k8s.gcr.io/pause", image, newTag)
```
weyfonk, line 162: This doesn't seem to be used anywhere, so perhaps we could `s/previousImageTag/imageTag/`.

e2e/single-cluster/imagescan_test.go
```diff
@@ -92,3 +68,211 @@ var _ = Describe("Image Scan", Label("infra-setup"), func() {
...
+		})
+	})
+	AfterEach(func() {
+		_, _ = k.Delete("gitrepo", "imagescan")
+		_, _ = k.Delete("secret", "git-auth")
+	})
+})
+
+func getFleetControllerRestarts(k kubectl.Command) int {
+	out, err := k.Namespace("cattle-fleet-system").Get("pods", "-l", "app=fleet-controller", "--no-headers")
```
weyfonk, line 186: If we didn't want to deal with indices, we could do something like this:
```suggestion
	out, err := k.Namespace("cattle-fleet-system").Get("pods", "-l", "app=fleet-controller", "--no-headers", "-o", "custom-columns=RESTARTS:.status.containerStatuses[0].restartCount")
```

e2e/assets/imagescan/pre-releases-ignored/fleet.yaml
```diff
@@ -0,0 +1,21 @@
+imageScans:
+# specify the policy to retrieve images, can be semver or alphabetical order
+- policy:
+    # if range is specified, it will take the latest image according to semver order in the range
+    # for more details on how to use semver, see https://github.com/Masterminds/semver
+    # in this test case we're going to use prerelease versions.
+    # as we're specifying * semver will ignore new tags but should not crash the fleet controller
```
weyfonk, line 7: Having this here is nice for us, the Fleet team, but I think we'd also need an update to our [fleet-docs](https://github.com/rancher/fleet-docs) repo to add info relative to semver and prereleases to [this page](https://fleet.rancher.io/imagescan).

weyfonk (approved):
LGTM, with a few comments :) Thanks!

e2e/single-cluster/imagescan_test.go
0xavi0: changed

e2e/single-cluster/imagescan_test.go
0xavi0: I'm not 100% sure [this](https://www.civo.com/learn/ttl-sh-your-anonymous-and-ephemeral-docker-image-registry) is a reliable source, but it states that the default expiration time is 1h and maximum is 24h. 

As we need to specify pre-release tags we can't use the tag to specify the expiration time, so I assume they expire in 1h.

e2e/single-cluster/imagescan_test.go
0xavi0: `previousImageTag` is checked here
```go
Eventually(func() string {
				out, _ := k.Get("bundles", "imagescan-examples", "-o", "yaml")
				return out
			}).Should(
				MatchRegexp(fmt.Sprintf(`image: %s # {"\$imagescan": "test-scan"}`, previousImageTag)))
``` 
`imageTag` is no longer used, so technically we could avoid storing `imageTag` and use the previous value.

I did it this way to make it clear that `previousImageTag` refers to `0.0.0-40` and that `imageTag` refers to the latest pushed tag (`0.0.0-50` in this case)

e2e/single-cluster/imagescan_test.go
0xavi0: The `fleet-controller` has 3 containers so... are we sure 0 is going te be the container we're looking for? 
As I wasn't sure I used a global restart count.

e2e/assets/imagescan/pre-releases-ignored/fleet.yaml
0xavi0: sure! I'll add some documentation about pre-releases in there.

e2e/single-cluster/imagescan_test.go
weyfonk: Nice, thanks for checking. That makes sense, 1h should be enough for our use case.

e2e/single-cluster/imagescan_test.go
weyfonk: Yes, sorry, what I meant is that `imageTag` is computed with the new tag but not used.
On the other hand, I like the extra clarity provided by `previousImageTag` :+1:

weyfonk (approved):

e2e/assets/imagescan/pre-releases-ignored/fleet.yaml
0xavi0: https://github.com/rancher/fleet-docs/pull/114

e2e/single-cluster/imagescan_test.go
weyfonk: For some reason this has been undone :thinking:

e2e/single-cluster/imagescan_test.go
```diff
@@ -92,3 +68,244 @@ var _ = Describe("Image Scan", Label("infra-setup"), func() {
...
+	Expect(err).NotTo(HaveOccurred())
+	out = strings.TrimSuffix(out, "\n")
+	boolValue, err := strconv.ParseBool(out)
+	Expect(err).NotTo(HaveOccurred())
+	return boolValue
+}
+
+func getFleetControllerContainerIndexInPod(k kubectl.Command, container string) (int, bool) {
+	// the fleet controller pod runs 3 containers.
+	// we need to know what is the index to the fleet-controller container inside de pod.
```
weyfonk, line 215:
```suggestion
	// we need to know the index of the fleet-controller container inside the pod.
```

weyfonk (approved):

Issue #2181, opened 2024-02-27
[backport][v0.9][SURE-7469] Imagescan breaks fleet controller


0xavi0:
## SURE-7469
backport of: https://github.com/rancher/fleet/issues/2096

### Issue description:
Fleet controller in Rancher upstream clusters crashes if the imagen tag does not include the semver range expected when using [ImageScan](https://fleet.rancher.io/imagescan) in the fleet.yaml file.

~~~
="2024-01-17T18:44:44Z" level=debug msg="DesiredSet - No change(2) gitjob.cattle.io/v1, Kind=GitJob fleet-default/project-test2 for gitjobs fleet-default/project-test2"
E0117 18:44:44.351563       1 runtime.go:79] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 1491 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic({0x22b04a0?, 0x3e240f0})
    /go/pkg/mod/k8s.io/apimachinery@v0.25.12/pkg/util/runtime/runtime.go:75 +0x99
k8s.io/apimachinery/pkg/util/runtime.HandleCrash({0x0, 0x0, 0xfffffffe?})
    /go/pkg/mod/k8s.io/apimachinery@v0.25.12/pkg/util/runtime/runtime.go:49 +0x75
panic({0x22b04a0, 0x3e240f0})
    /usr/lib64/go/1.20/src/runtime/panic.go:884 +0x213
github.com/Masterminds/semver/v3.(*Version).Original(...)
    /go/pkg/mod/github.com/!masterminds/semver/v3@v3.2.1/version.go:259
github.com/rancher/fleet/internal/cmd/controller/controllers/image.semverLatest({0x3d70a50?, 0xc0026ff260?}, {0xc0058ff200, 0x2f, 0x0?})
    /go/src/github.com/rancher/fleet/internal/cmd/controller/controllers/image/image.go:501 +0x105
github.com/rancher/fleet/internal/cmd/controller/controllers/image.latestTag({0xc0075afc90?, 0xc0075afca0?}, {0xc0058ff200?, 0xc009d1a01d?, 0x44?}) 
~~~

### Business impact:
Fleet goes into a crash and the GitRepos are not updated. 

### Troubleshooting steps:

From the customer: 
[...]

PR #2183, opened 2024-02-27
[v0.9] Fixes panic when imagescan uses prereleases and *

0xavi0:
When using pre-releases in imagescan combined with
```yaml
semver:
  range: *
```
it failed because `semver` ignores pre-releases unless they're explicitly described in the `semver.range`.

For example:
```yaml
semver:
  range: ">= 0.0.0-40"
```
works fine because it is explicitly using a pre-release.

This fixes the `*` case by checking if there's any version found (lastVersion != nil)

It also adds `e2e` tests for the following imagescans cases:

* Deploy gitrepo with imagescan using `semver.range = ">= 0.0.0-40"` deploying image `0.0.0-40` and after that pushing image `0.0.0-50`
* Same as above but for the case that whis branch fixed (using `semver.range = "*"` with pre-releases.

Refers to: https://github.com/rancher/fleet/issues/2181

### DIFF
--- a/e2e/single-cluster/imagescan_test.go
+++ b/e2e/single-cluster/imagescan_test.go
@@ -1,12 +1,20 @@
 package singlecluster_test
 
 import (
+	"fmt"
 	"os"
+	"os/exec"
 	"path"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"time"
 
+	"github.com/go-git/go-git/v5"
 	"github.com/rancher/fleet/e2e/testenv"
 	"github.com/rancher/fleet/e2e/testenv/githelper"
 	"github.com/rancher/fleet/e2e/testenv/kubectl"
+	"k8s.io/apimachinery/pkg/util/uuid"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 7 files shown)


e2e/single-cluster/imagescan_test.go
```diff
@@ -92,3 +68,244 @@ var _ = Describe("Image Scan", Label("infra-setup"), func() {
...
+			Expect(err).NotTo(HaveOccurred())
+			basedir := filepath.Join(clonedir, "examples")
+			b, err := os.ReadFile(filepath.Join(basedir, "deployment.yaml"))
+			Expect(err).NotTo(HaveOccurred())
+			Expect(string(b)).Should(ContainSubstring(newTag))
+		})
+	})
+
+	// this test didn't pass before adding fix for issue #2096
+	When("deploy imagescan setup with pre-release images and * semver range", func() {
```
weyfonk, line 136: nit:
```suggestion
	When("deploying imagescan setup with pre-release images and * semver range", func() {
```

weyfonk (approved):

e2e/single-cluster/imagescan_test.go
0xavi0: OMG, that thing again..... :) I think I fixed it... AGAIN

PR #2042, opened 2024-01-04
Improve container security

weyfonk:
This ensures that controller and agent containers are run as non-root users, to reduce their attack surfaces.

Note: changes to the Windows-based agent image have not been tested, as Fleet does not build on Windows:
```
$ ./dev/build-fleet
+ '[' '!' -d ./cmd/fleetcontroller ']'
+ export GOARCH=amd64
+ GOARCH=amd64
+ export CGO_ENABLED=0
+ CGO_ENABLED=0
+ git diff --quiet HEAD origin/master -- pkg/apis/fleet.cattle.io/v1alpha1
+ go generate
./pkg/apis
pkg\apis\fleet.cattle.io
pkg\apis\fleet.cattle.io\v1alpha1
pkg\apis\fleet.cattle.io\v1alpha1\bundle_types.go
pkg\apis\fleet.cattle.io\v1alpha1\bundledeployment_types.go
pkg\apis\fleet.cattle.io\v1alpha1\bundlenamespacemapping_types.go
pkg\apis\fleet.cattle.io\v1alpha1\cluster_types.go
pkg\apis\fleet.cattle.io\v1alpha1\clustergroup_types.go
pkg\apis\fleet.cattle.io\v1alpha1\clusterregistration_types.go
pkg\apis\fleet.cattle.io\v1alpha1\clusterregistrationtoken_types.go
pkg\apis\fleet.cattle.io\v1alpha1\content_types.go
pkg\apis\fleet.cattle.io\v1alpha1\doc.go
pkg\apis\fleet.cattle.io\v1alpha1\fleetyaml.go
pkg\apis\fleet.cattle.io\v1alpha1\generic_map.go
pkg\apis\fleet.cattle.io\v1alpha1\generic_map_test.go
pkg\apis\fleet.cattle.io\v1alpha1\gitrepo_types.go
pkg\apis\fleet.cattle.io\v1alpha1\gitreporestriction_types.go
pkg\apis\fleet.cattle.io\v1alpha1\groupversion_info.go
pkg\apis\fleet.cattle.io\v1alpha1\imagescan_types.go
pkg\apis\fleet.cattle.io\v1alpha1\zz_generated.deepcopy.go
Removing pkg\apis\fleet.cattle.io\v1alpha1\zz_generated.deepcopy.go
pkg\apis\go.mod
pkg\apis\go.sum
go: downloading sigs.k8s.io/controller-tools v0.12.0
go: downloading golang.org/x/tools v0.14.0
[...]

PR #2185, opened 2024-02-27
Fix drone building on windows

thardeck:
This should prevent the following Drone Windows error when building Docker images:
Invoke-WebRequest : Access to the path 'C:\git.zip' is denied.

e.g. https://drone-publish.rancher.io/rancher/fleet/3178/3/2

We could most likely change the directory but we are moving away from Drone building short term so it is not worth it.

### DIFF
--- a/package/Dockerfile-windows.agent
+++ b/package/Dockerfile-windows.agent
@@ -9,7 +9,8 @@ RUN pushd c:\; \
     $URL = 'https://github.com/git-for-windows/git/releases/download/v2.42.0.windows.2/MinGit-2.42.0.2-64-bit.zip'; \
     \
     Write-Host ('Downloading git from {0} ...' -f $URL); \
-    curl.exe -sfL $URL -o c:\git.zip; \
+    [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12; \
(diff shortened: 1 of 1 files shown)


package/Dockerfile-windows.agent
```diff
@@ -2,15 +2,15 @@ ARG SERVERCORE_VERSION
 FROM mcr.microsoft.com/windows/servercore:${SERVERCORE_VERSION}
 ARG RELEASES
 ARG VERSION
-USER ContainerUser
```
weyfonk, line 5: Was this incompatible with `Invoke-WebRequest`? We added it as part of #2042.

weyfonk: I see. Anyway, this is most likely a short-term fix as we may well end up cross-building Windows-based images from Linux hosts, meaning that download/unzip steps would happen from another image. We could then (but only then) keep `USER ContainerUser`.

TL;DR: this is fine, at least as a short-term fix. My first comment may have been a bit too hasty :þ

weyfonk (approved):

package/Dockerfile-windows.agent
manno: Maybe we can move it to the bottom? So the powershell commands are still allowed to write to `C:\`.. :shrug:

Issue #2195, opened 2024-03-01
[v0.9][SURE-7311] GitRepo fails with error "namespace not found" when using namespaceLabels: in existing namespaces


manno:
backport of https://github.com/rancher/fleet/issues/1994

QA Template: https://github.com/rancher/fleet/issues/1994#issue-2023502758

PR #2204, opened 2024-03-04
Fix Existing Namespace Not Found By Label

manno:
Refers to #2195 

Filter by `kubernetes.io/metadata.name` label

Use the label `kubernetes.io/metadata.name`, which is added by k8s to all
namespaces, instead of the helm "name" label. The helm "name" label
might be missing from pre-existing namespaces.

Switch `fetchNamespace` to Get instead of list

Was using List with a label, that's set by k8s to the name of the
namespace. Only returned the first result.
Testdata assumed label and namespace name could be different.

### DIFF
--- a/internal/cmd/agent/controllers/bundledeployment/controller_test.go
+++ b/internal/cmd/agent/controllers/bundledeployment/controller_test.go
@@ -33,13 +33,14 @@ func TestSetNamespaceLabelsAndAnnotations(t *testing.T) {
 			}},
 			ns: corev1.Namespace{
 				ObjectMeta: metav1.ObjectMeta{
-					Labels: map[string]string{"name": "test"},
+					Name:   "namespace1234",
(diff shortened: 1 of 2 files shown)


internal/cmd/agent/controllers/bundledeployment/controller_test.go
```diff
@@ -100,11 +106,8 @@ func TestSetNamespaceLabelsAndAnnotations(t *testing.T) {
 			u, _ := runtime.DefaultUnstructuredConverter.ToUnstructured(&test.ns)
 			// Resource will be called twice, one time for UPDATE and another time for LIST
```
weyfonk, line 107:
```suggestion
			// Resource will be called twice, one time for UPDATE and another time for GET
```

weyfonk (approved):

PR #2207, opened 2024-03-04
Remove drone and dapper scripts

manno:
Refers to https://github.com/rancher/fleet/issues/2121


dev/README.md
```diff
@@ -147,7 +147,7 @@ remedy this. By default, the setup script will use `/usr/bin/helm`.
 
 ## Different Script Folders
 
-Our CIs, dapper/drone and github actions, use a different set of scripts. CI
+Our CIs and github actions, use a different set of scripts. CI
```
weyfonk, line 150: nit: maybe that's just me, but I don't understand what that comma does there :thinking: 
```suggestion
Our CIs and github actions use a different set of scripts. CI
```

scripts/agent-dev
```diff
@@ -1,28 +0,0 @@
-#!/bin/bash
```
weyfonk, line 1: nit: Since we're removing all scripts, I think we could also remove [this line](https://github.com/rancher/fleet/blob/master/.github/workflows/rancher-integration.yml#L21).

scripts/agent-dev
```diff
@@ -1,28 +0,0 @@
...
-set -e
```
weyfonk, line 2: We should also update [DEVELOPING.md](https://github.com/rancher/fleet/blob/master/DEVELOPING.md) which still has references to `make agent-dev`.

weyfonk (commented):
Looks good :) I think we could take this opportunity to update a couple of files, see comments.

weyfonk (approved):

Issue #2214, opened 2024-03-08
[0.9] Bundles with no matching targets makes Content objects to be constantly recreated


aruiz14:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

[Customizing targets](https://github.com/rancher/fleet/issues/1552) using `doNotDeploy` or `overrideTargets` could cause the deployment to not match any cluster, hence not creating any `BundleDeployment` objects.
However, a `Content` object is always created based on the Bundle resources, but since this is not referenced by any `BundleDeployment`, the cleanup controller will prune it, just to be recreated again by the Bundles handler.
This could cause unnecessary CPU usage in the Fleet controller or space in the Kubernetes database.

### Expected Behavior

`Content` object is not created if no targets are matched.

### Steps To Reproduce

The simplest is to set `targetCustomizations` in `fleet.yaml` so that `doNotDeploy: true` is set for all clusters.

### Environment

```markdown
- Fleet Version: 0.9.0
```

aruiz14:
## Additional QA
 
### Solution

I moved the code that stores the encoded manifests (ensures that a `Content` object exists with that content) after the targets are calculated, skipping that step if no targets were found.
 
### Testing

 - No regressions should be added due to this change.
 - Creating a `bundle` whose targetCustomizations (in `fleet.yaml`) match no clusters should prevent the relevant `Content` objects from being constantly created and deleted.

thehejik:
### Test report
This seems to be fixed in `fleet:103.1.3+up0.9.3-rc.2` on rancher v2.8-head

On `fleet-0.9.2` I was unable to verify that the resource is getting deleted by a cleanup process and recreated automatically afterwards but I verified the resources are not created at all when the target doesn't exist on `fleet-0.9.3` so it should prevent the described problem.

#### Before:
* I could see on previous `fleet-0.9.2` version that even when a `gitRepo` is pointing to a non-existing target its `contents.fleet.cattle.io` resource is always created and recreated when deleted manually.

#### After:
* On the other hand the on `fleet-0.9.3` the resource is only created when the target exists.

Note:
I wasn't using fleet.yaml and `targetCustomizations` but only rancher UI to select desired Target for my testing gitRepo (No Clusters / All clusters in the Workspace).

Commands used for testing:
```
$ kubectl get contents.fleet.cattle.io  -A -w
# To verify content of contents.fleet.cattle.io resource
[...]

PR #2215, opened 2024-03-08
Prevent creating Content objects if Bundle does not match any target

aruiz14:
Refers to #2214


manno:
I forward ported this to main: https://github.com/rancher/fleet/pull/2239

manno (approved):

PR #2216, opened 2024-03-10
Make AlphabeticalPolicy string check case insensitive

aditditto:
Previously if users wanted to use image scan with descending alphabetical policy, they had to use uppercase `DESC`, as shown below:
```yaml
imageScans:
# specify the policy to retrieve images, can be semver or alphabetical order
- policy:
    # can use ascending or descending order
    alphabetical:
      order: DESC
```
This is not very intuitive, because the [documentation](https://fleet.rancher.io/imagescan) gives an example in lowercase:
```yaml
imageScans:
# specify the policy to retrieve images, can be semver or alphabetical order
- policy:
    # if range is specified, it will take the latest image according to semver order in the range
    # for more details on how to use semver, see https://github.com/Masterminds/semver
    semver:
      range: "*"
    # can use ascending or descending order
    alphabetical:
      order: asc
```
This PR makes the string check case insensitive, removing the need to use uppercase `DESC` in the `fleet.yaml`

### DIFF
--- a/internal/cmd/controller/imagescan/tagscan_job_test.go
+++ b/internal/cmd/controller/imagescan/tagscan_job_test.go
@@ -1,0 +1,48 @@
+package imagescan
+
+import (
+	"fmt"
+	"testing"
+
+	fleet "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
+)
+
+var alphabeticalVersions = []string{"a", "b", "c"}
+func TestLatestTag(t *testing.T) {
+	alphabeticalAscPolicyLowercase := fleet.ImagePolicyChoice{
+		Alphabetical: &fleet.AlphabeticalPolicy{Order: "asc"},
+	}
+	alphabeticalAscPolicyUppercase := fleet.ImagePolicyChoice{
+		Alphabetical: &fleet.AlphabeticalPolicy{Order: "ASC"},
+	}
+	alphabeticalDescPolicyLowercase := fleet.ImagePolicyChoice{
+		Alphabetical: &fleet.AlphabeticalPolicy{Order: "desc"},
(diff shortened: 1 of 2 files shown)


internal/cmd/controller/imagescan/tagscan_job_test.go
```diff
@@ -0,0 +1,58 @@
...
+			},
+			want: "c",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := latestTag(tt.policy, alphabeticalVersions)
+			if err != nil {
+				t.Errorf("Error calling latestTag: %v", err)
```
bigkevmcd, line 50: This isn't expected to return an error in these tests, so the next comparison isn't really valid?

I think this should be `t.Fatal()` see https://go.dev/wiki/TestComments#keep-going

> If you’re using subtests (and you’re inside a call to t.Run), then t.Fatal ends the current subtest and allows your test case to progress to the next subtest, so use t.Fatal.

bigkevmcd (commented):
Looks good to me, nice tidy fix.

internal/cmd/controller/imagescan/tagscan_job_test.go
aditditto: Thank you for the review. I am still new to Go, so I may make some mistakes. 

What do you mean by the next comparison isn't really valid? Does that mean checking for the `err` from `latestTag` is not necessary?

aditditto: I see. Thank you for the correction.

weyfonk (dismissed):

weyfonk (approved):

Issue #1972, opened 2023-11-27
Convert Git/Targeting controller to controller-runtime


manno:
Relies on https://github.com/rancher/fleet/issues/1732 as we don't convert the agent registration controllers.
The agent was converted in https://github.com/rancher/fleet/issues/1734

The merged PR  switches the targeting and gitrepo controllers to use controller-runtime instead of wrangler to set up watches:
  * bundle controller
  * gitrepo controller
  * display controller
  * imagescan

Additionally: 
* Changes imagescan to use a scheduler based on the quartz module. Imagescan previously used `EnqueueAfter`.
* Refactor: split code in target pkg into separate files.
* Purge bundledeployments from gitrepo reconciler when triggering. The previous cross-namespace ownership mechanism was implemented by wrangler's `Apply` and cannot be used anymore as bd resources are not created with `Apply`.
* Bundle reconciler deletes bundle deployments if bundle is not found.
* Gitrepo reconciler reports errors in status, as it did previously, however the implementation changed and status might be reported later or earlier.
* Moves the cluster status update code from agentmanagement to new cluster reconciler, as it's not related to agentmanagement.
* Controller supports  a configurable bind addresses for health probes and metrics.

Possible follow-up stories:
- [ ] add log level 4 ("trace") and more logging
- [ ] tune event filters and cache options, after baseline performance benchmark exists
- [ ] replace more "tickers" (drift detect, git job) with quartz module?

PR #2223, opened 2024-03-11
Propagate bundle and bundledeployment status updates to GitRepo

weyfonk:
A filter on status updates in the GitRepo reconciler prevents updates to statuses of bundles and bundle deployments from being reflected in their owning GitRepo's status. In turn, this leads to outdated and inconsistent GitRepo statuses.

This updates those filters to ensure that bundle status updates are taken into account. End-to-end tests are updated as well to prevent future regressions.

Refers to #1972
Follow-up to #2018


p-se (approved):
lgtm

Issue #1572, opened 2023-06-02
Add option to disable polling for Git repos


manno:
When this option, e.g. "disableGitPolling", is set, the gitjob will not check the repo every syncInterval seconds. This allows users to build workflows that only use webhooks.

Normally Fleet checks the repo regularly, even when using Webhooks, to make sure it can recover from a failed Webhook http request.

## Notes

Currently Fleet/gitjob will [interpret a syncInterval of 0 as 15 minutes](https://github.com/rancher/gitjob/blob/master/pkg/controller/gitjob/gitjobs.go#L68-L72).
Furthermore, if webhooks are used and the interval is at "0", the default, the interval [will be changed to one hour](https://github.com/rancher/gitjob/blame/master/pkg/webhook/webhook.go#L235-L242).

## Open Questions

How does this interact with the "pause" option from the GitRepo spec?

> Paused this cause changes in Git to not be propagated down to the clusters but instead mark resources as OutOfSync

Maybe paused should be propagated to gitjob instead of creating a new option?

manno:
Note, there is an (unused?) `Provider: polling` field in the gitjob spec: https://github.com/rancher/gitjob/blob/master/pkg/apis/gitjob.cattle.io/v1/types.go#L65

0xavi0:
# QA Template

## Solution
When new option `disablePolling` is set to `true` git repositories are not scanned periodically

## Testing
Apply a `fleet.yaml` file which refers to a git repository and don't set the `disablePolling` option. 
That should scan the repository every 15 seconds. (as of now)
Push something to the repo in that interval of time and it should be scanned and applied by Fleet.

Add the `disablePolling` option set to true and push some more changes to the repository.
Fleet should not scan the repository and changes should not be applied.

## Additional info

sbulage:
**System Information** 
* Rancher version: v2.9-f26d9888
* Fleet Version: 104.0.0+up0.10.0-rc.13
---
**TEST RESULT**

Scenarios
|Scenario|Test Case| Result | QASE ID added
|---|---|---|---|
|1|As an admin user, add `GitRepo` with disablePolling option set to `false` which will update GitRepo with latest changes made in GitHub repository.| :white_check_mark: |:ballot_box_with_check:
|2|As an admin user, add `GitRepo` with disablePolling option set to `true` which will not sync latest changes to the GitRepo with from GitHub repository.| :white_check_mark: |:ballot_box_with_check:
[...]

PR #2226, opened 2024-03-14
Adds option to disable polling for Git repos

0xavi0:
Adds the `disableGitPolling` option in order to disable polling for git repos. 

It also ports gitrepos polling to use the `quartz` library to be inline with `imagescan`, that is already using it.

Changes the unit tests to be implemented using the Ginkgo testing framework.

**NOTE**
Bumps `quartz` library to `0.11.2`, which fixes https://github.com/reugn/go-quartz/issues/119

Refers to: https://github.com/rancher/fleet/issues/1572


internal/quartzretry/quartzretry.go
```diff
@@ -0,0 +1,42 @@
...
+		time.Sleep(1 * time.Millisecond)
+	}
+	return nil, err
+}
+
+func DeleteScheduledJobWithRetries(scheduler quartz.Scheduler, key *quartz.JobKey) error {
+	retries := maxSchedulerOperationRetries
+	var err error
+	for retries > 0 {
+		err = scheduler.DeleteJob(key)
```
bigkevmcd, line 34: You could ignore `quartz.ErrJobNotFound` in this case?

internal/quartzretry/quartzretry.go
```diff
@@ -0,0 +1,42 @@
...
+package quartzretry
+
+import (
+	"time"
+
+	"github.com/reugn/go-quartz/quartz"
+)
+
+const maxSchedulerOperationRetries = 3
+
```
bigkevmcd, line 14: I think the two exported functions should have Doc comments? https://go.dev/wiki/CodeReviewComments#doc-comments

Also, I wonder if instead of this bespoke mechanism, this could use https://pkg.go.dev/k8s.io/client-go/util/retry#OnError ?

pkg/apis/fleet.cattle.io/v1alpha1/gitrepo_types.go
```diff
@@ -129,6 +129,9 @@ type GitRepoSpec struct {
 
 	// CorrectDrift specifies how drift correction should work.
 	CorrectDrift *CorrectDrift `json:"correctDrift,omitempty"`
+
+	// Disables git polling. When enabled only webhooks will be used.
+	DisableGitPolling bool `json:"disableGitPolling,omitempty"`
```
bigkevmcd, line 134: Could this be shorted to `DisablePolling` as it doesn't do any other form of polling?

pkg/git/poll/gitrepopolljob.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+type GitRepoPollJob struct {
+	sem     *semaphore.Weighted
+	client  client.Client
+	GitRepo v1alpha1.GitRepo
+	fetcher GitFetcher
+}
+
+func gitRepoPollDescription(gitRepo v1alpha1.GitRepo) string {
```
bigkevmcd, line 31: I wonder if this couldn't just be `String() string` and return that in the Description?

pkg/git/poll/gitrepopolljob.go
```diff
@@ -0,0 +1,86 @@
...
+
+func (j *GitRepoPollJob) Description() string {
+	return gitRepoPollDescription(j.GitRepo)
+}
+
+func (j *GitRepoPollJob) fetchLatestCommitAndUpdateStatus(ctx context.Context) {
+	logger := ctrl.Log.WithName("git-latest-commit-poll-watch")
+	commit, err := j.fetcher.LatestCommit(ctx, &j.GitRepo, j.client)
+	if err != nil {
+		logger.Error(err, "error fetching commit", "gitrepo name", j.GitRepo.Name)
```
bigkevmcd, line 68: I'd recommend using `gitrepo_name` to make this easier to query for in log storage.

pkg/git/poll/handler.go
```diff
@@ -24,70 +32,96 @@ type Watcher interface {
...
-		watch.UpdateGitRepo(gitRepo)
-
-		gitRepoSyncInterval := 0.0
-		if pi := gitRepo.Spec.PollingInterval; pi != nil {
-			gitRepoSyncInterval = pi.Seconds()
+		if gitRepo.Spec.DisableGitPolling {
+			// if polling is disabled, just delete the job from the scheduler
+			err = quartzretry.DeleteScheduledJobWithRetries(h.scheduler, gitRepoPollKey)
+			if err != nil {
+				h.log.Error(err, fmt.Sprintf("Error deleting job %s: %v", gitRepoPollKey.String(), err))
```
bigkevmcd, line 69: Isn't this going to log out the error _twice_ in the message?

pkg/git/poll/handler.go
```diff
@@ -24,70 +32,96 @@ type Watcher interface {
...
+			return
 		}
-
-		if oldSyncInterval != gitRepoSyncInterval {
-			watch.Restart(ctx)
+		job := scheduledJob.JobDetail().Job()
+		gitRepoPollJob, ok := job.(*GitRepoPollJob)
+		if !ok {
+			h.log.Error(fmt.Errorf("invalid job"),
+				fmt.Sprintf("Error getting Gitrepo poll job, the scheduled job: %s is not a GitRepoPollJob", job.Description()))
```
bigkevmcd, line 77: If you implement `.String()` (see earlier comment) you can just pass the `job` here?

pkg/git/poll/handler.go
```diff
@@ -3,17 +3,25 @@ package poll
...
+	"github.com/rancher/fleet/internal/quartzretry"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	maxSchedulerOperationRetries = 3
+	defaultSyncInterval          = 15
```
bigkevmcd, line 22: You can write this as `defaultSyncInterval = 15 * time.Second` which clarifies the scale of the number, you can then simplify the code below to remove the `* time.Second` when waiting?

pkg/git/poll/handler.go
```diff
@@ -24,70 +32,96 @@ type Watcher interface {
...
+	job := NewGitRepoPollJob(h.client, h.fetcher, gitRepo)
+	if runBefore {
+		// ignoring error because that is only used by the quartz library to implement its retries mechanism
+		// The GitRepoPollJob always returns nil
+		_ = job.Execute(ctx)
+	}
+	err := h.scheduler.ScheduleJob(quartz.NewJobDetail(job, jobKey),
+		quartz.NewSimpleTrigger(calculateSyncInterval(gitRepo)))
+	if err != nil {
+		h.log.Error(err, fmt.Sprintf("Error scheduling job %s: %v", jobKey, err))
```
bigkevmcd, line 125: Again, this will log out the error twice?

pkg/git/poll/handler.go
```diff
@@ -24,70 +32,96 @@ type Watcher interface {
...
+			_ = quartzretry.DeleteScheduledJobWithRetries(h.scheduler, gitRepoPollKey)
+			h.scheduleJob(ctx, gitRepoPollKey, gitRepo, true)
 		}
 	}
 }
 
-// CleanUpWatches removes all watches whose gitrepo is not present in the cluster.
-func (h *Handler) CleanUpWatches(ctx context.Context) {
+// CleanUpGitRepoPollJobs removes all poll jobs whose gitrepo is not present in the cluster.
+func (h *Handler) CleanUpGitRepoPollJobs(ctx context.Context) {
```
bigkevmcd, line 91: I don't see any tests for this code?

internal/quartzretry/quartzretry.go
0xavi0: Changed to use `OnError` as suggested. 
I've kept the `quartzretry` package so it's easier to delete once the fix is released.

We'll need to delete the whole package and change:
```go
err = quartzretry.DeleteScheduledJobWithRetries(h.scheduler, gitRepoPollKey)
```
to
```go
err = h.scheduler.DeleteJob(gitRepoPollKey)
```

and 
```go
scheduledJob, err := quartzretry.GetScheduledJobWithRetries(h.scheduler, key)
```
to
```go
scheduledJob, err := h.scheduler.GetScheduledJob(key)
```

pkg/git/poll/gitrepopolljob.go
0xavi0: yup, done.

pkg/git/poll/gitrepopolljob.go
0xavi0: changed

pkg/git/poll/handler.go
0xavi0: `job` is `quartz.Job` and `String()` would be undefined, right?
I guess it would be better that a `quartz.Job` had `String()` instead of `Description()` but...

pkg/git/poll/handler.go
0xavi0: yeah, good point, original code didn't have any and forgot to add those., thanks for spotting! 
I've added some tests now

0xavi0:
@bigkevmcd I bumped `quartz` to 0.11.2, which includes the fix for the concurrency issue and that package for doing retries is no longer needed

bigkevmcd:
@0xavi0 Awesome :-)

pkg/git/poll/gitrepopolljob.go
manno: This is a left-over from following the advice in https://github.com/kubernetes-sigs/controller-runtime/blob/main/TMP-LOGGING.md#logging-structured-values-key-value-pairs

> Use lower-case, space separated keys. For example object for objects, api version for APIVersion

I think, most of the code (see `WithValues`) has been switched to camelCase.

pkg/git/poll/handler.go
manno: This looks strange to me. Just a nit, but isn't `errors.New` preferred when we don't format the error?

More importantly, it looks like the package is using the [structured logger](https://github.com/uber-go/zap?tab=readme-ov-file#quick-start), so I'd suggest something like this for all values:

```suggestion
				"Error getting Gitrepo poll job, the scheduled job is not a GitRepoPollJob", "job", job.Description()))
```

pkg/git/poll/handler.go
```diff
@@ -24,70 +31,96 @@ type Watcher interface {
...
-		watch.UpdateGitRepo(gitRepo)
-
-		gitRepoSyncInterval := 0.0
-		if pi := gitRepo.Spec.PollingInterval; pi != nil {
-			gitRepoSyncInterval = pi.Seconds()
+		if gitRepo.Spec.DisablePolling {
+			// if polling is disabled, just delete the job from the scheduler
+			err = h.scheduler.DeleteJob(gitRepoPollKey)
+			if err != nil {
+				h.log.Error(err, fmt.Sprintf("Error deleting job %s", gitRepoPollKey))
```
manno, line 68: This might be a problem, the signature is
`func (l logr.Logger) Error(err error, msg string, keysAndValues ...any)`

manno (approved):

PR #2227, opened 2024-03-15
Fix Helm chart version computation in release workflow

weyfonk:
The app version is now computed from the git tag triggering the workflow, instead of being unbound.
This also fixes replacements in values files, ensuring that only prefixes are replaced for repositories.

Follow-up to #2168.


.github/workflows/release-fleet.yml
```diff
@@ -91,19 +91,22 @@ jobs:
           repo: "ghcr.io/rancher/fleet"
           tag: "${GITHUB_REF_NAME}"
         run: |
+          # Replace rancher/fleet, rancher/fleet-agent and rancher/gitjob image names, but not eg. rancher/kubectl
           sed -i \
-              -e "s@repository:.*@repository: $repo@" \
-              -e "s/tag:.*/tag: ${VERSION}/" \
+              -e "s@repository: rancher/\(\(fleet.*\|gitjob\)\).*@repository: ${repo/fleet/\\1}@" \
```
manno, line 96:
```suggestion
              -e "s@repository: rancher/\(fleet.*\|gitjob\).*@repository: $repo/fleet/\\1@" \
```

.github/workflows/release-fleet.yml
```diff
@@ -91,19 +91,22 @@ jobs:
...
+              -e "s/tag:.*/tag: ${tag}/" \
               charts/fleet/values.yaml
 
           sed -i \
-              -e "s@repository:.*@repository: $repo@" \
-              -e "s/tag: dev/tag: ${VERSION}/" \
+              -e "s@repository: rancher/\(\(fleet.*\|gitjob\)\).*@repository: ${repo/fleet/\\1}@" \
```
manno, line 101:
```suggestion
              -e "s@repository: rancher/\(fleet.*\|gitjob\).*@repository: $repo/fleet/\\1@" \
```

.github/workflows/release-fleet.yml
```diff
@@ -91,19 +91,22 @@ jobs:
...
+              -e "s/tag: dev/tag: ${tag}/" \
```
manno, line 102: `${}` looks important, but is useless since / can't be part of the variable name.
So, I'd suggest:
```suggestion
              -e "s/tag: dev/tag: $tag/" \
```

To be honest, I find shell variables in regex confusing. This isn't much more readable, though:
```
              -e 's/tag: dev/tag: '"$tag"'/' \
```

manno (changes requested):

.github/workflows/release-fleet.yml
weyfonk: Thanks for spotting the extra pair of parentheses, I'll remove them.
I find the braces still to be needed for string substitution though, omitting them would yield a diff such as:
```
diff --git a/charts/fleet/values.yaml b/charts/fleet/values.yaml
index 8ff51483..ba702362 100644
--- a/charts/fleet/values.yaml
+++ b/charts/fleet/values.yaml
@@ -1,11 +1,11 @@
 image:
-  repository: rancher/fleet
-  tag: dev
+  repository: ghcr.io/rancher/fleet/fleet/fleet
+  tag: v42.0.0
   imagePullPolicy: IfNotPresent
 
 agentImage:
-  repository: rancher/fleet-agent
-  tag: dev
+  repository: ghcr.io/rancher/fleet/fleet/fleet-agent
+  tag: v42.0.0
   imagePullPolicy: IfNotPresent
 
 # For cluster registration the public URL of the Kubernetes API server must be set here
@@ -89,8 +89,8 @@ leaderElection:
   renewDeadline: 25s
 
 gitjob:
-  repository: rancher/gitjob
-  tag: dev
+  repository: ghcr.io/rancher/fleet/fleet/gitjob
+  tag: v42.0.0
   imagePullPolicy: IfNotPresent
   nodeSelector:
     kubernetes.io/os: linux
```

.github/workflows/release-fleet.yml
weyfonk: It is useless, apart from emphasizing the presence of a shell variable :upside_down_face:

.github/workflows/release-fleet.yml
manno: I see, the input is `ghcr.io/rancher/fleet`. 
I'd suggest to change that to  `ghcr.io/rancher` and get rid of the bash parameter expansion inside the regex. Seems redundant.

weyfonk: Oh that makes sense! Done, thanks.

manno (approved):

PR #2229, opened 2024-03-15
Use portable API for bind mounts in integration tests

weyfonk:
This eliminates linter [errors](https://github.com/rancher/fleet/actions/runs/8291590248/job/22691555805).
Those were not caught when bumping `testcontainers/testcontainers-go` because when running against pull requests, `golangci-lint` only runs against the diff which in that case only affected the go mod and sum files.

### DIFF
--- a/integrationtests/gitcloner/clone_test.go
+++ b/integrationtests/gitcloner/clone_test.go
@@ -17,6 +17,8 @@ import (
 	"github.com/rancher/fleet/cmd/gitcloner/cmd"
 	"github.com/rancher/fleet/cmd/gitcloner/gogit"
 
+	dockercontainer "github.com/docker/docker/api/types/container"
+	dockermount "github.com/docker/docker/api/types/mount"
 	"github.com/go-git/go-git/v5"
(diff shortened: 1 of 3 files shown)


integrationtests/gitcloner/clone_test.go
```diff
@@ -17,6 +17,8 @@ import (
 	"github.com/rancher/fleet/cmd/gitcloner/cmd"
 	"github.com/rancher/fleet/cmd/gitcloner/gogit"
 
+	docker_container "github.com/docker/docker/api/types/container"
```
manno, line 20: Aliases with underscores seem uncommon to me. We only do that for `_test`.

Couldn't find advice for import aliases, but for packages:

>  Another convention is that the package name is the base name of its source directory; the package in src/encoding/base64 is imported as "encoding/base64" but has name base64, not encoding_base64 and not encodingBase64. 
https://go.dev/doc/effective_go#package-names

weyfonk: Thanks, underscores removed.

manno (approved):

PR #2239, opened 2024-03-19
Only create Content resource if Bundle matches any target

manno:
Refers to https://github.com/rancher/fleet/issues/2214

### DIFF
--- a/internal/manifest/store.go
+++ b/internal/manifest/store.go
@@ -22,21 +22,21 @@ type ContentStore struct {
 	client client.Client
 }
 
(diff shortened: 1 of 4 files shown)


internal/manifest/store.go
```diff
@@ -24,19 +24,19 @@ type ContentStore struct {
 
 // Store stores the manifest as a content resource and returns the name.
 // It copies the resources from the bundle to the content resource.
```
0xavi0, line 26: I guess we should update this doc as `Store` is no longer returning the name

0xavi0 (approved):
lgtm

PR #2259, opened 2024-03-22
Duplicate integrationtests scripts for devs

manno:
Devs modify these scripts to run tests locally. Protect the CI scripts from accidential changes, as these are overlooked easily in PR reviews.

### DIFF
--- a/.github/scripts/run-integration-tests.sh
+++ b/.github/scripts/run-integration-tests.sh
@@ -6,11 +6,8 @@ SETUP_ENVTEST_VER=${SETUP_ENVTEST_VER-v0.0.0-20221214170741-69f093833822}
 ENVTEST_K8S_VERSION=${ENVTEST_K8S_VERSION-1.25}
 
 # install and prepare setup-envtest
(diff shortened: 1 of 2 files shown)


.github/scripts/run-integration-tests.sh
```diff
@@ -6,11 +6,8 @@ SETUP_ENVTEST_VER=${SETUP_ENVTEST_VER-v0.0.0-20221214170741-69f093833822}
 ENVTEST_K8S_VERSION=${ENVTEST_K8S_VERSION-1.25}
```
aruiz14, line 6: I know it's not nice to do this kind of change in this PR, but I think we should use a newer version. The `0.9` branch uses:
```
SETUP_ENVTEST_VER=${SETUP_ENVTEST_VER-v0.0.0-20240115093953-9e6e3b144a69}
ENVTEST_K8S_VERSION=${ENVTEST_K8S_VERSION-1.28}
```

aruiz14 (approved):

Issue #2128, opened 2024-02-07
BundleDeployoments Status not being properly updated in Fleet local with Rancher 2.8-devel


mmartin24:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When deploying [fleet-examples ](https://github.com/rancher/fleet-examples) with path `simple` on local fleet cluster on Rancher 2.8-devel, it can take more than 20 mins to get clusters in Ready state. This does not seem only an UI issue as the backend displays the error on:

```
➜  ~ k get gitrepos -A -w
NAMESPACE     NAME               REPO                                        COMMIT                                     BUNDLEDEPLOYMENTS-READY   STATUS
fleet-local   time-health-test   https://github.com/rancher/fleet-examples   44f4634747e3dd6c9d1ad6c9d402b430f7aae20b   0/0                       
fleet-local   time-health-test   https://github.com/rancher/fleet-examples   44f4634747e3dd6c9d1ad6c9d402b430f7aae20b   0/0                       
fleet-local   time-health-test   https://github.com/rancher/fleet-examples   44f4634747e3dd6c9d1ad6c9d402b430f7aae20b   0/0                       WaitApplied(1) [Bundle time-health-test-simple]
fleet-local   time-health-test   https://github.com/rancher/fleet-examples   44f4634747e3dd6c9d1ad6c9d402b430f7aae20b   0/1                       WaitApplied(1) [Bundle time-health-test-simple]
fleet-local   time-health-test   https://github.com/rancher/fleet-examples   44f4634747e3dd6c9d1ad6c9d402b430f7aae20b   0/1                       NotReady(1) [Bundle time-health-test-simple]
fleet-local   time-health-test   https://github.com/rancher/fleet-examples   44f4634747e3dd6c9d1ad6c9d402b430f7aae20b   0/1                       NotReady(1) [Bundle time-health-test-simple]; deployment.apps default/frontend [progressing] Deployment does not have minimum availability., Available: 0/3; deployment.apps default/redis-master [progressing] Deployment does not have minimum availability., Available: 0/1; deployment.apps default/redis-slave [progressing] Deployment does not have minimum availability., Available: 0/2
[...]

PR #2267, opened 2024-04-02
Adds extra e2e test to check status display

0xavi0:
Adding an extra test that deploys a bit more complex scenario so it exercises the Watcher area and events that were filtered in PR: https://github.com/rancher/fleet/pull/2031 

This test should fail from `v0.9.1-rc.2` to `v0.9.1-rc.6` and was fixed after reverting the above PR in `v0.9.2-rc.1`

It also introduces the following changes in CI:
* ~bumps `golangci-lint` to the last version~
* ~adds the `skip-pkg-cache: true` to `golangci-lint`~
* ~Uses typos `v1.19.0` as latest version (v1.20.1) adds false positives~ → see #2269 and #2273 

Related to: https://github.com/rancher/fleet/issues/2128

### DIFF
--- a/e2e/single-cluster/single_cluster_test.go
+++ b/e2e/single-cluster/single_cluster_test.go
@@ -3,6 +3,7 @@ package singlecluster_test
 import (
 	"strings"
 
+	"github.com/asaskevich/govalidator"
 	"github.com/rancher/fleet/e2e/testenv"
 	"github.com/rancher/fleet/e2e/testenv/kubectl"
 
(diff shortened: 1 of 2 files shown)


e2e/single-cluster/single_cluster_test.go
```diff
@@ -96,5 +97,48 @@ var _ = Describe("Single Cluster Deployments", func() {
...
+		// This test fails from v0.9.1-rc.2 to v0.9.1-rc.6 due to issue:
+		// https://github.com/rancher/fleet/issues/2128
+		Context("containing multiple deployments", func() {
+			BeforeEach(func() {
+				asset = "single-cluster/helm-status-check.yaml"
+			})
+
+			It("all deployments are ready and status shown is Ready", func() {
+				Eventually(func() bool {
+					frontendReady := checkNReplicasAreReady(k, "frontend", "2")
```
manno, line 109: As discussed: Ideally we would be able to use less heavy deployments with smaller image sizes and no dependencies on physical volumes, etc.

0xavi0: Yup, moving this to draft so it's not merged by mistake

0xavi0: I've changed the test and now it's a simple `k8s.gcr.io/pause` container which should be less than 250kb and just by adding a init container sleeping 5 seconds I was able to recreate the issue with the affected RCs.

manno (approved):

PR #2268, opened 2024-04-02
CI bumps golangci-lint and uses typos 1.19.0

0xavi0:
Bumping `goglangci-lint` to 4.0.0 and 1.57.2 and adding `skip-pkg-cache: true`

Using `typos` 1.19.0 as 1.20.0 and 1.20.1 introduced false errors.
False positive introduced in 1.20.1 is:
```
error: `ro` should be `to`
  --> ./fleet/cmd/codegen/hack/patch_crd_descriptions.sh:27:59
   |
27 |     container_id=$(docker create --rm -i -v ${PWD}:${PWD}:ro -w ${PWD} -v ${tmpdir}:${tmpdir} ${image} yq $@ )
   |                                                           ^^
   |
 ```

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -27,13 +27,15 @@ jobs:
           export PATH=$PATH:/home/runner/go/bin/
 
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@v3.7.0
+        uses: golangci/golangci-lint-action@v4.0.0
         with:
           # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
-          version: v1.56
+          version: v1.57.2
 
           args: --timeout=10m --config=.golangci.json
 
           # Optional: show only new issues if it's a pull request. The default value is `false`.
           # The condition sets this to true for PR events.
           only-new-issues: "${{ github.event_name == 'pull_request'}}"
+
+          skip-pkg-cache: true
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: .github/workflows/typos.yaml [hunk 1]
```diff
@@ -10,5 +10,5 @@ jobs:
       uses: actions/checkout@v4
 
     - name: Check spelling of file.txt
-      uses: crate-ci/typos@master
+      uses: crate-ci/typos@v1.19.0
```
weyfonk, line 13: nit: we could of course find this info again via eg. `git blame`, but leaving a comment here explaining the choice for this specific version would make sense. WDYT?

weyfonk (approved):

.github/workflows/golangci-lint.yml
```diff
@@ -27,13 +27,15 @@ jobs:
           export PATH=$PATH:/home/runner/go/bin/
 
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@v3.7.0
+        uses: golangci/golangci-lint-action@v4.0.0
         with:
           # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
-          version: v1.56
+          version: v1.57.2
```
manno, line 33: nit: we could allow patch updates
```suggestion
          version: v1.57
```

.github/workflows/typos.yaml
manno: We do instruct dependabot to update our github actions: https://github.com/rancher/fleet/blob/main/.github/dependabot.yaml#L38-L44

So, eventually we would get a green PR?

0xavi0: Anyway... I've added a note.
My plan is to re-test this next week as `1.20.1` was released fast after `1.20.0` and I suspect more releases will come shortly. 
It looks like `1.20.0` introduced way to many false positives.

manno (approved):

Issue #1606, opened 2023-06-19
"SIGSEGV: segmentation violation" while trying to deploy Elemental via fleet


e-minguez:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Trying to deploy elemental via fleet , the gitrepo is failing as:
```
$ kubectl get gitrepo elemental -n fleet-local -o yaml
...
  - lastUpdateTime: "2023-06-16T15:43:03Z"
    message: "panic: runtime error: invalid memory address or nil pointer dereference\n[signal
      SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0x155764b]\n\ngoroutine
      52 [running]:\nhelm.sh/helm/v3/pkg/registry.(*Client).Tags(0x0, {0xc0000a0186?,
      0xc0004753f8?})\n\t/go/pkg/mod/github.com/rancher/helm/v3@v3.10.3-fleet1/pkg/registry/client.go:602
      +0x12b\nhelm.sh/helm/v3/pkg/downloader.(*ChartDownloader).getOciURI(0xc000475910,
      {0xc0000a0180, 0x60}, {0x0, 0x0}, 0xc0008377a0)\n\t/go/pkg/mod/github.com/rancher/helm/v3@v3.10.3-fleet1/pkg/downloader/chart_downloader.go:154
      +0x129\nhelm.sh/helm/v3/pkg/downloader.(*ChartDownloader).ResolveChartVersion(0xc000475910,
      {0xc0000a0180, 0x60}, {0x0, 0x0})\n\t/go/pkg/mod/github.com/rancher/helm/v3@v3.10.3-fleet1/pkg/downloader/chart_downloader.go:199
      +0x13df\nhelm.sh/helm/v3/pkg/downloader.(*ChartDownloader).DownloadTo(0xc000475910,
      {0xc0000a0180, 0x60}, {0x0?, 0x20063a0?}, {0xc00083b338, 0x14})\n\t/go/pkg/mod/github.com/rancher/helm/v3@v3.10.3-fleet1/pkg/downloader/chart_downloader.go:90
      +0x5b\ngithub.com/rancher/fleet/pkg/bundlereader.downloadOCIChart({0xc0000a0180,
      0x60}, {0x0, 0x0}, {0xc00083b338, 0x14}, {{0x0, 0x0}, {0x0, 0x0}, ...})\n\t/go/src/github.com/rancher/fleet/pkg/bundlereader/loaddirectory.go:192
[...]

PR #2311, opened 2024-04-09
Catch OCI version errors and return them

manno:
Refers to #1606 

The module we use would panic if version wasn't set or invalid.

With this change, an explicit version is not required for OCI charts. The module will reach out to the registry and fetch the tags and will resolve them to a version.
 
```
		// If empty, try to get the highest available tag
		// If exact version, try to find it
		// If semver constraint string, try to find a match
```
https://github.com/helm/helm/blob/release-3.14/pkg/downloader/chart_downloader.go#L144-L175

### DIFF
--- a/internal/bundlereader/loaddirectory_test.go
+++ b/internal/bundlereader/loaddirectory_test.go
@@ -350,30 +350,60 @@ func TestGetContentOCI(t *testing.T) {
 		source  string
 		version string
 
+		result      []string
 		expectedErr string
 	}{
+		// Note: These tests rely on external hosts and DNS resolution.
+		// We could just test the helm registry client is initialized
+		// correctly, however for now these tests document different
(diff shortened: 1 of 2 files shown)


internal/bundlereader/loaddirectory_test.go
```diff
@@ -344,6 +344,91 @@ func TestGetContent(t *testing.T) {
...
+func TestGetContentOCI(t *testing.T) {
+	cases := []struct {
+		name    string
+		source  string
+		version string
+
+		result      []string
+		expectedErr string
+	}{
+		// Note: These tests rely on external hosts and DNS resolution.
```
manno, line 356: Alternatively I can change the code, so I can mock the registry client. I'd then delete these test (because they mostly test the helm code) and if the client is initialized and if error wrapping works.

PR #2321, opened 2024-04-11
Improve release scripts

thardeck:
to make sure that the generated patch still applies after Gitjob bumps.

This [pr](https://github.com/rancher/charts/pull/3772) is the result and was created by this [action run](https://github.com/rancher/fleet/actions/runs/8644320049).

And verify version availability before submitting the release to rancher/rancher. Example result  for our [latest version](https://github.com/rancher/charts/pull/3772) which is not yet in rancher/charts: https://github.com/rancher/fleet/actions/runs/8647775689/job/23710019166

### DIFF
--- a/.github/scripts/release-against-rancher.sh
+++ b/.github/scripts/release-against-rancher.sh
@@ -24,6 +24,13 @@ if [ ! -e ~/.gitconfig ]; then
     git config --global user.email fleet@suse.de
 fi
 
+# Check if version is available online
+CURRENT_RANCHER_VERSION=$(git rev-parse --abbrev-ref HEAD | cut -d'/' -f 2)
+if ! curl -s --head --fail "https://github.com/rancher/charts/raw/dev-${CURRENT_RANCHER_VERSION}/assets/fleet/fleet-${NEW_CHART_VERSION}+up${NEW_FLEET_VERSION}.tgz" > /dev/null; then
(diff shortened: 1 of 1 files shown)


.github/scripts/release-against-rancher.sh
```diff
@@ -24,6 +24,13 @@ if [ ! -e ~/.gitconfig ]; then
     git config --global user.email fleet@suse.de
 fi
 
+# Check if version is available online
```
manno, line 27: This check can't be done in the github action, before calling release? I guess, because we sometimes have to run the script manually?

.github/scripts/release-against-charts.sh
```diff
@@ -36,6 +36,12 @@ for i in fleet fleet-crd fleet-agent; do
     yq --inplace ".${i} += [\"${NEW_CHART_VERSION}+up${NEW_FLEET_VERSION}\"]" release.yaml
 done
 
+# Adapt Gitjob version in generated patch
+if grep -q '^-  version: ' ./packages/fleet/fleet/generated-changes/patch/Chart.yaml.patch; then
```
manno, line 40: This check is needed to skip the check for fleet >0.10.x where gitjob is included?

.github/scripts/release-against-rancher.sh
manno: I just like the workflows where validations are run first and are separate from the "doer"-scripts :)

manno (approved):

Issue #2238, opened 2024-03-19
Imagescan: x509: certificate signed by unknown authority


mverleun:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When imageScan tries to connect to a private registry (Harbor) which is configured with a private CA the tls connection fails.

### Expected Behavior

This is expected behaviour because Imagescan does not allow to configure a private CA, only a secretRef to connect to a (private) registry.

### Steps To Reproduce

Create an Imagescan object in a cluster that is using a private CA. Try to connect to a registry which is also using a private CA, but a different one.
Check the logs of `fleet-controller` or look at the status of the object:

```
apiVersion: fleet.cattle.io/v1alpha1
kind: ImageScan
metadata:
...
spec:
  gitrepoName: image-tooling
  image: harbor.local/image-name
  interval: 0s
  policy:
    alphabetical:
      order: asc
    semver:
      range: '*'
  tagName: comm-docs
status:
  conditions:
    - lastUpdateTime: '2024-03-19T11:55:22Z'
      message: >-
        Get "https://harbor.local/v2/": tls: failed to verify
        certificate: x509: certificate signed by unknown authority; Get
        "http://harbor.local/v2/": dial tcp 10.1.1.1:80: i/o
        timeout
      reason: Error
      status: 'False'
      type: ImageScanned
  lastScanTime: null
```
  

### Environment

```markdown
- Architecture: x64
- Fleet Version: 0.9.0
- Cluster:
  - Provider: Rancher, K3S
  - Options: Installed as part of Rancher
  - Kubernetes Version: 1.27.9+k3s2
```

georgeyaodi:
I was blocked with the same error on my validation environment, I am validating fleet with image scanning function to automate the CD work. I configured secret for the harbor, since my harbor used the private CA, the GitRepo can not be created as expected, Rancher prompted "certificate signed by unknown authority."
![image](https://github.com/rancher/fleet/assets/28613988/2aa30d43-c61c-4f23-a27b-958e1ad8f7da)

kkaempf:
imagescan is deprecated, closing.

PR #2322, opened 2024-04-11
Support custom CAs in the tag scanner.

bigkevmcd:
Refers to https://github.com/rancher/fleet/issues/2238

Create an installation of any image registry (harbor, distribution, quay.io etc) and place it behind a custom-CA signed TLS certificate

Image scanning will fail with an error, this fixes the error by allowing the admin to provide the CA to use to establish the TLS connection.

This adds a new field .spec.caBundleRef which can reference a ConfigMap with a `ca.crt` field containing a PEM encoded CA bundle.

If this is configured, the CA is used to authenticate requests when listing tags.


pkg/apis/fleet.cattle.io/v1alpha1/imagescan_types.go
```diff
@@ -62,6 +62,13 @@ type ImageScanSpec struct {
 	// +nullable
 	SecretRef *corev1.LocalObjectReference `json:"secretRef,omitempty"`
 
+	// CABundleRef can be given the name of a ConfigMap containing a TLS CA to
+	// be used when connecting to the Image Registry.
+	//
+	// The ConfigMap should have a key `ca.crt` with a PEM encoded CA bundle.
+	// +nullable
+	CABundleRef *corev1.LocalObjectReference `json:"caBundleRef,omitempty"`
```
manno, line 70: We could use https://github.com/rancher/fleet/blob/main/pkg/apis/fleet.cattle.io/v1alpha1/bundledeployment_types.go#L285 since it's already there.

The GitRepo uses a []byte for the PEM:
https://github.com/rancher/fleet/blob/main/pkg/apis/fleet.cattle.io/v1alpha1/gitrepo_types.go#L89-L91

A separate config map seems better, as it's easier to create. Using the same approach as the Gitrepo would be consistent? Not sure if that's important here. Ideally they would use the same
config map?

bigkevmcd: I followed the SecretRef above which also uses the corev1.LocalObjectReference ?

bigkevmcd: Yeah, as I said in chat, all repositories within the same registry would presumably use the same CA, duplicating this across all `ImageScan` resources feels wasteful, never mind that it adds maintenance cost.

I think we could add support to `GitRepo` to allow you to reference the same `ConfigMap`.

smallteeths:
Hi @bigkevmcd,

Could you share what led to the decision to close this pr? I’ve tried using a GitRepo with caBundle for the ImageScanner, but I’m still hitting a 503 error. Is there a different approach recommended, or plans to revisit this feature?

截屏2025-06-24 下午3 10 20

Thanks for any guidance

PR #2323, opened 2024-04-11
Add static sharding support

weyfonk:
This adds support for static sharding to Fleet controllers, as follows:
* Fleet can be deployed with `--set shards="<comma-separated IDs>"`, which will create one controller per shard, named after that shard, plus the usual (unsharded) Fleet controller
* Each sharded Fleet controller is labeled with `shard=<shard ID>`
* Any invalid shard ID, which would not fit in a k8s resource name, will lead to the corresponding Fleet controller not being deployed

Then, a `GitRepo` can be deployed with label `fleet.cattle.io/shard: <shard ID>`, meaning that only the Fleet controller with that label will reconcile it and its children (bundles, bundle deployments). This is end-to-end tested with `GitRepo` creation, but should also work when creating bundles directly.

Any unsharded resource will be processed by the unsharded controller.
Any resource with an unknown shard ID, ie not configured at Fleet deployment time, will not be reconciled.

A controller's leader election ID now contains its shard ID. This prevents controllers from stepping on each other's toes and, while leader election itself is not of much use for as long as each shard ID only maps to a single Fleet controller, this configuration should enable enhancements relying on multiple controllers per shard ID.

Refers to #1740.

### DIFF
--- a/charts/fleet/templates/deployment.yaml
+++ b/charts/fleet/templates/deployment.yaml
@@ -15,6 +15,7 @@ spec:
     metadata:
       labels:
         app: fleet-controller
+        shard: "{{ . }}"
     spec:
       containers:
       - env:
--- a/charts/fleet/templates/deployment.yaml
+++ b/charts/fleet/templates/deployment.yaml
@@ -6,7 +6,7 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: "fleet-controller{{if . }}-shard-{{ . }}{{end}}" # TODO normalise strings or use indexes!
+  name: "fleet-controller{{if . }}-shard-{{ . }}{{end}}"
 spec:
   selector:
     matchLabels:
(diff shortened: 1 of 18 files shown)


charts/fleet/templates/deployment.yaml
```diff
@@ -18,55 +24,59 @@ spec:
...
         command:
         - fleetcontroller
-        {{- if not .Values.gitops.enabled }}
+        {{- if not $.Values.gitops.enabled }}
         - --disable-gitops
         {{- end }}
-        {{- if .Values.debug }}
+        {{- if . }}
+        - --shard-id
+        - {{ . }}
```
manno, line 74:
```suggestion
        - {{ quote . }}
```

Or we document id has to start with a letter.

manno (approved):

charts/fleet/templates/deployment.yaml
weyfonk: Good point, fixed, thanks!

manno (approved):

PR #2326, opened 2024-04-12
Update dev/README Dockerfile for act container image

p-se:
This update enables certain updated Github Actions to run, like `actions/setup-go@v5`.

### DIFF
--- a/dev/README.md
+++ b/dev/README.md
@@ -258,9 +258,11 @@ fleet.
 FROM ubuntu:22.04
 
 ARG BUILDARCH=amd64
+ARG NVM_VERSION=v0.39.7
+ARG NODE_VERSION=20
 
 RUN apt update && apt upgrade -y
-RUN apt install -y wget curl git jq nodejs
+RUN apt install -y wget curl git jq zstd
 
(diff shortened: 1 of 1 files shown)


dev/README.md
```diff
@@ -281,6 +281,13 @@ RUN curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/s
         chmod +x ./kubectl && \
         mv ./kubectl /usr/local/bin/kubectl
 
+RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.7/install.sh | bash && \
```
bigkevmcd, line 284: You could maybe use an [ARG](https://docs.docker.com/reference/dockerfile/#arg) with a default for this version?

Calls out the dependency for the build better?

p-se: Hmm, the real dependency is the latest LTS version of Node.

p-se: I like the idea of making the versions visible at the beginning of the script.

weyfonk (approved):

PR #2329, opened 2024-04-15
Ignore not-found errors when deleting bundle deployments

weyfonk:
When deleting a GitRepo, Fleet takes care of deleting bundle deployments deployed from that GitRepo, both:
* from the GitRepo reconciler
* from the bundle reconciler, when bundle deletion is triggered by reconciling the GitRepo

The latter would previously error, as the bundle reconciler would try to delete bundle deployments which would already have been deleted by the GitRepo reconciler.
This commit instructs the bundle reconciler not to consider a bundle deployment not being found as an error when deleting it, thereby preventing confusing reconciliation error messages in the Fleet controller.

### DIFF
--- a/internal/cmd/controller/reconciler/gitrepo_controller.go
+++ b/internal/cmd/controller/reconciler/gitrepo_controller.go
@@ -270,9 +270,9 @@ func purgeBundles(ctx context.Context, c client.Client, gitrepo types.Namespaced
 			return err
 		}
 
-		err = purgeBundleDeployments(ctx, c, types.NamespacedName{Namespace: bundle.Namespace, Name: bundle.Name})
-		if client.IgnoreNotFound(err) != nil {
-			return err
+		nn := types.NamespacedName{Namespace: bundle.Namespace, Name: bundle.Name}
+		if err = purgeBundleDeployments(ctx, c, nn); err != nil {
+			return client.IgnoreNotFound(err)
 		}
 	}
 


bigkevmcd:
Can we add tests for this behaviour?

weyfonk:
> Can we add tests for this behaviour?

Good point, I've just added checks to `GitRepo` end-to-end tests, looking for errors in Fleet controller logs. Those checks fail if we revert the first commit of this PR, and are deliberately vague to catch as many errors as possible in the future.

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -61,7 +61,8 @@ func (r *BundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	err := r.Get(ctx, req.NamespacedName, bundle)
 	if apierrors.IsNotFound(err) {
 		logger.V(1).Info("Bundle not found, purging bundle deployments")
-		if err := purgeBundleDeployments(ctx, r.Client, req.NamespacedName); err != nil {
+		err := purgeBundleDeployments(ctx, r.Client, req.NamespacedName)
+		if client.IgnoreNotFound(err) != nil {
 			return ctrl.Result{}, err
```
bigkevmcd, line 66: You can simplify this to...

```go
if err := purgeBundleDeployments(ctx, r.Client, req.NamespacedName); err != nil {
        return ctrl.Result{}, client.IgnoreNotFound(err)
}
```

weyfonk: Simplified, thanks.

internal/cmd/controller/reconciler/gitrepo_controller.go
```diff
@@ -271,7 +271,7 @@ func purgeBundles(ctx context.Context, c client.Client, gitrepo types.Namespaced
 		}
 
 		err = purgeBundleDeployments(ctx, c, types.NamespacedName{Namespace: bundle.Namespace, Name: bundle.Name})
-		if err != nil {
+		if client.IgnoreNotFound(err) != nil {
 			return err
```
bigkevmcd, line 275: Sorry, should've pointed out you can simplify this one too....

Basically `client.IgnoreNotFound` returns nil if err is a not found error, else it returns the error.

But this is definitely a nitpick.

bigkevmcd (approved):

internal/cmd/controller/reconciler/gitrepo_controller.go
weyfonk: Makes sense, I could have thought of it TBH... On the other hand, I was having doubts as that line is already long enough, but... Done :)

PR #2336, opened 2024-04-17
Switch to tini

manno:
catatonit is statically compiled and 10x larger


weyfonk:
Do we even need to install tini? I've just found this in its README:
> NOTE: If you are using Docker 1.13 or greater, Tini is included in Docker itself. This includes all versions of Docker CE. To > enable Tini, just [pass the --init flag to docker run](https://docs.docker.com/engine/reference/commandline/run/).

weyfonk (approved):

manno:
> Edit: not sure we'd have access to that flag when running images in k8s, plus compatibility, so scratch this.


Yes, different container runtime 🤷

weyfonk (approved):

Issue #2345, opened 2024-04-22
Gitjob is not logging webhook payload anymore


thehejik:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

In `rancher:v2.9-head-c9be13b` with preinstalled `fleet:104.0.0+up0.10.0-rc.11` there are no logs entries for webhook payload anymore.

I was testing with private azure https gitrepo with created secret.

The webhook payload messages are not appearing even when `--debug` is specified in gitjob deployment.

#### Reproducer:
* Install rancher v2.9-head
* add a azure gitrepo with `spec.pollingInterval: 24h` to disable polling into `fleet-local` namespace
* enable webhook on dev.azure.com for that repo
* Then trigger the webhook by pushing a commit to repository (you may also use `Test` button in azure webhook settings) 
* At the same time observe `kubectl logs -n cattle-fleet-system -lapp=gitjob -f` output. There wont be anything appearing but the resources will be updated by fleet.

rubhanazeem:
@thehejik can you please double check if this is still the case?

I am currently on fleet `v0.10.0-rc.13` and I can see webhook payload in gitjob logs.

```
webhook    Webhook payload    {"payload": {"ref":"refs/heads/master","before":"","after":"7b30bce854ffcd27b976d5673aa2ef8d66fe08ed","created":false,"del │
│ eted":false,"forced":false,"base_ref":null,"compare":"","commits":null,"head_commit":{"id":"","node_id":"" ...
```

thehejik:
I still don't the payload in the logs, I used `spec.disablePolling: true` this time in gitrepo and github webhook - the webhook is working fine as the resources were provisioned on push to my github repo. Maybe it is somehow related to rancher.

Tested on local rancher 2.9-head with `fleet:104.0.0+up0.10.0-rc.13`

thehejik:
Payload still now appearing in rancher:v2.9-head with `fleet:104.0.0+up0.10.0-rc.14`

@rubhanazeem can we unblock this please?

thehejik:
Got it ... problem is that the standalone `dev/setup-fleet` script is using extra helm options `--set debug=true --set debugLevel=1`.

Without them the payload is not visible, when I use them the payload is appearing in gitjob logs.
```
helm repo add fleet https://rancher.github.io/fleet-helm-charts/
[...]

PR #2367, opened 2024-04-24
Fix gitjob log messages for controller and webhook

manno:
Refers to #2345

### DIFF
--- a/internal/cmd/controller/reconciler/cluster_controller.go
+++ b/internal/cmd/controller/reconciler/cluster_controller.go
@@ -173,7 +173,7 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 			"namespace", cluster.Namespace,
 			"name", cluster.Name,
 			"ready", cluster.Status.ResourceCounts.Ready,
-			"DesiredReady", cluster.Status.ResourceCounts.DesiredReady,
+			"desiredReady", cluster.Status.ResourceCounts.DesiredReady,
 		)
 
 		// Counts from gitrepo are out of sync with bundleDeployment state, retry in a number of seconds.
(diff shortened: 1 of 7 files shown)


internal/cmd/controller/reconciler/cluster_controller.go
```diff
@@ -170,8 +169,12 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	}
 
 	if allReady && cluster.Status.ResourceCounts.Ready != cluster.Status.ResourceCounts.DesiredReady {
-		logrus.Debugf("Cluster %s/%s is not ready because not all gitrepos are ready: %d/%d, enqueue cluster again",
-			cluster.Namespace, cluster.Name, cluster.Status.ResourceCounts.Ready, cluster.Status.ResourceCounts.DesiredReady)
+		logger.V(1).Info("Cluster is not ready, because not all gitrepos are ready",
+			"namespace", cluster.Namespace,
+			"name", cluster.Name,
+			"ready", cluster.Status.ResourceCounts.Ready,
+			"DesiredReady", cluster.Status.ResourceCounts.DesiredReady,
```
weyfonk, line 176: Just curious: is there a rationale for upper- vs lowercase first chars of log fields here?
Are they deliberately inconsistent between themselves for the sake of keeping `DesiredReady` consistent with the state name it refers to?

pkg/git/lsremote.go
```diff
@@ -158,14 +163,14 @@ func (g *git) remoteSHAChanged(branch, sha string) (bool, error) {
 
 	client, err := g.httpClientWithCreds()
 	if err != nil {
-		logrus.Warnf("Problem creating http client to check git remote sha of repo [%v]: %v", g.URL, err)
+		g.log.Error(err, "Problem creating http client to check git remote sha of repo", "url")
```
weyfonk, line 166: Do we need an additional `"url"` parameter there, given that it's already provided in `g.log`? Plus, the signature of `Error` suggests that we should provide key/value _pairs_. Or am I missing something?
```suggestion
		g.log.Error(err, "Problem creating http client to check git remote sha of repo")
```

internal/cmd/controller/reconciler/cluster_controller.go
manno: Right, should be `desiredReady`.

manno:
```suggestion
			"desiredReady", cluster.Status.ResourceCounts.DesiredReady,
```

weyfonk (approved):

PR #2377, opened 2024-04-26
Embed controller-runtimes SchemeBuilder

manno:
Users of our apis package should not need to import controller-runtime


manno:
Turns out Rancher is using the exported type SchemeBuilder. Need to revert.

PR #2378, opened 2024-04-26
Remove go.mod from infra test setup command

manno:
infra uses the same modules as the other binaries.

We lose the ability to use another k8s version and switch to the helm fork.


dev/README.md
```diff
@@ -171,10 +171,8 @@ The local infra setup creates pods for:
 To build and run the infra setup command do:
 
 ```
-pushd e2e/testenv/infra
-  go build -o . ./...
-popd
-./e2e/testenv/infra/infra setup
+go build -o infra ./e2e/testenv/infra/
+./infra setup
```
weyfonk, line 175: This would become inconsistent with `dev/setup-{single,multi}-cluster`, so we should either edit them too or keep this as is.

manno: Thanks, I didn't know those (and missed them in my search)

weyfonk (approved):

Issue #2359, opened 2024-04-23
[SURE-8270] Gitjob controller security context


manno:
The fleet-controller and fleet-agent dissallow privilege escalation and use a read only filesystem. Gitjob controller should do the same.

* https://github.com/rancher/fleet/blob/main/charts/fleet/templates/deployment.yaml#L172-L178
* https://github.com/rancher/fleet/blob/main/charts/fleet-agent/templates/deployment.yaml#L30-L36

See https://github.com/rancher/fleet/blob/main/charts/fleet/templates/deployment_gitjob.yaml
We will need to add empty dirs for the writablel directories.

 
These must be disabled in debug mode, so we can attach debuggers and profilers.

sbulage:
QA Observation:

While checking the expected Security Context block under the GitJob deployment, below information was present:

<details>
<summary> Gitjob Deployment </summary>

![image](https://github.com/rancher/fleet/assets/12134754/f9cc1a54-90d3-4f02-9496-49ae12e9cded)
<details>

sbulage:
### **System Information**
**Before upgrade:**
- Rancher Version: `2.8.5-rc3`
- Fleet Version: `v0.9.5`
---
**After upgrade:**
- Rancher Version: `2.9.0-alpha5`
- Fleet Version: `v0.10.0-rc.16`
---
**Below Steps performed:**

- Created GitRepo in Rancher `2.8.5-rc3`.
- Wait for the application to deploy
- Check Gitjob pod for `securityContext` values: `readOnlyRootFilesystem`, `allowPrivilegeEscalation` and `drop all capabilities`.
- Upgrade Rancher to `2.9.0-alpha5`
- Wait for the Fleet to upgrade to `0.10.0-rc.16`
- Wait for the GitRepo is ready.
- Verify that Gitjob pod has `securityContext` values: `readOnlyRootFilesystem`, `allowPrivilegeEscalation` and `drop all capabilities`.
---
<details>
<summary><b>See screenshot for more details.</b></summary>

![Screenshot from 2024-06-17 19-32-50](https://github.com/rancher/fleet/assets/12134754/68ef9086-4ee0-47bd-9350-f61e47a2f45b)

</details>

PR #2379, opened 2024-04-26
Add readOnlyRootFilesystem to deployment_gitjob.yaml

thardeck:
Refers to #2359

I have added tests for ssh connections with a `known_hosts` file, because this operation requires writing files.

But since gitcloner runs in a different environment then the Gitjob controller it is not affected by the new file system restrictions.

### DIFF
--- a/integrationtests/gitcloner/clone_test.go
+++ b/integrationtests/gitcloner/clone_test.go
@@ -581,8 +581,9 @@ func getKnownHostEntry(container testcontainers.Container) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	algorithm := strings.Split(string(publicHostKeyBytes), " ")[0]
-	hostKey := strings.Split(string(publicHostKeyBytes), " ")[1]
+	fields := strings.Split(string(publicHostKeyBytes), " ")
+	algorithm := fields[0]
+	hostKey := fields[1]
 
 	return fmt.Sprintf("[localhost]:%s %s %s", port, algorithm, hostKey), nil
 }


charts/fleet/templates/deployment_gitjob.yaml
```diff
@@ -59,4 +59,5 @@ spec:
         runAsNonRoot: true
         runAsUser: 1000
         runAsGroup: 1000
+        readOnlyRootFilesystem: true
```
weyfonk, line 62: Do we want to also, as done in other deployments:
* run containers unprivileged?
* prevent privilege escalation?
* drop all POSIX capabilities?
If not, why not?

integrationtests/gitcloner/clone_test.go
```diff
@@ -504,6 +566,27 @@ func createAndAddKeys() (string, error) {
...
+	publicHostKey, err := container.CopyFileFromContainer(context.Background(), "/data/ssh/ssh_host_ecdsa_key.pub")
+	if err != nil {
+		return "", err
+	}
+	publicHostKeyBytes, err := io.ReadAll(publicHostKey)
+	if err != nil {
+		return "", err
+	}
+	algorithm := strings.Split(string(publicHostKeyBytes), " ")[0]
+	hostKey := strings.Split(string(publicHostKeyBytes), " ")[1]
```
weyfonk, line 585: nit: why split the public host key twice?
```suggestion
        fields := strings.Split(string(publicHostKeyBytes), " ")
	algorithm := fields[0]
	hostKey := fields[1]
```

weyfonk (dismissed):

integrationtests/gitcloner/clone_test.go
```diff
@@ -26,6 +26,7 @@ import (
 	cp "github.com/otiai10/copy"
 	"github.com/testcontainers/testcontainers-go"
 	"golang.org/x/crypto/ssh"
+	"golang.org/x/crypto/ssh/knownhosts"
```
manno, line 29: Good to have this test for the git cloner, however it's unrelated to the change in charts/fleet/templates/deployment_gitjob.yaml

PR #2387, opened 2024-04-29
Clean up metrics tests

manno:
* remove boolean argument
* do less http requests
* `Context("When..."` -> `When("..."`
* don't clean up loadbalancer when using env var


weyfonk (approved):
LGTM

Issue #2271, opened 2023-10-09
Feature Request: Ping webhook reply with GitHub webhooks setup


snovak7:
### Is your feature request related to a problem?

When testing out if gitjob webhook endpoint works, it reports error 500, which isn't really a problem in itself

### Solution you'd like

Ping reply with status 200

### Alternatives you've considered

Ignoring the error

manno:
Transferring back to fleet as gitjob is now included.

Tommy12789:
Pull request content for Rancher QA verification.

## Additional QA
### Issue: <link the issue or issues this PR resolves here>

#2395
 
### Problem
No handler for github test (ping) webhook in the source code 
 
### Solution
Added the case into the webhook.go file ("X-Github-Event" == "ping")

### Testing
Send a test webhook from Github and verify the reply from fleet (should be 200)

### Engineering Testing
#### Manual Testing
N/A

#### Automated Testing
./pkg/webhook/... unit test

### QA Testing Considerations
- Test with a ping webhook
- Test the other types of webhooks to see if they are still well handled

#### Regressions Considerations
N.A

mmartin24:
Test preparation detailed here: https://confluence.suse.com/display/~mamartin/Configure+Github+Webhook+and+Fleet
Tested in `v2.8-68e5d43e449fcd3ef73baf1caee98da6f0104eed-head`

- Prepared Github webhook on https://github.com/fleetqa/webhook-private-test/blob/main/deployment.yaml following guide's configuration

I checked:

- Modified replicas number over deployment and confirm inmediate change
![image](https://github.com/user-attachments/assets/6909d4e6-2759-4921-91cb-c52de224cb1f)

- Tested 200 response on Github. Confirmed `X-GitHub-Event: push`
![image](https://github.com/user-attachments/assets/946b4c83-9d87-4e8e-8a80-b2384c9bd075)


Headers
```
Request URL: https://34.48.112.6.nip.io/
Request method: POST
Accept: */*
Content-Type: application/json
User-Agent: GitHub-Hookshot/ef2935b
X-GitHub-Delivery: 80e8764e-5018-11ef-9f6e-4bda091e0e85
X-GitHub-Event: push
X-GitHub-Hook-ID: 493444396
X-GitHub-Hook-Installation-Target-ID: 836698502
X-GitHub-Hook-Installation-Target-Type: repository
X-Hub-Signature: sha1=7648d1680cdd3564325780926ee2d9f69af497d6
[...]

PR #2395, opened 2024-04-30
Creation of a New Webhook endpoint for Github test Webhook

Tommy12789:
Refers to #2271

Added a new case into the webhook.go to handle ping webhooks from Github.

go test ./pkg/webhook/... unit test


pkg/webhook/webhook_test.go
```diff
@@ -96,3 +100,83 @@ func (r *responseWriter) Write([]byte) (int, error) {
...
+
+	// Verify the response status code is correct
+	if status := rr.Code; status != http.StatusOK {
+		t.Errorf("handler returned wrong status code: got %v want %v", status, http.StatusOK)
+	}
+
+	// Verify the response message is correct
+	expectedResponse := "Webhook received successfully"
+	if rr.Body.String() != expectedResponse {
+		t.Errorf("handler returned unexpected body: got %v want %v", rr.Body.String(), expectedResponse)
```
bigkevmcd, line 180: And you can simplify this:

```go
t.Errorf("handler returned unexpected body: got %v want %v", rr.Body, expectedResponse)
```

Because https://pkg.go.dev/fmt#hdr-Printing

> 5. If an operand implements method String() string, that method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any).

bigkevmcd (commented):
Nice work in this.

pkg/webhook/webhook_test.go
Tommy12789: Thank you Kevin, I changed it

bigkevmcd (approved):

pkg/webhook/webhook.go
```diff
@@ -159,6 +159,10 @@ func (w *Webhook) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
 	//Gogs needs to be checked before Github since it carries both Gogs and (incompatible) Github headers
 	case r.Header.Get("X-Gogs-Event") != "":
 		payload, err = w.gogs.Parse(r, gogs.PushEvent)
+	case r.Header.Get("X-Github-Event") == "ping":
+		rw.WriteHeader(http.StatusOK)
+		_, _ = rw.Write([]byte("Webhook received successfully"))
+		return
 	case r.Header.Get("X-GitHub-Event") != "":
 		payload, err = w.github.Parse(r, github.PushEvent)
```
bigkevmcd, line 167: Having read more of this...

```go
        payload, err = w.github.Parse(r, github.PushEvent, github.PingEvent)
        if _, ok := payload.(github.PingPayload); ok {
            _, _ = rw.Write([]byte("Webhook received successfully"))
            return
        }
```
You don't need the `rw.WriteHeader(http.StatusOK)` as that's the default.

manno (approved):

PR #2400, opened 2024-05-02
Clustergroup controller migration

manno:
## Merge clustergroup handler with clustergroup reconciler

Both update the status of the clustergroup, it's more consistent when we
do the calculation in one place and only update the status once.
Especially since they depended on each other, the handler computed the
summary fields, which were used in the reconciler to compute the display
fields.

## Use setCondition/updateErrorStatus in two more controllers

This should provide similar information to the the previous status handling code:

https://github.com/rancher/fleet/blob/release/v0.9/pkg/generated/controllers/fleet.cattle.io/v1alpha1/cluster.go#L106-L117

## Predicates to filter events

`For(...)` and `Watch(...)` can have different filters, the predicates from `WithEventFilter(...)` apply to all events. This is useful, since `For` and `Watch` operate on different types.


weyfonk (commented):
LGTM, but I'd be more comfortable with merging this if we had tests around cluster and cluster group status updates, perhaps as integration tests.

manno:
> LGTM, but I'd be more comfortable with merging this if we had tests around cluster and cluster group status updates, perhaps as integration tests.

I started tests in a [separate PR](https://github.com/rancher/fleet/pull/2416/files#diff-ba6d79cc5de5a9bf2a4858d9f0e6c59c3517aeaa9cf44222058d5dd4a2f69313R114), but it's a lot to back fill those tests.

internal/cmd/controller/reconciler/gitrepo_controller.go
```diff
@@ -244,19 +266,9 @@ func (r *GitRepoReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 				return []ctrl.Request{}
 			}),
+			builder.WithPredicates(bundleStatusChangedPredicate()),
```
weyfonk, line 269: Nice, this means that we could probably skip the check on the object being a bundle within `bundleStatusChangedPredicate`.

weyfonk (approved):

Issue #2344, opened 2024-04-22
metrics: review and optimize collection calls in reconciliation loops


p-se:
It is likely that not all status changes to resources are reflected in metrics. Also, no tests for those exist yet. Perhaps a generalization would also be possible, for instance in conjunction with using `defer`.

PR #2406, opened 2024-05-03
metrics: optimize collection calls in reconciliation loops

p-se:
Those changes will account for updates in the status of GitRepo resources,
which previously would not have been accounted for. This means updates to the
status of a GitRepo may not always have been reflected in metrics.

Changes in other controllers where metrics are being collected were not
necessary:

- Cluster
- ClusterGroup
- Bundle
- BundleDeployment

Refers to #2344


weyfonk (dismissed):
LGTM

Issue #2407, opened 2024-05-03
Feature Request: Controller-side secret decryption at bundle preprocessing via vals


rajiteh:
### Is your feature request related to a problem?

Certain input values for a release defined in a `fleet.yaml` can contain sensitive data that should not be commited to git. A safer strategy would be to specify a reference to the secret in the fleet yaml for readability but resolve it from a secure location at run time.

### Solution you'd like

Ability for the fleet controller to parse and decrypt secret references in the `fleet.yaml` at deploy time when it's preparing bundles. A generic secret processing backend such as https://github.com/helmfile/vals is idea for this as it supports configuration via environment variables. 

Then the fleet controller deployment needs to be enriched with the required environment variables for vals library to resolve secrets from the appropriate backends. (https://github.com/helmfile/vals?tab=readme-ov-file#supported-backends)

framctr:
There is already #591

PR #2408, opened 2024-05-03
feat: use vals for secret decryption

rajiteh:
Refers to #2407 

Introduces the ability for the `fleet.yaml` to specify vals secret strings, which then will be decrypted at bundle processing time by the fleet operator.

Fleet operator deployment is responsible for configuring the appropriate environment variables for `vals` to decrypt secrets using the correct backend.

See #2407  for more context.


weyfonk (commented):
Thank you for this contribution :)
Leaving a couple of comments. One of our main concerns here is having all Helm values sent to an external library, perhaps we could filter values to only pre-process those bearing the known vals pattern.
We are currently assessing possible security implications of this feature.

charts/fleet/values.yaml
```diff
@@ -81,3 +80,6 @@ propagateDebugSettingsToAgents: true
 
 migrations:
   clusterRegistrationCleanup: true
+
+## Additional environment variables to be injected to the fleet-controller container
+additionalEnvs: {}
```
manno, line 85: I think I would prefer an approach similiar to how [nginx](https://artifacthub.io/packages/helm/bitnami/nginx?modal=values) handles env vars from secrets:

```
## @param extraEnvVarsSecret Secret with extra environment variables
##
extraEnvVarsSecret: ""
```

However, the secret would have to be created manually in this case. Which is not as automatic/user friendly?

Additionally we're adding `extraEnvs`, like Rancher, to allow users to specify non-secret env vars as helm values.

rajiteh:
@weyfonk I think you raise a valid concern. I could imagine we introduce a new gotpl helper that wraps the secret ref as an argument? 

'${ decodeSecretRef "ref+foo://my/secret" }'

weyfonk:
We have discussed this approach and will not be merging this in Fleet, because storing confidential values in etcd by means of custom resources (bundles in this case) is not secure by default.

An alternative to this could be using Kubernetes secrets, as @manno noted [here](https://github.com/rancher/fleet/pull/2408#discussion_r1617023162).

manno:
> An alternative to this could be using Kubernetes secrets, as @manno noted [here](https://github.com/rancher/fleet/pull/2408#discussion_r1617023162).

That's the first part, to have the credentials used to access the vals provider in a secret. Second part is to keep the response from vals out of the bundledeployment options and in a secret instead. We want to avoid storing sensitive material in our CRDs.  secrets. 
Doing so would make it more difficult to harden the cluster. E.g. https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/ would need to include `bundledeployments.fleet.cattle.io`. However, the [RKE2 guide](https://docs.rke2.io/security/secrets_encryption) only encrypts secrets.


We are still discussing a possible solution in https://github.com/rancher/fleet/pull/2375. That involves creating a secret in the bundledeployment's namespace (see "cluster namespace" in docs) and allowing the downstream agent to fetch it: 

> When targeting, it also clones the secret containing all the OCI values, just changing the namespace to match the bundledeployment's namespace. (This way the agents will be able to obtain the OCI values) It will also flag the bundledeployment to state that the content is stored in the OCI registry instead to a content resource.

We could do the same for sensitive helm values?

Issue #2404, opened 2024-05-03
Gitrepos based on repositories with default branch different to master fail to apply when revision is used


0xavi0:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

This issue is related to https://github.com/rancher/fleet/issues/2403

Apply the followin `gitrepo` (The important bit is that we're using a revision and that the git repo's default branch is `main`)
```yaml
apiVersion: fleet.cattle.io/v1alpha1
kind: GitRepo
metadata:
  name: sample
  namespace: fleet-local
spec:
  repo: "https://github.com/0xavi0/fleet-experiments"
  revision: v0.0.0
  paths:
  - helm-test-tags
 ```
 
 The `gitrepo` won't deploy successfully and we can observe the following error in `gitjob`:
 ```bash
[...]

Issue #2403, opened 2024-05-03
Gitrepos based on revision show the wrong commit in Status


0xavi0:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I'm using the following `gitrepo`:
```yaml
apiVersion: fleet.cattle.io/v1alpha1
kind: GitRepo
metadata:
  name: sample
  namespace: fleet-local
spec:
  repo: "https://github.com/0xavi0/fleet-experiments-master"
  revision: v0.0.0
  paths:
  - helm-test-tags
 ```

The git repository has 3 tags
* v0.0.0
* v0.0.1
* v0.0.2

After applying the `gitrepo` (which references tag v0.0.0) we can see the following `commit` in the `gitrepo`
```bash
$ kubectl get gitrepo -n fleet-local sample                                                                                                      
NAME     REPO                                                 COMMIT                                     BUNDLEDEPLOYMENTS-READY   STATUS
sample   https://github.com/0xavi0/fleet-experiments-master   b91254d06622d8e0ace1f76c0926b95af2efb63a   1/1
```

But `b91254d06622d8e0ace1f76c0926b95af2efb63a` is the hash for the repository's HEAD, not the commit for tag `v0.0.0`

```bash
$ git ls-remote https://github.com/0xavi0/fleet-experiments-master                                                                                
b91254d06622d8e0ace1f76c0926b95af2efb63a	HEAD
b91254d06622d8e0ace1f76c0926b95af2efb63a	refs/heads/master
b8a525671bfd03f32991b4ccec56517e74b3d7b0	refs/tags/v0.0.0
8e186ca6abd9930e90a129954b4bd0c630f5a7ab	refs/tags/v0.0.0^{}
b2f7eacdeda55833e299efdd6955abb68f581547	refs/tags/v0.0.1
a08281305a4320ba4df9cc877e20a295d0c128a4	refs/tags/v0.0.1^{}
ffad72d77b26152ad74053d736cc1b96c82b7c99	refs/tags/v0.0.2
b91254d06622d8e0ace1f76c0926b95af2efb63a	refs/tags/v0.0.2^{}
```

The issue is happening because the `LatestCommit` function in the [fetch.go](https://github.com/rancher/fleet/blob/44513ca8f61619bad7cab1bc0843a21c7dbdf407/pkg/git/fetch.go#L36:L39) file is using the provided branch to get the last commit or `master` if none is provided.
[...]

PR #2410, opened 2024-05-06
Return commit for revision instead of HEAD

0xavi0:
Changing the commit retrieved when applying `gitrepo`s with revisions. It will show the commit for that revision/tag instead to the latest commit for the default branch (`master`)

It also refactors the code of the `pkg/git` package in order to make it more testable. (Now more than 95% of the code is covered by unit tests)

It splits the previous code into different entities and functionalities. 
* remote.go: this file holds the main functionalities to obtain the latest commit for branches and tags. 
* fetch.go: this file already existed. It is the entry point to the package
* netutils.go: this file holds the functionalities to obtain HTTP clients and transports with the given credentials stored in a secret. (Functions are exported as they could be useful for other packages, maybe?)
* vendor.go: This file contains functions for retrieving the latest commit from a branch faster than running a whole `ls-remote` but that are tied to vendors and their APIs, it is not generic and right now it only supports `github` and `git.rancher.io` 
* validate.go: Previously existing file that contains validations for branches, commits and basic url checks.

It also changes how the latest commit is obtained to use the function using the vendor's API if available and falling back to `ls-remote`. 
Reason why is because the function was already called with the following Headers:
* "Accept" - "application/vnd.github.v3.sha" : This is still used to retrieve just the latest commit.
* "If-None-Match" - "referenceCommit": This was used before which made the query return code 200 AND the latest commit if it was different to the given reference and code 304 (not modified) otherwise. Now it always returns the latest commit, no matter if it's equal to the reference one or not. This way we can save the `ls-remote` query which is more expensive as we already know the latest commit for that branch. 

It also changes the error messages when a commit could not be retrieved, showing the branch or revision used.
This way the use is able to see, for example, that Fleet's defaults to `master` as default branch if the user didn't provide any branch in the `gitrepo` definition and his/her repo has `main` as the default branch.

Refers to: https://github.com/rancher/fleet/issues/2403
Refers to: https://github.com/rancher/fleet/issues/2404


pkg/git/vendor_test.go
```diff
@@ -0,0 +1,115 @@
...
+	"net/http"
+	"net/http/httptest"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("git's vendor specific functions tests", func() {
+	Context("When using not valid url", func() {
```
manno, line 14:
```suggestion
	When("using invalid url", func() {
```

manno (approved):

Issue #2355, opened 2024-04-23
metrics: make sure metrics work well with sharding


p-se:
We need to ensure that metrics are collected against resources reconciled not just by the unsharded Fleet controller, but also by sharded controllers.

~~This could be an opportunity to improve shard label naming, for instance by supporting label-based filtering on the unsharded controller (currently not possible with an empty `shard` value). For instance, we could use `~` instead of an empty string, or simply add a label such as `fleet.cattle.io/empty-shard=true` or similar.~~

A new label `shard-default` with value `true` for the unsharded controller has been implemented in https://github.com/rancher/fleet/pull/2172/commits/436a81de5f452183e607e3983cffd7f02ee88bd0 as part of the PR with the implementation of metrics https://github.com/rancher/fleet/pull/2172. Metrics still does not work with sharded fleet controllers. This is to be remedied.

This is a follow-up on both #1740 and #1408.

weyfonk:
Note: [this documentation](https://gist.github.com/bigkevmcd/41c897580e05475b42d0472c0f8e0bc9) on how Flux had planned to do it may be useful.

PR #2420, opened 2024-05-08
metrics: make sure metrics work well with sharding

p-se:
Refers to #2355

### DIFF
--- a/.github/workflows/nightly-ci.yml
+++ b/.github/workflows/nightly-ci.yml
@@ -111,6 +111,7 @@ jobs:
 
           # 2. Run tests for metrics
           ginkgo e2e/metrics
+          SHARD=shard1 ginkgo e2e/metrics
 
           # 3. Run tests requiring only the git server
           e2e/testenv/infra/infra setup --git-server=true
--- a/charts/fleet/templates/service.yaml
+++ b/charts/fleet/templates/service.yaml
@@ -1,8 +1,13 @@
 {{- if .Values.metrics.enabled }}
+{{ $shards := list "" }}
+{{ if .Values.shards }}
+{{ $shards = concat $shards .Values.shards | uniq }}
+{{ end }}
+{{ range $shards }}
 apiVersion: v1
 kind: Service
 metadata:
-  name: monitoring-fleet-controller
+  name: "monitoring-fleet-controller{{if . }}-shard-{{ . }}{{end}}"
   labels:
     app: fleet-controller
 spec:
@@ -14,4 +19,11 @@ spec:
     name: metrics
   selector:
     app: fleet-controller
+    {{- if empty . }}
+    shard-default: "true"
+    {{- else }}
+    shard: "{{ . }}"
+    {{- end }}
+---
+{{- end }}
 {{- end }}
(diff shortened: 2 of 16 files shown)


manno (dismissed):
Making the whole service template conditional.

dev/README.md
```diff
@@ -415,12 +417,17 @@ spec:
...
-
+      shard-default: true
+      # or:
+      # shard: <shard-id>
 ```
 
-This configures Prometheus to scrape the metrics from the fleet-controller. By
-accessing the Prometheus UI, you can now see the metrics from the
-Fleet-Controller. They are all prefixed with `fleet_`, which you will need to
+This configures Prometheus to scrape the metrics from the (unshared)
```
weyfonk, line 425: nit:
```suggestion
This configures Prometheus to scrape the metrics from the (unsharded)
```

p-se: Thank you!

charts/fleet/templates/service.yaml
```diff
@@ -1,8 +1,13 @@
-{{- if .Values.metrics.enabled }}
+{{ $shards := list "" }}
+{{ if .Values.shards }}
+{{ $shards = concat $shards .Values.shards | uniq }}
+{{ end }}
+{{ range $shards }}
+{{- if $.Values.metrics.enabled }}
```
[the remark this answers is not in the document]
p-se: Yes, absolutely right. I fixed it. Thanks!

e2e/assets/metrics/fleetcontroller_service.yaml
```diff
@@ -2,10 +2,18 @@ apiVersion: v1
...
+    app: fleet-controller
+    env: test
 spec:
   selector:
     app: fleet-controller
-    shard-default: "true"
+    {{- if .IsDefaultShard }}
+    shard-default: "{{ .IsDefaultShard }}"
+    {{- end }}
+    {{- if ne .Shard "" }}
```
weyfonk, line 14: :thinking: isn't this equivalent to `{{- if not .IsDefaultShard }}`, so to an `else` block in this context?
Do we use `IsDefaultShard` to keep a non-empty default shard as a possibility?

.github/workflows/nightly-ci.yml
```diff
@@ -111,6 +111,7 @@ jobs:
 
           # 2. Run tests for metrics
           ginkgo e2e/metrics
+          SHARD=shard1 ginkgo e2e/metrics
```
weyfonk, line 114: I find it interesting that we explicitly run tests twice here and in `e2e-ci.yml`, separating between sharded and unsharded contexts, which made me wonder: would we want to eventually expose the relevant shard (ID or empty value) in labels for metrics? How much sense could it make to users?

e2e/metrics/cluster_test.go
```diff
@@ -87,6 +88,9 @@ func expectMetrics(name string, expectedMetrics MetricsSelector, expectExist boo
 
 var _ = Describe("Cluster Metrics", Label("cluster"), func() {
 	It("should have metrics for the local cluster resource", func() {
+		if os.Getenv("SHARD") != "" {
```
weyfonk, line 91: I am not following this... Why not check the `shard` variable directly? Why would this not be set for cluster tests if it is for the whole test suite?

e2e/assets/metrics/fleetcontroller_service.yaml
p-se: No, a non-empty default shard is no possibility. This should be simplified.

.github/workflows/nightly-ci.yml
weyfonk: Ah nice, this is helpful already, thanks!

Issue #2415, opened 2024-05-07
Worker count for fleet-controller should be configurable


manno:
The controller runtime reconcilers can be configured with `MaxConcurrentReconciles`. We should expose this setting as a a separate env var for each reconciler. Similar to https://fleet.rancher.io/ref-configuration#environment-variables we can add the values to the helm values file.

PR #2430, opened 2024-05-14
Add a Parameter to Configure Worker count for fleet-controller

Tommy12789:
Refers to https://github.com/rancher/fleet/issues/2415

Added a new environment var for each type of reconciler (gitrepo, bundle, bundledeployment) to configure their worker count (in values.yaml)

No test for this, we need a large scale fleet to test to test it

### DIFF
--- a/charts/fleet/templates/deployment.yaml
+++ b/charts/fleet/templates/deployment.yaml
@@ -65,15 +65,15 @@ spec:
           value: "true"
         {{- end }}
         {{- if $.Values.controller.reconciler.workers.gitrepo }}
-        - name: CONTROLLER_GITREPO_RECONCILER_WORKERS
+        - name: GITREPO_RECONCILER_WORKERS
           value: {{ quote $.Values.controller.reconciler.workers.gitrepo }}
         {{- end }}
         {{- if $.Values.controller.reconciler.workers.bundle }}
-        - name: CONTROLLER_BUNDLE_RECONCILER_WORKERS
+        - name: BUNDLE_RECONCILER_WORKERS
           value: {{ quote $.Values.controller.reconciler.workers.bundle }}
         {{- end }}
         {{- if $.Values.controller.reconciler.workers.bundledeployment }}
-        - name: CONTROLLER_BUNDLEDEPLOYMENT_RECONCILER_WORKERS
+        - name: BUNDLEDEPLOYMENT_RECONCILER_WORKERS
           value: {{ quote $.Values.controller.reconciler.workers.bundledeployment }}
         {{- end }}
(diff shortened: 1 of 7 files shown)


charts/fleet/templates/deployment.yaml
```diff
@@ -64,6 +64,18 @@ spec:
         - name: CATTLE_DEV_MODE
           value: "true"
         {{- end }}
+        {{- if $.Values.controller.reconciler.workers.gitrepo }}
+        - name: CONTROLLER_GITREPO_RECONCILER_WORKERS
```
manno, line 68:
```suggestion
        - name: GITREPO_RECONCILER_WORKERS
```

I'd remove the controller prefix on all envs, for brevity.

charts/fleet/values.yaml
```diff
@@ -90,3 +90,12 @@ leaderElection:
...
   renewDeadline: 25s
+
+## Fleet controller configuration
+controller:
+  reconciler:
+    # The number of workers that are allowed to each type of reconciler
+    workers:
+      gitrepo: "1"
+      bundle: "2"
+      bundledeployment: "3"
```
manno, line 101:
```suggestion
      bundledeployment: "1"
```

I'd default them all to 1, I think that's the upstream default, too.

Tommy12789: Of course, I forgot to reset them !

charts/fleet/templates/deployment.yaml
Tommy12789: Just changed this TY Mario !

internal/cmd/controller/root.go
```diff
@@ -54,6 +54,12 @@ type LeaderElectionOptions struct {
 	RetryPeriod *time.Duration
 }
 
+type ControllerReconcilerWorkers struct {
+	Gitrepo          int
```
manno, line 58: I just noticed it's GitRepo everywhere else

```suggestion
	GitRepo          int
```

internal/cmd/controller/root.go
```diff
@@ -54,6 +54,12 @@ type LeaderElectionOptions struct {
...
+	Bundle           int
+	Bundledeployment int
```
manno, line 60:
```suggestion
	BundleDeployment int
```

manno (changes requested):

internal/cmd/controller/root.go
Tommy12789: changed !

internal/cmd/controller/root.go
Tommy12789: changed !

manno (approved):

Issue #1980, opened 2023-11-30
Fleet controller uses Finalizer to Orphan BundleDeployments


manno:
For the controller-runtime controllers, replace the cleanup controller with finalizers.

* removal of bd by fleetcontroller would be blocked until agent does clean up and removes finalizer
* needs webhooks to add finalizer to user created resources. Maybe we can avoid webhooks, when cleanup uses automatically generated resources only?

weyfonk:
[WIP branch](https://github.com/weyfonk/fleet/tree/1980-cleanup-e2e-wip) with an e2e test case for bundle cleanup, in case it might be helpful.

PR #2431, opened 2024-05-14
Partially replace cleanup with finalizers

weyfonk:
This implements [finalizers](https://book.kubebuilder.io/reference/using-finalizers) on GitRepos, bundles and bundle deployments, removing the corresponding cleanup controller logic for those resources. That cleanup is no longer necessary, as those resources can no longer be orphaned, ie deleted when no Fleet controller is running.

This is tested by end-to-end tests, by temporarily scaling down the (unsharded) `fleet-controller` deployment and checking that deleting one of those resources has no effect other than setting their deletion timestamps.

In general, adding and removing finalizers on a resource is handled by that resource's reconciler, with the following exception: the `GitRepo` reconciler takes care of removing bundle deployment finalizers, when purging bundle deployments. This prevents the agent from dealing with finalizers, which would require a change in permissions and thereby a possible security risk.

### Out of scope

This does not add finalizers to RBAC resources created for GitRepos (eg. service accounts, roles, role bindings, config maps), as those are native k8s resources which can be created in more use cases than GitRepo lifecycle management (eg. agent management), therefore  cleanup logic is still needed there.

Content resources do not bear finalizers either, because cleaning them up responds to a different process according to reference counts, ie how many bundles refer to a given Content resource.

Refers to #1980


0xavi0 (commented):
Just a couple of questions. And, aside from preventing the user to delete `bundledeployments` I don't see why we need to set a finalizer in those.

In the version without finalisers the user is able to delete a `bundledeployment` which is is almos instantly recreated by Fleet. 

With this the user can't delete `bundledeployments` as the DeleteTimestamp is set but the finaliser is never removed. 
I don't know if keeping the `bundledeployment` with the delete timestamp set and alive is the right thing to do. 

There is no problem deleting `bundles` or `gitrepos` (I've tested also those cases with this branch)

e2e/single-cluster/finalizers_test.go
```diff
@@ -0,0 +1,296 @@
...
+		_, _ = k.Delete("bundle", fmt.Sprintf("%s-%s", gitrepoName, path))
+		bdNamespace, err := k.Get(
+			"ns",
+			"-o=jsonpath='{.items[?(@."+
+				`metadata.annotations.fleet\.cattle\.io/cluster-namespace=="fleet-local"`+
+				")].metadata.name}'",
+		)
+		Expect(err).ToNot(HaveOccurred())
+
+		_, _ = env.Kubectl.Namespace(bdNamespace).Delete(
```
0xavi0, line 56: Why do we need to "manually" delete the `bundledeployment`?
It should be gone after deleting the `bundle`, right?

In fact this should be part of the test... checking that when deleting a `bundle` or `gitrepo` the `bundledeployments` related are also gone

e2e/single-cluster/finalizers_test.go
```diff
@@ -0,0 +1,296 @@
...
+				"--replicas=1",
+				"--timeout=5s",
+			)
+			Expect(err).ToNot(HaveOccurred())
+
+			_, err = k.Delete("gitrepo", gitrepoName)
+			Expect(err).NotTo(HaveOccurred())
+
+			// These resources should be deleted when the GitRepo is deleted.
+			By("checking that the auxiliary resources don't exist anymore")
```
0xavi0, line 154: We could also check that the `bundle`s and `bundledeployment`s are deleted

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -150,6 +189,11 @@ func (r *BundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		// NOTE we don't use the existing BundleDeployment, we discard annotations, status, etc
 		// copy labels from Bundle as they might have changed
 		bd := target.BundleDeployment()
+
+		if bd.DeletionTimestamp.IsZero() {
+			controllerutil.AddFinalizer(bd, bundleDeploymentFinalizer)
+		}
```
0xavi0, line 195: In what case is this going to be different to zero?
In my understanding the `BundleDeployment` struct was just instantiated and it's still not added to the cluster.

I guess that we could omit the `if` statement if that's correct

0xavi0 (commented):
Just a couple of questions. And, aside from preventing the user to delete `bundledeployments` I don't see why we need to set a finalizer in those.

In the version without finalisers the user is able to delete a `bundledeployment` which is is almos instantly recreated by Fleet. 

With this the user can't delete `bundledeployments` as the DeleteTimestamp is set but the finaliser is never removed. 
I don't know if keeping the `bundledeployment` with the delete timestamp set and alive is the right thing to do. 

There is no problem deleting `bundles` or `gitrepos` (I've tested also those cases with this branch)

e2e/single-cluster/finalizers_test.go
weyfonk: This does not depend on finalizers though, but rather on relationships between a `GitRepo`, bundles and bundle deployments. We cover some of it [here](https://github.com/rancher/fleet/blob/main/e2e/single-cluster/gitrepo_test.go#L78).

internal/cmd/controller/reconciler/bundle_controller.go
weyfonk: You are right, that condition was here as a precaution; I have replaced it with a comment explaining precisely that.

e2e/single-cluster/finalizers_test.go
weyfonk: I have removed this, as the bundle deployment will indeed have been deleted by each test case, either manually or through previous cleanup steps, ie scaling the Fleet controller back up, which would trigger `GitRepo` and bundle deletion.

0xavi0 (approved):

0xavi0 (approved):

Issue #2343, opened 2024-04-19
Gitjob complains that `Operation cannot be fulfilled on gitrepos.fleet.cattle.io`


thehejik:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

In `rancher:v2.9-head-c9be13b` with preinstalled `fleet:104.0.0+up0.10.0-rc.11` there are new suspicious info log messages from  `gitjob` pod that were not present in previous versions. Not sure if it has some impact on functionality yet.

I was testing with private azure https gitrepo but it is appearing also for public github gitrepo.

```json
$ kubectl logs -n cattle-fleet-system -lapp=gitjob -f
{
  "level": "info",
  "ts": "2024-04-19T15:51:17Z",
  "logger": "gitjob-reconciler",
  "msg": "conflict updating status",
  "message": "Operation cannot be fulfilled on gitrepos.fleet.cattle.io \"gh-https-2\": the object has been modified; please apply your changes to the latest version and try again"
}
```

#### Reproducer:
Install rancher v2.9-head and add a gitrepo to fleet-local namespace, then observe the gitjob logs.

thehejik:
During daily we discussed it is maybe related to another loglevel set or due to "zapr" logger.

bigkevmcd:
The code around this https://github.com/rancher/fleet/blob/b996ca8280080b0dae263472bbcb4a5cb9349ae7/internal/cmd/controller/gitops/reconciler/gitjob_controller.go#L169 could maybe be updated to `.Patch()` rather than `.Update()` and probably reload the status from the kube-apiserver before calculating the patch to apply.

Tommy12789:
## Additional QA
### Issue: <link the issue or issues this PR resolves here>
2343
 
### Problem
When applying a GitRepo, gitjob complaining about  "Operation cannot be fulfilled on gitrepos.fleet.cattle.io" because gitrepo controller conflicting with gitjob controller.

### Solution
Added the use of retryonconflict method within the updateStatusMethod of the gitjob reconciler to retry when a conflict happens
 
### Testing
Repro steps are valid 

### Engineering Testing
#### Manual Testing
Repro steps from the issue 

#### Automated Testing
N/A

### QA Testing Considerations
- Apply multiple gitrepo in multiple namespace to see if any error appears 
 
#### Regressions Considerations
N/A

PR #2434, opened 2024-05-16
Refactoring CreateJob and UpdateStatus methos in Gitjob Reonciler

Tommy12789:
Refers to #2343 

Changed the Update method into the updateStatus method of the gitjob reconciler into the controller to the Patch method.

./internal/cmd/controller/gitops/reconciler/gitjob_test.go


internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -166,19 +166,28 @@ func (r *GitJobReconciler) createJob(ctx context.Context, gitRepo *v1alpha1.GitR
 		}
 		gitRepoFromCluster.Status.ObservedGeneration = gitRepoFromCluster.Generation
 
-		return r.Status().Update(ctx, &gitRepoFromCluster)
```
aruiz14, line 169: I'm not convinced of this change. `Update` is immediately happening after `Get`, so it's weird that it finds a conflict (meaning another client wrote to the same object in the meantime).

In fact, I'm not sure this even correct, and there could be a race condition:

1. `createJob` receives a GitRepo, and does something based on that information
2. At the end, it updates the ObservedGeneration field in the status. However, the value being used is the one that was just read from the API, which may differ from the one used as an input.

The `Generation` field is used to represent a revision of the object `Spec` has changed. My understanding is that it `ObservedGeneration` should be set to the last revision being processed, so that if `ObservedGeneration != Generation`, it means that the a reconciliation is needed.

manno: Ah, good find. 
So, I think this is a left-over from when we had a separate gitjob resource.
When the `createJob` func returns, there is another function which updates the same status.
I think we can remove this.

That being said: 
* retryOnConflict only makes sense, when we use the outer value to update the freshly retrieved resource. It should be `gitRepoFromCluster.Status.ObservedGeneration = gitRepo.Generation`
* `gitRepoFromCluster` is a bit verbose, `t` is fine, the variable is not that important


Regarding this whole file:
* we might want to rename this reconciler to `GitOpsReconciler` and fix the comment which mentions a CronJobReconciler
* `newJob`, `computeJobSpec` and `generateInitContainer` do a good job at hiding what they actually create: containers for "fleet gitclone" and "fleet apply".
* actually this has a lot of overlap with our gitrepo reconciler in the other container, both update the same status. However both trigger on different events.

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -166,19 +166,28 @@ func (r *GitJobReconciler) createJob(ctx context.Context, gitRepo *v1alpha1.GitR
...
-	})
+		// Create a strategic patch to update only modified fields
+		patch := client.MergeFrom(&gitRepoFromCluster)
+		if err := r.Status().Patch(ctx, &gitRepoFromCluster, patch); err != nil {
+			return err
+		}
 
+		return nil
+	})
```
manno, line 176: return err
```suggestion
	return err
```

manno:
After our discussion, I'd suggest the following changes:

- [ ] `createJob` can just return the error from `err = r.Create(ctx, job)`, updating the observed generation in the status is already handled by the other controller.
- [ ] updateStatus in line 224 should use `retry.RetryOnConflict`. This is similar to https://github.com/rancher/fleet/blob/main/internal/cmd/controller/reconciler/gitrepo_controller.go#L216-L237. However it shouldn't need to update the `observedGeneration`. It should only update the fields it just calculated, like the conditions, not the whole status.

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -149,47 +149,30 @@ func (r *GitJobReconciler) createJob(ctx context.Context, gitRepo *v1alpha1.GitR
...
 	}
 
-	gitRepo.Status.GitJobStatus = result.Status.String()
-	for _, con := range result.Conditions {
-		condition.Cond(con.Type.String()).SetStatus(gitRepo, string(con.Status))
-		condition.Cond(con.Type.String()).SetMessageIfBlank(gitRepo, con.Message)
-		condition.Cond(con.Type.String()).Reason(gitRepo, con.Reason)
-	}
+	// Prepare the computed values for updating
+	gitJobStatus := result.Status.String() // General job status
```
manno, line 170: I don't think we need this variable anymore 🤷 
And the next line too

Tommy12789: Of course, I removed them

manno (approved):

Issue #2435, opened 2024-05-16
Race condition between gitops and gitrepo controller


manno:
The [gitops gitjob controller](https://github.com/rancher/fleet/blob/main/internal/cmd/controller/gitops/reconciler/gitjob_controller.go#L174) needs RBAC from the [gitrepo controller](https://github.com/rancher/fleet/blob/main/internal/cmd/controller/reconciler/gitrepo_controller.go#L185-L187).

Move from gitrepo to gitops controller:
- [ ] `AuthorizeAndAssignDefaults`
- [ ] [rbac creation](https://github.com/rancher/fleet/blob/main/internal/cmd/controller/reconciler/gitrepo_controller.go#L167-L191) 
- [ ] helm values check
- [ ] `NewTargetsConfigMap`

Basically, only the display/summary status updates remain in the gitrepo controller. 
* SetStatusFromResourceKey
* UpdateDisplayState
* SetStatusFromBundles
* SetStatusFromBundleDeployments
* gitrepo.Status.Display.ReadyBundleDeployments = 

Can we improve the condition updates, so both controllers don't conflict on the same condition?

manno:
Possible duplicate of https://github.com/rancher/fleet/issues/2343

0xavi0:
There is no longer race conditions between those 2 controllers, because https://github.com/rancher/fleet/pull/2537 merges the whole functionality into the `gitops` controller.
The `GitRepo` controller does not exist anymore.

PR #2451, opened 2024-05-22
Gitjob controller updates GitRepo with generation it is reconciling

manno:
The gitjob reconciler updates the gitrepo resource with information it was acting on.
* ObservedGeneration, this is also updated in the gitrepo reconciler. However the gitjob reconciler updates the status and should note the generation it worked on.
* Job termination message, if the gitrepo reconciler is conflicting with the gitjob reconciler while trying to update the status, the gitjob reconciler will store the message as it was when it triggered.

Also fixes several tests that were not waiting for setup to be complete, or trying to update outdated resources.

Refers to #2435

### DIFF
--- a/e2e/metrics/bundledeployment_test.go
+++ b/e2e/metrics/bundledeployment_test.go
@@ -3,7 +3,6 @@ package metrics_test
 import (
 	"fmt"
 	"math/rand"
-	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
@@ -24,25 +23,30 @@ var _ = Describe("BundleDeployment Metrics", Label("bundledeployment"), func() {
 		// objName is going to be "randomized" instead of using a dedicated and
 		// random namespace, like it is the case for the other tests.
 		objName string
+
+		r = rand.New(rand.NewSource(GinkgoRandomSeed()))
 	)
 
 	BeforeEach(func() {
 		k = env.Kubectl.Namespace(env.Namespace)
 		kw = k.Namespace(namespace)
-		objName = testenv.AddRandomSuffix(
-			"metrics",
-			rand.New(rand.NewSource(time.Now().UnixNano())),
-		)
+		objName = testenv.AddRandomSuffix("metrics", r)
+		targetNS := testenv.NewNamespaceName("metrics", r)
 
 		err := testenv.CreateGitRepo(
 			kw,
-			namespace,
+			targetNS,
 			objName,
 			branch,
 			shard,
 			"simple-manifest",
 		)
 		Expect(err).ToNot(HaveOccurred())
+
+		Eventually(func() string {
+			out, _ := kw.Get("gitrepo", objName, "-o", "jsonpath={.status.readyClusters}")
+			return out
+		}).Should(ContainSubstring("1"))
 
 		DeferCleanup(func() {
 			out, err := k.Delete("gitrepo", objName)
--- a/internal/cmd/controller/gitops/reconciler/gitjob_controller.go
+++ b/internal/cmd/controller/gitops/reconciler/gitjob_controller.go
@@ -164,6 +164,29 @@ func (r *GitJobReconciler) updateStatus(ctx context.Context, gitRepo *v1alpha1.G
 		return err
 	}
 
+	terminationMessage := ""
+	if result.Status == status.FailedStatus {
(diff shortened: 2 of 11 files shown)


e2e/metrics/bundledeployment_test.go
```diff
@@ -24,29 +23,38 @@ var _ = Describe("BundleDeployment Metrics", Label("bundledeployment"), func() {
 		// objName is going to be "randomized" instead of using a dedicated and
 		// random namespace, like it is the case for the other tests.
 		objName string
+
+		r = rand.New(rand.NewSource(GinkgoRandomSeed()))
 	)
 
 	BeforeEach(func() {
 		k = env.Kubectl.Namespace(env.Namespace)
 		kw = k.Namespace(namespace)
```
p-se, line 32: I don't understand how this code could have passed the tests. `kw` does not have `targetNS` as namespace.

e2e/metrics/bundledeployment_test.go
```diff
@@ -24,26 +23,31 @@ var _ = Describe("BundleDeployment Metrics", Label("bundledeployment"), func() {
 		// objName is going to be "randomized" instead of using a dedicated and
 		// random namespace, like it is the case for the other tests.
 		objName string
+
+		r = rand.New(rand.NewSource(GinkgoRandomSeed()))
```
p-se, line 27: GinkgoRandomSeed() is unique, even among parallelized tests, which means when running those tests in parallel, they will get the same seed, generate the same randomization source and create the same random names, thereby breaking parallelism of tests.

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -172,37 +195,18 @@ func (r *GitJobReconciler) updateStatus(ctx context.Context, gitRepo *v1alpha1.G
...
 		currentGitRepo.Status.GitJobStatus = result.Status.String()
+		currentGitRepo.Status.ObservedGeneration = gitRepo.Generation
 
 		for _, con := range result.Conditions {
 			condition.Cond(con.Type.String()).SetStatus(currentGitRepo, string(con.Status))
 			condition.Cond(con.Type.String()).SetMessageIfBlank(currentGitRepo, con.Message)
 			condition.Cond(con.Type.String()).Reason(currentGitRepo, con.Reason)
 		}
 
-		if result.Status == status.FailedStatus {
```
weyfonk, line 182: Now that we compute the termination message outside `RetryOnConflict`, that termination message may be outdated by the time the status is updated. The PR description suggests that this is intentional, but why? :thinking:

e2e/metrics/bundledeployment_test.go
manno: Do you have any source for that?
I quickly tested it and it seems to work:

```
package metrics_test

import (
	"fmt"
	"math/rand"

	. "github.com/onsi/ginkgo/v2"
)

var _ = FDescribe("Bundle Metrics", Label("bundle"), func() {
	var r = rand.New(rand.NewSource(GinkgoRandomSeed()))

	It("a", func() {
		fmt.Printf("a random: %d\n", r.Intn(100))
		fmt.Printf("a random: %d\n", r.Intn(100))
	})
	It("b", func() {
		fmt.Printf("b random: %d\n", r.Intn(100))
		fmt.Printf("b random: %d\n", r.Intn(100))
	})
	It("c", func() {
		fmt.Printf("c random: %d\n", r.Intn(100))
		fmt.Printf("c random: %d\n", r.Intn(100))
	})
})
```

However, what you cannot do, is re-create the source from the same seed in `BeforeEach`.

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
manno: Mainly for consistency. It's the job we pass into `updateStatus`. It's the one we computed the status for. I don't think the reconciler should switch "inputs" only because it hits a conflict with other controller. If we want to do that, we need to repeat all the work (status computation), not just parts (termination message) of it. 

That's alsow the reason for setting the generation to the value of the gitrepo that triggered the reconcile.
Gitrepo reconciler might get triggered for that change, too, so it will be consistent eventually.

weyfonk: I see, that makes sense to me. Thanks for clarifying :)

PR #2452, opened 2024-05-22
Remove trailing comma from bootstrap template

Danil-Grigorev:
Refers to #XXX

This causes `Error("trailing comma", line: 16, column: 13)` when using serde_json to deserialize the value and do something with apiServerURL and apiServerCA. Unfortunately this rust framework does not tolerate invalid json content.


weyfonk (approved):

Issue #2448, opened 2024-05-22
cluster.desiredReady is wrong if bundledeployment was deleted


manno:
https://github.com/rancher/fleet/blob/main/internal/cmd/controller/reconciler/cluster_controller.go#L101

manno:
## Additional QA


### Problem

Deleting a bundle deployment may end up with incorrect numbers on the Cluster summary and display status fields.

This could happen, when re-targeting a bundle, so a cluster is no longer receiving the bundle. The UI would show wrong numbers, e.g. for `desiredReady` and the state counters.

 
### Solution

We exclude the deleted bundledeployment from status calculations.
 
### Testing

### Engineering Testing
#### Manual Testing

This was found in a code review.

#### Automated Testing


No testing was done, as the status tests are currently being developed.

### QA Testing Considerations

There is a possibility this can't be tested, as another controller might trigger and the values might eventually end up right.
 
#### Regressions Considerations

-

mmartin24:
I paired with Mario and gave me some hints about how to possibly check this, unfortunately it seems I was not successfull. Basically tried to trigger a possible malfunctioning on some counts targeting cluster groups first and changing selectors later, but as commented I am not sure I saw anything strange. Please correct me if I am wrong.


### Test setup:

- Deployed Rancher `v2.9-03e149351b6877d02ffe6f32be8b896708132f08-head` with fleet `104.0.0+up0.10.0-rc.14` 
- Deployed 1 downstream cluster
- Deployed this Girepo on ds cluster `85`
```
https://github.com/rancher/fleet-test-data/
Branch: master
qa-test-apps/nginx-app
```
Content:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-keep
  labels:
    app: nginx
spec:
  replicas: 1
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:1.14.2
        ports:
        - containerPort: 80
```
- Created cluster group `cluster-group-1` with the following selector:
```yaml
spec:
  selector:
    matchExpressions: []
    matchLabels:
      env: test
```
- Bound the cluster `85` with labels `env: test` to the cluster group `cluster-group-1`

---
##### Step 1: Deploy GitRepo on downstream cluster and check bundle output. 

Short summary:
```json
   "summary": {
      "desiredReady": 1,
      "ready": 1
    },
```

<details>
[...]

PR #2454, opened 2024-05-22
Do not count deleted bundledeployments in cluster status

manno:
Refers to https://github.com/rancher/fleet/issues/2448

### DIFF
--- a/internal/cmd/controller/reconciler/cluster_controller.go
+++ b/internal/cmd/controller/reconciler/cluster_controller.go
@@ -94,15 +94,17 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	if err != nil {
 		return ctrl.Result{}, r.updateErrorStatus(ctx, req.NamespacedName, cluster.Status, err)
 	}
+
+	deleted := map[types.UID]bool{}
 	for _, bundle := range cleanup {
-		for _, bundleDeployment := range bundleDeployments.Items {
-			if bundleDeployment.Labels[fleet.BundleLabel] == bundle.Name &&
-				bundleDeployment.Labels[fleet.BundleNamespaceLabel] == bundle.Namespace {
-				logger.V(1).Info("cleaning up bundleDeployment not matching the cluster", "bundledeployment", bundleDeployment)
-				err := r.Delete(ctx, &bundleDeployment) // nolint:gosec // does not store pointer
+		for _, bd := range bundleDeployments.Items {
+			if bd.Labels[fleet.BundleLabel] == bundle.Name && bd.Labels[fleet.BundleNamespaceLabel] == bundle.Namespace {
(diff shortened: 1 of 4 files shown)


internal/cmd/controller/reconciler/cluster_controller.go
```diff
@@ -94,15 +94,20 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 	for _, bundle := range cleanup {
-		for _, bundleDeployment := range bundleDeployments.Items {
-			if bundleDeployment.Labels[fleet.BundleLabel] == bundle.Name &&
-				bundleDeployment.Labels[fleet.BundleNamespaceLabel] == bundle.Namespace {
-				logger.V(1).Info("cleaning up bundleDeployment not matching the cluster", "bundledeployment", bundleDeployment)
-				err := r.Delete(ctx, &bundleDeployment) // nolint:gosec // does not store pointer
+		for _, bd := range bundleDeployments.Items {
+			if deleted[bd.GetUID()] {
+				continue
+			}
```
0xavi0, line 103: Question: when is this happening? 
is it possible to have a `bundledeployment` that matches the condition to be deleted:
```go
bd.Labels[fleet.BundleLabel] == bundle.Name && bd.Labels[fleet.BundleNamespaceLabel] == bundle.Namespace
```
more than once?

internal/cmd/controller/reconciler/cluster_controller.go
```diff
@@ -94,15 +94,20 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+			if bd.Labels[fleet.BundleLabel] == bundle.Name && bd.Labels[fleet.BundleNamespaceLabel] == bundle.Namespace {
+				logger.V(1).Info("cleaning up bundleDeployment not matching the cluster", "bundledeployment", bd)
+				err := r.Delete(ctx, &bd) // nolint:gosec // does not store pointer
```
0xavi0, line 106: Nit: I THINK we don't need this `// nolint` statement?. (I known it's not part of the changes of the PR, but just saw it and remembered that I've seen it in other files) 
I've tested running `golangci-lint` locally without the `// nolint` and it didn't complain.

internal/cmd/controller/reconciler/cluster_controller.go
manno: Ah, I just wanted to skip the whole inner loop. But you're right, the existing `if` condition is probably not  significantly slower than `deleted[bd.GetUID()]`.

internal/cmd/controller/reconciler/cluster_controller.go
manno: Ah, nice. Last time I checked I still got warnings. I'll add a commit, but leave the ones in gitrepo controller to you, so we don't conflict.

0xavi0 (approved):
LGTM

PR #2461, opened 2024-05-27
dev: Make setup-single-cluster capable of setting up multiple clusters

p-se:
By using different git worktrees with different environment variables for configuring the single-cluster environment, specifically the `FLEET_E2E_CLUSTER` variable, you can use `./dev/setup-single-cluster` to start independent k3d clusters next to each other.

Please note that, at the time of creating a cluster, there is a possibility of a conflict still. While clusters can be run at the same time with `dev/setup-single-cluster`, they cannot be created simultaneously (yet).

- [ ] Perhaps add a entry in dev/README.

### DIFF
--- a/dev/build-fleet
+++ b/dev/build-fleet
@@ -18,6 +18,17 @@ fi
 
 export GOOS=linux
 
+# The name of the container image created here is a potential source of conflict
+# when fleet is set up simultaneously in multiple sets (i.e. 1 upstream + 0 to n
+# downstream). build-fleet always creates a container image with the same name
+# (`rancher/fleet`, rancher/fleet-agent`) and tag (`dev`). Conflicts can simply
+# be avoided by making sure that the images `rancher/fleet:dev` and
+# `rancher/fleet-agent:dev` are rebuilt from the right context (e.g. git clone,
+# checkout or worktree) before being imported into clusters created for testing
+# that context.  This can be easily achieved by ensuring that, for instance, at
+# most one instance of `./dev/setup-single-cluster` or
+# `./dev/setup-multi-cluster` runs at any given point in time.
+
 # fleet
 go build -gcflags='all=-N -l' -o "bin/fleet-linux-$GOARCH" ./cmd/fleetcli
 go build -gcflags='all=-N -l' -o bin/fleetcontroller-linux-"$GOARCH" ./cmd/fleetcontroller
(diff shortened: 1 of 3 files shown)


dev/build-fleet
```diff
@@ -25,4 +25,8 @@ docker build -f package/Dockerfile -t rancher/fleet:dev --build-arg="ARCH=$GOARC
 
 # fleet agent
 go build -gcflags='all=-N -l' -o "bin/fleetagent-linux-$GOARCH" ./cmd/fleetagent
+
+# The name of the container image created here is a potential source of
+# conflicts when the clusters are created simultaneously. It might be worth thinking about
+# how to make the image name unique for each test run.
```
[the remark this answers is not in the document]
p-se: Thank for taking the time to explain to me why this was confusing! Please let me know if the updated phrasing (and position of the comment) is not a sufficient improvement!

dev/build-fleet
```diff
@@ -18,6 +18,14 @@ fi
 
 export GOOS=linux
 
+# The name of the container image created here is a potential source of conflict
+# when fleet is set up simultaneously, as several scripts are calling
```
weyfonk, line 22: This `simultaneously` alone may leave the reader wondering what simultaneity we're talking about here.
```suggestion
# when fleet is set up simultaneously in multiple sets (ie 1 upstream +
# 0 to n downstream) of clusters, as several scripts are calling
```

dev/build-fleet
```diff
@@ -18,6 +18,14 @@ fi
...
+# build-fleet and it always creates a container image with the same name
+# (`rancher/fleet`, rancher/fleet-agent`) and tag (`dev`). Conflicts can simply
+# be avoided by making sure this script is not called sequentially, e.g. by
+# creating multiple clusters on a single host sequentially rather than
+# simultaneously.
```
weyfonk, line 27: I'm not sure what calling a script "sequentially" means.
What would you think of something like this? How does it compare to what you have/had in mind?
```suggestion
# be avoided by making sure that images `rancher/fleet:dev` and
# `rancher/fleet-agent:dev` are rebuilt from the right context (eg. git clone,
# checkout or worktree) before being imported into clusters created for testing
# that context.
# This can be easily achieved by ensuring that, for instance, at most one instance
# of `./dev/setup-single-cluster` or `./dev/setup-multi-cluster` runs at any given
# point in time.
```

weyfonk (approved):

weyfonk (approved):

PR #2467, opened 2024-05-29
Fix `fleet apply` command description

weyfonk:
That command does not render bundles into Kubernetes resources, but rather creates bundles from directories.


weyfonk:
* [ ] Create [fleet-docs](https://github.com/rancher/fleet-docs) PR reflecting these changes once approved

0xavi0:
It also applies the `bundle` if the output is not specified, right?

0xavi0:
> Indeed it does. How should we specify in the command description that this _may_ happen, without going to deep into the details? Would this work? `Create bundles from directories, and output them or apply them on a cluster`

Works for me. Maybe just omitting the first `them` in the sentence.

0xavi0 (approved):
LGTM

Issue #2432, opened 2024-05-15
Allow Force Update and initial clone for a gitrepo with `spec.disablePolling: true` set


thehejik:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

In fleet:104.0.0+up0.10.0-rc.13 on rancher:v2.9-head there is a new feature `spec.disablePolling` for gitrepo which prevents automatic polling of the gitrepo, ref. https://github.com/rancher/fleet/issues/1572

With this feature enabled it would be nice to do initial git repo checkout and provisioning of k8s resources on the target cluster. 

Also it should allow updating of resources after clicking on `Force Update` button (not sure about enabling the update also after pressing `Pause/Unpause`). 

Only way how to provision a gitrepo or update its resources with this flag enabled is to trigger a webhook event (also according to [docs](https://github.com/0xavi0/fleet-docs/blob/main/docs/ref-gitrepo.md?plain=1#L94 ))

I guess this feature should behave similarly as `spec.pollingInterval` which does initial checkout and also `Force Update` is possible.

Tommy12789:
Pull request content for Rancher QA verification.

## Additional QA
### Issue: <link the issue or issues this PR resolves here>
#2432
 
### Problem
GitRepos with "disablePolling = true"  were not initially checked out and ressources weren't applied to the cluster at their creation.
 
### Solution
Update the commit of the gitrepo when it is applied so it is checked out and its ressources are applied
 
### Testing
apply a gitrepo with the "disablePolling = true" option in its specs. Verify the ressources are deployed in the cluster.

### Engineering Testing
#### Manual Testing
N/A

#### Automated Testing
integrationtests/gitjob/controller/controller_test.go
e2e/single-cluster/gitrepo_polling_disabled_test.go

### QA Testing Considerations
- testing the initial checkout and ressources applications
- forceUpdate the gitrepo and verify all ressources are updated and commit updated in the gitrepo status
- Try webhooks 
- Try commit to the gitrepo to verify the ressources aren't updated in the cluster without webhook or forceUpdate 
 
#### Regressions Considerations
N/A

PR #2469, opened 2024-05-29
GitRepo Initial Checkout when DisablePolling is True

Tommy12789:
Refers to #2432

implemented an initial checkout for the gitrepos applied without polling and commit update while force updating 

./integrationtests/gitjob/controller/controller_test.go


internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -103,9 +104,27 @@ func (r *GitJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+	if errors.IsNotFound(err) {
+		if gitRepo.Status.Commit != "" {
+			if err := r.createJob(ctx, &gitRepo); err != nil {
+				return ctrl.Result{}, fmt.Errorf("error creating git job: %v", err)
+			}
+		} else {
+			if gitRepo.Status.GitJobStatus == status.FailedStatus.String() {
+				return ctrl.Result{}, fmt.Errorf("error creating git job: %v", err)
+			}
+
```
0xavi0, line 116: Why are we checking  for the `GitJobStatus` if the job was not found? 
Also... the log message seems wrong...

0xavi0 (commented):
As discussed offline, there is a bug when doing a "Force Update" because the contents are re-deployed but the `commit` is not updated.

That works because the completed job remains in the cluster and the 
```go
	} else if gitRepo.Status.Commit != "" {
		if err = r.deleteJobIfNeeded(ctx, &gitRepo, &job); err != nil {
			return ctrl.Result{}, fmt.Errorf("error deleting git job: %v", err)
		}
	}
```
code is executed (which recreates the job and fetches the latest from the repository).

But this part:
```go
                        fetcher := git.NewFetcher()
			commit, err := fetcher.LatestCommit(ctx, &gitRepo, r.Client)
			if err != nil {
				return ctrl.Result{}, fmt.Errorf("error fetching commit: %v", err)
			}
			gitRepo.Status.Commit = commit
			logger.V(1).Info("Updating GitRepo status", "commit", gitRepo.Status.Commit)
			if err := r.Status().Update(ctx, &gitRepo); err != nil {
				return ctrl.Result{}, fmt.Errorf("error updating git repo status: %v", err)
			}
			return ctrl.Result{}, nil
``` 
is no longer executed and the `commit` is not updated in the `gitrepo`.

It is not executed because it deletes the job and then it executes this branch:
```go
     if errors.IsNotFound(err) {
		if gitRepo.Status.Commit != "" {
			if err := r.createJob(ctx, &gitRepo); err != nil {
				return ctrl.Result{}, fmt.Errorf("error creating git job: %v", err)
			}
		}
```
as `gitRepo.Status.Commit` is no longer `""` like it was for the first time.

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
Tommy12789: Of course, I removed these lines

0xavi0 (commented):
I guess it would be cool to also test the case that had the bug before, but I guess that requires an `e2e` test because you need to create a new commit and these tests are relying on the `fleet` repo in github. 
Up to you if you want to experiment and create that.


Aside from the code simplification and the lines that belong to other PRs... LGTM

integrationtests/gitjob/controller/controller_test.go
```diff
@@ -217,6 +217,28 @@ var _ = Describe("GitJob controller", func() {
...
+		It("should update the commit from the actual repo", func() {
+			gitRepo = createGitRepoWithDisablePolling(gitRepoName)
+			Expect(k8sClient.Create(ctx, &gitRepo)).To(Succeed())
+
+			By("verifying the commit is updated")
+			Eventually(func() string {
+				Expect(k8sClient.Get(ctx, types.NamespacedName{Name: gitRepoName, Namespace: gitRepoNamespace}, &gitRepo)).To(Succeed())
+				return gitRepo.Status.Commit
+			}, "30s", "1s").ShouldNot(Equal(""))
+		})
```
0xavi0, line 240: I think this test could also check that the job is eventually created, like other tests in this same file.

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -103,9 +104,29 @@ func (r *GitJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+			}
+		} else {
+			if err := r.updateCommit(ctx, &gitRepo); err != nil {
+				if errors.IsConflict(err) {
+					logger.V(1).Info("conflict updating commit, retrying", "message", err)
+					return ctrl.Result{Requeue: true}, nil // just retry, but don't show an error
+				}
+				return ctrl.Result{}, fmt.Errorf("error updating commit: %v", err)
+			}
+			return ctrl.Result{}, nil
```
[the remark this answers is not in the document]
Tommy12789: Of course I"ll change that

integrationtests/gitjob/controller/controller_test.go
Tommy12789: I"ll do it !

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -172,6 +205,8 @@ func (r *GitJobReconciler) updateStatus(ctx context.Context, gitRepo *v1alpha1.G
 		}
 
 		currentGitRepo.Status.GitJobStatus = result.Status.String()
+		currentGitRepo.Status.ObservedGeneration = gitRepo.Generation
+		currentGitRepo.Status.UpdateGeneration = gitRepo.Status.UpdateGeneration
```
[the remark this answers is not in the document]
Tommy12789: Yes I had some conflicts so I added

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -103,9 +104,29 @@ func (r *GitJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		return ctrl.Result{}, fmt.Errorf("error retrieving git job: %v", err)
 	}
 
-	if errors.IsNotFound(err) && gitRepo.Status.Commit != "" {
-		if err := r.createJob(ctx, &gitRepo); err != nil {
-			return ctrl.Result{}, fmt.Errorf("error creating git job: %v", err)
+	if errors.IsNotFound(err) {
+		if gitRepo.Status.Commit != "" {
```
manno, line 108: Is there an easier way to express this logic which uses not as many nested ifs?

Maybe 

```
if commit == "" || disablePolling { updateCommit }
if commit != "" { createJob }
```

pkg/git/fetch.go
```diff
@@ -18,6 +18,10 @@ const (
 
 type Fetch struct{}
 
+func NewFetcher() *Fetch {
```
manno, line 21: I would have expected the name of the struct.

```suggestion
func NewFetch() *Fetch {
```

e2e/single-cluster/gitrepo_polling_disabled_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+		_, err := k.Delete("gitrepo", gitrepoName)
+		Expect(err).ToNot(HaveOccurred())
+
+		Eventually(func() string {
+			out, _ := k.Get("bundledeployments", "-A")
+			return out
+		}).ShouldNot(ContainSubstring(gitrepoName))
+
+		out, err := k.Namespace("cattle-fleet-system").Logs(
```
0xavi0, line 63: Nit: This could be flagged as a failure if any previous test failed.

0xavi0 (approved):
LGTM
Thanks for adding the extra `e2e` test

0xavi0 (approved):

PR #2475, opened 2024-05-31
Moves functionality from gitrepo to gitjob controller

0xavi0:
Moves the following from `gitrepo` controller to `gitjob` controller:
* AuthorizeAndAssignDefaults
* RBAC resources creation
* Helm secrets check
* NewTargetsConfigMap

The `gitrepo` controller still purges `bundles` and `bundledeployments` on `gitrepo` deletion and handles the status coming from `bundles` and `bundledeployments`

Some functionality that was shared between both controllers has been moved to a common package to avoid code repetition.

Refers to: https://github.com/rancher/fleet/issues/2435

### DIFF
--- a/charts/fleet/templates/rbac_gitjob.yaml
+++ b/charts/fleet/templates/rbac_gitjob.yaml
@@ -43,6 +43,26 @@ rules:
       - 'events'
     verbs:
       - '*'
+  - apiGroups:
+      - ""
+    resources:
+      - serviceaccounts
+    verbs:
+      - "create"
+  - apiGroups:
+      - rbac.authorization.k8s.io
+    resources:
+      - roles
+    verbs:
+      - escalate
+      - create
+      - bind
+  - apiGroups:
+      - rbac.authorization.k8s.io
+    resources:
+      - rolebindings
+    verbs:
+      - create
 
 ---
 apiVersion: rbac.authorization.k8s.io/v1
--- a/integrationtests/gitjob/controller/controller_test.go
+++ b/integrationtests/gitjob/controller/controller_test.go
@@ -1,13 +1,19 @@
 package controller
 
 import (
+	"strings"
+	"time"
+
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	v1alpha1 "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
 	"github.com/rancher/wrangler/v2/pkg/name"
 
 	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -43,6 +49,24 @@ var _ = Describe("GitJob controller", func() {
 
 			Expect(job.Spec.Template.Spec.Containers).To(HaveLen(1))
 			Expect(job.Spec.Template.Spec.Containers[0].Args).To(ContainElements("fleet", "apply"))
+
+			// it should create RBAC resources for that gitRepo
+			Eventually(func() bool {
+				saName := name.SafeConcatName("git", gitRepo.Name)
+				ns := types.NamespacedName{Name: saName, Namespace: gitRepo.Namespace}
+
+				if err := k8sClient.Get(ctx, ns, &corev1.ServiceAccount{}); err != nil {
+					return false
+				}
+				if err := k8sClient.Get(ctx, ns, &rbacv1.Role{}); err != nil {
+					return false
+				}
+				if err := k8sClient.Get(ctx, ns, &rbacv1.RoleBinding{}); err != nil {
+					return false
+				}
+
+				return true
+			}).Should(BeTrue())
 		})
 
 		When("a job completes successfully", func() {
@@ -54,7 +78,8 @@ var _ = Describe("GitJob controller", func() {
 				// simulate job was successful
 				Eventually(func() error {
 					err := k8sClient.Get(ctx, types.NamespacedName{Name: jobName, Namespace: gitRepoNamespace}, &job)
-					Expect(err).ToNot(HaveOccurred())
+					// We could be checking this when the job is still not created
+					Expect(client.IgnoreNotFound(err)).ToNot(HaveOccurred())
 					job.Status.Succeeded = 1
 					job.Status.Conditions = []batchv1.JobCondition{
 						{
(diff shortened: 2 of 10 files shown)


charts/fleet/templates/rbac_gitjob.yaml
```diff
@@ -43,6 +43,21 @@ rules:
...
+      - ""
+    resources:
+      - serviceaccounts
+    verbs:
+      - "create"
+  - apiGroups:
+      - rbac.authorization.k8s.io
+    resources:
+      - clusterroles
+      - clusterrolebindings
```
aruiz14, line 56: Are these really being used?

0xavi0: you're right... those 2 are not needed... ( :star: the wonders of copy/paste :star2: ). Thanks for catching it!

charts/fleet/templates/rbac_gitjob.yaml
```diff
@@ -43,6 +43,19 @@ rules:
...
+      - serviceaccounts
+    verbs:
+      - "create"
+  - apiGroups:
+      - rbac.authorization.k8s.io
+    resources:
+      - roles
+      - rolebindings
+    verbs:
+      - "*"
```
weyfonk, line 58: Wouldn't this be enough?
```suggestion
      - "create"
```

integrationtests/gitjob/controller/controller_test.go
```diff
@@ -226,6 +251,112 @@ var _ = Describe("GitJob controller", func() {
 			}).Should(BeTrue())
 		})
 	})
+
+	When("creating a gitRepo that references a non existing helm secrets", func() {
```
weyfonk, line 255: nit:
```suggestion
	When("creating a gitRepo that references a nonexistent helm secret", func() {
```

integrationtests/gitjob/controller/controller_test.go
```diff
@@ -226,6 +251,112 @@ var _ = Describe("GitJob controller", func() {
...
+
+		Context("helmSecretNameForPaths secret does not exist", func() {
+			BeforeEach(func() {
+				helmSecretNameForPaths = "secret-does-not-exist"
+				helmSecretName = ""
+			})
+			It("logs an error about HelmSecretNameForPaths not being found", func() {
+				Eventually(func() bool {
+					strLogs := logsBuffer.String()
+					return strings.Contains(strLogs, "failed to look up HelmSecretNameForPaths, error: Secret \\\"secret-does-not-exist\\\" not found")
```
weyfonk, line 287: nit: this would probably also work (same remark about L327):
```suggestion
					return strings.Contains(strLogs, `failed to look up HelmSecretNameForPaths, error: Secret \"secret-does-not-exist\" not found`)
```

integrationtests/gitjob/controller/controller_test.go
```diff
@@ -226,6 +251,112 @@ var _ = Describe("GitJob controller", func() {
...
+				Consistently(func() bool {
+					saName := name.SafeConcatName("git", gitRepo.Name)
+					ns := types.NamespacedName{Name: saName, Namespace: gitRepo.Namespace}
+
+					if err := k8sClient.Get(ctx, ns, &corev1.ServiceAccount{}); !errors.IsNotFound(err) {
+						return false
+					}
+					if err := k8sClient.Get(ctx, ns, &rbacv1.Role{}); !errors.IsNotFound(err) {
+						return false
+
```
weyfonk, line 301: nit:
```suggestion
```

integrationtests/gitjob/controller/suite_test.go
```diff
@@ -59,14 +62,18 @@ var _ = BeforeSuite(func() {
...
+	mgr, err := ctrlruntime.NewManager(cfg, ctrlruntime.Options{
 		Scheme: scheme.Scheme,
 	})
 	Expect(err).ToNot(HaveOccurred())
 
 	ctlr := gomock.NewController(GinkgoT())
 	gitPollerMock := mocks.NewMockGitPoller(ctlr)
 
+	// redirect logs to a buffer that we can read in the tests
+	GinkgoWriter.TeeTo(&logsBuffer)
```
weyfonk, line 74: Nice one :rocket:

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -141,6 +152,67 @@ func generationOrCommitChangedPredicate() predicate.Predicate {
...
+	_, err = controllerutil.CreateOrUpdate(ctx, r.Client, configMap, func() error {
+		configMap.BinaryData = data
+		return nil
+	})
+
+	return err
+}
+
+func (r *GitJobReconciler) validateExternalSecretExist(ctx context.Context, gitrepo *v1alpha1.GitRepo) error {
+	// Validate external secrets exist
```
weyfonk, line 203: Is this comment still necessary, given the name of the method? ;)

weyfonk (commented):
This almost LGTM, pending clarification on needed verbs for permissions on roles and role bindings. The rest is mostly nitpicks.
Thanks for this :)

charts/fleet/templates/rbac_gitjob.yaml
0xavi0: I tried that, but it was not enough. 
I even tried adding verbs one by one. 
I get the following error, I guess because later it tries to grand RBAC permissions that are different to "create".

```bash
Reconciler error	{"controller": "gitrepo", "controllerGroup": "fleet.cattle.io", "controllerKind": "GitRepo", "GitRepo": {"name":"sample","namespace":"fleet-local"}, "namespace": "fleet-local", "name": "sample", "reconcileID": "2ad237ee-5d4a-49c9-9cb9-b38b1273ceb4", "error": "roles.rbac.authorization.k8s.io \"git-sample\" is forbidden: user \"system:serviceaccount:cattle-fleet-system:gitjob\" (groups=[\"system:serviceaccounts\" \"system:serviceaccounts:cattle-fleet-system\" \"system:authenticated\"]) is attempting to grant RBAC permissions not currently held:\n{APIGroups:[\"fleet.cattle.io\"], Resources:[\"bundles\"], Verbs:[\"get\" \"create\" \"update\" \"list\" \"delete\"]}\n{APIGroups:[\"fleet.cattle.io\"], Resources:[\"imagescans\"], Verbs:[\"get\" \"create\" \"update\" \"list\" \"delete\"]}"}
``` 

I even tried with 
* create
* delete
* deletecollection
* get
* list
* patch
* update
* watch 

and I got the same error.

This is still a mistery to me.
I ended up adding `*` because the original RBAC had exactly the same, more than happy to hold this PR to discuss this and understand why.

integrationtests/gitjob/controller/controller_test.go
0xavi0: changed as suggested

integrationtests/gitjob/controller/controller_test.go
0xavi0: changed as suggested

integrationtests/gitjob/controller/controller_test.go
0xavi0: line deleted

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
0xavi0: :bowtie: good point. comment deleted

integrationtests/gitjob/controller/controller_test.go
```diff
@@ -226,6 +251,110 @@ var _ = Describe("GitJob controller", func() {
...
+					return errors.IsNotFound(err)
+				}, time.Second*5, time.Second*1).Should(BeTrue())
+			})
+		})
+		Context("helmSecretName secret does not exist", func() {
+			BeforeEach(func() {
+				helmSecretNameForPaths = ""
+				helmSecretName = "secret-does-not-exist"
+			})
+			It("logs an error about HelmSecretNameForPaths not being found", func() {
```
weyfonk, line 323: Shouldn't this be:
```suggestion
			It("logs an error about HelmSecretName not being found", func() {
```
?

integrationtests/gitjob/controller/controller_test.go
```diff
@@ -226,6 +251,110 @@ var _ = Describe("GitJob controller", func() {
...
+				Eventually(func() bool {
+					strLogs := logsBuffer.String()
+					return strings.Contains(strLogs, `failed to look up HelmSecretNameForPaths, error: Secret \"secret-does-not-exist\" not found`)
```
weyfonk, line 326:
```suggestion
					return strings.Contains(strLogs, `failed to look up HelmSecretName, error: Secret \"secret-does-not-exist\" not found`)
```

charts/fleet/templates/rbac_gitjob.yaml
weyfonk: Maybe something like this then?
```
    verbs:
      -  'get'
      - 'create'
      - 'update'
      - 'list'
      - 'delete'
```

0xavi0: I tried that combination as well and I got the same error. :confused:

integrationtests/gitjob/controller/controller_test.go
0xavi0: Thanks for catching this copy/paste evil thing!

The test was passing because that was the previous value checked and was still in the logs buffer. 
I fixed the name and error message check AND I'm cleaning the `logsBuffer` in the `AfterEach` statement for those tests so the logs are clean and we don't get entries from previous tests.

charts/fleet/templates/rbac_gitjob.yaml
0xavi0: As discussed offline, I had to add `escalate` and `bind` to the `roles`.
It ended up like this:
```yaml
  - apiGroups:
      - ""
    resources:
      - serviceaccounts
    verbs:
      - "create"
  - apiGroups:
      - rbac.authorization.k8s.io
    resources:
      - roles
    verbs:
      - escalate
      - create
      - bind
  - apiGroups:
      - rbac.authorization.k8s.io
    resources:
      - rolebindings
    verbs:
      - create
```

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -104,6 +104,17 @@ func (r *GitJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	}
 
 	if errors.IsNotFound(err) && gitRepo.Status.Commit != "" {
+		if err := r.validateExternalSecretExist(ctx, &gitRepo); err != nil {
+			nsname := types.NamespacedName{Namespace: gitRepo.Namespace, Name: gitRepo.Name}
+			return ctrl.Result{}, grutil.UpdateErrorStatus(ctx, r.Client, nsname, gitRepo.Status, err)
+		}
+		logger.V(1).Info("Creating Git job resources")
+		if err := r.createJobRBAC(ctx, &gitRepo); err != nil {
+			return ctrl.Result{}, err
```
weyfonk, line 113: nit: in the same way that we format errors when failing to create a git job (or even better: using the `%w` [placeholder](https://pkg.go.dev/fmt#Errorf) instead of `%v`), we could format errors happening when failing to create RBAC resources or the configmap, to make them easier to understand for users, eg:
```suggestion
			return ctrl.Result{}, fmt.Errorf("failed to create RBAC resources for git job: %w", err)
```

weyfonk (approved):

charts/fleet/templates/rbac_gitjob.yaml
weyfonk: Nice find :+1: Thanks!

integrationtests/gitjob/controller/suite_test.go
```diff
@@ -59,14 +62,18 @@ var _ = BeforeSuite(func() {
 	Expect(err).NotTo(HaveOccurred())
 	Expect(k8sClient).NotTo(BeNil())
 
-	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
+	mgr, err := ctrlruntime.NewManager(cfg, ctrlruntime.Options{
```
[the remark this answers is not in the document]
0xavi0: Oh, yeah, I forgot to change that back to `ctrl`. 
There was already a variable around called `ctrl` and they were clashing.

It should be back to the standard `ctrl` now

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
0xavi0: yup, I've changed as suggested, and also... as I didn't want to have `%v`s and `%w`s in the same file... I ended up changing all `%v`s.

weyfonk (approved):

Issue #2474, opened 2024-05-30
Force Update on GitRepo is creating multiple job workloads that fill-up the entire pod limit.


sbulage:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

`Gitjob` controller is creating too many job workload pods in the cluster when perform `Force Update` on `GitRepo`.

GitRepo creation works as expected, resources are created without any error. But as soon as Force Update action performed on GitRepo it's status also changes from `Active` to `Reconciling`.

### Expected Behavior

`Force Update` should work as expected on GitRepo when performed.

### Steps To Reproduce

- Install Rancher 2.9-head which installs latest Fleet RC version: `fleet:v0.10.0-rc.14`
- Create a GitRepo with single resource.
- Wait for resource to be created in the cluster.
- Perform `Force Update` on GitRepo.
- Navigate to Local cluster or imported cluster.
- Watch the GitJob pods logs. (See below logs for more information.)

### Environment

```markdown
- Architecture:
- Fleet Version: fleet:v0.10.0-rc.14
- Cluster:
  - Provider: k3s
  - Options: 1
  - Kubernetes Version:
```

### Logs

<details>
<summary>GitJob Pod logs</summary>

```
{"level":"info","ts":"2024-05-30T14:26:07Z","logger":"gitjob","msg":"job deletion triggered because of ForceUpdateGeneration","controller":"gitrepo","controllerGroup":"fleet.cattle.io","controllerKind":"GitRepo","GitRepo":{"name":"test-repo","namespace":"fleet-local"},"namespace":"fleet-local","name":"test-repo","reconcileID":"e1b79be3-141c-4cb5-aa46-95d2a1665b31","generation":3,"commit":"323ef01da577d4f0fcdd1ad68cf295737859cb6b"}
{"level":"info","ts":"2024-05-30T14:26:07Z","logger":"gitjob","msg":"job deletion triggered because of ForceUpdateGeneration","controller":"gitrepo","controllerGroup":"fleet.cattle.io","controllerKind":"GitRepo","GitRepo":{"name":"test-repo","namespace":"fleet-local"},"namespace":"fleet-local","name":"test-repo","reconcileID":"474b69ab-5788-495b-b3f2-1f6a449d5463","generation":3,"commit":"323ef01da577d4f0fcdd1ad68cf295737859cb6b"}
[...]

PR #2477, opened 2024-05-31
Fixes infinite job generation after ForceSyncGeneration changed

0xavi0:
`gitrepo`s `Status.UpdateGeneration` value was changed but not updated, which made the controller to enter in an infinite loop regenerating the job.

Refers to: https://github.com/rancher/fleet/issues/2474


weyfonk (approved):
Nice one! LGTM

Issue #478, opened 2021-08-04
Fleet is not cleaning up objects on cluster delete


dramich:
After a cluster is deleted multiple objects are left:
ClusterRegistration
Role
RoleBinding

These should be deleted when the cluster is deleted.

thedadams:
This seems to cause problems with the following scenario:
1. Create an RKE2 cluster.
2. Wait for the cluster to become active.
3. Delete the cluster.
4. Create a cluster with the same name as the first.
The cluster will get provisioned, but will not become active. There will be an Unauthorized error in the UI.

It seems that there is a role leftover from the first cluster with name `request-*****` that is not getting deleted. Deleting this role before recreating the cluster allowed it to come up as active.

nickgerace:
What is likely happening is that the `NoDeleteGVK` is including cluster namespaces. Since they have the same name, it is re-using data from the previous namespace.

We need to keep `fleet-system` alive for 2.5.x to 2.6.x upgrades, so the solution to this will likely involve only protecting GVR corresponding to `fleet-system`, and not protecting other namespaces from wrangler applies.

bootstrapping and pre-controller registration - https://github.com/rancher/fleet/pull/468 (https://github.com/rancher/fleet/blob/96c8c497e7a44e349200c08375203e3044e7df7b/pkg/controllers/controllers.go#L78)
import scenarios - https://github.com/rancher/fleet/pull/475

nickgerace:
Worked out a potential solution with @StrongMonkey. Co-assigning for the time being after offline discussion

davidnuzik:
**My checks Failed**
 
**Reproduction Steps:**

Not required.
 
**Validation Steps:**

Rancher version: v2.6.0-rc2 (8/10/21) (rc2 is `9f6398367`)
Rancher cluster type: single-node docker
Docker version: 19.03
 
Downstream cluster type: rke2
Downstream K8s version: v1.21.3-rc4+rke2r2
 
Validation steps:

1. Provision Rancher Server. Use kubectl to query the local cluster for all Pods, ClusterRegistration, Role, and RoleBinding. This is out "starting point"
2. Provision a downstream rke2 cluster. It should not matter, but in my case I created a 1-node cluster with the DigitalOcean node driver. Wait for the cluster to fully provision - give it a minute or two (just in case fleet is wrapping anything up).
3. Delete the cluster. Give this a minute or two to fully complete (just in case fleet is wrapping anything up).
[...]

PR #2486, opened 2024-06-05
[SURE-8456] fix: add owner reference to cluster registration request

pratikjagrut:
Refers to [SURE-8456](https://jira.suse.com/browse/SURE-8456)

This pull request adds the owner reference to cluster registration request

### DIFF
--- a/internal/cmd/controller/agentmanagement/controllers/cluster/controller.go
+++ b/internal/cmd/controller/agentmanagement/controllers/cluster/controller.go
@@ -107,26 +107,7 @@ func clusterNamespace(clusterNamespace, clusterName string) string {
 }
 
 func (h *handler) OnClusterChanged(cluster *fleet.Cluster, status fleet.ClusterStatus) (fleet.ClusterStatus, error) {
-	logrus.Debugf("OnClusterChanged for cluster status %s, checking cluster registration, updating namespace in status", cluster.Name)
-	if cluster.DeletionTimestamp != nil {
-		// cluster is being deleted, clean up the cluster registrations
-		clusterRegistrations, err := h.clusterRegistrations.List(cluster.Namespace, metav1.ListOptions{})
-		if err != nil {
-			return status, err
-		}
-		for _, clusterRegistration := range clusterRegistrations.Items {
-			if clusterRegistration.Status.ClusterName == cluster.Name {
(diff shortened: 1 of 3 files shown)


manno:
From looking at related issues, like https://github.com/rancher/fleet/issues/478 and https://github.com/rancher/fleet/issues/1615 it seems clusterregistrations should be cleaned up.

The code at https://github.com/rancher/fleet/blob/main/internal/cmd/controller/agentmanagement/controllers/cluster/controller.go#L109-L134 is never executed. The handler doesn't trigger for delete events?

However regarding this PR:
* I would add the owner earlier, when OnChange created the cluster resource. At that point the cluster registration can be an owner.
* OnChange needs to update the request to actually store the owner, calling `SetOwnerReferences` is not enough.
* The [existing clean up code](https://github.com/rancher/fleet/blob/main/internal/cmd/controller/agentmanagement/controllers/cluster/controller.go#L109-L134) doesn't work, so it should also be removed. Or do we still need it for clusterregistrations that don't have a cluster yet?

pratikjagrut:
> From looking at related issues, like #478 and #1615 it seems clusterregistrations should be cleaned up.
> 
> The code at https://github.com/rancher/fleet/blob/main/internal/cmd/controller/agentmanagement/controllers/cluster/controller.go#L109-L134 is never executed. The handler doesn't trigger for delete events?
> 
It's not deleting the cluster registration request.
> However regarding this PR:
> 
>  >   * I would add the owner earlier, when OnChange created the cluster resource. At that point the cluster registration can be an owner.
> 
> >    * OnChange needs to update the request to actually store the owner, calling `SetOwnerReferences` is not enough.
> 
> >    * The [existing clean up code](https://github.com/rancher/fleet/blob/main/internal/cmd/controller/agentmanagement/controllers/cluster/controller.go#L109-L134) doesn't work, so it should also be removed. Or do we still need it for clusterregistrations that don't have a cluster yet?

I updated the code as you suggested, but I'm still unable to test it, having some issues with the dev environment. As soon as I'm done testing I'll update it here.

pratikjagrut:
@manno I tested the PR manually, and everything is working fine. Let me know if you need me to add tests to cover the changes.

weyfonk:
> @manno I tested the PR manually, and everything is working fine. Let me know if you need me to add tests to cover the changes.

Would you mind expanding a bit on what has been tested, and results of those tests? :)

weyfonk (approved):

PR #2490, opened 2024-06-06
Refactor Gitops Controllers

manno:


internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -77,54 +77,54 @@ func (r *GitJobReconciler) SetupWithManager(mgr ctrl.Manager) error {
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.15.0/pkg/reconcile
 func (r *GitJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	logger := log.FromContext(ctx).WithName("gitjob")
-	var gitRepo v1alpha1.GitRepo
+	gitrepo := &v1alpha1.GitRepo{}
```
weyfonk, line 80: Renaming aside, I liked having this variable defined as a value instead of a pointer, which made eg. `Get` and `updateStatus` calls more explicit about mutating `gitRepo`, since we would reference it.

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -77,54 +77,54 @@ func (r *GitJobReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 		return ctrl.Result{}, fmt.Errorf("error retrieving git job: %w", err)
 	}
 
-	if errors.IsNotFound(err) && gitRepo.Status.Commit != "" {
-		if err := r.validateExternalSecretExist(ctx, &gitRepo); err != nil {
-			nsname := types.NamespacedName{Namespace: gitRepo.Namespace, Name: gitRepo.Name}
-			return ctrl.Result{}, grutil.UpdateErrorStatus(ctx, r.Client, nsname, gitRepo.Status, err)
+	if errors.IsNotFound(err) && gitrepo.Status.Commit != "" {
+		if err := r.validateExternalSecretExist(ctx, gitrepo); err != nil {
+			nsname := types.NamespacedName{Namespace: gitrepo.Namespace, Name: gitrepo.Name}
```
bigkevmcd, line 108: I think you could use `client.ObjectKeyFromObject(gitrepo)` here.

https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#ObjectKeyFromObject

internal/cmd/controller/reconciler/gitrepo_controller.go
```diff
@@ -113,6 +97,22 @@ func (r *GitRepoReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+			if err := r.Get(ctx, req.NamespacedName, gitrepo); err != nil {
+				return err
+			}
+
+			controllerutil.AddFinalizer(gitrepo, gitRepoFinalizer)
+
+			return r.Update(ctx, gitrepo)
+		})
+
+		if client.IgnoreNotFound(err) != nil {
```
bigkevmcd, line 111: You could simplify this to `return client.IgnoreNotFound(err)` because `IgnoreNotFound` returns `nil` if the error is a _NotFound_ error?

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
manno: Interesting function, but here we can just use the existing `req.NamespacedName`.

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
manno: I didn't copy the commit messages into the PR description, but the short form with a pointer is just so much more common in my opinion.
Sure, mutation becomes more obvious, but there are lot's of functions, e.g. `createJobRBAC`, which expect a pointer, but don't mutate. While get and updatestatus always do mutate, as indicated by their name.

internal/cmd/controller/reconciler/gitrepo_controller.go
manno: If I understand correctly, this code block should not exit the reconciler when the resource is not found. Not sure if that makes sense, as line 122 will exit the reconcile for an empty gitrepo? 

So, this is probably what we want? When we can't add the finalizer we don't reconcile. However, when the resource disappeared, it's not a re-queue and we don't error.
```
                 if err != nil {
                         return ctrl.Result{}, client.IgnoreNotFound(err)
                 }
```

weyfonk: I think it makes sense to exit the reconcile loop if adding a finalizer to the resource failed because the resource was not found. This could be symptomatic of the resource having been manually deleted before Fleet had a chance to add a finalizer to claim control over its deletion. Unlikely, but still possible? I agree with Mario's suggestion.

In any case, I don't think that we want to return/exit unconditionally here, as this would short-circuit further status updates at least until the next reconcile loop. Not sure if this is what Kevin suggested.

Apart from that, when the resource is not found, the reconcile should have been exited [at L61](https://github.com/rancher/fleet/pull/2490/files#diff-05c10d2aa87a39d27654c853a7b5535ca09ec533a1918197a3531abceeac75a9R61), so I wonder if L122 is still relevant.

manno: > Apart from that, when the resource is not found, the reconcile should have been exited [at L61](https://github.com/rancher/fleet/pull/2490/files#diff-05c10d2aa87a39d27654c853a7b5535ca09ec533a1918197a3531abceeac75a9R61), so I wonder if L122 is still relevant.

I think this is just about the "small" time window between the initial get and the end of the retry statement. So it's still relevant. 

This would also be fine (that's maybe what you meant), as line 61 will cover it in the follow-up reconcile to the error:
```
  if err != nil {
                         return ctrl.Result{}, err
 }
```

weyfonk: > This would also be fine (that's maybe what you meant), as line 61 will cover it in the follow-up reconcile to the error:

I hadn't thought of that, but this also makes sense, although we could probably save a reconcile loop by simply ignoring not-found errors.

> I think this is just about the "small" time window between the initial get and the end of the retry statement. So it's still relevant.

So yes, this is the small time window I had in mind, but if deletion happens during that window and the resource is no longer found, then the reconciler will exit based on L111-113, ie the block where this thread is anchored (whether or not we treat not-found errors separately). In that case, L122 would never be reached, would it?
Or is that line a separate validation step, preventing reconciles on invalid (but existing) `GitRepo`s?

0xavi0 (approved):
LGTM

PR #2494, opened 2024-06-07
Fix example values.yaml for enabling OCI storage

bigkevmcd:
Refers to #XXX

Try installing the chart with the example in the values.yaml uncommented, this will not apply cleanly.

This changes the example in the values.yaml for enabling the OCI storage to be valid for insertion into the "env" for a container.

Previously, it was a boolean, which would lead to a failure to unmarshal a bool to a string.


0xavi0 (approved):
Thanks for catching.

Issue #2456, opened 2024-05-23
Enable node selection for shards


weyfonk:
Fleet recently got basic sharding support through #1740, which allows a set of shards to be configured at deployment time.

This would be more flexible if users could configure which shard(s) would be handled by which node in their cluster, by means of a `nodeSelector` which would be set on each shard's `fleet-controller` deployment.

In turn, this means that the currently basic configuration/values format used to set shards should evolve to support a node for each shard.

Tommy12789:
## Additional QA
### Issue: <link the issue or issues this PR resolves here>
2456
 
### Problem
N/A 
### Solution
Implemented a way to add nodeselector for controllers shards 
 
### Testing
manual test only 

### Engineering Testing
#### Manual Testing
Installing fleet with shards and nodeSelector then verify each shard is well deployed in the good node
#### Automated Testing
N/A
### QA Testing Considerations
- Testing basic functionment
- Testing with duplicate shards id
- Testing a shard without nodeSelector to verify it's also deployed without the parameter 
#### Regressions Considerations
N/A

weyfonk:
Thanks @mmartin24 and @thehejik for raising this!
This looks like an omission in gitjob deployments, ie we most likely need to take the shard's node selector into account in [gitjob deployments](https://github.com/rancher/fleet/blob/main/charts/fleet/templates/deployment_gitjob.yaml) as we do for [fleet-controller deployments](https://github.com/rancher/fleet/blob/main/charts/fleet/templates/deployment.yaml#L235-L238).

mmartin24:
To clarify the above pasting @weyfonk  comment on Slack over this issue:

>There was a bug with scheduling of gitjob pods, but we don't specify any scheduling for jobs created by Fleet.
>I think it would make sense for node selection to be present there as well, for the sake of consistency.
>This could work by detecting which shard (if any) a gitjob controller applies to, and applying the corresponding node selector [here](https://github.com/rancher/fleet/blob/main/internal/cmd/controller/gitops/reconciler/gitjob_controller.go#L603).

mmartin24:
# Test 1 - Basic shard deployment with nodeSelector works
```json
{
  "debug": true,
  "debugLevel": 1,
  "shards": [
    {
      "id": "shard0",
      "nodeSelector": {
        "kubernetes.io/hostname": "k3d-3-server-3-agents-cluster-agent-0"
      }
    },
    {
      "id": "shard1",
      "nodeSelector": {
[...]

PR #2505, opened 2024-06-11
Adding NodeSelection for Controller Shards

Tommy12789:
Refers to #2456

Changed the shards format in helm Values to add a new parameter NodeSelector to chose the node for each controller shard.

You can set them directly into the value.yaml or pass them as argument ti setup-fleet script (json formating) as you can see in the setup-single-cluster script


dev/setup-single-cluster
```diff
@@ -12,7 +12,7 @@ source ./dev/setup-cluster-config
 ./dev/setup-k3d
 ./dev/build-fleet
 ./dev/import-images-k3d
-./dev/setup-fleet upstream shard1,shard2,shard3
+./dev/setup-fleet upstream '[{"id":"shard0","nodeSelector":{"kubernetes.io/hostname":"k3d-upstream-server-0"}},{"id":"shard1"},{"id":"shard2","nodeSelector":{"kubernetes.io/hostname":"k3d-upstream-server-2"}}]'
```
weyfonk, line 15: Would this work over several lines, eg. something like this, which would be more readable?
```suggestion
./dev/setup-fleet upstream '[
    {"id":"shard0","nodeSelector":{"kubernetes.io/hostname":"k3d-upstream-server-0"}},
    {"id":"shard1"},
    {"id":"shard2","nodeSelector":{"kubernetes.io/hostname":"k3d-upstream-server-2"}}
    ]'
```

charts/fleet/templates/service.yaml
```diff
@@ -1,13 +1,14 @@
-{{- if .Values.metrics.enabled }}
```
weyfonk, line 1: I think we still want this condition here. Why was it removed?

charts/fleet/templates/deployment.yaml
```diff
@@ -1,12 +1,14 @@
-{{ $shards := list "" }}
-{{ if .Values.shards }}
-{{ $shards = concat $shards .Values.shards | uniq }}
-{{ end }}
-{{ range $shards }}
+{{- $shards := list (dict "id" "" "nodeSelector" dict) -}}
+{{- if .Values.shards -}}
+  {{- range .Values.shards -}}
+    {{- $shards = append $shards . -}}
```
weyfonk, line 4: What if a user specifies the same shard (or at least the same shard ID, possibly with different node selectors) multiple times in `.Values.shards`?

charts/fleet/templates/deployment.yaml
```diff
@@ -224,11 +226,13 @@ spec:
       {{- end }}
 
       serviceAccountName: fleet-controller
-      nodeSelector: {{ include "linux-node-selector" . | nindent 8 }}
-{{- if $.Values.nodeSelector }}
```
weyfonk, line 228: We seem to be omitting this now: if a node selector is present in the values, it will not be reflected in the deployment anymore :thinking:

weyfonk (commented):
Nice one, thanks :)
Leaving a few comments and doubts.

charts/fleet/templates/deployment.yaml
Tommy12789: An error, I need to work on this, I tried several implementations not working, I'll take time to do it

charts/fleet/templates/service.yaml
Tommy12789: An error TY

dev/setup-single-cluster
Tommy12789: I'll try thx

charts/fleet/templates/deployment.yaml
```diff
@@ -1,12 +1,29 @@
...
+      {{- $found = true -}}
+      {{- break -}}
+    {{- end -}}
+  {{- end -}}
+  {{- if not $found -}}
+    {{- $uniqueShards = append $uniqueShards $shard -}}
+  {{- end -}}
+{{- end -}}
+
+{{ range $shard := $uniqueShards }}
```
weyfonk, line 22: This works, but there is a much simpler way ;)
```suggestion
{{ $shards = $shards | uniq }}

{{ range $shard := $shards }}
```

charts/fleet/templates/service.yaml
```diff
@@ -1,13 +1,30 @@
...
+      {{- $found = true -}}
+      {{- break -}}
+    {{- end -}}
+  {{- end -}}
+  {{- if not $found -}}
+    {{- $uniqueShards = append $uniqueShards $shard -}}
+  {{- end -}}
+{{- end -}}
+
+{{ range $shard := $uniqueShards }}
```
weyfonk, line 23:
```suggestion
{{ $shards = $shards | uniq }}

{{ range $shard := $shards }}
```

charts/fleet/templates/deployment.yaml
Tommy12789: Already tried and it do not work because if two shards have the same id but not the same node selector it will not delete them, we just wan't the id to not be duplicated

Tommy12789: This compare the whole list item

Tommy12789: Yes, just updated it, thank you

weyfonk (approved):

manno (dismissed):

manno (approved):

PR #2510, opened 2024-06-12
Bump go releaser config

thardeck:
I have migrated the config following: https://goreleaser.com/blog/goreleaser-v2/

Since we do not seem to use any deprecated options, I have just had to add the version to the config file.

I have ran the goreleaser check with the following output:

```
$ goreleaser check
  • checking                                 path=.goreleaser.yaml
  • 1 configuration file(s) validated
  • thanks for using goreleaser!
```


manno (approved):

weyfonk (dismissed):
LGTM

Issue #2515, opened 2024-06-13
[0.9] [SURE-8550] drift detection is generating secrets without cleaning 


kkaempf:
### Issue description:
When enabling Self Healing (drift detection) Fleet will generate a new secret every time drift is detected. To a point where it might exhaust Rancher.
Fleet 0.9.4

### Business impact:
For the customer Rancher went down due to too many secrets being cached

### Troubleshooting steps:
Disabling self healing will clean the secrets

### Repro steps:

* create a git repo with a simple deployment (e.g. https://github.com/rbreddy/bundledependency/tree/main/hello-world)
* enable self healing
* scale the Deployment resource up (e.g. from 1 to 2 replicas), and observe how it's automatically reverted.
* multiple secrets (type `helm.sh/release.v1`) will be created in the namespace for the deployment
  * This can also be checked using `helm history` commands in the target namespace and specifying the Helm release name.

### Workaround:
Is a workaround available and implemented? yes
What is the workaround: disable self healing (disabling self healing also remove all the secrets)

### Actual behavior:
Multiple secrets are created for a single "correction", and old ones are preserved.

### Expected behavior:
Only 1 secret is created per "correction", while keeping the total number of Helm releases at a maximum of just 2.

### Additional notes:
~~~
helm history  test-fastweb-hello-world -n hello
REVISION	UPDATED                 	STATUS    	CHART                   	APP VERSION	DESCRIPTION
164     	Wed Jun 12 15:06:08 2024	superseded	nginx-rancherhello-0.0.1	0.0.0      	Upgrade complete
165     	Wed Jun 12 15:06:09 2024	superseded	nginx-rancherhello-0.0.1	0.0.0      	Upgrade complete
166     	Wed Jun 12 15:06:24 2024	superseded	nginx-rancherhello-0.0.1	0.0.0      	Rollback to 165
167     	Wed Jun 12 15:06:31 2024	deployed  	nginx-rancherhello-0.0.1	0.0.0      	Rollback to 166
~~~

rancherbot:
Not creating port issue, milestone v0.10 does not exist or is not an open milestone

weyfonk:
## Additional QA
 
### Problem
Correcting drift on Fleet-deployed resources would create a new Helm release, and a new `sh.helm.<ID>` secret every time, leading to an expanding set of stored secrets and Helm history items. This could lead to performance issues.
 
### Solution
[...]

PR #2517, opened 2024-06-14
[0.9] Honor Helm's MaxHistory when using drift correction

aruiz14:
Refers to #2515

* Drift correction (self-healing) uses Helm's [Rollback action](https://pkg.go.dev/helm.sh/helm/v3/pkg/action#Rollback), which has a `MaxHistory` field. This value takes precedence over the config value. Unlike the [`Upgrade` action](https://github.com/rancher/fleet/blob/68597ce1aca9bfac58547f8b92ad060b7aa90cf5/internal/helmdeployer/install.go#L163-L166), we omitted to set this value, so the releases will only shrink when running an upgrade but not rollbacks`.
* I noticed that sometimes, a regular upgrades would be deployed (creating a new). Initially I thought this would happen if all the releases in the remaining history were rollbacks, but realized the problem is that the BundleDeployment status has a `release` field, referencing the latest release installed. This value is not being updated on rollbacks, hence making the status differ from what `helm history` reflects.

### DIFF
--- a/internal/helmdeployer/deployer.go
+++ b/internal/helmdeployer/deployer.go
@@ -246,17 +246,17 @@ func (h *Helm) Deploy(bundleID string, manifest *manifest.Manifest, options flee
 
 // RemoveExternalChanges does a helm rollback to remove changes made outside of fleet.
 // It removes the helm history entry if the rollback fails.
-func (h *Helm) RemoveExternalChanges(bd *fleet.BundleDeployment) error {
+func (h *Helm) RemoveExternalChanges(bd *fleet.BundleDeployment) (*Resources, error) {
 	logrus.Infof("Drift correction: rollback BundleDeployment %s", bd.Name)
 
 	_, defaultNamespace, releaseName := h.getOpts(bd.Name, bd.Spec.Options)
 	cfg, err := h.getCfg(defaultNamespace, bd.Spec.Options.ServiceAccount)
 	if err != nil {
-		return err
+		return nil, err
 	}
 	currentRelease, err := cfg.Releases.Last(releaseName)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	r := action.NewRollback(&cfg)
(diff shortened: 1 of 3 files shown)


internal/helmdeployer/deployer.go
```diff
@@ -246,30 +246,40 @@ func (h *Helm) Deploy(bundleID string, manifest *manifest.Manifest, options flee
...
+		}
+		return nil, removeFailedRollback(cfg, currentRelease, err)
 	}
 
-	return err
+	release, err := cfg.Releases.Last(releaseName)
+	if err != nil {
+		return nil, err
+	}
+	return releaseToResources(release)
```
weyfonk, line 282: This involves converting the full contents of the release manifest into YAML objects; I wonder if that might be a bit overkill, when all we need is the release ID. It could make sense to extract release name computation into a separate function instead, and call that.

weyfonk (commented):
Thanks for troubleshooting and fixing this bug :)
Ideally we'd have tests for this, which in this case most likely mean end-to-end tests.

internal/helmdeployer/deployer.go
aruiz14: Good point, addressed in 235d80be

Makefile
```diff
@@ -12,4 +12,19 @@ $(TARGETS): .dapper
...
-.PHONY: $(TARGETS)
+debug:
+	k3d image import rancher/fleet:${TAG} -c st-upstream
+	k3d image import rancher/fleet-agent:${TAG} -c st-upstream
+
+debug-agent:
+	k3d image import rancher/fleet-agent:${TAG} -c st-upstream
+
+restart-agent:
+	KUBECONFIG=/Users/aruiz/projects/github/moio/scalability-tests/terraform/main/k3d/config/upstream.yaml kubectl -n cattle-fleet-local-system scale deploy/fleet-agent --replicas=0
```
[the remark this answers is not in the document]
aruiz14: Oops, thanks :)

weyfonk (approved):

weyfonk (approved):

Issue #2521, opened 2024-06-14
Drift detection/correction should omit status fields


aruiz14:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Sometimes will detect false positives about modified resources, reporting differences inside the `status` field.
This situation is impossible to remediate (`status` is normally a subresource and is not part of the deployment specs), which also causes a redeployment loop when drift correction (self-healing) is active.

However, this does not happen for all resources, e.g. `deployments`, but I experienced when installing a chart that contained CRDs.

### Expected Behavior

Changes in `status` fields are not reported as modified resources.

### Steps To Reproduce

1. Install Fleet
2. Add a GitRepo whose `fleet.yaml` is like:
```
defaultNamespace: longhorn-system
helm:
  chart: longhorn-crd
  repo: https://charts.rancher.io
  releaseName: longhorn-crd
  version: "103.3.1+up1.6.2"
```
3. Check how it will stay as `NotReady`, reporting some CRDs were modified, e.g.:
```
  modifiedStatus:
  - apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    name: engineimages.longhorn.io
    patch: '{"status":{"acceptedNames":{"kind":"","plural":""},"conditions":[],"storedVersions":[]}}'
  - apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    name: engines.longhorn.io
    patch: '{"status":{"acceptedNames":{"kind":"","plural":""},"conditions":[],"storedVersions":[]}}'
  - apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    name: instancemanagers.longhorn.io
    patch: '{"status":{"acceptedNames":{"kind":"","plural":""},"conditions":[],"storedVersions":[]}}'
  - apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    name: nodes.longhorn.io
    patch: '{"status":{"acceptedNames":{"kind":"","plural":""},"conditions":[],"storedVersions":[]}}'
  - apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    name: replicas.longhorn.io
    patch: '{"status":{"acceptedNames":{"kind":"","plural":""},"conditions":[],"storedVersions":[]}}'
  - apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    name: settings.longhorn.io
[...]

PR #2522, opened 2024-06-14
Ignore status fields during drift detection

aruiz14:
Refers to #2521

Use a new `StatusNormalizer` when comparing deployed resources against the desired spec, since `status` is normally not part of it.

### DIFF
--- a/e2e/drift/drift_test.go
+++ b/e2e/drift/drift_test.go
@@ -184,6 +184,16 @@ var _ = Describe("Drift", Ordered, func() {
 					`{"name":"http","port":1234,"protocol":"TCP","targetPort":"http-web-svc"}]}}`))
 			})
 		})
+
+		Context("Resource manifests containing status fields", func() {
+			It("Is marked as not ready", func() {
+				bundleName := "drift-correction-test-drift-ignore-status"
+				Eventually(func() bool {
+					b := getBundle(bundleName, k)
+					return b.Status.Summary.Ready == 1
+				}).Should(BeTrue())
+			})
+		})
 	})
 
 	When("Drift correction is enabled with force", func() {
--- a/e2e/drift/drift_test.go
+++ b/e2e/drift/drift_test.go
@@ -186,7 +186,8 @@ var _ = Describe("Drift", Ordered, func() {
 		})
 
 		Context("Resource manifests containing status fields", func() {
-			It("Is marked as not ready", func() {
+			// Status must be ignored for drift correctiong, despite being part of the manifests
+			It("Is marked as ready", func() {
 				bundleName := "drift-correction-test-drift-ignore-status"
 				Eventually(func() bool {
 					b := getBundle(bundleName, k)
--- a/e2e/drift/drift_test.go
+++ b/e2e/drift/drift_test.go
@@ -186,7 +186,7 @@ var _ = Describe("Drift", Ordered, func() {
 		})
 
 		Context("Resource manifests containing status fields", func() {
-			// Status must be ignored for drift correctiong, despite being part of the manifests
+			// Status must be ignored for drift correction, despite being part of the manifests
 			It("Is marked as ready", func() {
 				bundleName := "drift-correction-test-drift-ignore-status"
 				Eventually(func() bool {
--- a/internal/cmd/agent/deployer/normalizers/status.go
+++ b/internal/cmd/agent/deployer/normalizers/status.go
@@ -1,0 +1,13 @@
+package normalizers
+
+import (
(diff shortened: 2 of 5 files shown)


e2e/drift/drift_test.go
```diff
@@ -18,6 +19,9 @@ var _ = Describe("Drift", func() {
 		k          kubectl.Command
 		namespace  string
 		bundleName string
+
+		interval = 2 * time.Second
+		duration = 30 * time.Second
```
weyfonk, line 24: This feels like a long time, on the other hand we know that end-to-end tests can be slow on Github runners.

weyfonk (approved):
Thanks for this fix :)

e2e/drift/drift_test.go
aruiz14: I know, but I just copied it from somewhere else. I'm fact, this is not a timeout, but used for a `Consistently()` statement. It's the time that during which it will continue checking if the condition is still true.
Please let me know if you still want to change this to 10 or 15 seconds.

weyfonk: No worries, let's leave it at this for now; what I meant is that precisely because of slowness on Github runners, it makes sense for `Consistently` to run over a longer duration, just to be on the safe side.

e2e/drift/drift_test.go
```diff
@@ -184,6 +184,16 @@ var _ = Describe("Drift", Ordered, func() {
...
+
+		Context("Resource manifests containing status fields", func() {
+			It("Is marked as not ready", func() {
+				bundleName := "drift-correction-test-drift-ignore-status"
+				Eventually(func() bool {
+					b := getBundle(bundleName, k)
+					return b.Status.Summary.Ready == 1
+				}).Should(BeTrue())
+			})
+		})
```
aruiz14, line 196: I tested this with an older version, and checked that it fails. The bundle never reaches a ready status:
```
❯ kubectl get bundle -n fleet-local -w drift-correction-test-drift-ignore-status
NAME                                        BUNDLEDEPLOYMENTS-READY   STATUS
drift-correction-test-drift-ignore-status   0/1                       Modified(1) [Cluster fleet-local/local]; deployment.apps drift-ignore-status/nginx-deployment modified {"status":{"collisionCount":0}}
```

internal/cmd/agent/deployer/normalizers/status.go
```diff
@@ -0,0 +1,13 @@
+package normalizers
+
+import (
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
```
bigkevmcd, line 5: Feels like we could write a test for this code here?

Also, I wonder about something like this in `internal/cmd/agent/deployer/internal/diff/diff.go` which would simplify this (and the noopNormalizer).


```go
// The NormalizerFunc type is an adapter to allow the use of ordinary functions
// as Normalizers. If f is a function with the appropriate signature,
// NormalizerFunc(f) is a Normalizer that calls f.
type NormalizerFunc func(*unstructured.Unstructured) error

// Normalize calls f(un).
func (f NormalizerFunc) Normalize(un *unstructured.Unstructured) error {
    return f(un)
}
```

```go
// GetNoopNormalizer returns normalizer that does not apply any resource modifications
func GetNoopNormalizer() Normalizer {
    return NormalizerFunc(func(*unstructured.Unstructured) error {
        return nil 
    })  
}
```

Which could even be simplified further to...

```go
// NoopNormalizer is a normalizer that does not apply any resource modifications
var NoopNormalizer = NormalizerFunc(func(*unstructured.Unstructured) error {
    return nil 
})  
```

Where the StatusNormalizer would become...

```go
// StatusNormalizer removes a top-level "status" fields from the object, if present
func NormalizeStatus(un *unstructured.Unstructured) error {
    unstructured.RemoveNestedField(un.Object, "status")
    return nil 
}
```

aruiz14: While I agree with you, I believe such refactoring belongs to a separate PR. I prefer keeping things as independent as possible.

To be honest, I didn't even know these code existed before investigating this issue, so maybe the same happens for other team members and they would like to review it more deeply and apply any needed refactorings.

About adding a test, I didn't do it initially because I thought it wouldn't provide much value, since it would be like testing `RemoveNestedField` instead, but I'll add it since it doesn't hurt either and should be easy. Thanks.

bigkevmcd (dismissed):

weyfonk (approved):
LGTM :)

weyfonk (approved):

PR #2524, opened 2024-06-14
Keep schemebuilder type

manno:
Refers to https://github.com/rancher/rancher/pull/45271 and https://github.com/rancher/fleet/pull/2377

rancher/rancher and rancher/shepherd are expecting `SchemeBuilder` to be a `apimachinery.SchemeBuilder`, we want to use the [one from controller runtime](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/scheme/scheme.go) internally. Embedding the c-r code, so consumers don't inherit the controller-runtime dependency from our API pkg. Hasn't changed for years.

### DIFF
--- a/pkg/apis/internal/scheme.go
+++ b/pkg/apis/internal/scheme.go
@@ -1,0 +1,93 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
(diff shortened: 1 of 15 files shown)


pkg/apis/internal/scheme.go
```diff
@@ -0,0 +1,93 @@
...
+//	func init() {
+//		SchemeBuilder.Register(&MyType{}, &MyTypeList)
+//	}
+//	var (
+//		scheme *runtime.Scheme = runtime.NewScheme()
+//	)
+//
+// This also true of the built-in Kubernetes types.  Then, in the entrypoint for
+// your manager, assemble the scheme containing exactly the types you need,
+// panicing if scheme registration failed. For instance, if our controller needs
```
aruiz14, line 40: @manno this reached `main` and is now reported for every PR.

weyfonk: A commit like [this one](https://github.com/rancher/fleet/pull/2507/commits/585fd48cfbe6c140898fdc8d01e6a7361b037499) fixes it, in case we don't want to edit this embedded code.

aruiz14: Thanks! I missed that part. I've changed the PR to completely ignore this file then

Issue #2450, opened 2024-05-22
metrics: Add metrics to gitops controller


p-se:
Several metrics have been added to various controllers. The Gitops controller is not among them. This needs to be remedied.

Since monitoring of the other controllers is based on the status of the custom resources they handle and the gitops controller does not have a custom resource it handles, there is nothing to monitor in the sense of what the other controllers do.

While monitoring can still be implemented, we would need to think about what exactly would be supposed to be monitored. The `controller-runtime` framework we use provides generic metrics. Those metrics are already exported by the other controllers and collected and stored by Prometheus. Those metrics enable the creation or use of dashboards to visualize performance of those controllers. To be able to visualize the performance of the gitops controller, it

1. needs to be checked whether gitops controller does provide metrics already. If not, it needs to adapted to provide those.
2. A Kubernetes service needs to be created when the fleet (controller) chart is installed. The service needs to point to the gitops controller(s)'s port for metrics. If sharding is used, several services need to be created.
3. A `ServiceMonitoring` needs to be added to the `rancher-monitoring` chart to configure Prometheus to scrape the metrics from the gitops controller.

p-se:
Since monitoring of the other controllers is based on the status of the custom resources they handle and the gitops controller does not have a custom resource it handles, there is nothing to monitor in the sense of what the other controllers do.

While monitoring can still be implemented, we would need to think about what exactly would be supposed to be monitored. The `controller-runtime` framework we use provides generic metrics. Those metrics are already exported by the other controllers and collected and stored by Prometheus. Those metrics enable the creation or use of dashboards to visualize performance of those controllers. To be able to visualize the performance of the gitops controller, it

1. needs to be checked whether gitops controller does provide metrics already. If not, it needs to adapted to provide those.
2. A Kubernetes service needs to be created when the fleet (controller) chart is installed. The service needs to point to the gitops controller(s)'s port for metrics. If sharding is used, several services need to be created.
3. A `ServiceMonitoring` needs to be added to the `rancher-monitoring` chart to configure Prometheus to scrape the metrics from the gitops controller.

PR #2530, opened 2024-06-17
Add metrics to gitops controller

Tommy12789:
Refers to #2450 

Introduced a new service for gitops metrics

### DIFF
--- a/charts/fleet/templates/service_gitops_metrics.yaml
+++ b/charts/fleet/templates/service_gitops_metrics.yaml
@@ -1,0 +1,15 @@
+apiVersion: v1
+kind: Service
+metadata:
+    name: "monitoring-gitops-controller"
+    labels:
+        app: gitjob
+spec:
+    type: ClusterIP
+    ports:
+        - port: 8081
+          targetPort: 8081
+          protocol: TCP
+          name: metrics
+    selector:
+        app: gitjob
--- a/internal/cmd/controller/gitops/operator.go
+++ b/internal/cmd/controller/gitops/operator.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"os"
 	"time"
 
 	command "github.com/rancher/fleet/internal/cmd"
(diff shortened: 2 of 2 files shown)


charts/fleet/templates/service_gitops_metrics.yaml
```diff
@@ -0,0 +1,31 @@
...
+    type: ClusterIP
+    ports:
+        - port: 8081
+          targetPort: 8081
+          protocol: TCP
+          name: metrics
+    selector:
+        app: gitjob
+        {{- if empty . }}
+        # fleet.cattle.io/shard-default: "true"
```
p-se, line 24: Doesn't that need to be enabled?

Tommy12789: It do not work for the moment as xavi didn't merge is PR

p-se: Hmm, but it will work eventually and doesn't do any harm in the meantime. The same is true for `fleet.cattle.io/shard-id: <shard-id>`, right?

So, we either make it compatible with the sharding functionality that will be there in the future, or we just make it work for the unsharded gitjob controller. Both ways will enable us to proceed with this PR before https://github.com/rancher/fleet/pull/2537 is merged.

Personally, I'm fine with having that feature in already, even if it does not work yet, though, removing the sharding functionality from this controller and waiting for the sharding of gitjob controller is probably less controversial.

@weyfonk @0xavi0 

Any opinions? We need this service to be merged and those changes to be put into a release candidate to have QA testing the [PR in rancher/charts](https://github.com/rancher/charts/pull/4101), which depends on this service.

p-se (approved):
Works as expected.

0xavi0:
Don't we also need to expose the port and name it in the `gitjob` deployment?

For example:
```yaml
spec:
      serviceAccountName: gitjob
      containers:
        - image: "{{ template "system_default_registry" $ }}{{ $.Values.image.repository }}:{{ $.Values.image.tag }}"
          name: gitjob
          {{- if $.Values.metrics.enabled }}
          ports:
          - containerPort: 8081
            name: metrics
          {{- end }}
          args:
          - fleetcontroller
          - gitjob
```

Issue #2115, opened 2024-01-31
[SURE-7365?] Update status fields to provide what UI needs for "resourceCount" 


manno:
## Background

The old key `BundleStatus.ResourceKey`  https://fleet.rancher.io/ref-crds#resourcekey works like this:
> ResourceKey lists resources, which will likely be deployed. The actual list of resources on a cluster might differ, depending on the helm chart, value templating, etc..

The [sum of all resources in bundles is stored](https://github.com/rancher/fleet/blob/2fb2ed57f19ea85ccebd9641e5af497bbd6a1554/internal/cmd/controller/gitrepo/resourcekey.go#L13) in the GitRepo.
This seems to multiply the estimate.

We want to use the accurate resource lists and counts.
The `BundleDeploymentStatus.Resources` lists the actual deployed resources from Helm:
> Resources lists the metadata of resources that were deployed according to the helm release history.

### Open Questions

- [ ] what views in the UI use the resource list? what makes sense?
  - 1 gitrepo -> n bundles -> n clusters 
  -  resources per cluster
  - resources per gitrepo 
- [x] can the UI use `bundledeployment.status.resources`  directly? - It does.
- [ ] is there a view for resources/bundles of a gitrepo per cluster?
- [ ] should work bundlenamespacemapping

### Possible Follow Ups:

-  document meaning of conditions. Is "Active" a failure state?
-  remove `helm template`  fleet controller in a later release (like 2.10)

## TODO

- [x] populate `gitrepo.status.resources` from `bundledeployment.status.resources` instead, [see fromResourceKey](https://github.com/rancher/fleet/blob/b7fdd8306cb1d766cb90b40f88c76fd6de866e85/internal/cmd/controller/gitrepo/resourcekey.go#L37). [PR #2523](https://github.com/rancher/fleet/pull/2523)
- [x] `bundle.status.resourceKey` is still computed from `helm template`, for compatibility. We need to check if UI dashboard uses this. Once it's no longer used, we can stop populating the field and remove `helm template`.
- [x] status derived from new `bundledeployment.status.resources` in bundle reconciler [PR #2535](https://github.com/rancher/fleet/pull/2535)
- ~[x] status derived from new resources key  in gitrepo reconciler~
- ~[x] resource count calculation in cluster reconciler~
- ~[x] resource count calculation in clustergroup reconciler?~
- ~[x] deprecate resourceKey and related outdated status fields~ 

Also see:
* https://github.com/rancher/fleet/issues/1101

[...]

PR #2535, opened 2024-06-18
Populate bundle status from bundleDeployment status resources

rubhanazeem:
Refers to #2115
- [x] status derived from new `bundledeployment.status.resources` in bundle reconciler

### DIFF
--- a/integrationtests/controller/bundle/status_test.go
+++ b/integrationtests/controller/bundle/status_test.go
@@ -70,10 +70,19 @@ var _ = Describe("Bundle Status Fields", func() {
 				if err != nil {
 					return err
 				}
+				resources := []v1alpha1.BundleDeploymentResource{}
+				resource := v1alpha1.BundleDeploymentResource{
+					Kind:       "ConfigMap",
+					APIVersion: "v1",
+					Namespace:  namespace,
+					Name:       "app-config",
+				}
+				resources = append(resources, resource)
 				bd.Status.Display.State = "Ready"
 				bd.Status.AppliedDeploymentID = bd.Spec.DeploymentID
 				bd.Status.Ready = true
 				bd.Status.NonModified = true
+				bd.Status.Resources = resources
 				return k8sClient.Status().Update(ctx, bd)
 			}).ShouldNot(HaveOccurred())
 
@@ -81,6 +90,7 @@ var _ = Describe("Bundle Status Fields", func() {
 			Expect(err).NotTo(HaveOccurred())
 			Expect(bd.Status.Display.State).To(Equal("Ready"))
 			Expect(bd.Status.AppliedDeploymentID).To(Equal(bd.Spec.DeploymentID))
+			Expect(len(bd.Status.Resources)).To(Equal(1))
 
 			Eventually(func() bool {
 				err = k8sClient.Get(ctx, types.NamespacedName{Namespace: namespace, Name: "name"}, bundle)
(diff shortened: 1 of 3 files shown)


internal/cmd/controller/reconciler/bundle_status.go
```diff
@@ -41,64 +32,26 @@ func updateDisplay(status *fleet.BundleStatus) {
...
 			continue
 		}
-
-		for _, obj := range objs {
-			m, err := meta.Accessor(obj)
-			if err != nil {
-				return err
-			}
+		if target.Deployment.Namespace == "" {
+			logrus.Infof("Skipping bundledeployment with empty namespace %v", target.Deployment)
```
weyfonk, line 43: I find this log message a bit confusing, as if `%v` were a placeholder for an empty namespace...
Also, what information is really relevant for logging here? Does a full bundle deployment body make sense, or should we instead log its name?
How about this? Happy to discuss.
```suggestion
			logrus.Infof("Skipping bundledeployment %q with empty namespace", target.Deployment.Name)
```

integrationtests/controller/bundle/status_test.go
```diff
@@ -70,25 +70,42 @@ var _ = Describe("Bundle Status Fields", func() {
...
 				bd.Status.NonModified = true
+				bd.Status.Resources = resources
 				return k8sClient.Status().Update(ctx, bd)
 			}).ShouldNot(HaveOccurred())
 
 			err = k8sClient.Get(ctx, types.NamespacedName{Namespace: namespace, Name: "name"}, bd)
 			Expect(err).NotTo(HaveOccurred())
 			Expect(bd.Status.Display.State).To(Equal("Ready"))
 			Expect(bd.Status.AppliedDeploymentID).To(Equal(bd.Spec.DeploymentID))
+			Expect(bd.Status.Resources).ToNot(BeNil())
```
weyfonk, line 93: This looks safer to me:
```suggestion
			Expect(len(bd.Status.Resources)).To(Equal(1))
```

integrationtests/controller/bundle/status_test.go
```diff
@@ -70,25 +70,42 @@ var _ = Describe("Bundle Status Fields", func() {
...
 
+			bdResource := bd.Status.Resources[0]
```
weyfonk, line 95: nit: I'd move this closer to where it is needed, in this case to L105.

integrationtests/controller/bundle/status_test.go
```diff
@@ -70,25 +70,42 @@ var _ = Describe("Bundle Status Fields", func() {
...
 			Eventually(func() bool {
 				err = k8sClient.Get(ctx, types.NamespacedName{Namespace: namespace, Name: "name"}, bundle)
 				Expect(err).NotTo(HaveOccurred())
 				return bundle.Status.Summary.Ready == 1
 			}).Should(BeTrue())
+			bundleResourceKey := bundle.Status.ResourceKey[0]
```
weyfonk, line 101: Same remark as above. It would make sense to have this next to the line defining `bdResource`, and after what is currently L104 (the check on `ResourceKey` not being nil/empty).

integrationtests/controller/bundle/status_test.go
```diff
@@ -70,25 +70,42 @@ var _ = Describe("Bundle Status Fields", func() {
...
 			Expect(bundle.Status.Summary.DesiredReady).To(Equal(1))
 			Expect(bundle.Status.Display.ReadyClusters).To(Equal("1/1"))
+			Expect(bundle.Status.ResourceKey).ToNot(BeNil())
```
weyfonk, line 104:
```suggestion
			Expect(len(bundle.Status.ResourceKey)).To(Equal(1))
```

weyfonk (commented):
Thanks for adding a test case! Leaving a few comments and suggestions.

internal/cmd/controller/reconciler/bundle_status.go
rubhanazeem: Agree. This is just to log when bundle deployment is skipped. I've applied the suggested changes

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -165,11 +163,7 @@ func (r *BundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		return ctrl.Result{}, err
 	}
 
-	if bundle.Status.ObservedGeneration != bundle.Generation {
```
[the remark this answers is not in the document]
rubhanazeem: This check is not needed anymore because before adding new changes the values in `bundle.Status.ResourceKey` were populated by running helm template. This condition prevents running `setResourceKey` multiple times. We need this method to run everytime the bundle reconciler runs because `bundleDeployment.Status.Resources` update retriggers bundle reconciler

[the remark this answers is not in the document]
weyfonk: I see, now that we rely on actual resources from bundle deployments, we need to update resource keys regardless of whether the bundle itself has changed. Thanks for clarifying.

weyfonk (approved):

PR #2537, opened 2024-06-19
Merges gitrepo and gitops controllers

0xavi0:
Moves all the functionality implemented in the `gitrepo` controller to the `gitopts` controller.

Refers to: https://github.com/rancher/fleet/issues/2513


internal/cmd/controller/operator.go
```diff
@@ -121,21 +121,6 @@ func start(
 
 	sched := quartz.NewStdScheduler()
 
-	if !disableGitops {
```
manno, line 124: We can also remove the variable, command line argument and switch in the deployment.yaml.
All git ops functionality is now in the gitjob deployment?

0xavi0: Stuff related with disabling gitops in `fleet-controller` deleted.

internal/cmd/controller/finalizeutil/finalizeutil.go
```diff
@@ -0,0 +1,153 @@
+package finalizeutil
```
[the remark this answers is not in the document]
0xavi0: changed as suggested

internal/cmd/controller/finalizeutil/finalizeutil.go
```diff
@@ -0,0 +1,153 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+const (
+	GitRepoFinalizer          = "fleet.cattle.io/gitrepo-finalizer"
+	BundleFinalizer           = "fleet.cattle.io/bundle-finalizer"
+	BundleDeploymentFinalizer = "fleet.cattle.io/bundle-deployment-finalizer"
+)
+
+func PurgeBundles(ctx context.Context, c client.Client, gitrepo types.NamespacedName) error {
```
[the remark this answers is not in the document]
0xavi0: yup, I've added some documentation

internal/cmd/controller/grutil/status.go
```diff
@@ -99,6 +107,64 @@ func SetStatusFromBundleDeployments(ctx context.Context, c client.Client, gitrep
...
+			for _, podStatus := range podList.Items[len(podList.Items)-1].Status.ContainerStatuses {
+				if podStatus.Name != "step-git-source" && podStatus.State.Terminated != nil {
+					terminationMessage += podStatus.State.Terminated.Message
+				}
+			}
+		}
+	}
+
+	gitRepo.Status.GitJobStatus = result.Status.String()
+	gitRepo.Status.ObservedGeneration = gitRepo.Generation
```
weyfonk, line 147: Do we deliberately not set the `UpdateGeneration` here? If so, why?

internal/cmd/controller/grutil/status.go
```diff
@@ -99,6 +107,64 @@ func SetStatusFromBundleDeployments(ctx context.Context, c client.Client, gitrep
 	return nil
 }
 
+// SetStatusFromGitjob sets the status fields relative to the given job in the gitRepo
+func SetStatusFromGitjob(ctx context.Context, c client.Client, gitRepo *fleet.GitRepo, job *batchv1.Job) error {
```
weyfonk, line 111: This function's body seems to mostly come from `updateStatus` in `gitjob_controller.go`, which is no longer called and should then be deleted.

weyfonk (commented):
Nice one! :sunglasses:

e2e/single-cluster/status_test.go
```diff
@@ -87,44 +97,46 @@ var _ = Describe("Checks status updates happen for a simple deployment", func()
...
-				}
-
-				expectedReadyBD := "\"readyBundleDeployments\":\"0/0\""
-				if !strings.Contains(out, expectedReadyBD) {
-					return fmt.Errorf("expected %q not found in %q", expectedReadyBD, out)
-				}
-
-				return nil
-			}).ShouldNot(HaveOccurred())
+			Eventually(func(g Gomega) {
```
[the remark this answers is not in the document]
0xavi0: I think it was because I had issues like the one described in https://github.com/aws/amazon-vpc-cni-k8s/issues/2495

I changed it to: 
```go
Eventually(func(g Gomega) {
     out, err := k.Delete("bundle", "my-gitrepo-helm-verify", "-n", "fleet-local")
     g.Expect(err).ToNot(HaveOccurred(), out)
}).Should((Succeed()))
```

and it also works. 
So that `Eventually` is no longer nested.

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -365,6 +512,15 @@ func (r *GitJobReconciler) newGitJob(ctx context.Context, obj *v1alpha1.GitRepo)
 		},
 		Spec: *jobSpec,
 	}
+	// if the repo references a shard, add the same label to the job
+	// this avoids call to Reconcile for controllers that are not matching
```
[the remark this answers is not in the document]
0xavi0: changed as suggested

internal/cmd/controller/grutil/status.go
0xavi0: I moved that to the place where it is checked:

```go
func (r *GitJobReconciler) deleteJobIfNeeded(ctx context.Context, gitRepo *v1alpha1.GitRepo, job *batchv1.Job) error {
	logger := log.FromContext(ctx)
	// if force delete is set, delete the job to make sure a new job is created
	if gitRepo.Spec.ForceSyncGeneration != gitRepo.Status.UpdateGeneration {
		gitRepo.Status.UpdateGeneration = gitRepo.Spec.ForceSyncGeneration
		logger.Info("job deletion triggered because of ForceUpdateGeneration")
		if err := r.Delete(ctx, job, client.PropagationPolicy(metav1.DeletePropagationBackground)); err != nil && !errors.IsNotFound(err) {
			return err
		}
	}
```

I think it makes more sense to set the value where it is checked so the logic is at the same scope. 

In fact, I did the same for `gitRepo.Status.ObservedGeneration = gitRepo.Generation`
I moved it to the place where it is checked also.

This way the `SetStatusFromGitjob` function sets only fields related to the Gitjob.

internal/cmd/controller/grutil/status.go
0xavi0: yup, deleted. I wonder why `golang-ci` didn't complain... Maybe we should enable more linters

weyfonk (approved):
LGTM :rocket:

manno (approved):

PR #2538, opened 2024-06-19
Added Mac support for the sed command

Tommy12789:
No ID 

Added mac support to the upgrade fleet in rancher script


manno:
We don't run the CI on Mac, no need to make it more complicated by adding another `if` switch to the scripts.
 
You can solve this on your machine, e.g. by setting the path in an .envrc file:
```
export PATH="/opt/homebrew/opt/gnu-sed/libexec/gnubin:$PATH"
export PATH="/opt/homebrew/opt/python-yq/bin:$PATH"
export PATH="/opt/homebrew/opt/grep/bin/ggrep:$PATH"
```

PR #2539, opened 2024-06-19
Fix flaky e2e tests

weyfonk:
This is an attempt to fix flaky end-to-end tests and ease troubleshooting of failures detected here and there.

### DIFF
--- a/e2e/drift/drift_test.go
+++ b/e2e/drift/drift_test.go
@@ -2,6 +2,7 @@ package examples_test
 
 import (
 	"encoding/json"
+	"fmt"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
@@ -50,10 +51,12 @@ var _ = Describe("Drift", Ordered, func() {
 				kw := k.Namespace(namespace)
 				out, err := kw.Patch(
 					"service", "nginx-service",
+					"-o=json",
 					"--type=json",
 					"-p", `[{"op": "replace", "path": "/spec/externalName", "value": "modified"}]`,
 				)
 				Expect(err).ToNot(HaveOccurred(), out)
+				GinkgoWriter.Print(out)
 			})
 
 			It("Bundle is modified", func() {
@@ -81,10 +84,12 @@ var _ = Describe("Drift", Ordered, func() {
 				kw := k.Namespace(namespace)
 				out, err := kw.Patch(
 					"service", "nginx-service",
+					"-o=json",
 					"--type=json",
 					"-p", `[{"op": "replace", "path": "/spec/externalName", "value": "modified"}]`,
 				)
 				Expect(err).ToNot(HaveOccurred(), out)
+				GinkgoWriter.Print(out)
 			})
 
 			It("Drift is corrected", func() {
@@ -107,10 +112,12 @@ var _ = Describe("Drift", Ordered, func() {
 				kw := k.Namespace(namespace)
 				out, err := kw.Patch(
 					"deployment", "nginx-deployment",
+					"-o=json",
 					"--type=json",
 					"-p", `[{"op": "replace", "path": "/spec/template/spec/containers/0/image", "value": "nginx:modified"}]`,
 				)
 				Expect(err).ToNot(HaveOccurred(), out)
+				GinkgoWriter.Print(out)
 			})
 
 			It("Drift is corrected", func() {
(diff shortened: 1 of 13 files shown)


e2e/single-cluster/gitrepo_polling_disabled_test.go
```diff
@@ -67,7 +68,17 @@ var _ = Describe("GitRepoPollingDisabled", Label("infra-setup"), func() {
 			"fleet-controller",
 		)
 		Expect(err).ToNot(HaveOccurred())
-		Expect(out).ToNot(ContainSubstring("ERROR"))
+
+		// Errors about bundles or bundle deployments not being found at deletion time should be ignored.
+		isError, err := regexp.MatchString(
```
manno, line 73: I had trouble understanding this until I looked at the commit message. I still think the code comment here doesn't match the commit message?

> Checking logs for a single `ERROR` keyword may lead to false positives,
for instance caused by other test suites. Instead, this uses a regular
expression to look for logs related to missing bundles or bundle
deployments.


Maybe getting rid of one more error check with `regexp.MustCompile` or:

```
Expect(out).ToNot(MatchRegexp(`ERROR.*Reconciler error.*Bundle(Deployment)?.fleet.cattle.io \\".*\\" not found`))
```

e2e/single-cluster/gitrepo_test.go
```diff
@@ -87,7 +89,17 @@ var _ = Describe("Monitoring Git repos via HTTP for change", Label("infra-setup"
...
 		Expect(err).ToNot(HaveOccurred())
-		Expect(out).ToNot(ContainSubstring("ERROR"))
+
+		// Errors about bundles or bundle deployments not being found at deletion time should be ignored.
+		isError, err := regexp.MatchString(
+			`ERROR.*Reconciler error.*Bundle(Deployment)?.fleet.cattle.io \\".*\\" not found`,
+			out,
+		)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(isError).To(BeFalse())
```
manno, line 99:
```suggestion
		Expect(out).ToNot(MatchRegexp(`ERROR.*Reconciler error.*Bundle(Deployment)?.fleet.cattle.io \\".*\\" not found`))
```

e2e/drift/drift_test.go
```diff
@@ -32,13 +33,24 @@ var _ = Describe("Drift", Ordered, func() {
...
+			if r := recover(); r != nil {
+				bundle := getBundle(bundleName, k)
+				GinkgoWriter.Printf("bundle status: %v", bundle.Status)
+			}
+		}()
 	})
 
 	AfterEach(func() {
-		out, err := k.Namespace(env.Namespace).Delete("-f", testenv.AssetPath(asset))
+		out, err := k.Namespace(env.Namespace).Delete("-f", testenv.AssetPath(asset), "--wait")
```
manno, line 46: We are waiting on some namespace deletions, but not on others? I guess this corresponds to random/static name, or is it by usage?

e2e/single-cluster/imagescan_test.go
```diff
@@ -279,9 +279,23 @@ func tagAndPushImage(baseImage, image, tag string) string {
...
+		}
+		GinkgoWriter.Printf(
+			"failed to 'docker pull %s' (attempt #%d): %w; output: %s",
+			i,
+			baseImage,
+			err,
+			string(out),
+		)
+
+		time.Sleep(100 * time.Millisecond)
```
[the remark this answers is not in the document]
weyfonk: So, the idea here is to work around transient network issues which may cause failures to pull `k8s.gcr.io/pause`, eg. [here](https://github.com/rancher/fleet/actions/runs/9658017571/job/26638371373).
Apart from rewriting these tests to rely on a different image, I don't know of another check we can run to work around sleeping between retries. We `break` the loop once a `pull` is successful.

e2e/drift/drift_test.go
weyfonk: In this case, we wait for complete deletion of a `gitrepo`, to be absolutely sure that no deletion will still be in progress while subsequent test cases run, to limit possible conflicts/interferences between test cases. The idea is to prevent failures like [this one](https://github.com/rancher/fleet/actions/runs/9655199796/job/26630714091), with bundles sharding names between test cases.

e2e/single-cluster/gitrepo_polling_disabled_test.go
weyfonk: Thanks, comment fixed and regex check simplified.

e2e/drift/drift_test.go
manno: Ah, we can't use a random namespace, because the cluster must be in the same namespace.

e2e/single-cluster/imagescan_test.go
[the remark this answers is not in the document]
manno: Ah, now I see. But then I'd wait  longer. Maybe 10s for the network to be fine?
I think `Eventually(ACTUAL, (TIMEOUT), (POLLING_INTERVAL), (context.Context)).Should(MATCHER)` would also work?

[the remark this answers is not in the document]
weyfonk: Much better indeed :) Suggestion applied, thanks.

manno (approved):

Issue #2520, opened 2024-06-14
Problems with Gitrepos with wrong URLs


mmartin24:
### Issue:
Some logs and performance degradation seem to occur on  Rancher `v2.9-e0af6cccbff0210e1538004dfc67f10f40597c20-head`  with  `fleet:v0.10.0-rc.15` when creating a GitRepo with spaces in URL and trying to fix. 
There are more then one issue here, but I try to summarize as this: 

- Issue 1: Error Status not shown when adding an URL with space; Insted permanent `Git Updating`.
- Issue 2: When correcting the URL, it takes more time than in `Rancher v2.8.4` with `fleet:103.1.5+up0.9.5`to settle things.

This can be easier visualized as in this video of rancher `v2.9-e0af6cccbff0210e1538004dfc67f10f40597c20-head` | `fleet:v0.10.0-rc.15` :

https://github.com/rancher/fleet/assets/37271841/3968d434-5828-41e3-b8ee-c34017bd5b5f

As opposed to this one in `v2.8.4` | ` fleet:103.1.5+up0.9.5` where things look smoother:

https://github.com/rancher/fleet/assets/37271841/03edfffb-366c-44b4-a6ca-d7c927f08411

Attaching logs of gitjob: [gitjob_logs.json](https://github.com/user-attachments/files/15837756/gitjob_logs.json)

<details>
<summary>Logs summary here</summary>

{"level":"error","ts":"2024-06-14T08:01:43Z","logger":"git-latest-commit-poll-watch","msg":"error fetching commit","gitrepo":{"kind":"GitRepo","apiVersion":"fleet.cattle.io/v1alpha1","metadata":{"name":"test29","namespace":"fleet-local","uid":"234426b7-3e2c-4e4d-9a │
│ {"level":"error","ts":"2024-06-14T08:01:58Z","logger":"git-latest-commit-poll-watch","msg":"error fetching commit","gitrepo":{"kind":"GitRepo","apiVersion":"fleet.cattle.io/v1alpha1","metadata":{"name":"test29","namespace":"fleet-local","uid":"234426b7-3e2c-4e4d-9a │
│ {"level":"info","ts":"2024-06-14T08:02:01Z","logger":"gitjob","msg":"job deletion triggered because of generation change","controller":"gitrepo","controllerGroup":"fleet.cattle.io","controllerKind":"GitRepo","GitRepo":{"name":"test29","namespace":"fleet-local"},"na │
[...]

PR #2542, opened 2024-06-20
Fixes error display and improves user experience in git poller

0xavi0:
When the gitpoller job returns an error it is added to the `gitrepo` conditions so it can be properly displayed by the UI.

It also reschedules the gitpoller job when the `gitrepo` spec is changed so the job is immediately executed.

Refers to: https://github.com/rancher/fleet/issues/2520

### DIFF
--- a/pkg/git/poll/handler.go
+++ b/pkg/git/poll/handler.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	v1alpha1 "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
+	"github.com/rancher/fleet/pkg/durations"
 	"github.com/rancher/fleet/pkg/git"
 	"github.com/reugn/go-quartz/quartz"
 
@@ -77,9 +78,13 @@ func (h *Handler) AddOrModifyGitRepoPollJob(ctx context.Context, gitRepo v1alpha
 			return
 		}
 		previousInterval := gitRepoPollJob.GitRepo.Spec.PollingInterval
+		previousGeneration := gitRepoPollJob.GitRepo.Generation
 		gitRepoPollJob.GitRepo = gitRepo
-		if gitRepo.Spec.PollingInterval != previousInterval {
-			// ignoring the error because we're going to schedule immediately
+		if (previousGeneration != gitRepo.Generation) ||
+			!durations.Equal(gitRepo.Spec.PollingInterval, previousInterval) {
+			// Spec or polling interval changed
+			// Reschedule so the job is immediately executed
+			// (otherwise it'll wait until next timeout)
 			_ = h.scheduler.DeleteJob(gitRepoPollKey)
 			h.scheduleJob(ctx, gitRepoPollKey, gitRepo, true)
 		}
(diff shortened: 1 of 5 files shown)


pkg/git/poll/handler.go
```diff
@@ -77,9 +78,13 @@ func (h *Handler) AddOrModifyGitRepoPollJob(ctx context.Context, gitRepo v1alpha
 			return
 		}
 		previousInterval := gitRepoPollJob.GitRepo.Spec.PollingInterval
+		previousGeneration := gitRepoPollJob.GitRepo.Generation
 		gitRepoPollJob.GitRepo = gitRepo
-		if gitRepo.Spec.PollingInterval != previousInterval {
-			// ignoring the error because we're going to schedule immediately
+		if (previousGeneration != gitRepo.Generation) ||
+			(!reflect.DeepEqual(gitRepo.Spec.PollingInterval, previousInterval)) {
```
weyfonk, line 84: Why do we need to switch to using `reflect.DeepEqual` here? Did the previous, simpler comparison not work as expected?

pkg/git/poll/handler_test.go
```diff
@@ -164,6 +167,28 @@ var _ = Describe("Gitrepo pooling tests", func() {
 				fetcher.EXPECT().LatestCommit(gomock.Any(), gomock.Any(), gomock.Any()).Return("commit", nil).Times(1)
 			},
 		),
+
+		Entry("gitrepo present, same polling interval, disablePolling=false, generation changed",
+			10*time.Second, false, true, func(gitRepo v1alpha1.GitRepo, pollingInterval time.Duration) {
+				gitRepoCopy := gitRepo
+				gitRepoCopy.Spec.PollingInterval = &metav1.Duration{Duration: 10 * time.Second}
```
weyfonk, line 174: If we're in a `same polling interval` case, does this line really belong here?

pkg/git/poll/handler.go
0xavi0: When working on https://github.com/rancher/fleet/pull/2537 I started to get weird errors when applying repos.
It was continuously creating the job, and it was because of that comparison.

aruiz14: Because that field is a pointer:
```
type GitRepoSpec struct {
    ...
	// PollingInterval is how often to check git for new updates.
	PollingInterval *metav1.Duration `json:"pollingInterval,omitempty"`
```
I share @weyfonk's concerns here, I prefer avoiding reflection when possible. Unfortunately, unlike `metav1.Time`, `metav1.Duration` does not provide an `Equal` method, but we could use it as inspiration to implement our own:
https://github.com/kubernetes/apimachinery/blob/v0.30.2/pkg/apis/meta/v1/time.go#L74-L83

pkg/git/poll/handler_test.go
0xavi0: It doesn't break the test because the `gitRepo` passed from `DescribeTable` has already the same duration, but, better to delete the line for clarity. 
Line deleted.

pkg/git/poll/handler.go
0xavi0: I've added an `Equal` function to compare pointers to `v1.Duration` in the `durations` package.
I thought it makes sense to have that accesible to other packages

weyfonk (approved):

PR #2553, opened 2024-06-25
Refactor reconcilers and remove foldin func

manno:
### Move SetupWithManager before Reconcile

We often lookup what triggers a reconcile, though receiver funcs are not
alphabetically sorted anymore, this should be easier to find now.

### Inline foldInDeployments

Targets returns a list of targets and an error.
foldInDeployments cleverly mutated the returned list. No more.


0xavi0 (approved):
LGTM

Issue #2171, opened 2024-02-22
SURE-7666


kkaempf:
Just a placeholder for SURE-7666

kkaempf:
closing

PR #2556, opened 2024-06-26
Extend strict TLS mode to `fleet-agent` container

weyfonk:
Strict TLS mode must be applied not only in the `fleet-agent-register` init container, but also in the main agent container itself, which also reaches out to the upstream cluster.

Refers to #2171.
Follow-up to #2507.

### DIFF
--- a/internal/cmd/agent/register/register.go
+++ b/internal/cmd/agent/register/register.go
@@ -59,6 +59,24 @@ func Get(ctx context.Context, namespace string, cfg *rest.Config) (*AgentInfo, e
 		return nil, err
 	}
 
+	configMap, err := k8s.Core().V1().ConfigMap().Get(namespace, config.AgentConfigName, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		logrus.Warnf("Cannot find %s config map", config.AgentConfigName)
+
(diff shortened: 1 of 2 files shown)


internal/cmd/agent/register/register.go
```diff
@@ -59,6 +59,24 @@ func Get(ctx context.Context, namespace string, cfg *rest.Config) (*AgentInfo, e
...
+
+		return nil, err
+	}
+
+	agentConfig, err := config.ReadConfig(configMap)
+	if err != nil {
+		logrus.Warnf("Cannot parse %s config map into agent config", config.AgentConfigName)
+
+		return nil, err
+	}
```
manno, line 74: `runRegistration` loads it like this:

```suggestion
        cfg, err := config.Lookup(ctx, namespace, config.AgentConfigName, k8s.Core().V1().ConfigMap())
	if err != nil {
		return nil, fmt.Errorf("failed to look up client config %s/%s: %w", namespace, config.AgentConfigName, err)
	}

```

weyfonk: That's much better :) Thanks, done.

manno (approved):

Issue #2492, opened 2024-06-06
Submodules aren't cloned recursively anymore since Fleet 0.9.x


puffitos:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

After updating to Rancher 2.8.x (and thus getting Fleet 0.9.x), we noticed that some gitrepos which dependend on submodules being recursively cloned as well don't get updated anymore. The Gitjob pods fail with an error in the lines of "failed to download <path-to-local-submodule>".

This happens apparently, because the new `gitjob gitcloner` command that took the place of the tekton containers doesn't clone recursively:

https://github.com/rancher/gitjob/blob/1495df9fee27de43ec8104faece17028a96c7459/cmd/gitcloner/gogit/cloner.go#L71C1-L83C1

This results in our bundles being stuck and having to apply some workaround (like moving our charts/manifests from the local folder to a remote location - which is OK for the time being).

### Expected Behavior

The functionality of recursively cloning submodules is still present in the current fleet version.

### Steps To Reproduce

1. Install Fleet 0.8.x (before gitjob became the default cloner for gitrepos)
2. Deploy a Gitrepo that targets a repository that uses submodules, which include a helm chart (for example, just add a the https://github.com/bitnami/charts as a submodule and instruct fleet to deploy the `chart: ./charts/bitnami/nginx/` chart)
3. See the pod get deployed
4. Update to fleet 0.9.x
5. Fleet gitjob for the GitRepo should fail after a force update, because submodule isn't pulled and cannot be found

### Environment

```markdown
- Architecture: amd64
- Fleet Version: 0.9.2
- Cluster: 
  - Provider: seems irrelevant
  - Options: seems irrelevant
  - Kubernetes Version: 1.27.13
```

sbulage:
QA Template followed: https://github.com/rancher/fleet/pull/2557#issue-2375658491

sbulage:
System Information| Before Upgrade | After Upgrade |
|--------|--------|--------|
Rancher Version| 2.8.5 | 2.9.0-alpha7 |
Fleet Version| 0.9.5 | 0.10.0-rc.18 | 

---

### **Steps Performed:**
**Before upgrade.**
1. Created GitRepo using @manno's [fleet-experiments](https://github.com/manno/fleet-experiments) with path: `fleet-test-data/qa-test-apps/nginx-app`.
2. Wait for the resources to be deployed.
[...]

PR #2557, opened 2024-06-26
Gitcloner recurses submodules

manno:
Refers to https://github.com/rancher/fleet/issues/2492

Pull request content for Rancher QA verification.

### Problem

Since switching to go-git, Fleet would no longer clone submodules.
 
### Solution

According to the go-git documentation this is how to enable submodule cloning.
 
#### Manual Testing

Tested with fleet-test-data submodule in https://github.com/manno/fleet-experiments

```
kind: GitRepo
metadata:
  name: crd-experiment
spec:
  repo: https://github.com/manno/fleet-experiments
  branch: main
  paths:
    - fleet-test-data/simple-chart
  targets:
    - clusterSelector: {}
    
 ```

### QA Testing Considerations

We would need another repo like rancher/fleet-test-data to include as a submodule.

Too many nested repos are a problem for performance.

### DIFF
--- a/internal/cmd/agent/register/register.go
+++ b/internal/cmd/agent/register/register.go
@@ -326,6 +326,7 @@ func createClientConfigFromSecret(secret *corev1.Secret, trustSystemStoreCAs boo
 	token := string(data[Token])
 
 	if trustSystemStoreCAs { // Save a request to the API server URL if system CAs are not to be trusted.
+		// NOTE(manno): client-go will use the system trust store even if a CA is configured. So, why do this?
 		if _, err := http.Get(apiServerURL); err == nil {
 			apiServerCA = nil
 		}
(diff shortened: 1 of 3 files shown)


weyfonk (commented):
Do we want tests for this, eg. integration tests? This would help us detect any regression here.

Otherwise LGTM.

manno:
> Do we want tests for this, eg. integration tests? This would help us detect any regression here.

We would need another git repo, so we have one with a submodule. I though about adding fleet-example to fleet-test-data, but that would increase the clone time for all tests 🤷 
I guess, I could argue the actual cloning is done by go-git and our unit tests make sure we pass the right parameter to go-git.

weyfonk:
> the actual cloning is done by go-git and our unit tests make sure we pass the right parameter to go-git.

This is good enough for me :+1:

Issue #2464, opened 2024-05-28
Improve Content resources cleanup


weyfonk:
This is a follow-up to #1980.

With finalizers in place for bundle deployments, we have one additional reconcile loop available to run cleanup tasks when a bundle deployment is deleted.
This opportunity could be used to act on Content resources, preventing their batch deletion by our [cleanup controller](https://github.com/rancher/fleet/blob/main/internal/cmd/controller/cleanup/content/purge.go).

Possible approaches to baking this logic into the bundle deployment reconciler:
**A.** When deleting a bundle deployment, check referenced Content resources and, for each Content resources, check how many bundle deployments reference it (might be expensive/cumbersome): if no other bundle deployment than the one we are deleting references it anymore, then the Content resource is safe to delete. We could increase timer for the content purge logic, or make it configurable, since garbage collection now happens as needed.
    * This approach would be backwards compatible

**B.** Add finalizers to Content resources, eg. one per bundle deployment referencing it? A given finalizer would be deleted when the bundle deployment it refers to is itself deleted. These finalizers could act as reference counts, enabling a Content resource to be deleted as soon as its finalizers have all been deleted.

- [x] Implement new Content resources cleanup logic
- [x] Remove Content purge logic linked above

manno:
Moving to the next milestone, because of the ongoing code freezes.

PR #2566, opened 2024-07-01
Clean up content resources through finalizers

weyfonk:
Content resources may be used by one or more bundle deployments, which now each add their own finalizer on a content resource to indicate their reliance on it, and delete that finalizer when that reliance no longer applies.

This eliminates the need for batch cleanup of content resources.

Refers to #2464

Open points:
- should the edge case of a `GitRepo` transitioning from etcd- to OCI-stored contents be automatically tested here, or could this be taken into account through QA efforts?
- is this backwards compatible? Pending Fleet upgrade tests on my end, hence leaving this in draft state in the meantime.


internal/cmd/controller/finalize/finalize.go
```diff
@@ -96,6 +98,47 @@ func PurgeBundleDeployments(ctx context.Context, c client.Client, bundle types.N
...
+		}
+	}
+
+	hasOtherFinalizers := (len(content.Finalizers) > 1) ||
+		((len(content.Finalizers) == 1) && content.Finalizers[0] != name)
+	if !hasOtherFinalizers {
+		logrus.Infof("Deleting content resource %s", contentID)
+		if err := c.Delete(ctx, content); err != nil {
+			return err
+		}
```
0xavi0, line 141: Question: why checking exactly for 1 finalizer? What is the other finalizer that the Content resource has? 
Should not be deleted when number of finalizers reaches 0?

0xavi0 (commented):
LGTM, just 1 question regarding the check before deleting the `Content` resource

internal/cmd/controller/finalize/finalize.go
weyfonk: An alternative approach to this could be to eagerly mark a Content resource for deletion, as soon as it is referenced by at least one bundle deployment even when we therefore know that deleting it doesn't make sense. _That_ would allow it to be transparently deleted as soon as its finalizers count reaches 0.

0xavi0 (approved):
Looks good to me

0xavi0 (approved):

PR #2574, opened 2024-07-02
Fix linting errors

weyfonk:
This refactors some code [deemed](https://github.com/rancher/fleet/actions/runs/9691013848/job/26741822654) too complex by `gocyclo`.

### DIFF
--- a/internal/cmd/controller/reconciler/bundle_controller.go
+++ b/internal/cmd/controller/reconciler/bundle_controller.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/go-logr/logr"
 	"github.com/rancher/fleet/internal/cmd/controller/finalize"
 	"github.com/rancher/fleet/internal/cmd/controller/summary"
 	"github.com/rancher/fleet/internal/cmd/controller/target"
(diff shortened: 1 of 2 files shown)


internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -249,35 +255,10 @@ func (r *BundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
-
-		updated := bd.DeepCopy()
-		op, err := controllerutil.CreateOrUpdate(ctx, r.Client, bd, func() error {
-			bd.Spec = updated.Spec
-			bd.Labels = updated.GetLabels()
-			return nil
-		})
-		if err != nil && !apierrors.IsAlreadyExists(err) {
-			logger.Error(err, "Reconcile failed to create or update bundledeployment", "bundledeployment", bd, "operation", op)
+		bd, err := r.createBundle(ctx, logger, target, contentsInOCI)
```
manno, line 258: Since createBundle can return a nil bd, do we still need the `continue`. Apparently it's only passed to createDeploymentOCISecret, so if that can handle nil, we probably don't need `continue`..?

internal/cmd/controller/reconciler/cluster_controller.go
```diff
@@ -145,7 +145,7 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		for _, bd := range bundleDeployments.Items {
 			if bd.Labels[fleet.BundleLabel] == bundle.Name && bd.Labels[fleet.BundleNamespaceLabel] == bundle.Namespace {
 				logger.V(1).Info("cleaning up bundleDeployment not matching the cluster", "bundledeployment", bd)
-				err := r.Delete(ctx, &bd)
+				err := r.Delete(ctx, &bd) // nolint: gosec // does not store pointer
```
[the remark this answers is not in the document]
weyfonk: Correct. Fixed my local setup and reverted that commit, thanks.

internal/cmd/controller/reconciler/bundle_controller.go
weyfonk: What I do see now from reading your comment and thinking about it is that since `createBundle` will return a `nil` bundle with a `nil` error in cases where we would previously `continue`, it would make sense to add `bd != nil` to the condition on L263, to avoid calling `createDeploymentOCISecret` in such cases.
While `createDeploymentOCISecret` can perhaps handle `nil` bundle deployments, it may error when getting a secret or setting ownership of a secret on a `nil` bundle, so it would make sense to avoid that.

How close is this to what you meant?

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -249,35 +255,10 @@ func (r *BundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+		if err != nil {
 			return ctrl.Result{}, err
 		}
-		logger.V(1).Info(upper(op)+" bundledeployment", "bundledeployment", bd, "operation", op)
 
 		if contentsInOCI {
```
manno, line 263:
```suggestion
		if bd != nil && contentsInOCI {
```

manno (approved):

Issue #2138, opened 2024-02-09
[SURE-7572] gitjob robustness problem: connectivity problem that ends up in a lock not being released


kkaempf:
### Issue description:
gitjob pod is still running but seems to be in a hung state showing this error:

time="2024-01-25T17:45:23Z" level=fatal msg="leaderelection lost for gitjob"

Since that timestamp, none of the updates in their Git Repos are being pushed to configured Rancher managed clusters. And new Git Repos are not being pushed to configured Rancher managed clusters.

### Business impact:
This prevented Fleet CD from pushing any of the updates their devs were making in their Git Repos

### Workaround:
Is workararound available and implemented? yes/no
What is the workaround: restart the gitjob pod

### Actual behavior:
gitjob pod hangs  with a a fatal leader election lost error and breaks fleet CD

### Expected behavior:
If the gitjob pod has a fatal leader election lost issue, it should restart, not hang.

### Files, logs, traces:
see JIRA

manno:
Gitjob uses an old lease mechanism, called "ConfigMapsLeasesResourceLock". It consists of a lease resource and a config map.

After checking that there is exactly one gitjob controller running, the lease resource can be deleted. Leader election should renew the lock [within the minute](https://github.com/rancher/wrangler/blob/8a1858d3883c7e0eed81204b614397472449627f/pkg/leader/leader.go#L53C1-L55C34).


Check the output of `kubectl get lease -A` for possible lease resources.

manno:
In this case the gitjob controller was unable to reach the API server.

```


E0125 17:40:57.357995       7 leaderelection.go:330] error retrieving resource lock cattle-fleet-system/gitjob: Get "https://192.168.0.1:443/api/v1/namespaces/cattle-fleet-system/configmaps/gitjob": dial tcp 192.168.0.1:443: connect: connection refused
```

The controller must be able to reach the k8s API to do any leaderelection.

dnoland1:
@manno Agree, but shouldn't the controller either retry or exit instead of hang indefinitely?

manno:
Let's see if we can exit after a  long time.

PR #2576, opened 2024-07-02
Re-enable gitjob leader election

p-se:
Part of #2138


internal/cmd/controller/options/leaderelection.go
```diff
@@ -0,0 +1,50 @@
+package options
```
manno, line 1: That options package is for bundledeployment options, I think.
> Package options merges the BundleDeploymentOptions, so that targetCustomizations take effect.

internal/cmd/controller/options/leaderelection.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"fmt"
+	"os"
+	"time"
+)
+
+type LeaderElectionOptions struct {
```
manno, line 9: This can probably be simplified to `LeaderElection`.

> All references to names in your package will be done using the package name, so you can omit that name from the identifiers. For example, if you are in package chubby, you don’t need type ChubbyFile, which clients will write as chubby.ChubbyFile. Instead, name the type File, which clients will write as chubby.File. Avoid meaningless package names like util, common, misc, api, types, and interfaces. See https://go.dev/doc/effective_go#package-names and https://go.dev/blog/package-names for more.

https://go.dev/wiki/CodeReviewComments#package-names

manno: Maybe just put this next to https://github.com/rancher/fleet/blob/main/internal/cmd/debug.go in the cmd package?

manno (changes requested):

internal/cmd/controller/options/leaderelection.go
p-se: > go.dev/wiki/CodeReviewComments

Nice document!

p-se: > Maybe just put this next to [`main`/internal/cmd/debug.go](https://github.com/rancher/fleet/blob/main/internal/cmd/debug.go?rgh-link-date=2024-07-02T16%3A04%3A53Z) in the cmd package?

Yeah, that was the other thought. I put it there.

manno (approved):

Issue #2586, opened 2024-07-03
Bundle is not removed after removing GitRepo configured with multiple paths


sbulage:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

After deleting GitRepo pointing to multiple paths(Service and ConfigMap), configmap getting removed but service is not being removed instead it is re-created.
Also, observed that bundle not removed after deleting GitRepo, may causing this issue. Same service app is created under Apps which is being re-created again and again.

### Expected Behavior

Once GitRepo removed no leftover should be present.

### Steps To Reproduce

1. Create a GitRepo which is pointing to multiple paths. (I used: https://github.com/rancher/fleet-test-data and paths: multiple-paths)
2. Wait for applications to deployed on each cluster (if there are more than one)
3. Now delete GitRepo
4. Check the applications are removed correctly or not.
5. Check the bundle it is still there and didn't get removed.

### Environment

```markdown
- Architecture:
- Fleet Version:fleet:v0.10.0-rc.18
- Cluster:
  - Provider:k3s
  - Options:
  - Kubernetes Version:
```

### **Logs**

<details>
<summary>GitJob Logs</summary>

```
[...]

PR #2591, opened 2024-07-03
Deletes `Bundles` after deleting all its `BundleDeployments`

0xavi0:
When testing https://github.com/rancher/fleet/issues/2586 the `Bundle` reconciler found that the `Bundle` was gone as was trying to delete the same `BundleDeployments` in parallel.

Deleting the `Bundle` after deleting all its `BundleDeployments` seems to help to prevent that race condition between both reconcilers

Refers to: https://github.com/rancher/fleet/issues/2586

### DIFF
--- a/internal/cmd/controller/finalize/finalize.go
+++ b/internal/cmd/controller/finalize/finalize.go
@@ -50,14 +50,14 @@ func PurgeBundles(ctx context.Context, c client.Client, gitrepo types.Namespaced
 	}
 
 	for _, bundle := range bundles.Items {
-		err := c.Delete(ctx, &bundle)
-		if client.IgnoreNotFound(err) != nil {
-			return err
-		}
-
 		nn := types.NamespacedName{Namespace: bundle.Namespace, Name: bundle.Name}
 		if err = PurgeBundleDeployments(ctx, c, nn); err != nil {
 			return client.IgnoreNotFound(err)
+		}
+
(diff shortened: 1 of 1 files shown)


manno (approved):

0xavi0:
Tested with the following `GitRepo`

```yaml
kind: GitRepo
metadata:
  name: sample
  namespace: fleet-local
spec:
  repo: "https://github.com/rancher/fleet-test-data"
  branch: master
  paths:
  - multiple-paths
```

and the following bash script:
```bash
#!/bin/bash

i=1

while true
do
  echo "---------------------------"
  kubectl apply -f test-2586.yaml

  OUT=""
  while [ "$OUT" != "2" ];
  do
   OUT=`kubectl get gitrepo sample -n fleet-local -o jsonpath='{.status.summary.ready}'`
  done

  kubectl delete gitrepo sample -n fleet-local
  kubectl wait --for=delete gitrepo/sample --timeout=60s -n fleet-local

  kubectl get bundles -A
  BUNDLES_LEFT=`kubectl get bundles -A | grep sample | wc -l`

  if [ "$BUNDLES_LEFT" == "       0" ];
  then
    echo "ALL GOOD attempt number: ${i}"
    echo "---------------------------"
    i=$(($i+1))
  else
    echo "FAILED"
    echo "---------------------------"
    exit 1
  fi
done
```

PR #2600, opened 2024-07-04
Extend bundle-diff tests

manno:
Needed https://github.com/rancher/fleet-test-data/pull/20

Bundle diff tests were not very effective, as the patches were missing a namespace and were never applied to the deployed resources.

Fixing several issues that came up during testing, as well as refactoring the "diff" code tree after reading it and considering to update it. "Refactor normalizers" could be in a different PR, together with that update. 🤷


internal/cmd/agent/deployer/applied/diff.go
```diff
@@ -81,9 +82,18 @@ func Diff(plan apply.Plan, bd *fleet.BundleDeployment, ns string, objs ...runtim
 	return plan, nil
 }
 
-func normalizers(live objectset.ObjectByGVK, bd *fleet.BundleDeployment) (diff.Normalizer, error) {
+// newNormalizers creates a normalizer that removes fields from resources.
+// The normalizer is composed of:
+//
+//   - StatusNormalizer
+//   - MutatingWebhookNormalizer
+//   - ValidatingWebhookNormalizer
```
weyfonk, line 90: Where do those 3 come from?

charts/fleet-agent/templates/deployment.yaml
```diff
@@ -59,6 +59,9 @@ spec:
             drop:
             - ALL
         {{- end }}
+        volumeMounts:
```
weyfonk, line 62: Why do we need this?

internal/cmd/agent/deployer/internal/resource/ignore.go
```diff
@@ -1,4 +1,4 @@
-// extracted from argoproj/argo-cd/pkg/apis/application/v1alpha1/types.go
+// +vendored argoproj/argo-cd/pkg/apis/application/v1alpha1/types.go
```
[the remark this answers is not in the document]
manno: Not yet, but it's easier to grep for and might allow us to experiment with markers: https://pkg.go.dev/sigs.k8s.io/controller-tools/pkg/markers

internal/cmd/agent/deployer/applied/diff.go
manno: Prepended by the fleet normalizer package when calling `normalizers.New` in the end:
https://github.com/rancher/fleet/blob/main/internal/cmd/agent/deployer/normalizers/norm.go#L22-L39

charts/fleet-agent/templates/deployment.yaml
manno: To get rid of:
`fleet-agent-0 fleet-agent I0705 14:38:47.514247       1 cached_discovery.go:93] failed to write cache to /.kube/cache/discovery/10.43.0.1_443/batch/v1/serverresources.json due to mkdir /.kube: permission denied?`

weyfonk (approved):

PR #2605, opened 2024-07-05
Configurable git client timeout

### LINKED ISSUE #2188: [SURE-7783] Fleet 0.9.0 having trouble with bitbucket syncing

### Current Behavior

Gitjob in cattle-fleet-system namespace is giving following error for all the gitjobs after Rancher upgrade to 2.8.2

time="2024-02-28T11:00:35Z" level=error msg="Error fetching latest commit: Get \"https://xxx/xposer.git/info/refs?service=git-upload-pack\": context deadline exceeded"

If I do force resync it becomes active but after sometime it again goes to failed state with the above error

### Expected Behavior

repo should sync without any error

### Steps To Reproduce

_No response_

### Environment

```markdown
- Architecture:
- Fleet Version: 0.9.0
- Cluster:
  - Provider: EKS
  - Options:
  - Kubernetes Version: 1.26
```

### Logs

_No response_

### Anything else?

_No response_

p-se:
Fixes #2188

Also fixes an issue in the config reconciler which has prevented the
config from being reloaded when it was changed.

### DIFF
--- a/integrationtests/gitjob/controller/controller_test.go
+++ b/integrationtests/gitjob/controller/controller_test.go
@@ -9,13 +9,6 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-	"sigs.k8s.io/controller-runtime/pkg/client"
-
-	"github.com/rancher/fleet/integrationtests/utils"
-	v1alpha1 "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
-	"github.com/rancher/wrangler/v3/pkg/genericcondition"
-	"github.com/rancher/wrangler/v3/pkg/name"
-
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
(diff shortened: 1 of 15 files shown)


integrationtests/cli/cleanup/suite_test.go
```diff
@@ -49,7 +49,7 @@ func (g *getter) GetNamespace() string {
 
 func TestFleetCLICleanUp(t *testing.T) {
 	RegisterFailHandler(Fail)
-	RunSpecs(t, "Fleet CLI Cleanup Suite")
+	RunSpecs(t, "GitJob Controller Suite")
```
manno, line 52: This seems wrong.
```suggestion
	RunSpecs(t, "Fleet CLI Cleanup")
```

integrationtests/gitjob/controller/controller_test.go
```diff
@@ -43,6 +44,9 @@ var _ = Describe("GitJob controller", func() {
 		)
 
 		JustBeforeEach(func() {
+			config.Set(&config.Config{
```
manno, line 47: This will affect all integration tests running after this one, I think? 
But that doesn't matter I guess, is this just for initialization? Instead of initializing the config in some tests, could we make the code work with an empty config instead?

integrationtests/cli/cleanup/suite_test.go
p-se: It was the other suite that needed correction.

integrationtests/gitjob/controller/controller_test.go
p-se: I've added the initialization to the suite instead of the tests where possible and initialized it empty where possible.

0xavi0 (approved):
looks good to me

Issue #2175, opened 2024-02-24
.fleetignore not being respected


skaven81:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I add a `.fleetignore` to the top level of my bundle:
```
tests/*
samples/*
suite.yaml
```

But the Bundle still contains all those files:

```
❯ kubectl -n fleet-drekar get bundle gkopa-constraints-dev-templates -o json | jq -r '.spec.resources[].name' | grep -c /tests/
140
❯ kubectl -n fleet-drekar get bundle gkopa-constraints-dev-templates -o json | jq -r '.spec.resources[].name' | grep -c /samples/
324
```

Interestingly, the `suite.yaml` files were removed, so it's at least partially working.  But the directory tree removal does not seem to work.

### Expected Behavior

As described in #1634, .fleetignore should prevent the ignored files from appearing in the Bundle resource.  The docs in #1634 say:

> A match may be found for a file _or directory_ at any level below
> the directory where a .fleetignore lives, ie foo.yaml will
> match ./foo.yaml as well as ./path/to/foo.yaml

emphasis added -- _or directory_ suggests that this should work.

### Steps To Reproduce

Create a fleet bundle that contains some subdirectories you want excluded

Add a `.fleetignore` that names those subdirectories

Examine the generated `Bundle` resource and see the directories were not excluded.

### Environment

```markdown
- Architecture: x86_64
- Fleet Version: v0.8.0 (Rancher 2.7.10)
- Cluster:
  - Provider: RKE1
  - Kubernetes Version: 1.26
```

skaven81:
It does seem that this works:
```
tests
samples
suite.yaml
```
But I would have expected the wildcard version of it to work (which would differentiate a directory named `tests` from a file named `tests`).

grbsystems:
I'm also seeing the same issue.  All the .fleetignore files are being ignored.  I have a directory with 2 bundles for nginx and neo4j.  The .fleetignore file has:

```
neo4j
nginx
```

in an effort to ignore the bundles I did not want installed.  The bundles always appear in the deployment list.

### Environment

Rancher Docker Image rancher/rancher:latest

```
- Rancher V2.8.2
- Dashboard v2.8.0
- Helm v2.16.8-rancher2
- Machine v0.15.0-rancher106
- Fleet Agent v0.9.0
```

mverleun:
Similar issue here.

A wild card at the highest level works.

```
# .fleetignore
*
```

[...]

PR #2617, opened 2024-07-08
Adds support for folder/* patterns in .fleetignore

0xavi0:
As of now, the `folder/*` was not supported by Fleet.

This PR adds support for it, giving the option to ignore folders (not files) with a given name.

For example:

This folder structure:

```
root
   dir1
     test
       file1
       file2
   file3
test
file4
```

with the following `.fleetignore` file:
```
test/*
```

would produce the following recource list:
```
root/file3
root/test
root/file4
```

`test` folder would be ignored but `test` file would not.

The difference with a `.fleetignore` file like this:
```
test
```

is that both the `test` flile and `dir1/test` folder would be ignored.

Refers to: https://github.com/rancher/fleet/issues/2175


weyfonk (approved):
Nice one :+1:

Issue #2606, opened 2024-07-05
Ensure jobs are scheduled on specified shard id through nodeSelectors


mmartin24:
Follow up on [this](https://github.com/rancher/fleet/issues/2456#issuecomment-2203507977) comment of node selection of shards.

## Issue
After defining shards through nodeSelectors, they gitjobs are correctly processed using the specified , but Jobs are scheduled randomly (details in link above)

Basically, given a label on git repo pointing to `shard-2`; namely `fleet.cattle.io/shard-ref: shard2` and this shard config...
```json
{
    "debug": true,
    "debugLevel": 1,
    "shards": [
      {
        "id": "shard0",
        "nodeSelector": {
          "kubernetes.io/hostname": "k3d-local-two-server-2-agent-agent-1"
        }
      },
      {
        "id": "shard1"
      },
      {
        "id": "shard2",
        "nodeSelector": {
          "kubernetes.io/hostname": "k3d-local-two-server-2-agent-agent-0"
        }
      }
    ]
  }
```
...the expectation is that  theJob scheduled would take place in node `k3d-local-two-server-2-agent-agent-0`. 
Instead it got scheduled on `k3d-local-two-server-2-agent-agent-1` as can be seen:

![Image](https://github.com/rancher/fleet/assets/37271841/514a0bfa-46bd-46c2-8911-a49c763cdb3e)

## Request
Pls ensure job scheduling takes place  where in the specific node lead by `nodeSelector`

sbulage:
### System Information
| Rancher Version | Fleet Version |
|--------|--------|
| v2.9-7bc90dd10e937ca1b7c03e217b884a863dcc8eb5-head | 	fleet:104.0.1+up0.10.1-rc.3 |


### Screenshots and Logs of sharding pods.
<details>
<summary><b>GitJob Shard0 Pod Logs</b></summary>

```
2024-08-14T09:56:23Z	DEBUG	gitjob	Reconciling GitRepo	{"controller": "gitrepo", "controllerGroup": "fleet.cattle.io", "controllerKind": "GitRepo", "GitRepo": {"name":"test-job-placement","namespace":"fleet-local"}, "namespace": "fleet-local", "name": "test-job-placement", "reconcileID": "b5db6d9d-9f8f-4396-9bb4-c9f3c59768fa", "generation": 2, "commit": ""}
[...]

PR #2620, opened 2024-07-08
gitjob creation on the same node as the controller shard

Tommy12789:
Refers to #2606

Create a gitrepo associated with a shard and verify the job created is on the same node than the sharded controller 

Added a way to retrieve controllers nodeSelector to put it in the gitjob specs

### DIFF
--- a/e2e/single-cluster/sharding_test.go
+++ b/e2e/single-cluster/sharding_test.go
@@ -79,26 +79,37 @@ var _ = Describe("Filtering events by shard", Label("sharding"), Ordered, func()
 
 				By("checking the gitjob pod has the same nodeSelector as the sharded controller")
 				Eventually(func() bool {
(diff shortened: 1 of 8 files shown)


e2e/single-cluster/sharding_test.go
```diff
@@ -79,27 +79,38 @@ var _ = Describe("Filtering events by shard", Label("sharding"), Ordered, func()
 
 				By("checking the gitjob pod has the same nodeSelector as the sharded controller")
 				Eventually(func() bool {
-					shardNodeSelector, err := k.Namespace("cattle-fleet-system").Get("pods", "-o=jsonpath={.items[0].spec.nodeSelector}", "-l", "app=fleet-controller", "-l", fmt.Sprintf("fleet.cattle.io/shard-id=%s", shard))
+					shardNodeSelector, err := k.Namespace("cattle-fleet-system").Get(
```
manno, line 82: nit: I'm fine with reading long wrapped lines. Especially if it's the equivalent of a shell command... 🤷

0xavi0 (approved):
lgtm, just a comment about the new test added.

0xavi0 (approved):

PR #2650, opened 2024-07-18
Fix extraEnv in gitjob controller template

manno:
Setting debug to false would put the additional env vars outside the env block.

```
% helm template --set 'extraEnv[0].name="foo"' --set 'extraEnv[0].value="bar"' fleet charts/fleet
Error: YAML parse error on fleet/templates/deployment_gitjob.yaml: error converting YAML to JSON: yaml: line 47: did not find expected key

Use --debug flag to render out invalid YAML
```

### DIFF
--- a/charts/fleet-agent/ci/default-values.yaml
+++ b/charts/fleet-agent/ci/default-values.yaml
@@ -1,0 +1,2 @@
+apiServerURL: "https://localhost"
+apiServerCA: "abc"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,15 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
+
+      -
+        name: Set up chart-testing
+        uses: helm/chart-testing-action@v2.3.1
+        with:
+          version: v3.3.0
+      -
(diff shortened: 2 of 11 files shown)


weyfonk (approved):
:eagle: :eye: !

manno:
I'm adding an initial setup for https://github.com/helm/chart-testing/blob/main/doc/ct_lint.md

charts/fleet-agent/ci/default-values.yaml
```diff
@@ -0,0 +1,2 @@
+apiServerURL: "https://localhost"
```
weyfonk, line 1: Would this file, and other files under `charts/fleet{-agent,}/ci` end up being packaged in our charts?
Should we add `.helmignore` files to those charts to prevent that?

Apart from that, I'm sligthly concerned about the maintenance overhead of those values files when we add new features involving such values: it looks likely to me that we may forget to update at least some of them to reflect the latest available values.

manno: Thanks, added those `.helmignore` files. The actual values in those files don't matter much. I think chart-testing just checks that the result is valid YAML.

weyfonk (approved):

Issue #2484, opened 2024-06-04
[SURE-8482] Misleading error message when trying to deploy cluster resources with targetNamespace


kkaempf:
### Issue description:
When deploying the sample code attached by the client (sample.zip), an error is thrown by Fleet :

~~~
Error while running post render on files: invalid cluster scoped object [name=kube-ops-view-2 kind=ClusterRole apiVersion=rbac.authorization.k8s.io/v1] found, consider using "defaultNamespace", not "namespace" in fleet.yaml
~~~

This is related to the following GitHub issue: https://github.com/rancher/fleet/issues/752.

When GitRepo.spec.targetNamespace is set, Fleet fails at creating cluster-scoped objects (as documented here: https://fleet.rancher.io/ref-crds#gitrepospec).

The error message is misleading and does not point to the actual problem when attempting to deploy cluster-wide resources and having targetNamespace present in the Git Repo.

### Repro steps:

Use the sample code (sample.zip) to create a repo
Apply the gitrepo.yaml file with the needed modifications
The error will appear on the UI in the Git Repos section of Continuous Delivery (along with a "Err Applied" state, see "Error message repo.png" for example screenshot)
Comment/remove GitRepo.spec.targetNamespace from the repo
App will be deployed without errors

### Workaround:
Is a workaround available and implemented? no

### Actual behavior:
The error message speaks about using the defaultNamespace, which is not the problem in practice.

### Expected behavior:
The error message in the UI should indicate that using targetNamespace is the cause of the deployment failure of a cluster-wide resource.

### Files, logs, traces:
See attachments for sample code and a screenshot of the error message in the UI

manno:
We should try to make the error message clearer.

PR #2660, opened 2024-07-22
Changes post render error when using targetNamespace or namespace

0xavi0:
Changes the error message when `targetNamespace` or `namespace` is specified and there are cluster wide resources to be deployed. 

I've tested also with `rancher` so we can see how it looks like in the UI (pretty large, but hopefully more explanatory) 

Screenshot 2024-07-22 at 17 39 05

Refers to: https://github.com/rancher/fleet/issues/2484

### DIFF
--- a/internal/helmdeployer/postrender.go
+++ b/internal/helmdeployer/postrender.go
@@ -99,7 +99,7 @@ func (p *postRender) Run(renderedManifests *bytes.Buffer) (modifiedManifests *by
 				}
 				if mapping.Scope.Name() == meta.RESTScopeNameRoot {
 					apiVersion, kind := gvk.ToAPIVersionAndKind()
-					return nil, fmt.Errorf("invalid cluster scoped object [name=%s kind=%v apiVersion=%s] found, consider using \"defaultNamespace\", not \"namespace\" in fleet.yaml", m.GetName(),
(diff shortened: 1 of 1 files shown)


internal/helmdeployer/postrender.go
```diff
@@ -99,7 +99,7 @@ func (p *postRender) Run(renderedManifests *bytes.Buffer) (modifiedManifests *by
 				}
 				if mapping.Scope.Name() == meta.RESTScopeNameRoot {
 					apiVersion, kind := gvk.ToAPIVersionAndKind()
-					return nil, fmt.Errorf("invalid cluster scoped object [name=%s kind=%v apiVersion=%s] found, consider using \"defaultNamespace\", not \"namespace\" in fleet.yaml", m.GetName(),
+					return nil, fmt.Errorf("invalid cluster scoped object [name=%s kind=%v apiVersion=%s] found. Your config uses targetNamespace or namespace and thus forbids cluster scoped resources. If you do not intend to disallow cluster scoped resources, you could switch to defaultNamespace", m.GetName(),
```
weyfonk, line 102: nit: we could use concatenation to make for shorter lines, but others may disagree :þ
```suggestion
					return nil, fmt.Errorf(
					        "invalid cluster scoped object [name=%s kind=%v apiVersion=%s] found. "+
					        "Your config uses targetNamespace or namespace and thus forbids cluster-scoped resources. "+
					        "If you do not intend to disallow cluster scoped resources, you could switch to defaultNamespace",
					        m.GetName(),
```

weyfonk (approved):

weyfonk (approved):

weyfonk (approved):

Issue #2441, opened 2024-05-18
Feature Request: Enable namespaceAnnotations (and namespaceLabels) as targetCustomizations 


EdKingscote:
### Is your feature request related to a problem?

#1627 introduced the ability to have namespaceLabels and namespaceAnnotations as part of the bundle deployment options.

This doesn't quite go far enough, as a single set of values can only be applied to all clusters, which means annotations and labels have to be identical across all environments.

I am specifically after namespaceAnnotations in the targetCustomizations, however I think it makes sense to look at both of these elements together to also resolve #2262 in the same feature given the similarities.

I believe this could also provide a quick path for enabling #304 via static configuration as an interim step too (which is my immediate use-case for this right now)

### Solution you'd like

Adding the necessary elements to the fleet.yaml doesn't appear to incur complaint, and from deeper analysis in the codebase, I believe the only missing piece of the jigsaw to enable this is in the [merge function](https://github.com/rancher/fleet/blob/c802a27262668d42e5a3c69c23c6dc343a65a12d/internal/cmd/controller/options/calculate.go#L24)? as well as the necessary testing and documentation.

This is only the second time ever that I've attempted to write anything in Go, but this feels close to what the missing code should be?

```
if result.NamespaceAnnotations == nil {
	result.NamespaceAnnotations = custom.NamespaceAnnotations
} else if custom.NamespaceAnnotations != nil {
	result.NamespaceAnnotations.Data = data.MergeMaps(result.NamespaceAnnotations.Data, custom.NamespaceAnnotations.Data)
}

if result.NamespaceLabels == nil {
	result.NamespaceLabels = custom.NamespaceLabels
} else if custom.NamespaceLabels != nil {
	result.NamespaceLabels.Data = data.MergeMaps(result.NamespaceLabels.Data, custom.NamespaceLabels.Data)
}
```

### Alternatives you've considered

I've tried to have my helm chart create the namespace with the appropriate annotations to place in a Rancher project via helm values and `takeOwnership: true`, however whilst the namespace lands correctly, Rancher doesn't assign appropriate permissions for non-administrative users to see everything else in the namespace, which puts me in a tricky spot!

I also considered using kustomize, but [this](https://github.com/rancher/fleet/issues/1461#issuecomment-1750706567) suggests it won't work when using helm charts for the applications to be deployed.

For now, I think I'm going to end up with the plethora of `fleet.yaml` files in multiple directories to force split things across my different clusters and work with the top level bundle deployment options.

[...]

Issue #2262, opened 2024-03-23
namespaceLabels does not work as a targetCustomizations


jhoblitt:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I attempted to add additional namespace labels as targetCustomizations[0].namespaceLabels and no change was made to the existing labels on the default namespace.

### Expected Behavior

I expected either the top level namespaceLabels to merged with targetCustomization namespaceLabels or overridden completely.  I did not expect nothing at all to happen as namespaceLabels is part of [`BundleDeploymentOptions`](https://github.com/rancher/fleet/blob/main/pkg/apis/fleet.cattle.io/v1alpha1/bundledeployment_types.go#L103).

### Steps To Reproduce

```
---
defaultNamespace: &name mimir
labels:
  bundle: *name
namespaceLabels:
  foo: bar
helm:
  chart: &chart mimir-distributed
  releaseName: *chart
  repo: https://grafana.github.io/helm-charts
  version: 5.2.0
  timeoutSeconds: 600
  waitForJobs: true
targetCustomizations:
  - name: ruka
    clusterName: ruka
    helm:
      valuesFiles:
        - overlays/ruka/values.yaml
    namespaceLabels:
      this.is/a: test
```

### Environment

```markdown
- Architecture: amd64
- Fleet Version: 0.90 / rancher 2.8.2
- Cluster:
  - Provider: rke[1]
  - Options: extra cheese
  - Kubernetes Version: v1.25.9
```

jhoblitt:
Is this possibly related to #1994?

jhoblitt:
Hmm.  It seems that updating labels doesn't work at all, targetCustomizations or not.

manno:
Closing this as we're tracking the feature in https://github.com/rancher/fleet/issues/2441

PR #2664, opened 2024-07-23
Simplify target customisations for namespace labels and annotations

weyfonk:
Follow-up to #2583.
Refers to #2262 and #2441.

This simplifies end-to-end tests and the implementation for using namespace labels and annotations as target customisations.

In bundle deployment options, maps are used directly instead of pointers to maps for namespace labels and namespace annotations.
Manual tests suggest that Kubernetes represents bundles deployments with identical YAML in both cases, ie whether pointers to maps or simple maps are used to represent those fields.

### DIFF
--- a/internal/cmd/controller/target/target.go
+++ b/internal/cmd/controller/target/target.go
@@ -60,18 +60,14 @@ func (t *Target) BundleDeployment() *fleet.BundleDeployment {
 	bd.Spec.Paused = t.IsPaused()
 
 	for _, bundleTarget := range t.Bundle.Spec.Targets {
-		if bundleTarget.NamespaceLabels != nil {
-			for key, value := range *bundleTarget.NamespaceLabels {
-				(*bd.Spec.Options.NamespaceLabels)[key] = value
-				(*bd.Spec.StagedOptions.NamespaceLabels)[key] = value
-			}
+		for key, value := range bundleTarget.NamespaceLabels {
+			(*bd.Spec.Options.NamespaceLabels)[key] = value
+			(*bd.Spec.StagedOptions.NamespaceLabels)[key] = value
 		}
 
-		if bundleTarget.NamespaceAnnotations != nil {
-			for key, value := range *bundleTarget.NamespaceAnnotations {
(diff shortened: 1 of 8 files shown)


internal/cmd/controller/target/target.go
```diff
@@ -58,22 +58,19 @@ func (t *Target) BundleDeployment() *fleet.BundleDeployment {
...
-	for _, bundleTargets := range t.Bundle.Spec.Targets {
-		if bundleTargets.NamespaceAnnotations != nil {
-			for key, value := range *bundleTargets.NamespaceAnnotations {
-				(*bd.Spec.Options.NamespaceAnnotations)[key] = value
-				(*bd.Spec.StagedOptions.NamespaceAnnotations)[key] = value
-			}
+
+		for key, value := range bundleTarget.NamespaceAnnotations {
+			(*bd.Spec.Options.NamespaceAnnotations)[key] = value
+			(*bd.Spec.StagedOptions.NamespaceAnnotations)[key] = value
```
manno, line 70: We should change this for bundledeployments, too.

internal/cmd/controller/target/target.go
```diff
@@ -58,22 +58,19 @@ func (t *Target) BundleDeployment() *fleet.BundleDeployment {
...
-			for key, value := range *bundleTarget.NamespaceLabels {
-				(*bd.Spec.Options.NamespaceLabels)[key] = value
-				(*bd.Spec.StagedOptions.NamespaceLabels)[key] = value
-			}
+		for key, value := range bundleTarget.NamespaceLabels {
+			(*bd.Spec.Options.NamespaceLabels)[key] = value
+			(*bd.Spec.StagedOptions.NamespaceLabels)[key] = value
 		}
-	}
-	for _, bundleTargets := range t.Bundle.Spec.Targets {
```
manno, line 69: 👍

internal/cmd/controller/target/target.go
weyfonk: Done :+1:

manno (approved):

Issue #2169, opened 2024-02-20
old release not removed during bundle update


aDisplayName:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When a bundle update in the Git Repository results in a new helm release name, the helm release installed by the previous version of the bundle definition is not being removed. It is then causing problem due to resource conflict if both release share the same namespace.

### Expected Behavior

When a bundle update results in a change of helm release name, the bundle deployment should make sure the old helm release is removed by fleet agent, and the new helm release is created according to the bundle definition

### Steps To Reproduce

1. Create a new GitRepo in Rancher Continuous Delivery with the following:

   ```yaml
   apiVersion: fleet.cattle.io/v1alpha1
   kind: GitRepo
   metadata:
     name: github-issue-20240220
   spec:
     correctDrift:
       enabled: true
     paths:
       - /20240220
     repo: https://github.com/aDisplayName/bugsamplecode.git
     revision: 681b822618a0e89b4c8dc731bc104deb89fa322d
   ```
   The link to the revision [681b822618a0e89b4c8dc731bc104deb89fa322d](https://github.com/aDisplayName/bugsamplecode/commit/681b822618a0e89b4c8dc731bc104deb89fa322d)
1. Once it is deployed, it will result in the following helm release in namespace `ver1`
   ![image](https://github.com/rancher/fleet/assets/19527258/bd95107b-bdd1-4289-853e-6cc8bf46f8b9)
1. Update the gitrepo with `revision` set to [cfd907e5a17c386bf369736c798e0cac1e00e7e6](https://github.com/aDisplayName/bugsamplecode/commit/cfd907e5a17c386bf369736c798e0cac1e00e7e6)
   
   Once deployed, a new helm release in namespace `ver2` is now created, but the old release `ver1` was not removed:
   
   ![image](https://github.com/rancher/fleet/assets/19527258/253d9f7b-80d4-4b25-b685-3dd52c19ba1b)

### Environment

```markdown
- Architecture: amd64
- Fleet Version: 0.8.1
- Cluster:
  - Provider: k3s
  - Options:
  - Kubernetes Version: 1.25
- Rancher Server: 2.7.9
```

### Logs

```markdown
2024-02-20T15:18:03.455296701Z time="2024-02-20T15:18:03Z" level=info msg="Deploying bundle cluster-fleet-default-c-5nxhd-7df993ee2063/fleet-agent-c-5nxhd"
[...]

PR #2665, opened 2024-07-23
Make Helm release garbage collection interval configurable

weyfonk:
Refers to #2169.

This also adds end-to-end tests validating that old releases are cleaned up when updating a bundle's namespace or release name.


e2e/assets/helm/repo/with-release-name/crds-only/crds-chart/templates/crds.yaml
```diff
@@ -0,0 +1,41 @@
...
+            foo:
+              description: 'Foo is important.'
+              nullable: true
+              type: string
+            bar:
+              description: 'Bar does not walk into a Foo. A Foo walks into a bar.'
+              nullable: true
+              type: string
+            kind:
+              description: 'Kindness is everything.'
```
0xavi0, line 33: :heart: the descriptions

internal/cmd/controller/agentmanagement/controllers/cluster/import.go
```diff
@@ -97,7 +97,10 @@ func (i *importHandler) onConfig(config *config.Config) error {
 
 		hasConfigChanged := config.APIServerURL != cluster.Status.APIServerURL ||
 			hashStatusField(config.APIServerCA) != cluster.Status.APIServerCAHash ||
-			config.AgentTLSMode != cluster.Status.AgentTLSMode
+			config.AgentTLSMode != cluster.Status.AgentTLSMode ||
+			(config.GarbageCollectionInterval.Duration != 0 && cluster.Status.GarbageCollectionInterval == nil) ||
+			(cluster.Status.GarbageCollectionInterval != nil &&
+				config.GarbageCollectionInterval.Duration != cluster.Status.GarbageCollectionInterval.Duration)
```
0xavi0, line 103: nit: maybe create a helper function that do those checks about `GarbageCollectionInterval`.
I found it hard to read as it is a pretty large expression, the other checks are pretty simple in comparison.

0xavi0 (approved):
lgtm.
Just a nit about a expression that I found too large, but not enough to not approve this.
Nice one

internal/cmd/controller/agentmanagement/controllers/cluster/import.go
weyfonk: Good point, thanks for raising this. I've moved this to a separate function.

0xavi0 (approved):

Issue #2134, opened 2024-02-07
fleet incorrectly reports pre-existing resources as missing when setting helm.takeOwnership: true


jhoblitt:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When settings `helm.takeOwernship: true` , the chart deployment will add the standard helm labels and annotations to an existing object.  E.g.:

```yaml
apiVersion: ceph.rook.io/v1
kind: CephFilesystem
metadata:
  annotations:
    meta.helm.sh/release-name: rook-ceph-conf
    meta.helm.sh/release-namespace: rook-ceph
    ...
  labels:
    app.kubernetes.io/managed-by: Helm
```

However, `fleet-agent` will complain about missing resources:

```
time="2024-02-07T23:16:25Z" level=error msg="bundle kueyen-only-kueyen-fleet-rook-ceph-conf: cephfilesystem.ceph.rook.io rook-ceph/obs-env missing"
```

As will the `bundle` and `bundleDeployment` resources:

```
NAME                                      DEPLOYED   MONITORED   STATUS
kueyen-only-kueyen-fleet-rook-ceph-conf   True       True        cephfilesystem.ceph.rook.io rook-ceph/obs-env missing
```
```
NAME                                      BUNDLEDEPLOYMENTS-READY   STATUS
kueyen-only-kueyen-fleet-rook-ceph-conf   0/1                       Modified(1) [Cluster fleet-default/kueyen]; cephfilesystem.ceph.rook.io rook-ceph/obs-env missing; cephnfs.ceph.rook.io rook-ceph/obs-env missing; cephobjectstore.ceph.rook.io rook-ceph/lfa missing
```

### Expected Behavior

I expect that "fleet" takes ownership of existing resources and does not report them as missing (when they exists!).

The resource is reported as missing because the pre-existing resources do not have the `objectset.rio.cattle.io/hash` label set. E.g.:

```yaml 
    labels:
      app.kubernetes.io/managed-by: Helm
      objectset.rio.cattle.io/hash: 563cb3553e5fba84eee5ba3054b9b537c653465a
```
Manually copying the value of `objectset.rio.cattle.io/hash` from another resource resolves the "missing" resource warning.

### Steps To Reproduce

1. Manually create a resource
2. Create a fleet bundle with `helm.takeOwnership: true` with a duplicate of that resource
3. Observe warnings on the `bundle`, `bundleDeployment` resources and in the `fleet-agent` logs.

### Environment

[...]

PR #2679, opened 2024-07-25
BundleDeployment status error message improvement 

p-se:
When a resource of a BundleDeployment is removed after the Bundle has been created, the BundleDeployment says it is missing. This is also the case if the resource is not removed but changes its owner. In that case, the message of a resource that is missing might be misleading. These changes adapt the message to say "<resource> is not owned by us" instead of saying that the "\<resource\> is missing".

Part of #2134

### DIFF
--- a/integrationtests/agent/adoption_test.go
+++ b/integrationtests/agent/adoption_test.go
@@ -1,0 +1,387 @@
+package agent
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	"github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
+)
+
+func init() {
+	resources["BundleDeploymentConfigMap"] = []v1alpha1.BundleResource{
+		{
+			Name: "configmap.yaml",
+			Content: `apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cm1
+data:
+  key: value
+`,
+			Encoding: "",
+		},
+	}
+}
+
+var _ = Describe("Adoption", Label("adopt"), func() {
+	var (
+		namespace string
+		env       *specEnv
+	)
+
+	createBundleDeployment := func(name string, takeOwnership bool) {
+		bundled := v1alpha1.BundleDeployment{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      name,
+				Namespace: clusterNS,
+			},
+			Spec: v1alpha1.BundleDeploymentSpec{
+				DeploymentID: "BundleDeploymentConfigMap",
+				Options: v1alpha1.BundleDeploymentOptions{
+					DefaultNamespace: namespace,
+					Helm: &v1alpha1.HelmOptions{
+						TakeOwnership: takeOwnership,
+					},
+				},
+			},
+		}
+
+		err := k8sClient.Create(context.TODO(), &bundled)
+		Expect(err).To(BeNil())
+		Expect(bundled).To(Not(BeNil()))
+		Expect(bundled.Spec.DeploymentID).ToNot(Equal(bundled.Status.AppliedDeploymentID))
+		Expect(bundled.Status.Ready).To(BeFalse())
+		Eventually(func() bool {
+			err := k8sClient.Get(context.TODO(), types.NamespacedName{Namespace: clusterNS, Name: name}, &bundled)
+			if err != nil {
+				return false
+			}
+			return bundled.Status.Ready
+		}).Should(BeTrue(), "BundleDeployment not ready: status: %+v", bundled.Status)
+		Expect(bundled.Spec.DeploymentID).To(Equal(bundled.Status.AppliedDeploymentID))
+	}
+
+	waitForConfigMap := func(name string) {
+		Eventually(func() error {
+			_, err := env.getConfigMap(name)
+			return err
+		}).Should(Succeed())
+	}
+
+	createConfigMap := func(data, labels, annotations map[string]string) *corev1.ConfigMap {
+		cm := &corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:        "cm1",
+				Namespace:   namespace,
+				Labels:      labels,
+				Annotations: annotations,
+			},
+			Data: data,
+		}
+		Expect(k8sClient.Create(ctx, cm)).To(Succeed())
+		waitForConfigMap("cm1")
+		return cm
+	}
+
+	// assertConfigMap checks that the ConfigMap exists and that it passes the
+	// provided validate function.
+	assertConfigMap := func(validate func(corev1.ConfigMap) error) {
+		cm := corev1.ConfigMap{}
+		var err error
+		Eventually(func() error {
+			err = k8sClient.Get(
+				ctx,
+				types.NamespacedName{Namespace: namespace, Name: "cm1"},
+				&cm,
+			)
+			if err != nil {
+				return err
+			}
+			return validate(cm)
+		}).Should(Succeed(), "assertConfigMap error: %v in %+v", err, cm)
+	}
+
+	// assertBundleDeployment checks that the BundleDeployment exists and that it
+	// passes the provided validate function.
+	assertBundleDeployment := func(name string, validate func(*v1alpha1.BundleDeployment) error) {
+		bd := v1alpha1.BundleDeployment{}
+		var err error
+		Eventually(func() error {
+			err = k8sClient.Get(
+				ctx,
+				types.NamespacedName{Namespace: clusterNS, Name: name},
+				&bd,
+			)
+			if err != nil {
+				return err
+			}
+			return validate(&bd)
+		}).Should(Succeed(), "assertBundleDeployment: error %v in %+v", err, bd)
+	}
+
+	// mapPartialMatch checks that the super map contains all the keys and values
+	// of the sub map. If the value in the sub map is an empty string, the key
+	// must exist in the super map but the value is not compared.
+	mapPartialMatch := func(super, sub map[string]string) error {
+		for k, v := range sub {
+			if v == "" {
+				if _, ok := super[k]; !ok {
+					return fmt.Errorf("key %s not found in %+v", k, super)
+				}
+			} else {
+				if v2, ok := super[k]; !ok || v2 != v {
+					return fmt.Errorf("key %s not found or value %s does not match %s in %+v", k, v2, v, super)
+				}
+			}
+		}
+		return nil
+	}
+
+	// configMapAdoptedAndMerged checks that the ConfigMap is adopted. It may
+	// need to be extended to check for more labels and annotations.
+	isConfigMapAdopted := func(cm *corev1.ConfigMap) error {
+		err := mapPartialMatch(cm.Labels, map[string]string{
+			"app.kubernetes.io/managed-by": "Helm",
+		})
+		if err != nil {
+			return err
+		}
+		err = mapPartialMatch(cm.Annotations, map[string]string{
+			"meta.helm.sh/release-name":      "",
+			"meta.helm.sh/release-namespace": "",
+		})
+		if err != nil {
+			return err
+		}
+		return nil
+	}
+
+	changeConfigMap := func(name string, change func(*corev1.ConfigMap)) {
+		cm := &corev1.ConfigMap{}
+		Eventually(func() error {
+			return k8sClient.Get(ctx, types.NamespacedName{Namespace: namespace, Name: name}, cm)
+		}).Should(Succeed())
+		change(cm)
+		Eventually(func() error {
+			return k8sClient.Update(ctx, cm)
+		}).Should(Succeed())
+	}
+
+	deleteConfigMap := func(name string) {
+		cm := &corev1.ConfigMap{}
+		Eventually(func() error {
+			return k8sClient.Get(ctx, types.NamespacedName{Namespace: namespace, Name: name}, cm)
+		}).Should(Succeed())
+		Eventually(func() error {
+			return k8sClient.Delete(ctx, cm)
+		}).Should(Succeed())
+	}
+
+	bundleDeploymentResourceMissing := func(bd *v1alpha1.BundleDeployment) error {
+		const msgPrefix = "BundleDeployment resource:"
+		for _, condition := range bd.Status.Conditions {
+			if condition.Type != v1alpha1.BundleDeploymentConditionReady {
+				continue
+			}
+			if condition.Status != corev1.ConditionFalse {
+				return fmt.Errorf("%s Status is not False", msgPrefix)
+			}
+			if condition.Reason != "Error" {
+				return fmt.Errorf("%s Reason is not Error", msgPrefix)
+			}
+			if !strings.Contains(condition.Message, "missing") {
+				return fmt.Errorf("%s Message does not contain 'missing'", msgPrefix)
+			}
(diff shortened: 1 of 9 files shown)


pkg/apis/fleet.cattle.io/v1alpha1/bundledeployment_types.go
```diff
@@ -407,6 +407,7 @@ type ModifiedStatus struct {
 	// +nullable
 	Name   string `json:"name,omitempty"`
 	Create bool   `json:"missing,omitempty"`
+	Exist  bool   `json:"exist,omitempty"`
```
manno, line 410:
```suggestion
        // Exist is true if the modified resource exists in the cluster. This can happen, when ...
	Exist  bool   `json:"exist,omitempty"`
```

p-se: > So this happens because we're missing annotations and the resource ends up in `plan.Create` instead of `plan.Update`?

It might not be accurate to say, that we're missing annotations. So far I was only able to reproduce the issue when two bundles have fought for ownership. When one has it and the other takes it, the first says "missing" instead of "isn't owned by me". This message is adapted in this PR.

> If we wrote our own `applied.DryRun`, could it return lists that fit our use case better? So that we don't have to do a an additional `Get` for every resource in `plan.Create`? I'm asking, because we do want to remove wrangler where possible.

Yes, I think we can do better when we write our own function to diff. However, I do not claim to have fully understood the diffing process.

And I only was able to reproduce the issue with two competing bundles, not with resources installed without a bundle and a bundle failing to properly adopt them. The newly create SURE might help to further investigate that.

internal/cmd/agent/deployer/monitor/updatestatus.go
```diff
@@ -262,12 +262,22 @@ func modified(plan apply.Plan, resourcesPreviousRelease *helmdeployer.Resources)
...
 
 			apiVersion, kind := gvk.ToAPIVersionAndKind()
+
+			obj := &unstructured.Unstructured{}
+			obj.SetGroupVersionKind(gvk)
+			key := client.ObjectKey{
+				Namespace: key.Namespace,
+				Name:      key.Name,
+			}
+			err := c.Get(context.Background(), key, obj)
```
manno, line 272: We need to distinguish client errors from not found errors.

manno (changes requested):
This adds 10 get requests per deployment if I'm right. Is that a performance problem?

p-se:
> This adds 10 get requests per deployment if I'm right. Is that a performance problem?

This adds _up to_ 10 requests per bundledeployment. One request for every resource that the dry-run reported to be missing (that it would create). There is no default requeue period set in the reconciler, so that those requests are only done once on an attempted deployment of the downstream cluster (or if the bundledeployment is changed). It does not look like a performance problem to me.

integrationtests/agent/adoption_test.go
```diff
@@ -0,0 +1,387 @@
...
+		}
+		return fmt.Errorf("does not match expected condition")
+	}
+
+	BeforeEach(func() {
+		namespace = createNamespace()
+		DeferCleanup(func() {
+			Expect(k8sClient.Delete(ctx, &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{Name: namespace}})).ToNot(HaveOccurred())
+		})
```
weyfonk, line 235: I like the intention, but is this really necessary, considering that integration tests don't run on an actual cluster and that no cleanup will happen based on ownership references, etc ([source](https://book.kubebuilder.io/reference/envtest#testing-considerations))?

integrationtests/agent/adoption_test.go
```diff
@@ -0,0 +1,387 @@
...
+	}
+}
+
+var _ = Describe("Adoption", Label("adopt"), func() {
+	var (
+		namespace string
+		env       *specEnv
+	)
+
+	createBundleDeployment := func(name string, takeOwnership bool) {
```
weyfonk, line 41: I think that this test suite would be more readable if these helper functions were moved _after_ the actual test case implementations, and perhaps outside this `Describe` block.

integrationtests/agent/adoption_test.go
```diff
@@ -0,0 +1,387 @@
...
+
+				return mapPartialMatch(cm.Labels, map[string]string{
+					"app.kubernetes.io/managed-by": "Helm",
+					"objectset.rio.cattle.io/hash": "0f3e1d9d146fa8b290c0de403881184751430e59",
+				})
+			})
+			changeConfigMap("cm1", func(cm *corev1.ConfigMap) {
+				cm.Labels = map[string]string{}
+			})
+			assertBundleDeployment("remove-metadata", bundleDeploymentNotOwnedByUs)
```
weyfonk, line 295: Just to be sure I understand this properly: in this case, does it still make sense to expect a bundle deployment not to be ready because the config map is still specified as part of its resources?

In other words, do we expect removing the config map from the bundle deployment (apart from restoring its labels) to be the way to go to enable the bundle deployment to reach ready state?

integrationtests/agent/adoption_test.go
```diff
@@ -0,0 +1,387 @@
...
+		})
+	})
+
+	When("a bundle deployment adopts a \"clean\" resource", Label("clean"), func() {
```
weyfonk, line 299: Would we want to clarify what "clean" means here, for future reference?

I understand it to mean that this resource does not bear labels indicating that it would belong to any bundle deployment. It is simply a resource with the same kind and name as, and living in the same namespace as, one that would be created from the bundle deployment. Is that correct?

integrationtests/agent/adoption_test.go
```diff
@@ -0,0 +1,387 @@
...
+		It("verifies that the ConfigMap is adopted and its content merged", func() {
+			createConfigMap(map[string]string{"foo": "bar"}, nil, nil)
+			createBundleDeployment("adopt-clean", true)
+			assertConfigMap(func(cm corev1.ConfigMap) error {
+				if err := isConfigMapAdopted(&cm); err != nil {
+					return err
+				}
+				return mapPartialMatch(cm.Data, map[string]string{"foo": "bar", "key": "value"})
```
weyfonk, line 307: btw, do we expect the config map data to contain any key/value pairs _other_  than those two? :thinking:

integrationtests/agent/adoption_test.go
```diff
@@ -0,0 +1,387 @@
...
+				map[string]string{"foo": "bar"},
+				map[string]string{"objectset.rio.cattle.io/hash": "234"},
+				map[string]string{"objectset.rio.cattle.io/id": "$#@"},
+			)
+			createBundleDeployment("adopt-invalid-wrangler-metadata", true)
+			assertConfigMap(func(cm corev1.ConfigMap) error {
+				if err := isConfigMapAdopted(&cm); err != nil {
+					return err
+				}
+				return mapPartialMatch(cm.Data, map[string]string{"foo": "bar", "key": "value"})
```
weyfonk, line 356: Should we check that annotations and labels do _not_ contain the ID and hash keys, respectively?

integrationtests/agent/adoption_test.go
```diff
@@ -0,0 +1,387 @@
...
+			})
+		})
+	})
+
+	When("a bundle deployment adopts a resource with random metadata", Label("random-metadata"), func() {
+		It("verifies that the ConfigMap is adopted and its content merged", func() {
+			createConfigMap(
+				map[string]string{"foo": "bar"},
+				map[string]string{"foo": "234"},
+				map[string]string{"bar": "xzy"},
```
weyfonk, line 366: nit: this may be a tad easier to read as something like:
```suggestion
				map[string]string{"foo": "bar"},
				map[string]string{"foo-label": "234"},
				map[string]string{"bar-annotation": "xzy"},
```

integrationtests/agent/adoption_test.go
```diff
@@ -0,0 +1,387 @@
...
+			)
+			createBundleDeployment("adopt-random-metadata", true)
+			assertConfigMap(func(cm corev1.ConfigMap) error {
+				if err := isConfigMapAdopted(&cm); err != nil {
+					return err
+				}
+				return mapPartialMatch(cm.Data, map[string]string{"foo": "bar", "key": "value"})
```
weyfonk, line 373: Same question as in the previous test case: do we want to validate that the random label and annotations are _not_ included in `cm`?

internal/cmd/agent/deployer/monitor/updatestatus.go
```diff
@@ -262,12 +266,38 @@ func modified(plan apply.Plan, resourcesPreviousRelease *helmdeployer.Resources)
...
+
+			exists := true
+			if apierrors.IsNotFound(err) {
+				exists = false
+			}
+
+			if exists {
+				logger.Info("BundleDeployment resource not owned by us",
```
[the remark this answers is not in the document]
p-se: What about `Resource of BundleDeployment not owned by us`? :)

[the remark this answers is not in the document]
weyfonk: That also works :+1:

integrationtests/agent/adoption_test.go
p-se: I think it's even worse than that. From your referenced source:

> EnvTest does not support namespace deletion. Deleting a namespace will seem to succeed, but the namespace will just be put in a Terminating state, and never actually be reclaimed.

For which reason every tests is assigned its own namespace.

I just didn't think much about it when I copied it from one of the other tests next to this one, but now that I do, I feel like I would have to "fix" it everywhere.

integrationtests/agent/adoption_test.go
p-se: I thought about putting them into specEnv, but wasn't confident enough to do that. Do you have an opinion about that?

integrationtests/agent/adoption_test.go
p-se: Yes. Do you think it would be better if that was different?

integrationtests/agent/adoption_test.go
p-se: Yes, that is correct. I'll add a comment to keep the description short.

integrationtests/agent/adoption_test.go
p-se: No, we only expect the config map data to contain the original data patched with whatever data the config map of the bundle deployment contains. Do you suggest to check then length of `cm.Data`, too?

weyfonk: We could check for length, or maybe even run a simple equality check between `cm.Data` and `map[string]string{"foo": "bar", "key": "value"}`.

integrationtests/agent/adoption_test.go
```diff
@@ -0,0 +1,387 @@
...
+		for k, v := range sub {
+			if v == "" {
+				if _, ok := super[k]; !ok {
+					return fmt.Errorf("key %s not found in %+v", k, super)
+				}
+			} else {
+				if v2, ok := super[k]; !ok || v2 != v {
+					return fmt.Errorf("key %s not found or value %s does not match %s in %+v", k, v2, v, super)
+				}
+			}
```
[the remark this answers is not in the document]
p-se: I was kind of happy with the mapPartialMatch function, but thinking about it, it has benefits to use Expect rather than returning values and having to handle them (in tests). So, I've rewritten it again and the result is a really nice representation of test failures. I hope you will ike it, too.

integrationtests/agent/adoption_test.go
p-se: I did the later.

integrationtests/agent/adoption_test.go
p-se: I went backwards here. I had the intention to make it simple, but also had the same doubt about readability. So I went back to using the corev1.ConfigMap instead, but setting name and namespace in helper functions.

integrationtests/agent/adoption_test.go
p-se: I added that. The result is interesting. As resources deployed from bundle deployments are added the hash as label and the ID as annotation, we do not expect those to be preserved. But all other labels and annotations should be preserved.

integrationtests/agent/adoption_test.go
```diff
@@ -317,71 +102,221 @@ var _ = Describe("Adoption", Label("adopt"), func() {
...
+			types.NamespacedName{Namespace: clusterNS, Name: name},
+			&bundled,
+		)).To(Succeed())
+		g.Expect(bundled.Status.Ready).To(BeTrue())
+	}).Should(Succeed(), "BundleDeployment not ready: status: %+v", bundled.Status)
+	Expect(bundled.Spec.DeploymentID).To(Equal(bundled.Status.AppliedDeploymentID))
+}
+
+func (e adoptEnv) waitForConfigMap(name string) {
+	time.Sleep(5 * time.Second)
```
weyfonk, line 210: Why add a `Sleep` statement here? Shouldn't `Eventually` already cover any possible delay?

integrationtests/agent/adoption_test.go
```diff
@@ -317,71 +102,221 @@ var _ = Describe("Adoption", Label("adopt"), func() {
...
+	var err error
+	Eventually(func(g Gomega) {
+		err = k8sClient.Get(
+			ctx,
+			types.NamespacedName{Namespace: e.namespace, Name: "cm1"},
+			&cm,
+		)
+		g.Expect(err).ToNot(HaveOccurred())
+		validate(g, cm)
+		g.Expect(err).ToNot(HaveOccurred())
```
weyfonk, line 237: The expectation on L237 would now be void, since `err` hasn't been updated since L235, unless:
```suggestion
		err = validate(g, cm)
		g.Expect(err).ToNot(HaveOccurred())
```

integrationtests/agent/adoption_test.go
```diff
@@ -317,71 +102,221 @@ var _ = Describe("Adoption", Label("adopt"), func() {
...
+func (e adoptEnv) createConfigMap(cm *corev1.ConfigMap) {
+	cm.ObjectMeta.Name = "cm1"
+	cm.ObjectMeta.Namespace = e.namespace
+	Expect(k8sClient.Create(ctx, cm)).To(Succeed())
+	e.waitForConfigMap("cm1")
+}
+
+// assertConfigMap checks that the ConfigMap exists and that it passes the
+// provided validate function.
+func (e adoptEnv) assertConfigMap(validate func(Gomega, corev1.ConfigMap)) {
```
weyfonk, line 226:
```suggestion
func (e adoptEnv) assertConfigMap(validate func(Gomega, corev1.ConfigMap) error) {
```

integrationtests/agent/adoption_test.go
```diff
@@ -35,276 +35,61 @@ data:
 var _ = Describe("Adoption", Label("adopt"), func() {
 	var (
 		namespace string
```
weyfonk, line 37: Where is this namespace set?

integrationtests/agent/adoption_test.go
p-se: Nice spotting. That shouldn't be there. I used it for debugging.

integrationtests/agent/adoption_test.go
p-se: validate runs expectations on it's on using the reference to Gomega, so there's no need to return an exception anymore. I'll remove the line in 237.

integrationtests/agent/adoption_test.go
p-se: In `BeforeEach` for each `It` a few lines below.

```
	BeforeEach(func() {
		namespace = createNamespace() # <-- here
		env = adoptEnv{namespace: namespace, env: &specEnv{namespace: namespace}}
	})
```

weyfonk: So it is :facepalm: thanks!

weyfonk (approved):
LGTM

weyfonk (approved):

manno (approved):

Issue #2633, opened 2024-07-12
Running 'fleet deploy --dry-run' fails against Helm Charts with Kubernetes version contraints


ampie:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

We are using the new fleet CLI subcommands 'apply', 'target' and 'deploy --dry-run' in our PR workflows on Github repos to verify Fleet apps. The workflow runs the the 'apply' and 'target' commands against an ephemeral VCluster (kube version 1.30.0). For 'deploy' it uses the '--dry-run' option specifically because it doesn't require a connection to the cluster and no assumptions can be made of the state of the ephemeral cluster. When running against the [kyverno chart](https://github.com/kyverno/kyverno/blob/main/charts/kyverno-policies/Chart.yaml) , Helm fails with "chart requires kubeVersion: >=1.25.0-0 which is incompatible with Kubernetes v1.20.0". 

### Expected Behavior

Helm should not fail and the Kubernetes manifests for the kyverno Helm chart should be generated successfully. Granted, the '--dry-run' option cannot determine a Kubernetes version as it executed without any connection to a Kubernetes cluster, but the hard-coded version of v1.20.0 which comes from [Helm's code](https://github.com/helm/helm/blob/main/pkg/chartutil/capabilities.go#L31-L49) is extreme and highly impractical. An even better option would be to mirror the behaviour of the 'helm template' command which takes the '--kube-version' argument which allows the user to test the Fleet app against a specific version of Kubernetes

### Steps To Reproduce

1. Create a fleet app that uses the [kyverno chart](https://github.com/kyverno/kyverno/blob/main/charts/kyverno-policies/Chart.yaml)
2. Run 'fleet apply' against any cluster using the fleet.yaml file
3. Create a matching Fleet cluster object in the cluster with labels that match the fleet.yaml file
4. Run 'fleet target' using the bundle file generated by step 2
5. Run 'fleet deploy --dry-run' using the bundledeployment file generated by step 4
6. Note the message "chart requires kubeVersion: >=1.25.0-0 which is incompatible with Kubernetes v1.20.0"

### Environment

```markdown
- Architecture: any
- Fleet Version: v0.10.0-rc20
- Cluster:
  - Provider: any
  - Options: 
  - Kubernetes Version: any
```

### Anything else?

[...]

PR #2686, opened 2024-07-26
KubeVersion fix for fleet deploy --dry-run

ampie:
Refers to #2633

This Pull Requests fixes the kubeVersion validation error by:
1.) Introducing a new argument `--kube-version` to allow users to override the target version Helm charts can be validated against. This follows the exact same semantics as the `--kube-version` argument of the `helm template` command.
2.) Where no `--kuberversion` is specified, a more reasonable default of 1.25.0 is provided.

An integration test, "Printing results with --dry-run where the chart specifies kubeVersion >= 1.26.0", was added  in integrationtests/cli/deploy/deploy_test.go to illustrate both the issue and the fix


manno (approved):
💚

PR #2691, opened 2024-07-30
Fix linting error

weyfonk:
Getting an agent StatefulSet from a manifest for unit tests always happens in namespace `fleet-system`, which is now a package-wide constant.
To prevent that constant from being used in the implementation package, unit tests now live in a separate `agent_test` package.

This should fix `golangci-lint` workflow failures such as [this one](https://github.com/rancher/fleet/actions/runs/10138467073/job/28030207568).


0xavi0 (approved):
lg2m

PR #2692, opened 2024-07-30
Clarify error message when e2e tests fail

weyfonk:
This should reduce likelihood of test flakiness and ease its troubleshooting, by:
* speeding up deployment of test workloads for drift correction
* improving failure reporting by mentioning the actual unexpected output, as opposed to a boolean (representing success) being false.

~Left in draft mode until https://github.com/rancher/fleet-test-data/pull/22 is merged, after which changes made here to `gitrepo.yaml` files used for drift end-to-end tests will be reverted.~


0xavi0 (approved):
lgtm

Issue #2674, opened 2024-07-24
Add k8s events to reconcilers


manno:
Fleet writes information about deployments and errors into the status of resources. To assist users with troubleshooting, e.g. when they don't have access to the logs, we can record k8s events.

* record events for critical/terminal failures
* record events for noteworthy deployment steps
* don't add too many events, keep logging separate from events
* events should use the past tense 

Log and debug messages still exists for details and tracing.

See:
* https://book-v1.book.kubebuilder.io/beyond_basics/creating_events

sbulage:
I can see the k8s events are generating for the Fleet operations i.e. GitRepo commit found, pod creation, deletion etc. See below screenshot for more details.

<details>
<summary><b>Screenshot showing Fleet events</b></summary>

![Screenshot from 2024-08-14 15-56-31](https://github.com/user-attachments/assets/0fa07116-a298-4d95-a63d-5438e222f0c4)

</details>

PR #2698, opened 2024-08-05
Adds k8s events to the gitopts controller

0xavi0:
Adds events for failures so the user gets information faster when there's no access to logs.

It also adds events when we get a new commit, a new job is created or deleted to inform the user of the status

Refers to: https://github.com/rancher/fleet/issues/2674


integrationtests/gitjob/controller/suite_test.go
```diff
@@ -97,6 +98,7 @@ var _ = BeforeSuite(func() {
 		Scheduler:  sched,
 		GitFetcher: fetcherMock,
 		Clock:      reconciler.RealClock{},
+		Recorder:   mgr.GetEventRecorderFor(fmt.Sprintf("gitjob-controller%s", "")),
```
weyfonk, line 101: nit: this could simply be:
```suggestion
		Recorder:   mgr.GetEventRecorderFor("gitjob-controller"),
```

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -141,8 +144,9 @@ func (r *GitJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 
 	// Restrictions / Overrides, gitrepo reconciler is responsible for setting error in status
 	oldStatus := gitrepo.Status.DeepCopy()
-	gitrepo, err := grutil.AuthorizeAndAssignDefaults(ctx, r.Client, gitrepo)
```
weyfonk, line 144: Was there an issue with overwriting `gitrepo` here?

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -141,8 +144,9 @@ func (r *GitJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+	_, err := grutil.AuthorizeAndAssignDefaults(ctx, r.Client, gitrepo)
 	if err != nil {
+		r.Recorder.Event(gitrepo, fleetevent.Warning, "Failed", err.Error())
```
weyfonk, line 149: The `EventRecorder` [documentation](https://github.com/kubernetes/client-go/blob/master/tools/record/event.go#L98) states that the `reason` behind an event should be short and _unique_ (my emphasis).
Does this mean that each of our calls to `r.Recorder.Event` should provide a different reason?
eg. this one could be `FailedToApplyRestrictions`, whereas the one on L184 could be something like `FailedToCheckPolling`... of course, depending on how short we want to make them, there may be better examples.

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -197,10 +206,14 @@ func (r *GitJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
 				gitrepo.Status.Commit = commit
 			}
+			if err != nil {
+				r.Recorder.Event(gitrepo, fleetevent.Warning, "Failed", err.Error())
+			}
 		}
 
 		if gitrepo.Status.Commit != "" {
 			if err := r.validateExternalSecretExist(ctx, gitrepo); err != nil {
+				r.Recorder.Event(gitrepo, fleetevent.Warning, "Failed", err.Error())
```
weyfonk, line 216: Which criteria were used to determine whether or not to generate events?
For instance, do we want to generate an event if failing to create git job resources or to delete a job?

internal/cmd/controller/gitops/reconciler/gitjob_test.go
```diff
@@ -418,6 +478,148 @@ func TestReconcile_Error_WhenGitrepoRestrictionsAreNotMet(t *testing.T) {
...
+	mockFetcher := gitmocks.NewMockGitFetcher(mockCtrl)
+	commit := "1883fd54bc5dfd225acf02aecbb6cb8020458e33"
+	mockFetcher.EXPECT().LatestCommit(gomock.Any(), gomock.Any(), gomock.Any()).Times(1).Return(commit, nil)
+
+	// we need to return a NotFound error, so the code tries to create it.
+	mockClient.EXPECT().Get(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Times(1).DoAndReturn(
+		func(ctx context.Context, req types.NamespacedName, job *batchv1.Job, opts ...interface{}) error {
+			return errors.NewNotFound(schema.GroupResource{}, "TEST ERROR")
+		},
+	)
```
weyfonk, line 576: Why do we need this? It seems that the comment aims to clarify it, but it's not yet clear to me :upside_down_face:

weyfonk (commented):
I like how concise this is :)
A few doubts arose, summed up here.

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
0xavi0: Yes, as we're using the `gitrepo` in the later Event call... `AuthorizeAndAssignDefaults` returns `nil` when something goes wrong. 
Anyway, when that happens we don't do anything else with that `gitrepo` because we return, but we need it in order to call `Event` and the object (`gitrepo` in this case, is the first parameter)

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
0xavi0: Yeah, good point, I also thought about it. 
I don't know if we should go further into detail we give to the user. 
I can try as you suggest...

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
0xavi0: Criteria in this case is to create the event in the case that the information we give to the user means something he/she can do something about.

For example, in this case we're checking for the secrets that the user creates.
If something goes wrong.. the user is able to fix it.

Later in the same file... we call `createJobRBAC`, for example.
I didn't add any event there because if that fails... it's an internal error and the user is not going to be able to do anything about it.

For those cases I was thinking on adding a more generic error, maybe.

internal/cmd/controller/gitops/reconciler/gitjob_test.go
0xavi0: that's because in the controller code... the branch that tries to create the job (the one that checks for the helm secrets) starts with:
```go
        // Gitjob handling
	if errors.IsNotFound(err) {
```

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
weyfonk: Ah that seems obvious now :) Thanks for clarifying.

internal/cmd/controller/gitops/reconciler/gitjob_test.go
weyfonk: Oh I see :bulb:
Makes sense, thanks.

integrationtests/gitjob/controller/controller_test.go
```diff
@@ -345,6 +367,30 @@ var _ = Describe("GitJob controller", func() {
 
 				return string(job.UID) != string(newJob.UID)
 			}).Should(BeTrue())
+			// it should log 2 events
+			// first one is to log the new commit from the poller
+			// second one is to inform that the job was created
```
weyfonk, line 372:
```suggestion
			// it should log 3 events
			// first one is to log the new commit from the poller
			// second one is to inform that the job was created
			// third one reports on the job being deleted because of ForceUpdateGeneration
```

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
weyfonk: Is the remaining `Failed` a leftover here?

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -175,7 +179,15 @@ func (r *GitJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
-	gitPollerWasExecuted, _ := r.checkPollingTask(ctx, gitrepo)
+	commitBefore := gitrepo.Status.Commit
+	gitPollerWasExecuted, err := r.checkPollingTask(ctx, gitrepo)
+	if err != nil {
+		r.Recorder.Event(gitrepo, fleetevent.Warning, "FailedToCheckCommit", err.Error())
+	} else {
+		if gitPollerWasExecuted && commitBefore != gitrepo.Status.Commit {
+			r.Recorder.Event(gitrepo, fleetevent.Normal, "GotNewCommit", gitrepo.Status.Commit)
+		}
+	}
```
weyfonk, line 193: nit: we could save a couple of lines and an indentation level:
```suggestion
	} else if gitPollerWasExecuted && commitBefore != gitrepo.Status.Commit {
	        r.Recorder.Event(gitrepo, fleetevent.Normal, "GotNewCommit", gitrepo.Status.Commit)
	}
```

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -386,27 +409,48 @@ func (r *GitJobReconciler) createJob(ctx context.Context, gitRepo *v1alpha1.GitR
...
 	return nil
 }
 
+func generationChanged(r *v1alpha1.GitRepo) bool {
+	// checks if generation changed.
+	// it ignores the case when Status.ObservedGeneration=0 because that's
+	// the initial value of a just created GitRepo and the initial value
+	// for Generation in k8s is 1.
+	// If we don't ignore we would be deleting the gitjob that was just created
+	// until later we reconcile ObservedGeneration with Generation
```
weyfonk, line 453: Good to know, thanks for documenting this :clap:

internal/cmd/controller/gitops/reconciler/gitjob_test.go
```diff
@@ -214,12 +260,19 @@ func TestReconcile_LatestCommitIsOkay(t *testing.T) {
 		},
 	).Times(1)
 
+	recorderMock := mocks.NewMockEventRecorder(mockCtrl)
+	recorderMock.EXPECT().Event(&gitRepoMatcher{gitRepo},
+		fleetevent.Normal,
+		"GotNewCommit",
+		"1883fd54bc5dfd225acf02aecbb6cb8020458e33")
```
weyfonk, line 267: nit: this kind of formatting is more idiomatic (same remark around L404 and L471):
```suggestion
	recorderMock.EXPECT().Event(
	        &gitRepoMatcher{gitRepo},
		fleetevent.Normal,
		"GotNewCommit",
		"1883fd54bc5dfd225acf02aecbb6cb8020458e33",
		)
```

weyfonk (commented):
Would LGTM apart from a doubt about a remaining `Failed` reason.

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
0xavi0: yup, it was a leftover. Changed now

integrationtests/gitjob/controller/controller_test.go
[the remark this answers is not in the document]
0xavi0: changed

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
0xavi0: changed as suggested

internal/cmd/controller/gitops/reconciler/gitjob_test.go
0xavi0: changed

weyfonk (approved):
LGTM, thanks for this :)

PR #2702, opened 2024-08-06
Remove content cleanup from bundle deployment reconciler

weyfonk:
Purging content resources happens when triggering bundle deployment deletion from a bundle deletion, or even a `GitRepo` deletion.
Direct bundle deployment deletion is not supported, and does not work because:
* a bundle deployment is not meant to be created on its own, ie without a bundle
* deleting a bundle deployment before deleting the bundle from which it was created will lead to that bundle deployment being re-created by the bundle reconciler.

Follow-up to #2566
Refers to #2464


0xavi0 (approved):
lgtm

PR #2727, opened 2024-08-12
Revert "Simplify target customisations for namespace labels and annotations"

manno:
Reverts rancher/fleet#2664 we want to hold off until v0.11 or v2.10


manno:
Not reverting as we pick issues for 0.10 and need a few commit from this PR anyhow.

PR #2756, opened 2024-08-20
Prevent `0` as value for gitClientTimeout

p-se:
A zero value for the gitClientTimeout setting would mean that the request would not have a client side timeout set, potentially resulting in an endless waiting state. Although it should not happen that this value is missing, which would lead to it being zero, it should be safer to fall back to 30s as in previous versions of Fleet, thereby preventing the user from disabling the git client timeout completely.

Follow-up for #2188

### DIFF
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -1,0 +1,34 @@
+package config_test
+
+import (
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	v1 "k8s.io/api/core/v1"
+
+	"github.com/rancher/fleet/internal/config"
+)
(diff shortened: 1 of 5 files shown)


manno (approved):

internal/config/config_test.go
```diff
@@ -0,0 +1,33 @@
+package config_test
+
+import (
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	v1 "k8s.io/api/core/v1"
```
manno, line 8:
```suggestion

	v1 "k8s.io/api/core/v1"
```

0xavi0 (approved):
lgtm

PR #2760, opened 2024-08-21
[v0.8] Bump Docker to v23.0.14

thardeck:
https://artifacthub.io/packages/helm/fleet/fleet/0.8.5


weyfonk (approved):
Approving for the sake of keeping our dependencies up-to-date, but with similar doubts as expressed on #2761.

PR #2763, opened 2024-08-21
Rename Fleet manager references as Fleet controller

weyfonk:
This removes manager references from non-controller Fleet chart descriptions, and renames the `FleetManager` struct as `FleetController` to enable correct Markdown documentation to be generated from it through `generate-cli-docs.go`.

This could be taken further by renaming eg. `Manager-initiated [cluster] registration` as `Controller-initiated registration` for more consistency.


kkaempf (dismissed):
:+1:

manno (approved):

PR #2764, opened 2024-08-21
Remove shard node selector from hidden flags

weyfonk:
That flag need not be hidden on cleanup and agent management commands, which parent command (`fleetcontroller`) does not define it. `fleetcontroller gitjob` defines it, where it therefore must not be hidden either.

Follow-up to #2620.
Refers to #2456,


manno (approved):
Makes me wonder why "shard-id" is on the controller struct and needs to be hidden.

weyfonk:
> Makes me wonder why "shard-id" is on the controller struct and needs to be hidden.

We still need it to run a Fleet (non-gitOps) controller, which logic does not need to know about node selectors.
Do you mean that we could/should specify shard IDs as environment variables to the `fleetcontroller` command line, as we do for eg. reconciler worker counts?

Issue #2739, opened 2024-01-26
Gitjob returns 500 to a webhook with wrong credentials instead of 401


thehejik:
Not so critical but it would be nice if gitjob returned 401 instead of 500 on a webhook POST with wrong credentials.

The credentials were set on local cluster by: ([doc](https://fleet.rancher.io/webhook))
```
$ kubectl create secret generic gitjob-webhook -n cattle-fleet-system --from-literal=azure-username=user --from-literal=azure-password=password
$ kubectl create secret generic gitjob-webhook2 -n cattle-fleet-system --from-literal=github=secretsecret
```

Noted on rancher:v2.8.1 / fleet 0.9.0 on RKE2:1.27 when testing webhooks from Azure Devops and Github with custom built rancher/gitjob image from `release/fleet/v0.9` branch.

This is from Azure WH test:
![image_720](https://github.com/rancher/gitjob/assets/12828077/205acd75-c3f7-4838-bffa-9fa9f17ba298)

mmartin24:
This issue seems fixed now in Rancher `v2.9-573d477bcf8b328f8951d6e48bd54a190d136cd3-head` with `fleet-agent:v0.10.2-rc.1`

---
Checked that invalid secret returns now a 401 error rather than a 500:

- Test in Azure webhook:
![image](https://github.com/user-attachments/assets/774e2a77-adf8-44a2-93d8-2264e3eed047)

- Test with Github webhook:
![image](https://github.com/user-attachments/assets/b9d2654c-f390-4610-bab9-aa843e45e3d4)

Testomg references for this: 
- https://confluence.suse.com/display/~thehejik/Fleet+and+Azure+DevOps+testing
- https://confluence.suse.com/display/~mamartin/Configure+Github+Webhook+and+Fleet

PR #2777, opened 2024-08-26
Return error code 401 for auth errors in webhooks

0xavi0:
Return error code 401 instead of 500 when the error is related to the secret or any kind of authentication.

It also adds tests for checking the basic error handling code and a full test for github that tests bad secrets and right secrets (checking also that the commit being updated is as expected)

Refers to https://github.com/rancher/fleet/issues/2739

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -298,8 +298,23 @@ func HandleHooks(ctx context.Context, namespace string, client client.Client, cl
 
 func (w *Webhook) logAndReturn(rw http.ResponseWriter, err error) {
 	w.log.Error(err, "Webhook processing failed")
-	rw.WriteHeader(500)
+	rw.WriteHeader(getErrorCodeFromErr(err))
 	_, _ = rw.Write([]byte(err.Error()))
+}
+
+func getErrorCodeFromErr(err error) int {
+	// check if the error is a verification of identity error
+	// secret check, or basic credentials or token verification
+	// depending on the provider
+	if err == gogs.ErrHMACVerificationFailed ||
+		err == github.ErrHMACVerificationFailed ||
+		err == gitlab.ErrGitLabTokenVerificationFailed ||
+		err == bitbucket.ErrUUIDVerificationFailed ||
+		err == bitbucketserver.ErrHMACVerificationFailed ||
+		err == azuredevops.ErrBasicAuthVerificationFailed {
+		return 401
+	}
+	return 500
 }
 
 // git ref docs: https://git-scm.com/book/en/v2/Git-Internals-Git-References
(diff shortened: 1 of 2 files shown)


pkg/webhook/webhook.go
```diff
@@ -298,10 +298,25 @@ func HandleHooks(ctx context.Context, namespace string, client client.Client, cl
...
+	// check if the error is a verification of identity error
+	// secret check, or basic credentials or token verification
+	// depending on the provider
+	if err == gogs.ErrHMACVerificationFailed ||
+		err == github.ErrHMACVerificationFailed ||
+		err == gitlab.ErrGitLabTokenVerificationFailed ||
+		err == bitbucket.ErrUUIDVerificationFailed ||
+		err == bitbucketserver.ErrHMACVerificationFailed ||
+		err == azuredevops.ErrBasicAuthVerificationFailed {
+		return 401
```
weyfonk, line 315: nit:
```suggestion
		return http.StatusUnauthorized
```

pkg/webhook/webhook.go
```diff
@@ -298,10 +298,25 @@ func HandleHooks(ctx context.Context, namespace string, client client.Client, cl
...
+	}
+	return 500
```
weyfonk, line 317: nit:
```suggestion
	return http.StatusInternalServerError
```

pkg/webhook/webhook_test.go
```diff
@@ -225,3 +235,200 @@ func TestGitHubPingWebhook(t *testing.T) {
...
+	tests := map[string]struct {
+		err               error
+		expectedErrorCode int
+	}{
+		"gogs-verification": {
+			err:               gogs.ErrHMACVerificationFailed,
+			expectedErrorCode: 401,
+		},
+		"gogs-no-verification": {
+			err:               gogs.ErrInvalidHTTPMethod,
```
weyfonk, line 249: nit: those cases about invalid HTTP methods (same comment about L289) could expect 405, or `http.StatusMethodNotAllowed`, if we wanted to be picky about this ;)

weyfonk (dismissed):
LGTM, thanks for this!

pkg/webhook/webhook.go
0xavi0: Changed as suggested

pkg/webhook/webhook_test.go
0xavi0: I didn't do it because those are 3rd party provider errors and there are quite a few and they are repeated in all packages, and I didn't want to add more complexity to this.
Also the user can't do much about those errors (I would expect Github or Gitlab to send valid payloads) 
The authentication or bad secret or token one is, maybe the only error that the user can do something about it to fix it. 

I've added the 405 case and the rest are still 500.

weyfonk: Agreed, git hosting providers should issue requests with valid methods and payloads. Thanks for this update, at least in the unlikely event that they don't, users will have access to more details about what's wrong.

PR #2782, opened 2024-08-28
Tolerate unitialized node taint

k0da:
In order to bootstrap cluster with fleet (to deploy CPI), we need gitjob to tolerate unitialized node taint

Refers to #XXX

Tolerating unitialized nodes taints helps with bootstraping CPI with fleet. Withouth that fleet produced jobs never completes


0xavi0:
Hi, thanks for your contribution!

We have a few questions about this changed proposed. 
For example: why you you need to let initialised nodes? This is intentional so nodes that are not fully initialised are not eligible, so we don't really understand why do you need this. 

Could you please explain a bit more?

k0da:
@0xavi0 Our use case:

Our use case is to deploy Vsphere Cloud Provider with Fleet along with other applications (cert-manager,ingress-nginx, cert-manager).  Those fleet.ymls are tested and well known from configuration point of view.

We're deploying RKE2 clusters with clusterAPI and don't want to configure CloudProvider credentials (as this will be solved with fleet stack). Now cluster api deploy us, 1 CP and 1 worker node, and awaits those nodes to be initialized from Cloud Provider POV.  We now just need to deploy fleet and add gitrepos there. We add tolerations for fleet itself via helmchart values. But redendering mainfest is failing or rather stuck in pending (nodes are still uninialized), hence no CPI could be installed to procceed with provisioning.

```
node.cloudprovider.kubernetes.io/uninitialized: When the kubelet is started with an "external" cloud provider, this taint is set on a node to mark it as unusable. After a controller from the cloud-controller-manager initializes this node, the kubelet removes this taint.
```
In most of the cases CPI just fetches node id from cloud and updates v1.Node ProvierID field. So a node is usable for pod run. (fleet-controller + fleet-agent already running)

Not sure that clarifies the ask.

k0da:
FleetAgent already supports that:
https://github.com/rancher/fleet/blob/main/charts/fleet-agent/values.yaml#L56-L60

manno (approved):

Issue #1640, opened 2023-07-07
Test Dev Fleet within released Rancher


manno:
Rancher could be  (nightly, rc, release).

## How to install Rancher with our fleet build?

* switching the rancher-charts **branch** with a helm value at installation time
  * we'd need a write token for test-fleet in rancher/charts
  * cannot use a different server, because of https://github.com/rancher/rancher/pull/40281
* using airgap mode ("bundled")  and building our own repo to **install with dev fleet**
  * [setup a minimal, local helm chart repo](https://github.com/rancher/rancher/pull/41523/commits/1aaba7da2187cec6301e7fe11891571558241771)
  * unless we run rancher outside the cluster, with a local repo dir, we need to ADD our repo over the[ clone in the rancher/rancher image](https://github.com/rancher/rancher/blob/e5deaaf81f8af41a1219d3470a0876d384119386/package/Dockerfile#L64C1-L74)
* Add another layer to the Rancher image, which contains our bundled fleet version in the airgap repo 

Running outside the cluster, without the rancher helm chart, could be interesting for vscode&integration tests
E2e  CI probably needs the modified image (or a branch in rancher/charts).
  
## What are existing upgrade tests in Fleet:
* [Upgrade Fleet in Rancher to Latest Release ](https://github.com/rancher/fleet/actions/workflows/rancher-upgrade-fleet.yml) - red 
  Upgrade fleet in specific Rancher versions to latest fleet release and run MC tests
* [E2E Rancher Fleet](https://github.com/manno/fleet/actions/workflows/rancher-integration.yml) - green (last run months ago)
  Upgrade fleet in latest Rancher to dev version and run MC tests
* [Upgrade Fleet in Rancher](https://github.com/rancher/fleet/actions/workflows/fleet-upgrade-in-rancher.yml) - disabled

## How effective are our existing rancher integration tests?

* Only testing simple, short-lived deployments 
* Existing multi-cluster tests
* Upgrade from one Rancher version to the next?
* Installation with special options?
* Does Fleet work in previous Rancher versions, too? (`fleet-upgrade-in-rancher.yaml` had a nice matrix)
* ...

# Problem

* Current tests only upgrade after the installation. This relies on a modified helm chart, with a very high version number.   
   Do we observe downgrades to the released fleet chart?
* Are current test runners sufficient? Move the tests to the AKS or GKE workflow?
* Tests could use our 'test infra' with registry and git server?
* Tests could profit from version pinning PR

[...]

PR #2804, opened 2024-09-04
Test Fleet in Rancher on self-hosted runner

weyfonk:
This creates a new `Test Fleet in Rancher` [workflow](https://github.com/rancher/fleet/actions/workflows/e2e-rancher-upgrade-fleet-to-head-ci.yml) to install Fleet through the latest Rancher. An example run can be found [here](https://github.com/rancher/fleet/actions/runs/10703546330/job/29674368579).
This workflow can be called from [this one](https://github.com/rancher/fleet/actions/workflows/release-against-test-charts.yml) to test any given Fleet commit against Rancher through multi-cluster tests.

### Open points
* since this workflow is scheduled to run at 1 PM every day, would we want to switch the scheduling to run [this one](https://github.com/rancher/fleet/actions/workflows/release-against-test-charts.yml) instead, which would effectively test Fleet's latest `main` commit against the latest Rancher release every day?
* if not, which defaults do we want to set for chart branches and repositories, which would be sensible for as many testing use cases as possible? Or should those values be computed to, say, fetch the latest charts branch name according to default configuration of the latest Rancher release?
* do we want to run this new workflow against each pull request? It seems to last around 10 min, including generating test Fleet charts.
* Each run of [this workflows](https://github.com/rancher/fleet/actions/workflows/release-against-test-charts.yml) creates a new `fleetrepoci/charts` branch, which is only useful for as long as Docker images pointed to by test charts exist. In practice, this means only [an hour](https://github.com/rancher/fleet/blob/main/.github/workflows/release-against-test-charts.yml#L73), after which those branches could, and probably should, be deleted. Do we want to automate that deletion to prevent cluttering of `fleetrepoci/charts` with obsolete test branches?

### Possible improvements

* Use this to run single-cluster end-to-end tests, including those requiring test infrastructure.

Refers to #1640.


dev/test-in-rancher
```diff
@@ -0,0 +1,73 @@
...
+branch=${CHARTS_BRANCH-""}
+
+# Build Rancher
+# This is not needed if all you want is to test Fleet against an existing Rancher release
+#pushd ../rancher
+#make quick
+#popd
+
+# XXX: this could be improved by using a single server instead of 3; calling this script is lazy but overkill
+./dev/setup-k3d
```
manno, line 14: I'm in favor of deleting that script. Just copy what you need here? Or remove the setup, creating k3d or re-using one can be left to the dev. If this script creates the k3d, does it also have to remove it?

dev/test-in-rancher
manno, line 1: I'd name this script something like 'setup-rancher-with-dev-fleet', as it doesn't run any tests and fits the other setup-thing scripts in the folder.

.github/workflows/e2e-rancher-upgrade-fleet-to-head-ci.yml
```diff
@@ -23,88 +57,62 @@ env:
...
 
 jobs:
-  rancher-fleet-integration:
-    runs-on: ubuntu-latest
+  rancher-integration:
+    runs-on: runs-on,runner=4cpu-linux-x64,run-id=${{ github.run_id }}
 
     steps:
-      -
+      - name: Add paths into PATH
```
manno, line 64: I find the style with the empty newline after `-` very readable, as it pronounces the blocks. We also use it nearly all other workflows.

.github/scripts/register-downstream-clusters.sh
```diff
@@ -41,14 +35,17 @@ userPrincipal:
...
+# Log into the 4th project listed by `rancher login`, which should be the local cluster's default project.
 echo -e "4\n" | rancher login "https://$public_hostname" --token "$token" --skip-verify
 
 rancher clusters create second --import
 until rancher cluster ls --format json | jq -r 'select(.Name=="second") | .ID' | grep -Eq "c-[a-z0-9]" ; do sleep 1; done
 id=$( rancher cluster ls --format json | jq -r 'select(.Name=="second") | .ID' )
 
 kubectl config use-context "$cluster_downstream"
-rancher cluster import "$id"
+kubectl create clusterrolebinding cluster-admin-binding --clusterrole cluster-admin --user $user
```
manno, line 46: Interesting, I wonder if it would help to pick an admin user:

https://github.com/manno/fleet-dev-tools/commit/f53ec4c96d9217a036d05bbb4b3a2cef4fdda31b#diff-713765b9c3a7b1fdfcfe622eba2300d80eeff5d04089de16342b77072e83efa3R23

dev/test-in-rancher
weyfonk: Good point, I've removed that call and documented cluster availability as a prerequisite to running this script.

dev/test-in-rancher
weyfonk: Done, thanks!

Issue #2768, opened 2024-08-22
After removing cluster from cluster group does not delete the installed resources from the cluster.


sbulage:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Resources are present even after removing cluster from the cluster group.

### Expected Behavior

It should delete the resource from the cluster when removed from the cluster group.

### Steps To Reproduce

1. Add labels to 3 downstream clusters.
2. Using those labels, create a cluster group which includes 3 clusters.
3. Create a GitRepo and Deploy To --> select the cluster group (created in previous step)
4. Wait for the application to be created on each cluster and GitRepo status Active.
5. Navigate to Continuous Delivery ---> Clusters
6. Remove the label from any of the cluster for e.g. second cluster.
7. Wait for the cluster group to update it's count from 3 --> 2.
8. GitRepo also shows the correct cluster count i.e. 2/2.
9. Navigate to the second cluster ---> Workloads --> Pods/Deployments

You can see the application/resource is still present and didn't get removed.

### Environment

```markdown
- Architecture:
- Fleet Version:0.10.0/0.10.1
- Cluster:
  - Provider:k3s/k3d (downstream clusters)
  - Options:
  - Kubernetes Version:
```

### Anything else?

See below video for detailed information: 

https://github.com/user-attachments/assets/1a54f46b-482d-494f-8935-293a8934be8f

sbulage:
### System Information
| Rancher Version | Fleet Version |
|--------|--------|
| 2.9-head | 104.0.2+up0.10.2-rc.4 | 
---

### Verification Steps
_**Note: Followed steps from description**_

- Verified that **after removing label** from **one cluster out of 3 clusters** able to see the **application is getting removed**.
- Also, verified that **re-adding label to the same cluster** **installs application to same cluster**.
---

<details>
<summary><b> Screenshot showing actual results </b></summary>

![Screenshot from 2024-09-11 22-15-24](https://github.com/user-attachments/assets/ec2ff57c-9d89-4270-8e4e-a8622aa63e3d)

</details>

PR #2820, opened 2024-09-10
Fixes finalizer deletion in bundle deployments.

0xavi0:
This issue is affecting the `BundleDeployment` when it is deleted after targetting discards it as a valid target.

The reconciler was not called when `DeletionTimestamp` was set and there was a typo that prevented the finalizer to be deleted.

Refers to: https://github.com/rancher/fleet/issues/2768

### DIFF
--- a/internal/cmd/controller/reconciler/bundledeployment_controller.go
+++ b/internal/cmd/controller/reconciler/bundledeployment_controller.go
@@ -75,7 +75,7 @@ func (r *BundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 
 				controllerutil.RemoveFinalizer(t, bundleDeploymentFinalizer)
 
-				return r.Update(ctx, bd)
+				return r.Update(ctx, t)
 			})
 			if err != nil {
 				return ctrl.Result{}, err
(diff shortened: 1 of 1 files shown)


internal/cmd/controller/reconciler/bundledeployment_controller.go
```diff
@@ -139,7 +139,7 @@ func bundleDeploymentStatusChangedPredicate() predicate.Funcs {
 			if n == nil || o == nil {
 				return false
 			}
-			return !reflect.DeepEqual(n.Status, o.Status)
+			return !reflect.DeepEqual(n.Status, o.Status) || !n.DeletionTimestamp.IsZero()
```
weyfonk, line 142: nit: swapping the conditions could prevent a call to `reflect.DeepEqual` if the bundle deployment has been marked for deletion (optimisation):
```suggestion
			return !n.DeletionTimestamp.IsZero() || !reflect.DeepEqual(n.Status, o.Status)
```

weyfonk (approved):
Nice catch! LGTM, with automated tests pending as discussed.

weyfonk (approved):

weyfonk (approved):

PR #2841, opened 2024-09-12
Adds e2e test to check that bundledeployments are deleted

0xavi0:
This adds a new test to check that when applying a gitrepo that targets based on labels it deploys/deletes the bundledeployment depending if the label for targetting is present or not in the downstream cluster

This is related to: https://github.com/rancher/fleet/pull/2820
Refers to: https://github.com/rancher/fleet/issues/2768


weyfonk (approved):
LGTM :)

PR #2842, opened 2024-09-12
Add crust-gather to e2e fail handlers

manno:
This replaces dump-failed-k3d.sh

* changing the job names to be fixed strings for branch protection
* updating k3s 1.30.4 for api fix
* updating k3d to support k3s 1.30.4


e2e/testenv/fail.go
```diff
@@ -0,0 +1,35 @@
...
+
+func FailAndGather(message string, callerSkip ...int) {
+	if _, err := exec.LookPath("crust-gather"); err != nil {
+		ginkgo.GinkgoWriter.Print("⛔ crust-gather is not available, not dumping cluster info")
+		ginkgo.Fail(message, callerSkip...)
+	}
+
+	pwd := os.Getenv("GITHUB_WORKSPACE")
+	if pwd != "" {
+		pwd = path.Join(pwd, "tmp", "upstream")
```
weyfonk, line 19: Is this why we call `crust-gather` explicitly for downstream clusters through:
```
kubectl config use-context k3d-downstream
crust-gather collect --exclude-namespace=kube-system --exclude-kind=Lease --duration=5s -f tmp/downstream
```
... but not for upstream clusters?
It looks pretty likely from the command run below, but just to be sure.

e2e/testenv/fail.go
```diff
@@ -0,0 +1,35 @@
...
+	} else {
+		pwd = path.Join(os.TempDir(), "fleet-gather")
+	}
+	path := path.Join(pwd, ginkgo.CurrentSpecReport().FullText())
+
+	ginkgo.GinkgoWriter.Printf("   Gathering cluster info for '%s' to '%s'...\n", ginkgo.CurrentSpecReport().FullText(), pwd)
```
weyfonk, line 25: I see squares here, presumably a fonts issue; what are these chars meant to represent? Perhaps we could use more common characters?

weyfonk (dismissed):

e2e/testenv/fail.go
manno: These are little bugs...

e2e/testenv/fail.go
manno: I didn't want to make the fail handler downstream cluster aware. Downstreams are just dumped in the workflows for now.

weyfonk (approved):

PR #2847, opened 2024-09-13
Fleet deploy CLI always outputs list of deployed resources

manno:
Previously `--dry-run` would output the list of resources and a deploy would output the resources struct, which contained the list of resources, too.

This also optimized Deployer.Deploy in the agent not to unpack (`objs, err := yaml.ToObjects`) the helm manifest needlessly.

### DIFF
--- a/internal/helmdeployer/template.go
+++ b/internal/helmdeployer/template.go
@@ -52,10 +52,10 @@ func Template(ctx context.Context, bundleID string, manifest *manifest.Manifest,
 	h.globalCfg.Log = logrus.Infof
 	h.globalCfg.Releases = storage.Init(mem)
 
-	resources, err := h.Deploy(ctx, bundleID, manifest, options)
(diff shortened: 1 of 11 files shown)


internal/helmdeployer/template.go
```diff
@@ -57,5 +57,5 @@ func Template(ctx context.Context, bundleID string, manifest *manifest.Manifest,
 		return nil, err
 	}
 
-	return resources.Objects, nil
+	return ReleaseToObjects(resources)
```
weyfonk, line 60: nit: renaming `resources` to `release` could clarify this code as already done around other calls to `Deploy`.

weyfonk (approved):

internal/helmdeployer/template.go
manno: Thanks:)

weyfonk (approved):

Issue #2716, opened 2024-08-08
"helm/resource-policy" was set to keep when installing certain helm charts via fleet, for unknown reason.


aDisplayName:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When using fleet to deploy the following fleet charts, workload resources are annotated with `helm/resource-policy: keep`, but we could not figure out why. By searching the source code of their helm charts, we can only see the crds are annotated with such setting.
| app | helm repo | helm charts | GitHub repository |
|-|-|-|-|
|argo workflows| https://argoproj.github.io/argo-helm | argo-workflows:0.41.14 | https://github.com/argoproj/argo-helm/tree/main/charts/argo-workflows |
| cert-manager | https://charts.jetstack.io | cert-manager:v1.15.2 | https://github.com/cert-manager/cert-manager/tree/master/deploy/charts/cert-manager | 
| bitnami/cert-manager |  oci://registry-1.docker.io/bitnamicharts/cert-manager | :1.3.16| https://github.com/bitnami/charts/blob/main/bitnami/cert-manager/ |

Using `argo-workflows:0.41.14` as an example, the only resources annotated with `helm.sh/resource-policy: keep` are those under `/templates/crds`: https://github.com/search?q=repo%3Aargoproj%2Fargo-helm+path%3A%2Fcharts%2Fargo-workflows%2F**+%22resource-policy%22&type=code

But when installing via fleet config: 
```yaml
defaultNamespace: argo

helm:
  # Use a custom location for the Helm chart. This can refer to any go-getter URL.
  # This allows one to download charts from most any location.  Also know that
  # go-getter URL supports adding a digest to validate the download. If repo
  # is set below this field is the name of the chart to lookup
  chart: argo-workflows

  repo: https://argoproj.github.io/argo-helm

  version: 0.41.14
  releaseName: argo-workflows
  values:
    fullnameOverride: argo-workflows
    server:
      extraArgs: [--auth-mode=server]
    images:
      pullPolicy: IfNotPresent
    crds:
      keep: false # avoid argo conflicts between different gitrepo due to ownership annotation
```

The resulted deployments are annotated with `helm.sh/resource-policy` set to `keep`
```yaml
metadata:
  name: argo-workflows-server
  annotations:
    deployment.kubernetes.io/revision: '1'
    helm.sh/resource-policy: keep
    meta.helm.sh/release-name: argo-workflows
    meta.helm.sh/release-namespace: argo
[...]

PR #2850, opened 2024-09-16
Fix helm.sh/resource-policy being added to everything and not just CRDs

SuperTux88:
Refers to #2716

When `deleteCRDResources` was disabled (default), the `helm.sh/resource-policy` annotation was added to more resources than just CRDs, because as soon as there was one CRD, the `annotations` variable was modified which also gets used by other resources in the loop (resources that were processed before the first CRD are fine). The fix is now to only modify a copy of the annotations that doesn't get reused for other resources.

This same bug also exists in the 0.9 branch in the [`deployer.go`](https://github.com/rancher/fleet/blob/release/v0.9/internal/helmdeployer/deployer.go#L170-L175). Should I also create a separate PR for that branch?

### DIFF
--- a/internal/helmdeployer/postrender.go
+++ b/internal/helmdeployer/postrender.go
@@ -83,12 +83,13 @@ func (p *postRender) Run(renderedManifests *bytes.Buffer) (modifiedManifests *by
 		if err != nil {
 			return nil, err
 		}
+		objAnnotations := mergeMaps(m.GetAnnotations(), annotations)
 		if !p.opts.DeleteCRDResources &&
 			obj.GetObjectKind().GroupVersionKind().Kind == CRDKind {
-			annotations[kube.ResourcePolicyAnno] = kube.KeepPolicy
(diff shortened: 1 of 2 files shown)


manno (approved):

SuperTux88:
Thanks for merging and the backport to 0.10.

> Do you need a backport to v0.9

We don't need a backport to 0.9, as our next step will be to upgrade to 0.10 anyway. I just wanted to mention that the same issue also exists in 0.9, as the original issue I references was made for 0.9, so maybe others would need a backport to 0.9?

aDisplayName:
> Thanks for merging and the backport to 0.10.
> 
> > Do you need a backport to v0.9
> 
> We don't need a backport to 0.9, as our next step will be to upgrade to 0.10 anyway. I just wanted to mention that the same issue also exists in 0.9, as the original issue I references was made for 0.9, so maybe others would need a backport to 0.9?

Thank you guys. It would be great if it can be backported to 0.9, as our production environment are not ready to move to rancher 2.9.x yet

PR #2854, opened 2024-09-16
Vendor apply.DryRun

manno:
Fleet's helmdeployer used the [wranger/apply](https://github.com/rancher/wrangler/tree/master/pkg/apply) module to do a "DryRun" to find modified objects.
Apply is very configurable, however Fleet only used a specific set of options. The relevant code paths to build the "plan" have been copied into the "deployer" packager. Several options, parameters, interfaces and switches have been removed.

* copy helmdeployer's pkgs from wrangler
* summary, had to move to the API pkg, as API cannot import fleet
* apply "patcher" was always replaced with an "append" function, patchers were never used
* delete is never used since ErrReplaced is never returned
* two structs implement "Apply" for the sugared interface, we only need
  one as "apply" is embedded in "desiredset"
* "original" is never used in patches
* rename applied package to desiredset
* fix/exclude linter warnings
* remove context from struct, pass in context directly
* replace logrus debug messages and increase verbosity

## Notes

* this affects how detect drift finds deleted resources
* this affects how modified resources are detected
* the `patch:` field is populated by the "dryrun" code, but always overwritten by the ["diff" code](https://github.com/rancher/fleet/blob/54f46c8396deba271511d866155b265eac0294b7/internal/cmd/agent/deployer/applied/diff.go#L76). It's only calculated to fill the `plan.Update` struct. That struct is similar to the `plan.Objects' struct and `Diff` already skips non modified resources. Data structures are different enough to make an optimization complicated, though. So removing `plan.Update` is not part of this PR.

```
  modifiedStatus:
  - apiVersion: v1
    kind: ConfigMap
    name: test-simple-manifest-config
    namespace: simple
    patch: '{"data":{"test":"value"}}'
```


p-se (dismissed):
lgtm

Issue #2858, opened 2024-09-17
Fix flaky drift correction tests


weyfonk:
Drift correction end-to-end tests are notoriously flaky. They currently get in the way of smoothly running CI workflows.

[Here](https://github.com/rancher/fleet/actions/runs/10875601151?pr=2850) is an example of such failures, which mostly involve [this test case](https://github.com/rancher/fleet/blob/main/e2e/drift/drift_test.go#L204).

weyfonk:
First hypothesis: the Fleet agent's garbage collection, acting on Helm releases missing a bundle deployment, might interfere with drift correction which installs releases, upgrades them and rolls them back.
This has been tested through #2861, by:
1. Deploying Fleet with a longer garbage collection interval, which was then expected to reduce the likelihood of forced drift correction failing
2. Disabling garbage collection altogether

In both cases, drift correction would still fail on the 1st or 2nd run (unlike locally, where it always passes), suggesting that garbage collection is not the culprit here.

Issue #2771, opened 2024-08-22
Convert end-to-end tests into more lightweight alternatives


weyfonk:
Some of our end-to-end tests suites could be made more lightweight, for instance by converting them into integration test suites. This would reduce CI runtime and likelihood of flakiness.
Candidates include:
* [x] [drift correction](https://github.com/rancher/fleet/tree/main/e2e/drift)
* [x] [finalizers](https://github.com/rancher/fleet/blob/main/e2e/single-cluster/finalizers_test.go)

PR #2861, opened 2024-09-17
Convert drift e2e tests into integration tests

weyfonk:
This commit is meant to test a hypothesis about flakiness of drift correction tests coming from a race condition between:
* the Fleet agent's Helm release cleanup mechanism, which deletes releases for which no bundle exists, at a configurable frequency (previously set to 1s for testing purposes)
* deletion of bundle deployments, re-created by Fleet controllers when a `GitRepo` is updated.

Release cleanup end-to-end tests, which depend on a short garbage collection interval, are skipped to prevent noise caused by unrelated test failures.

Refers to https://github.com/rancher/fleet/issues/2858
Refers to #2771.


manno (approved):
Please squash :)

Issue #2829, opened 2024-09-11
[v0.10] [SURE-8993] Backport of CABundle secret not created for gitjob


manno:
Backport of #2824 

Pull request content for Rancher QA verification.

 
### Problem

When a user specifies a CA in a GitRepo resource, the CA data should be put into a secret. The job that clones the repo and creates the bundle mounts the secret. 
The secret was never created. However, the secret was mounted on the job's pod, so the job would never start. K8s will emit an event like this:

`fleet-local          46s               Warning        FailedMount              pod/simple-aa0c4-4lrcx`

> MountVolume.SetUp failed for volume "additional-ca" : secret "simple-cabundle" not found

### Solution

The code to create the secret was re-added.
 
#### Manual Testing

We created the secret manually and observed the job to complete. We then added the code to create it, added a GitRepo with an arbitrary CA bundle and the job completed.
After deleting the GitRepo, the secret is removed, too.

#### Automated Testing

We only test that gitcloner works with a CA, apparently we don't have a test that makes sure the gitops controller creates the secret for the job.

mmartin24:
Tested via UI in Rancher `v2.9.2-alpha5` with Fleet `rancher/fleet:v0.10.2-rc.4`

--- 

Testing Steps:
- In Rancher, create local repo, for example:
```
URL: https://github.com/rancher/fleet-examples
Branch: master
Path: simple
```
- Create CA bundle by adding in option `Select TLS Certificate Verification -> Specified additional certificates to be accepted` and  example Certificate as [this](https://ospkibook.sourceforge.net/docs/OSPKI-2.4.7/OSPKI-html/sample-ca-cert.htm) one.
- Create git repo 

---

- Checked CABundle secret is created and Gitrepo successfully deployed both in local and downstream clusters. In the screenshot can be seen before / after fix:
**Before fix:**
![before_cabundle_fix](https://github.com/user-attachments/assets/50e42a0d-95e1-4709-9c65-0986567c54df)
**After fix:**
![ok_ca_bundle](https://github.com/user-attachments/assets/ec55eafb-ac40-428d-bf44-1e468f25a655)


- Checked secret is removed after Gitrepo deletion
- Checked after non-working example with fleet:v0.10.0 upgrading to `rancher/fleet:v0.10.2-rc.4` and forcing update it works.

--- 
Further steps:
Added this example into QASE under https://app.qase.io/case/FLEET-142 to be automated
[...]

PR #2863, opened 2024-09-17
[v0.10] Add integration test checks for CA bundle secret

weyfonk:
Backport of #2846 to `release/v0.10`.
Refers to #2829.

When a git job is created for a `GitRepo` with a configured CA bundle, integration tests now validate that a secret is created with the contents of that bundle and validate its contents against the `GitRepo` spec.

This also validates through integration tests that RBAC resources and the CA bundle secret are owned by their `GitRepo` and will therefore be deleted as soon as the `GitRepo` is.
Validation does not actually delete anything, due to a limitation [1] of the test framework.

[1]: https://book.kubebuilder.io/reference/envtest#testing-considerations


0xavi0 (approved):
lgtm

PR #2882, opened 2024-09-19
Copy SafeConcat from wrangler

manno:
Extend our existing name package.

### DIFF
--- a/internal/cmd/cli/apply/apply.go
+++ b/internal/cmd/cli/apply/apply.go
@@ -18,7 +18,7 @@ import (
 	"github.com/rancher/fleet/internal/client"
 	"github.com/rancher/fleet/internal/fleetyaml"
 	"github.com/rancher/fleet/internal/manifest"
-	name2 "github.com/rancher/fleet/internal/name"
+	"github.com/rancher/fleet/internal/names"
 	"github.com/rancher/fleet/internal/ociwrapper"
 	fleet "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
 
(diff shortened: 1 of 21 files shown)


internal/name/safeconcat.go
```diff
@@ -0,0 +1,25 @@
+package name
```
aruiz14, line 1: What about naming it `names` instead? It will avoid the collision between the import name and frequent `name` variables, in addition to follow the convention from the stdlib (`strings`, `maps`, `slices`).

internal/cmd/controller/grutil/targetsyaml.go
```diff
@@ -3,11 +3,10 @@ package grutil
 import (
 	"encoding/json"
 
+	"github.com/rancher/fleet/internal/name"
 	fname "github.com/rancher/fleet/internal/name"
```
aruiz14, line 7: Duplicated imports

internal/name/safeconcat.go
aruiz14: Sorry, I just realized the package already existed before 🙃

manno: The package exists already, but I like the idea.

internal/cmd/cli/apply/apply.go
```diff
@@ -18,7 +18,7 @@ import (
 	"github.com/rancher/fleet/internal/client"
 	"github.com/rancher/fleet/internal/fleetyaml"
 	"github.com/rancher/fleet/internal/manifest"
-	name2 "github.com/rancher/fleet/internal/name"
+	name2 "github.com/rancher/fleet/internal/names"
```
aruiz14, line 21: `s/name2/names/`?

internal/helmdeployer/deployer.go
```diff
@@ -12,7 +12,7 @@ import (
 	"helm.sh/helm/v3/pkg/storage"
 	"helm.sh/helm/v3/pkg/storage/driver"
 
-	name2 "github.com/rancher/fleet/internal/name"
+	name2 "github.com/rancher/fleet/internal/names"
```
aruiz14, line 15: `s/name2/names/`?

internal/cmd/cli/apply/apply.go
manno: Yes, I only looked at `fname`...

aruiz14 (approved):
LGTM!

PR #2892, opened 2024-09-24
Skip drift detection controller refresh if disabled

weyfonk:
Refreshing the drift mini controller does not make sense on a bundle deployment which does not have drift detection enabled, and just wastes computing resources.


weyfonk:
This would disable drift _detection_, which does not make much sense at this point as it would break status updates. Closing.

PR #2897, opened 2024-09-25
Increase worker count for reconcilers

manno:
This should reflect the old settings from 0.9


dev/setup-fleet
```diff
@@ -40,6 +40,9 @@ helm -n cattle-fleet-system upgrade --install --create-namespace --wait --reset-
   --set apiServerCA="$ca" \
   --set apiServerURL="$server" \
   $shards_settings \
+  --set controller.reconciler.workers.gitrepo=50 \
+  --set controller.reconciler.workers.bundle=50 \
+  --set controller.reconciler.workers.bundledeployment=50 \
```
weyfonk, line 45: If we need it here (do we, if those are the new default values?), then we'd need it in `.github/scripts/deploy-fleet.sh` as well so that it applies to our testing workflows.

manno: We don't need, but I wanted to expose if for testing. Should I remove that?

charts/fleet/templates/deployment.yaml
```diff
@@ -69,10 +69,6 @@ spec:
         - name: CATTLE_DEV_MODE
           value: "true"
         {{- end }}
-        {{- if $.Values.controller.reconciler.workers.gitrepo }}
```
weyfonk, line 72: Nice catch, this has indeed been moved over to the `gitjob` deployment :+1:

dev/setup-fleet
weyfonk: Oh I see, thanks for clarifying. No need then.

weyfonk (approved):

weyfonk (approved):

internal/cmd/controller/reconciler/cluster_controller.go
```diff
@@ -100,6 +101,7 @@ func (r *ClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			}),
 		).
 		WithEventFilter(sharding.FilterByShardID(r.ShardID)).
+		WithOptions(controller.Options{MaxConcurrentReconciles: 50}).
```
weyfonk, line 104: Now I think that at least this one should also be made configurable.

manno: Yes, that's a good follow-up story.

Issue #2870, opened 2024-09-18
Jobs are not cleaned up from local cluster


mikmatko:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

In Rancher local cluster, for each commit/change in each `GitRepo`, there is a `Job` started by Fleet. There is nothing to clean up these Jobs, so you will quickly end up with hundreds of lingering `Job` objects and their completed Pods.

I didn't notice this behavior in Fleet 0.9.x, so I assume something in 0.10.x introduced these Jobs. I was assuming this is related to automatic chart dependency update, but setting `disableDependencyUpdate` to `true` doesn't seem to affect.

### Expected Behavior

Unnecessary Job objects are cleaned up, e.g. by setting some sane default for `.spec.ttlSecondsAfterFinished`: https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/

### Steps To Reproduce

1. Install Rancher & Fleet
2. Add any GitRepo and make sure it deploys
3. Check rancher-local cluster. You now have lingering `Job` objects

### Environment

```markdown
- Architecture: x86
- Fleet Version: v0.10.2
- Cluster:
  - Provider: GKE
  - Options: Rancher 2.9.1
  - Kubernetes Version: v1.30.4-gke.1213000
```

manno:
This can help to identify completed jobs that have not been cleaned up:

`kubectl get jobs --all-namespaces -o custom-columns='NAMESPACE:.metadata.namespace,NAME:.metadata.name,OWNER:.metadata.ownerReferences[].name,STATUS:.status.succeeded'`

manno:
- [ ] we need to turn https://github.com/rancher/fleet/pull/2907 into a run-once job when #2903 is merged.
- [ ] make sure it's actually run when upgrading 0.10.2->0.10.4

0xavi0:
## Additional QA
### Problem
Fleet is not deleting the jobs related to `GitRepos`. 
We create a new job for every new commit we get in the git repository, which is a problem in systems with many `GitRepos` and many commits because we could reach the `etcd` limits. 

### Solution
* Fleet will create a new job when is needed and will delete it after it succeeds 
* In case of error the job won't be deleted (so we can describe the job, check the logs, etc)
* If a job is not finished and the user changes the Spec or force updates or a new commit is received, the job running will be deleted and a new one will be created. 

### Testing

Test a few scenarios so cover all the possible cases

[...]

PR #2903, opened 2024-09-26
Changes job handling in gitops controller

0xavi0:
Changes the job handling so jobs are deleted after they succeed. Jobs are created when they are needed, this avoids having always a minimum of 1 job per `gitrepo` resource.

It also deletes the Owns(job) statement from the reconciler setup as we're already setting the controller reference when creating the job and the Owns statement makes extra reconciler calls that are not needed.

Refers to https://github.com/rancher/fleet/issues/2870

### DIFF
--- a/integrationtests/gitjob/controller/controller_test.go
+++ b/integrationtests/gitjob/controller/controller_test.go
@@ -21,7 +21,6 @@ import (
 	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	"github.com/rancher/wrangler/pkg/name"
 	"github.com/rancher/wrangler/v3/pkg/genericcondition"
 
 	"github.com/rancher/fleet/integrationtests/utils"
@@ -174,9 +173,10 @@ var _ = Describe("GitJob controller", func() {
 					g.Expect(checkCondition(&gitRepo, "Accepted", corev1.ConditionTrue, "")).To(BeTrue())
 				}).Should(Succeed())
 
-				// it should log 2 events
+				// it should log 3 events
 				// first one is to log the new commit from the poller
 				// second one is to inform that the job was created
+				// third one is to inform that the job was deleted because it succeeded
 				Eventually(func(g Gomega) {
 					events, _ := k8sClientSet.CoreV1().Events(gitRepo.Namespace).List(context.TODO(),
 						metav1.ListOptions{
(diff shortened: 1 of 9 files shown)


integrationtests/gitjob/controller/controller_test.go
```diff
@@ -373,9 +377,51 @@ var _ = Describe("GitJob controller", func() {
...
+					{
+						Type:   "Complete",
+						Status: "True",
+					},
+				}
+				return k8sClient.Status().Update(ctx, &job)
+			}).Should(Not(HaveOccurred()))
+			// wait until the job has finished
+			Eventually(func() bool {
+				jobName = name.SafeConcatName(gitRepoName, name.Hex(repo+commit, 5))
```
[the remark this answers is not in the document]
0xavi0: changed

integrationtests/gitjob/controller/controller_test.go
```diff
@@ -183,7 +184,7 @@ var _ = Describe("GitJob controller", func() {
 							TypeMeta:      metav1.TypeMeta{Kind: "GitRepo"},
 						})
 					g.Expect(events).ToNot(BeNil())
-					g.Expect(len(events.Items)).To(Equal(2))
+					g.Expect(len(events.Items)).To(Equal(3))
```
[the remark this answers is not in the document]
0xavi0: yup, updated and added the new expected event

pkg/apis/fleet.cattle.io/v1alpha1/gitrepo_types.go
```diff
@@ -170,6 +170,9 @@ type GitRepoStatus struct {
 	// Commit is the Git commit hash from the last git job run.
 	// +nullable
 	Commit string `json:"commit,omitempty"`
+	// WebhookCommit is the Git commit hash from the last git commit received from a webhook
```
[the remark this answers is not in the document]
0xavi0: changed

pkg/apis/fleet.cattle.io/v1alpha1/gitrepo_types.go
```diff
@@ -170,6 +170,9 @@ type GitRepoStatus struct {
 	// Commit is the Git commit hash from the last git job run.
 	// +nullable
 	Commit string `json:"commit,omitempty"`
+	// WebhookCommit is the latest Git commit hash received from a webhook
+	// +nullable
```
weyfonk, line 174: In light of #2901, I wonder if we might want to make this field `+optional` instead of `+nullable`, to prevent it from appearing with an empty value on `GitRepo`s which rely solely on polling. WDYT?

0xavi0: I think that being a commit hash... it makes sense to have the value as `+nullable` rather than being an empty string. 
So either we have a valid commit hash or we have nothing.

weyfonk: > either we have a valid commit hash or we have nothing

Yes, precisely, if you meant to say that this field should either contain a valid value, or not appear at all; I think that this is what `+optional` achieves, not `+nullable`, as per [these docs](https://book.kubebuilder.io/reference/markers/crd-validation.html).

0xavi0: oh, right, in that case I'll make this one and the normal commit both non `+nullable`

PR #2917, opened 2024-09-30
Limit `Deployed bundle` logs

weyfonk:
This prevents a log message from being produced even when a bundle deployment and its Helm release were already installed in their latest available versions.

Related to #2869


manno (approved):

kkaempf:
Assuming this is planned for 2.10 ?! 🤔

Issue #2904, opened 2024-09-26
Skip `cabundle` secrets when Gitrepo does not include `GitRepo's .Spec.CABundle`


mmartin24:
## Description
On Rancher `2.9.-head` we have introduced automatic cabundle secret creation when any gitrepo is deployed (perhaps after  https://github.com/rancher/fleet/pull/2831?)

In this screenshot we can see how upon creation of 60 gitrepos without any certificate passed, it creates `60` cabundles secrets in `2.9.2` while `0` in `2.9.1`:

### Request
We should only create this cabundle secret when a certificate is passed

### Steps to reproduce
- Deploy Rancher 2.9-head (> `v2.9-3511f1655eaf53f86df3d87b9d4ff6f754d18d80-head`)
- Create normal gitrepo without special certificate. For example this gitrepo containing a `fleet.yaml` defining a `defaultNamespace` and a deployment file with `nginx` image: https://github.com/rancher/fleet-test-data/tree/master/qa-test-apps/nginx-app
- Observe how cabundle is deployed.

### The expected behavior would be
- not to deploy the cabundle secret in this case. @weyfonk , pointed out to perhaps skip secret creation if `GitRepo's .Spec.CABundle` field is empty
- be able to deploy it when an aditional certificate as [this](https://ospkibook.sourceforge.net/docs/OSPKI-2.4.7/OSPKI-html/sample-ca-cert.htm) pem file is passed along when creating the gitrepo

weyfonk:
## Additional QA
### Problem
Each git job would be created with a `*-cabundle`  secret for the `GitRepo`, even when that `GitRepo` did not specify any `.Spec.CABundle` field.

### Solution
Only create that secret when a non-empty `.Spec.CABundle` field exists.
 
### Testing

### Engineering Testing
#### Manual Testing
None.

#### Automated Testing
Updated integration tests to verify that a CA bundle secret is only created when the `GitRepo` has a non-empty `.Spec.CABundle` field.

### QA Testing Considerations
This should be tested following the reproduction steps above.
 
#### Regressions Considerations
N/A

mmartin24:
Verified in `v2.10-fe49760f4e50d0b78ca0102c8475bc93361336b0-head` with `fleet:105.0.0+up0.11.0-beta.3`  along [here](https://github.com/rancher/fleet/issues/2824#issuecomment-2444879410).

- ca-bundle secret IS NOT created when TLS cert is not added

PR #2918, opened 2024-09-30
Skip CA bundle secret creation with empty payload

weyfonk:
When no `caBundle` field is configured on a `GitRepo`, the gitOps controller now skips creation of the corresponding secret.

Refers to #2904.


0xavi0 (approved):
LGTM

Issue #594, opened 2021-11-05
If any clusters are offline, bundle status can get stuck


philomory:
If any clusters are offline/unavailable, the status of Bundles that get deployed to those clusters can get stuck with misleading/confusing error messages.

# Steps to reproduce:
1. Create a fleet workspace, and add at least two clusters. Make one cluster an e.g. single-node k3s cluster, so that it is easy to start/stop the cluster. We'll call that cluster "A".
2. Create a ClusterGroup that contains at least two of your clusters, including cluster A. We'll call this group "G".
3. Create a git repository containing the following code:
    ```yaml
    # fleet.yaml
    defaultNamespace: test-bundle
    helm:
      chart: ./chart
      releaseName: test-bundle

    # chart/Chart.yaml
    apiVersion: v2
    name: test-bundle
    version: 0.0.1

    # chart/templates/deployment.yaml
    apiVersion: apps/v1
    kind: Deployment
    metadata:
      name: test
    spec:
      selector:
        matchLabels:
          app: test
      template:
        metadata:
          labels:
            app: test
        spec:
          containers:
            - name: test
              image: paulbouwer/hello-kubernetes:1.10.1
              lifecycle:
                preStop:
                  exec:
                    command: ["sleep", "1"]
    ```
3. Create a GitRepo object pointing to this repository that applies to ClusterGroup "G". We'll call this GitRepo "R".
4. Let Fleet deploy the bundle from R to all clusters in G. Wait for them all to be Ready.
5. Push the following change to the git repository:
    ```
    --- a/chart/templates/deployment.yaml
    +++ b/chart/templates/deployment.yaml
    @@ -20,6 +20,6 @@ spec:
             - name: test
               image: paulbouwer/hello-kubernetes:1.10.1
               lifecycle:
    -            preStop:
    +            preStart:
                   exec:
                     command: ["sleep", "2"]
    ```
6. Wait for Fleet to attempt to deploy this change. The GitRepo, Bundle, and all relevant BundleDeployments should end up in the state `ErrApplied`, with an error message similar to `error validating "": error validating data: ValidationError(Deployment.spec.template.spec.containers[0].lifecycle): unknown field "preStart" in io.k8s.api.core.v1.Lifecycle'`.
7. Shut off cluster A (stop the only node in the cluster).
8. Wait until Cluster A shows as offline under Cluster Management in Rancher.
9. Push the following change to GitRepo G:
    ```
    --- a/chart/templates/deployment.yaml
[...]

PR #2933, opened 2024-10-07
Detect offline clusters

weyfonk:
This adds a cluster status monitor to the Fleet controller, which checks when each cluster last saw its agent online. If more than the expected interval elapses, that cluster is considered offline, and the monitor updates its bundle deployments' statuses to reflect that. This will trigger status updates to bundles, `GitRepo`s, clusters and cluster groups.

Refers to #594.

Open points:
* how far, and how fine-grained, do we want to make bundle deployment status updates for offline clusters? This currently takes a fairly basic approach, updating both `Ready` and `Monitored` conditions while clearing modified and non-ready statuses, to prevent outdated messages from appearing in a bundle deployment's display status and further up the chain of status updates (to bundles, then upwards to `GitRepo`s, etc)
* should we make the frequency of monitoring configurable?
* do we have a way to exclude the local/management cluster from agent-last-seen checks?

### Update 2026-03
* the cluster controller checks for offline bundle deployments, to mark the corresponding cluster as offline. It also supports the opposite, namely marking a cluster back online when its bundle deployments are detected as such.
* this is covered by integration tests, not involving the cluster monitor, while end-to-end tests validate this behaviour when the cluster monitor is involved.

### DIFF
--- a/internal/cmd/controller/clustermonitor/monitor.go
+++ b/internal/cmd/controller/clustermonitor/monitor.go
@@ -2,7 +2,7 @@ package clustermonitor
 
 import (
 	"context"
-	"fmt"
+	"errors"
 	"math"
 	"time"
 
@@ -15,6 +15,8 @@ import (
 	"github.com/rancher/fleet/internal/config"
 	"github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
 )
+
+const offlineMsg = "cluster is offline"
 
 // Run monitors Fleet cluster resources' agent last seen dates. If a cluster's agent was last seen longer ago than
 // a certain threshold, then Run updates statuses of all bundle deployments targeting that cluster, to reflect the fact
@@ -45,6 +47,9 @@ func Run(ctx context.Context, c client.Client, interval, threshold time.Duration
 	}
 }
 
+// UpdateOfflineBundleDeployments looks for offline clusters based on the provided threshold duration. For each cluster
+// considered offline, this updates its bundle deployments' statuses accordingly.
+// If a cluster's bundle deployments have already been marked as offline, they will be skipped.
 func UpdateOfflineBundleDeployments(ctx context.Context, c client.Client, threshold time.Duration) {
 	logger := ctrl.Log.WithName("cluster status monitor")
 
@@ -58,10 +63,6 @@ func UpdateOfflineBundleDeployments(ctx context.Context, c client.Client, thresh
 		lastSeen := cluster.Status.Agent.LastSeen
 
 		logger.Info("Checking cluster status", "cluster", cluster.Name, "last seen", lastSeen.UTC().String())
-
-		// XXX: do we want to run this more than once per cluster, updating the timestamp each time?
-		// Or would it make sense to keep the oldest possible timestamp in place, for users to know since when the
-		// cluster is offline?
 
 		// lastSeen being 0 would typically mean that the cluster is not registered yet, in which case bundle
 		// deployments should not be deployed there.
@@ -85,7 +86,19 @@ func UpdateOfflineBundleDeployments(ctx context.Context, c client.Client, thresh
 			continue
 		}
 
+	bd_update:
 		for _, bd := range bundleDeployments.Items {
+			for _, cond := range bd.Status.Conditions {
+				switch cond.Type {
+				case "Ready":
+					fallthrough
+				case "Monitored":
+					if cond.Message == offlineMsg {
+						break bd_update
+					}
+				}
+			}
+
 			logger.Info("Updating bundle deployment in offline cluster", "cluster", cluster.Name, "bundledeployment", bd.Name)
 			err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
 				t := &v1alpha1.BundleDeployment{}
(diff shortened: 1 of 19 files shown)


internal/cmd/controller/clustermonitor/monitor.go
```diff
@@ -0,0 +1,132 @@
...
+						// XXX: do we want to set Deployed and Installed conditions as well?
+						// XXX: should we set conditions to `Unknown`?
+					case "Monitored":
+						mc := monitor.Cond(v1alpha1.BundleDeploymentConditionMonitored)
+						mc.SetError(&bd.Status, "Cluster offline", fmt.Errorf("cluster is offline"))
+
+					}
+				}
+
+				return c.Status().Update(ctx, t)
```
0xavi0, line 116: Shouldn't it be setting the Error on `t` instead of `bd`?
The update call is done with `t`.

internal/cmd/controller/clustermonitor/monitor_test.go
```diff
@@ -0,0 +1,369 @@
...
+
+			if !strings.Contains(cond.Message, "offline") {
+				return false
+			}
+		} else if cond.Type == "Monitored" {
+			foundMonitored = true
+
+			if !strings.Contains(cond.Message, "offline") {
+				return false
+			}
```
0xavi0, line 53: Would be good to check that the condition's status is `False` when the `offline` string is found and `True` otherwise. 
Also would be good to check that the Reason in the Condition.

0xavi0 (commented):
I don't know if you want to merge this PR with the comments with open questions, maybe better to clarify first. 

Just a couple of observations, but overall LGTM

internal/cmd/controller/clustermonitor/monitor.go
weyfonk: Nice catch, thanks! This still behaved as expected because of references, but was clearly not ideal. Fixed.

internal/cmd/controller/clustermonitor/monitor_test.go
weyfonk: Thanks, checks added.

internal/cmd/controller/operator.go
```diff
@@ -238,6 +240,25 @@ func start(
 		return err
 	}
 
+	cfg := fleetcfg.Get()
+
+	if cfg.ClusterMonitor.Enabled && shardID == "" { // only one instance of the cluster status monitor needs to run.
+		setupLog.Info("starting cluster status monitor")
+		// No need to run a similar check on the threshold, since its minimum value will be a multiple of the agent check-in
+		// interval anyway.
+		if cfg.ClusterMonitor.Interval.Seconds() == 0 {
```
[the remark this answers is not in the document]
weyfonk: It _does_ compile, and work.

internal/cmd/controller/clustermonitor/monitor.go
```diff
@@ -0,0 +1,156 @@
...
+
+	bd_update:
+		for _, bd := range bundleDeployments.Items {
+			for _, cond := range bd.Status.Conditions {
+				switch cond.Type {
+				case "Ready":
+					fallthrough
+				case "Monitored":
+					if cond.Message == v1alpha1.ClusterOfflineMsg {
+						break bd_update
```
0xavi0, line 101: I'm not sure I understand this.
Isn't this break statement also breaking the outer loop?

internal/cmd/controller/reconciler/cluster_controller.go
```diff
@@ -222,6 +245,28 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+			}
+
+		} else if summary.MessageFromCondition(fleet.ClusterConditionReady, cluster.Status.Conditions) == fleet.ClusterOfflineMsg {
+			// bundle deployment online, mark offline cluster back online
+			for _, c := range offlineConds {
+				cond := monitor.Cond(c)
+				cond.SetError(&cluster.Status, "", nil)
+			}
+		}
 	}
```
0xavi0, line 270: nit: If there are 2 Bundle deployments and they are in the slice in the following order:

bdA (offline)
bdC (not offline)

bdA will be processed and will call `cond.SetError(&cluster.Status, "Cluster offline", errors.New(fleet.ClusterOfflineMsg))`

bdB will be then processed and will cleanup the previous status, right?


I understand this should be temporary and at some point both BDs will report as "cluster offline" if the cluster is really offline, but I guess that it's maybe better to set the status to "offline" if any of the BDs is offline.

0xavi0 (commented):
lgtm overall, just a couple of questions.

internal/cmd/controller/clustermonitor/monitor.go
weyfonk: Yes, I seem to recall the idea was that if _one_ of the bundle deployments was already marked offline, then we could assume that _all_ bundle deployments for that cluster had been marked offline by a previous iteration, in which case we could break out of the outer loop, as we wouldn't need to update the rest of bundle deployments.

Come to think of it, there's still a risk that one or more new bundle deployment(s) might have been created since the previous run of this loop, hence the above assumption might not hold.
Therefore, I've changed this `break` statement to a `continue`, which simply skips to the next bundle deployment.

internal/cmd/controller/reconciler/cluster_controller.go
weyfonk: Good point; I've simplified this to mark the cluster offline if _any_ bundle deployment is offline, and to only mark it back online if _all_ BDs are back online. This should prevent flapping cluster status updates.

weyfonk:
Closing this for now, as the current [approach](https://github.com/rancher/fleet/blob/main/internal/cmd/controller/summary/summary.go#L91-L95) of reporting deployed, but obsolete bundle deployments as `WaitApplied` instead of `ErrApplied` is considered more efficient than an additional offline cluster detector.

PR #2939, opened 2024-10-07
[v0.8] Bump gitjob version to v0.8.13

thardeck:


charts/fleet/charts/gitjob/values.yaml
```diff
@@ -1,10 +1,10 @@
 gitjob:
   repository: rancher/gitjob
-  tag: v0.8.12
+  tag: v0.8.13
 
 tekton:
   repository: rancher/tekton-utils
-  tag: v0.1.46
+  tag: v0.1.47
```
weyfonk, line 7: As discussed, for reference: while both of those tags point to the same commit, creating a new tag triggers a new build and release, using the latest base image with updated OS packages.

weyfonk (approved):

PR #2940, opened 2024-10-08
dev-scripts: Wait for ingress to appear before waiting for load balancer

p-se:
Prevents

```
Error from server (NotFound): the server could not find the requested resource
```

while deploying.


weyfonk (approved):

manno:
Wouldn't this also affect our CI? So should it go into the wait-for-loadbalancer script?

Issue #2916, opened 2024-09-30
Remove "driftdetect" bundledeployment update workaround from agent


manno:
Currently the agent needs to update the bundledeployment to trigger drift detection and correction. To reduce the number of events this event, sent from the "driftdetect" mini-controller to the controller-runtime controller will be converted into a channel.

PR #2942, opened 2024-10-08
Remove update workaround from drift detection 

manno:
Refers to https://github.com/rancher/fleet/issues/2916

This adds a new reconciler, which reacts to a channel source, to remove
the bd update workaround. The driftdetect mini controller previously
signaled change via the bundledeployment's status to the
controller-runtime based reconciler.
The bundledeployment reconciler is now split into two, one
reconciler to install bundles and another to handle drift.
Both can update the status of the bundledeployment resource with the
current state of the deployment.

### DIFF
--- a/integrationtests/agent/suite_test.go
+++ b/integrationtests/agent/suite_test.go
@@ -3,6 +3,7 @@ package agent_test
 import (
 	"context"
 	"fmt"
+	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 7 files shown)


weyfonk (commented):
Nice one, thanks for this!
LalmostGTM

weyfonk (dismissed):

weyfonk (approved):

PR #2945, opened 2024-10-09
Fix job deletion when generation is not updated on time

0xavi0:
Changes the way jobs are handled when `ForceSyncGeneration` or `Generation` change in the `GitRepo`.

It adds 2 new labels in the `job` with the actual values of `ForceSyncGeneration` and `Generation` in the `GitRepo`s spec. 
That way, if we encounter a race condition in which any of those 2 values are still not set in the database when a new reconcile is triggered, we can compare the value of the `job` against the expected generation values.

We would delete the job only when the values don't match.

### DIFF
--- a/internal/cmd/controller/gitops/reconciler/gitjob_controller.go
+++ b/internal/cmd/controller/gitops/reconciler/gitjob_controller.go
@@ -57,6 +57,8 @@ const (
 
 	defaultPollingSyncInterval = 15 * time.Second
 	gitPollingCondition        = "GitPolling"
+	generationLabel            = "generation"
+	forceSyncGenerationLabel   = "forceSyncGeneration"
 )
 
 var two = int32(2)
@@ -501,23 +503,33 @@ func (r *GitJobReconciler) deleteJobIfNeeded(ctx context.Context, gitRepo *v1alp
 	// the following cases imply that the job is still running but we need to stop it and
 	// create a new one
 	if gitRepo.Spec.ForceSyncGeneration != gitRepo.Status.UpdateGeneration {
-		jobDeletedMessage := "job deletion triggered because of ForceUpdateGeneration"
-		logger.Info(jobDeletedMessage)
-		if err := r.Delete(ctx, job, client.PropagationPolicy(metav1.DeletePropagationBackground)); err != nil && !errors.IsNotFound(err) {
-			return err, true
+		if forceSync, ok := job.Labels[forceSyncGenerationLabel]; ok {
+			strForceSync := fmt.Sprintf("%d", gitRepo.Spec.ForceSyncGeneration)
+			if strForceSync != forceSync {
+				jobDeletedMessage := "job deletion triggered because of ForceUpdateGeneration"
+				logger.Info(jobDeletedMessage)
+				if err := r.Delete(ctx, job, client.PropagationPolicy(metav1.DeletePropagationBackground)); err != nil && !errors.IsNotFound(err) {
+					return err, true
+				}
+				return nil, true
+			}
 		}
-		return nil, true
 	}
 
 	// k8s Jobs are immutable. Recreate the job if the GitRepo Spec has changed.
 	// Avoid deleting the job twice
 	if generationChanged(gitRepo) {
-		jobDeletedMessage := "job deletion triggered because of generation change"
-		logger.Info(jobDeletedMessage)
-		if err := r.Delete(ctx, job, client.PropagationPolicy(metav1.DeletePropagationBackground)); err != nil && !errors.IsNotFound(err) {
-			return err, true
+		if gen, ok := job.Labels[generationLabel]; ok {
+			strGen := fmt.Sprintf("%d", gitRepo.Generation)
+			if strGen != gen {
+				jobDeletedMessage := "job deletion triggered because of generation change"
+				logger.Info(jobDeletedMessage)
+				if err := r.Delete(ctx, job, client.PropagationPolicy(metav1.DeletePropagationBackground)); err != nil && !errors.IsNotFound(err) {
+					return err, true
+				}
+				return nil, true
+			}
 		}
-		return nil, true
 	}
 
 	// check if the job finished and was successful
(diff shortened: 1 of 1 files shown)


internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -57,6 +57,8 @@ const (
 
 	defaultPollingSyncInterval = 15 * time.Second
 	gitPollingCondition        = "GitPolling"
+	generationLabel            = "generation"
```
manno, line 60:
```suggestion
	generationLabel            = "fleet.cattle.io/gitrepo-generation"
```

I'd prefix with fleet and clarify what kind of generation is stored.

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -57,6 +57,8 @@ const (
...
+	forceSyncGenerationLabel   = "forceSyncGeneration"
```
manno, line 61:
```suggestion
	forceSyncGenerationLabel   = "fleet.cattle.io/force-sync-generation"
```

All our labels seem to use kebap-case

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -501,23 +503,33 @@ func (r *GitJobReconciler) deleteJobIfNeeded(ctx context.Context, gitRepo *v1alp
 	// the following cases imply that the job is still running but we need to stop it and
 	// create a new one
 	if gitRepo.Spec.ForceSyncGeneration != gitRepo.Status.UpdateGeneration {
-		jobDeletedMessage := "job deletion triggered because of ForceUpdateGeneration"
-		logger.Info(jobDeletedMessage)
-		if err := r.Delete(ctx, job, client.PropagationPolicy(metav1.DeletePropagationBackground)); err != nil && !errors.IsNotFound(err) {
-			return err, true
+		if forceSync, ok := job.Labels[forceSyncGenerationLabel]; ok {
+			strForceSync := fmt.Sprintf("%d", gitRepo.Spec.ForceSyncGeneration)
```
manno, line 507: I would not prepend the type. Since the scope is small, any short variable name will do.

```suggestion
			t := fmt.Sprintf("%d", gitRepo.Spec.ForceSyncGeneration)
```

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -501,23 +503,33 @@ func (r *GitJobReconciler) deleteJobIfNeeded(ctx context.Context, gitRepo *v1alp
...
 	// k8s Jobs are immutable. Recreate the job if the GitRepo Spec has changed.
 	// Avoid deleting the job twice
 	if generationChanged(gitRepo) {
-		jobDeletedMessage := "job deletion triggered because of generation change"
-		logger.Info(jobDeletedMessage)
-		if err := r.Delete(ctx, job, client.PropagationPolicy(metav1.DeletePropagationBackground)); err != nil && !errors.IsNotFound(err) {
-			return err, true
+		if gen, ok := job.Labels[generationLabel]; ok {
+			strGen := fmt.Sprintf("%d", gitRepo.Generation)
+			if strGen != gen {
```
manno, line 524: Again, no type prefix:
```suggestion
			if t != gen {
```

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -501,23 +503,33 @@ func (r *GitJobReconciler) deleteJobIfNeeded(ctx context.Context, gitRepo *v1alp
...
+				jobDeletedMessage := "job deletion triggered because of generation change"
+				logger.Info(jobDeletedMessage)
```
manno, line 526: I would log this in debug mode only:
```suggestion
				logger.V(1).Info(jobDeletedMessage)
```

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -501,23 +503,33 @@ func (r *GitJobReconciler) deleteJobIfNeeded(ctx context.Context, gitRepo *v1alp
...
 	if gitRepo.Spec.ForceSyncGeneration != gitRepo.Status.UpdateGeneration {
-		jobDeletedMessage := "job deletion triggered because of ForceUpdateGeneration"
-		logger.Info(jobDeletedMessage)
-		if err := r.Delete(ctx, job, client.PropagationPolicy(metav1.DeletePropagationBackground)); err != nil && !errors.IsNotFound(err) {
-			return err, true
+		if forceSync, ok := job.Labels[forceSyncGenerationLabel]; ok {
+			strForceSync := fmt.Sprintf("%d", gitRepo.Spec.ForceSyncGeneration)
+			if strForceSync != forceSync {
+				jobDeletedMessage := "job deletion triggered because of ForceUpdateGeneration"
+				logger.Info(jobDeletedMessage)
```
manno, line 510: I would log this in debug mode only, same for the other "job deleted" messages:
```suggestion
				logger.V(1).Info(jobDeletedMessage)
```

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
manno, line 524: What if strGen > gen? Does that happen?

0xavi0: Yes, that does happen. 
It is the case where the job will be deleted because it means the generation in the `gitrepo` spec is greater (something changed in the spec) than the generation value we do have in the `job` label.

It should never be less than because generation should be increasing always.

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -563,6 +575,10 @@ func (r *GitJobReconciler) newGitJob(ctx context.Context, obj *v1alpha1.GitRepo)
 				"generation": strconv.Itoa(int(obj.Generation)),
 				"commit":     obj.Status.Commit,
 			},
+			Labels: map[string]string{
+				forceSyncGenerationLabel: fmt.Sprintf("%d", obj.Spec.ForceSyncGeneration),
```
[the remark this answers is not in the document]
0xavi0: ForceSyncGeneration is an `int64`, it is no longer a pointer, so it should default to 0, right?

manno (approved):

PR #2952, opened 2024-10-10
[v0.10] Cleanup jobs back to cronjob

0xavi0:
In case of corner cases in which succeeded jobs remain for whatever reason (most likely events loss due to heavy loaded system) it is useful to still run the cleanup job daily.

This will still delete only succeeded jobs. Any job in error state or running will be kept.


mmartin24:
Checked upgrading from fleet `0.10.2` to `0.10.4-rc4`and working:

- Cleanup job set to @dayly. 
- Can be triggered in the moment crontab can be edited.
- It cleans succesful jobs
- It leaves faulty jobs:

[Screencast from 11-10-24 11:53:35.webm](https://github.com/user-attachments/assets/5d4aeff7-6f4e-4e57-ad96-417a198fdc0e)

PR #2959, opened 2024-10-17
Add more information on logging

manno:
* the majority of log statements uses kebab case for the logger name. Switch the camelCase and PascalCase cases to kebab


dev/LOGGING.png
weyfonk, line 1: Where is this image sourced from btw? Or did you create it?

internal/cmd/agent/deployer/monitor/updatestatus.go
```diff
@@ -81,7 +81,7 @@ func ShouldUpdateStatus(bd *fleet.BundleDeployment) bool {
 }
 
 func (m *Monitor) UpdateStatus(ctx context.Context, bd *fleet.BundleDeployment, resources *helmdeployer.Resources) (fleet.BundleDeploymentStatus, error) {
-	logger := log.FromContext(ctx).WithName("UpdateStatus")
+	logger := log.FromContext(ctx).WithName("update-status")
```
weyfonk, line 84: We have a few other CamelCased names, eg. `HelmDeployer`, `deleteByRelease`, `getNamespacesForBundle`... Is there a reason not to convert those?

manno: Those must have escaped my grep.

dev/LOGGING.png
manno: I created it. The original template was a pyramid, though.

weyfonk (approved):

PR #2977, opened 2024-10-22
Resource limits on controller

manno:
This adds resource limits as a helm chart option for the main controller.


0xavi0:
Do we want this only for the `fleet-controller`? 
What about the `gitops`, `helmops` and agent?

PR #2978, opened 2024-10-22
Remove tmate step from Fleet-in-Rancher test workflow

weyfonk:
That step uses a Github action which is no longer allowed in the Rancher org.

Refers to #1640
Follow-up to #2804.


weyfonk:
> We could also remove the tmate part in the `dev/README.md`. Besides that lgtm

Good catch, thanks! Mentions removed.

Issue #2969, opened 2024-10-21
New commits from webhook should trigger deployments, but disablePolling for gitrepo prevents new deployments


skanakal:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When disablePolling is enabled in a Git repository, Fleet fetches the latest commit but does not apply it to the deployment. The same behavior is observed with Azure DevOps.

### Expected Behavior

when disablePolling set in gitrepo, New commits from webhook should trigger deployments

### Steps To Reproduce

1. Install rancher 2.9.2 / fleet v0.10.3 (enable [debug](https://fleet.rancher.io/troubleshooting#enable-debug-logging-for-fleet-controller-and-fleet-agent) logging on fleet)
2. configure the [Fleet webhook ](https://fleet.rancher.io/webhook)
3. create a girepo using [example](https://github.com/rancher/fleet-examples.git) with `simple` path and set `disablePolling: true`
4. update the deployment (ex: change the replicas count to a higher number)
5. Notice the Fleet fetches the new commit in the gitrepo...
6. gitjob pod logs shows a webhook payload.  
~~~
2024-10-21T09:46:20Z	DEBUG	webhook	Webhook payload	{"payload": {"ref":"refs/heads/main","before":"bf810dec76b610ba01271b6899306b63189c6eba","after":"afa5f4f644233e16f65ee3251d0522358d72fc02","created":false,"deleted":false,"forced":false,"base_ref":null,"compare":"https://github.com/xxx/test/compare/bf810dec76b6...afa5f4f64423","commits":[{"sha":"","id":"afa5f4f644233e16f65ee3251d0522358d72fc02","node_id":"","tree_id":"7bbe0c768f77fda465980cc5e2597bd50615f11b","distinct":true,"message":"Update deployment.yaml","timestamp":"2024-10-21T15:16:18+05:30","url":"https://github.
~~~
7. until I force update the gitrepo, there is no job created with the latest commit 

### Environment

```markdown
- Architecture: x86_64
- Fleet Version: 0.10.3
- Cluster: local
  - Provider: custom
  - Options:
  - Kubernetes Version: 1.28
```

### Logs

```markdown
2024-10-21T10: 50: 26Z	DEBUG	webhook	Webhook payload	{
    "payload": {
        "ref": "refs/heads/main",
        "before": "afa5f4f644233e16f65ee3251d0522358d72fc02",
[...]

PR #2986, opened 2024-10-23
Adds predicate when webhook commit changes

0xavi0:
This adds a new predicate when the webhook commit is changed in the `GitRepo`. It also filters out `Job` events for creation/deletion, because they add extra reconcile loops that only increase the possibility of race conditions and that are not required.

The reconciler already knows when a `Job` is created/deleted because it is the owner and does not need to react upon those 2 events from the cluster.

Refers to: https://github.com/rancher/fleet/issues/2969

### DIFF
--- a/integrationtests/gitjob/controller/controller_test.go
+++ b/integrationtests/gitjob/controller/controller_test.go
@@ -577,9 +577,11 @@ var _ = Describe("GitJob controller", func() {
 
 	When("a new GitRepo is created with DisablePolling set to true", func() {
 		var (
-			gitRepo     v1alpha1.GitRepo
-			gitRepoName string
-			job         batchv1.Job
+			gitRepo               v1alpha1.GitRepo
+			gitRepoName           string
+			job                   batchv1.Job
+			webhookCommit         string
+			forceUpdateGeneration int
 		)
 
 		JustBeforeEach(func() {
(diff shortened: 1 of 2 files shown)


integrationtests/gitjob/controller/controller_test.go
```diff
@@ -591,6 +593,40 @@ var _ = Describe("GitJob controller", func() {
...
+					job.Status.Succeeded = 1
+					job.Status.Conditions = []batchv1.JobCondition{
+						{
+							Type:   "Complete",
+							Status: "True",
+						},
+					}
+					return k8sClient.Status().Update(ctx, &job)
+				}).Should(Not(HaveOccurred()))
+				// wait until the job has finished
```
weyfonk, line 614: nit:
```suggestion

				// wait until the job has finished
```

integrationtests/gitjob/controller/controller_test.go
```diff
@@ -616,6 +652,39 @@ var _ = Describe("GitJob controller", func() {
 				}, "30s", "1s").Should(Equal(stableCommit))
 			})
 		})
+
+		When("WebhookCommit changes and user clicks ForceUpdateGeneration", func() {
```
weyfonk, line 656: nit: a click is specific to the Rancher UI, but generation update may be forced through an API or `kubectl` call.
```suggestion
		When("WebhookCommit changes and user forces a redeployment", func() {
```

integrationtests/gitjob/controller/controller_test.go
```diff
@@ -616,6 +652,39 @@ var _ = Describe("GitJob controller", func() {
...
+				expectedCommit = stableCommit
+				// user clicks ForceUpdate 2 times
+				// This exercises possible race conditions in the reconciler
+				forceUpdateGeneration = 2
+			})
+
+			It("creates a new Job", func() {
+				Eventually(func() error {
+					jobName := names.SafeConcatName(gitRepoName, names.Hex(repo+webhookCommit, 5))
+					return k8sClient.Get(ctx, types.NamespacedName{Name: jobName, Namespace: gitRepoNamespace}, &job)
```
weyfonk, line 669: If more than one job were created, how would we detect it?

0xavi0: The job name depends on the repopath+commit, so unless the commit changes...we'll have the same job all the time. 
We can see from time to time `Already Exist` errors when trying to re-create the job under race condition circumstances.

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -101,9 +101,11 @@ func (r *GitJobReconciler) SetupWithManager(mgr ctrl.Manager) error {
 					predicate.GenerationChangedPredicate{},
 					predicate.AnnotationChangedPredicate{},
 					predicate.LabelChangedPredicate{},
+					webhookCommitChangedPredicate(),
 				),
 			),
 		).
+		Owns(&batchv1.Job{}, builder.WithPredicates(jobUpdatesOnly())).
```
manno, line 108: Maybe rename this to match the pattern of the other predicates

```suggestion
		Owns(&batchv1.Job{}, builder.WithPredicates(jobUpdatedPredicate())).
```

manno (approved):

Issue #2995, opened 2024-10-24
OverrideTargets and clustersDesiredReady Count mismatch, also targetCustomizations with doNoDeploy: true removes the bundleDeployment


manno:
Could be related to https://github.com/rancher/dashboard/issues/12232, https://github.com/rancher/rancher/issues/46916, https://github.com/rancher/dashboard/issues/11643

The calculation of gitrepo status fields has several problems:

It seems the `clustersDesiredReady` count does not work with the [overrideTargets](https://github.com/rancher/fleet/pull/1580) option. The result is set in the `gitrepo.Status.DesiredReadyClusters`.

Furthermore, the gitops controller, does not trigger for bundledeployment changes, but the `resourceKey` and  "Bundle" condition are computed from bundledeployments. The "Bundle" condition contains `gitrepo.Status.Summary`, which consists of the `ready` and `desiredReady` for bundledeployments. The code assumes a bundle always changes after a bundledeployment changes and triggers the status update code that way.

The original JIRA issue SURE-9111 has been resolved.

This can be worked on after https://github.com/rancher/fleet/issues/3186

- [ ]  Investigate if overrideTargets is reflected in the resource counts correctly
- [ ] Add a test if needed

sbulage:
## System Information
| Rancher Version | Fleet Version |
|--------|--------|
| v2.11-4db694b491abf4af48d23a0e84d68a0200906c6e-head | 106.0.1+up0.12.1-rc.1 | 

---

## QA Verification
### Scenarios

|Scenario|Test Case|Result|
|---|---|---|
|1| Test GitRepo shows `Status.Summary` and `Status.DesiredReadyClusters` when `overrideTargets` is set in the fleet.yaml| :white_check_mark:|
|2| Test bundleDeployment is removed when `doNotDeploy` is set to `true` with `targetCustomizations` in the fleet.yaml| :white_check_mark:|

**Scenario 1**
1. Add `overrideTargets` option to the fleet.yaml as mentioned [here](https://github.com/sbulage/test-fleet/blob/main/overrride-targets-donot-deploy/fleet.yaml)
2. Created a `GitRepo` in the UI (https://github.com/sbulage/test-fleet having path `overrride-targets-donot-deploy`)
3. Select `Target` `"No Clusters"` in the `Deploy To` section in the UI. 
4. Verify that no bundle or deployment is created because there is no target mentioned.
5. Verify that `gitrepo.Status.Summary`, shows count for `ready` and `desiredReady`.
[...]

PR #2997, opened 2024-10-24
Refactor grutils/gitops controller

manno:
Refers to https://github.com/rancher/fleet/issues/2995


internal/cmd/controller/gitops/reconciler/gitrepo_status.go
```diff
@@ -1,33 +1,53 @@
...
+	}
+
+	setResourceKey(ctx, c, gitrepo)
+
+	gitrepo.Status.Display.ReadyBundleDeployments = fmt.Sprintf("%d/%d",
+		gitrepo.Status.Summary.Ready,
+		gitrepo.Status.Summary.DesiredReady)
+
+	setCondition(&gitrepo.Status, nil)
+
```
0xavi0, line 46: this is also done in line 224, right after calling this function. I guess it remained when copying/pasting...

0xavi0 (approved):
LGTM, just a nit about one line that remained after copy/paste

internal/cmd/controller/gitops/reconciler/gitrepo_status.go
manno: Ah, right. The "setStatus" code doesn't even change the GitRepoAcceptCondition.

0xavi0 (approved):

PR #3011, opened 2024-10-25
[v0.9] Bump Go

thardeck:
The nil deference issue might be related to the following change (which potentially can happen in other untested code areas too):
`Functions that shrink the size of a slice (Delete, DeleteFunc, Compact, CompactFunc, and Replace) now zero the elements between the new length and the old length.
`
https://tip.golang.org/doc/go1.22


weyfonk (approved):
LGTM, those functions don't seem to be in use anywhere else in the codebase.

PR #3014, opened 2024-10-25
Avoid retries for bundle and content resources

manno:
I observed some log messages when the agents were down.
When two reconciles try to create the same content resource, one will error, however content resources are meant to be shared between bundles.

We also don't need to update the content for a finalizer, when no finalizer was added.


0xavi0 (approved):
lgtm

weyfonk (approved):

PR #3017, opened 2024-10-28
Remove Go Toolchain entry

thardeck:


go.mod
```diff
@@ -2,8 +2,6 @@ module github.com/rancher/fleet
 
 go 1.23.0
 
-toolchain go1.23.2
```
manno, line 5: I think we want to keep this one, as it helps with choosing the right toolchain to compile fleet.

The other one is just a library, which we never compile

manno (approved):

PR #3019, opened 2024-10-28
Use setup-go in release-against-charts.yaml

thardeck:
because the self-hosted seem to miss go?

https://github.com/rancher/fleet/actions/runs/11554501554/job/32157846848


weyfonk:
What's wrong with setting up Go? eg. [here](https://github.com/rancher/fleet/blob/main/.github/workflows/e2e-test-fleet-in-rancher.yml#L70)?

weyfonk (approved):

PR #3027, opened 2024-10-29
GitRepo resources list doesn't list resources multiple times

manno:
Refers to https://github.com/rancher/dashboard/issues/11643

This is a hotfix to avoid a conflict with UI code, which multiplies the resources list per cluster. Multiplying a GitRepo's resources from all clusters with the number of clusters potentially freezes the UI.

With this PR `gitrepo.Status.Resources` reports resources only once, even if they are deployed to multiple clusters. If any cluster has a nonready/modified status for the resource, it is still added to the `PerClusterState` field. 

That means, it's no longer possible to tell if a resource was not deployed to a cluster, e.g. because it was excluded via helm chart values.

Future code needs to respect that some resources are not present in each bundledeployment.
Users can for example use targetCustomization to deploy resources only to some clusters. Additionally, popular charts install different resources depending on the k8s version of the cluster. Some charts even allow to deploy "extraObjects" via a helm chart value.

* https://github.com/rancher/dashboard/blob/master/shell/models/fleet.cattle.io.gitrepo.js#L334-L339
* https://github.com/rancher/fleet/blob/main/internal/cmd/controller/gitops/reconciler/resourcekey.go#L38-L54

### DIFF
--- a/internal/cmd/controller/gitops/reconciler/resourcekey.go
+++ b/internal/cmd/controller/gitops/reconciler/resourcekey.go
@@ -9,34 +9,57 @@ import (
 )
 
 func setResources(list *fleet.BundleDeploymentList, gitrepo *fleet.GitRepo) {
-	state := bundleErrorState(gitrepo.Status.Summary)
-	gitrepo.Status.Resources, gitrepo.Status.ResourceErrors = fromResources(list, state)
-	gitrepo.Status = countResources(gitrepo.Status)
+	s := summaryState(gitrepo.Status.Summary)
+	r, errors := fromResources(list, s)
+	gitrepo.Status.ResourceErrors = errors
+	gitrepo.Status.ResourceCounts = countResources(r)
+	gitrepo.Status.Resources = merge(r)
 }
 
-func bundleErrorState(summary fleet.BundleSummary) string {
-	bundleErrorState := ""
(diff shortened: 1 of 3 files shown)


aruiz14 (approved):

internal/cmd/controller/gitops/reconciler/resourcekey.go
```diff
@@ -9,34 +9,61 @@ import (
...
+	s := summaryState(gitrepo.Status.Summary)
+	r, errors := fromResources(list, s)
+	gitrepo.Status.ResourceErrors = errors
+	gitrepo.Status.ResourceCounts = countResources(r)
+	gitrepo.Status.Resources = merge(r)
 }
 
-func bundleErrorState(summary fleet.BundleSummary) string {
-	bundleErrorState := ""
+func merge(resources []fleet.GitRepoResource) []fleet.GitRepoResource {
```
weyfonk, line 19:
```suggestion
// merge takes a list of GitRepo resources and deduplicates resources deployed to multiple clusters,
// ensuring that for such resources, the output contains a single resource entry with a field summarizing
// its status on each cluster.
func merge(resources []fleet.GitRepoResource) []fleet.GitRepoResource {
```

manno: Sorry, automation was faster :|

Issue #2859, opened 2024-09-17
[SURE-8794] Deploying ClusterGroup from GitRepo results in loop


p-se:
Deploying a ClusterGroup from a GitRepo which also contains accompanying other GitRepo resources that use those newly created ClusterGroups result in a loop.

This loop triggers ClusterGroups and appends a message to it that endlessly grows, until the limit of etcd is hit. In which case Fleet is supposedly blocked.

The issue can be reproduced by adding [this GitRepo resource](https://github.com/p-se/fleet-tests/blob/sure-8794/repo1.yaml) to the cluster. The issue was reproducible on the latest Fleet development version at the time and did not require a Rancher installation to reproduce. The cluster was prepared using `dev/setup-multi-cluster`.

rancherbot:
Not creating port issue, milestone  does not exist or is not an open milestone

mmartin24:
I tested this in `v2.10-d8667221a2eec48d4350d00a9d39aee54f00f810-head` with fleet `fleet:105.0.1+up0.11.1`.

I saw a significant improvement against `2.8.5` where the logs grow every few seconds. In a matter of minutes, I logs fill the page as can be seen in this screenshot:
![2 8 5](https://github.com/user-attachments/assets/050d0e04-6cec-46e8-80dc-d4f0c2883f18)


When I checked in `v2.10-d8667221a2eec48d4350d00a9d39aee54f00f810-head` with fleet `fleet:105.0.1+up0.11.1`, the log growth was significantly lower; however, it was still present, as can be seen in the screenshot after a few hours as can be seen in the screenshot:

![2 10-longlog](https://github.com/user-attachments/assets/dc9e40fb-91f7-4664-b0d8-6c7640417368)

@p-se  is this expected?

p-se:
> @p-se is this expected?

No, it is not expected to see the status ever-growing! That said, I'm not sure if the fix is in the versions you've used for testing.

mmartin24:
Tested in `v2.11-05a12b2b913325def4805dee2b83d18ba94ef426-head`with `fleet:106.0.0+up0.12.0-alpha.2` and working well.

Logs do not endlessly duplicate:

![Image](https://github.com/user-attachments/assets/5eba56cd-2274-4184-a36a-b2675eb711d6)

PR #3042, opened 2024-10-31
Deduplicate status messages

p-se:
Prevents fleet from crashing due to resources exceeding etcd's
configured size limit.

Deduplicate messages should only be necessary for edge cases which are
not officially supported by fleet but result in ever increasing message
sizes.

This is due to the messages being copied from one resource to another
and back again. Every resource adds its status to the message. This only
happens if a cluster group is deployed by a GitRepo, which results in a
bundle containing a cluster group. This bundle can only become ready if
the cluster group is ready, but if the cluster group points to the
cluster of the bundle that deployed the cluster group, this cannot ever
happen. The user is expected to fix this situation but deduplicating the
messages prevents the message from growing up to the point where etcd's
limit is reached and fleet crashes.

Deduplicating the messages also has the effect of not changing the
status of resources frequently, which results in less controllers being
triggered.

Refers to #2859.


internal/cmd/agent/deployer/summary/summary_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package summary_test
```
weyfonk, line 1: Just curious about the name of this file: why not `suite_test.go` as done elsewhere?

internal/cmd/agent/deployer/summary/summarizers_test.go
```diff
@@ -0,0 +1,31 @@
...
+							"status":  "False",
+							"message": "foo; bar; foo; foo; bar",
+						},
+					},
+				},
+			}
+			summary := kStatusSummarizer(obj, nil, fleetv1.Summary{})
+			Expect(summary.Message).To(Equal([]string{"foo; bar"}))
+			Expect(summary.Transitioning).To(BeTrue())
+			Expect(summary.Error).To(BeFalse())
```
weyfonk, line 28: I think we may want more test cases here, eg.:
* checking that messages without duplicated strings between `;` are left unaltered
* checking that using a different separator than `;` does not lead to deduplication
* checking what happens with an empty message, just to be safe

internal/cmd/agent/deployer/summary/summary_suite_test.go
p-se: I'm fine with `suite_test.go` as well. The name `summary_suite_test.go` came directly from `ginkgo bootstrap`.

weyfonk (approved):

PR #3052, opened 2024-11-06
Fix namespace target customization support with no defaults

weyfonk:
This initialises maps in bundle deployment options when those are empty due to no default configuration being provided.
As a results, this prevents panics when namespace labels or annotations are set as target customizations without default, all-targets configuration.

Refers to #2441
Depends on https://github.com/rancher/fleet-test-data/pull/23.


internal/cmd/controller/target/target.go
```diff
@@ -59,6 +59,8 @@ func (t *Target) BundleDeployment() *fleet.BundleDeployment {
 	}
 	bd.Spec.Paused = t.IsPaused()
 
+	initialiseOptionsMaps(bd)
```
manno, line 62: Since this is  a new bundledeployment, I would just initialize the maps, without a function and a nil check. Am I missing something?

weyfonk: The issue here is that those maps are set on [L58](https://github.com/rancher/fleet/pull/3052/files#diff-d81320afcb590e95a0b56577f6c5907ea071596efc14d1b09783b552ee5d1de9R58) through `t.Deployment.Spec`, which _may_ already contain values set at the root of `fleet.yaml` (which may then be overridden by `bundleTarget` values).
The panic only occurred. ie the maps were only empty, when no default, all-targets values are specified at the root of `fleet.yaml`. See https://github.com/rancher/fleet-test-data/pull/23 for an example.

manno (approved):

internal/cmd/controller/target/target.go
```diff
@@ -136,3 +138,24 @@ func (t *Target) state() fleet.BundleState {
 func (t *Target) message() string {
 	return summary.MessageFromDeployment(t.Deployment)
 }
+
+// initialiseOptionsMaps initialises options and staged options maps of namespace labels and annotations on bd, if
+// needed.
+// Assumes that bd is not nil.
```
manno, line 144: At this point we could just pass spec?

manno (approved):

internal/cmd/controller/target/target.go
weyfonk: Oh, indeed we could!

PR #3057, opened 2024-11-07
[backport v0.11] Fix namespace target customization support with no defaults (#3052)

manno:
Refers to https://github.com/rancher/fleet/issues/2441

Initialise options maps when empty

This prevents panics when namespace labels or annotations are configured as target customizations over nonexistent defaults.


kkaempf (changes requested):
GH spell check warnings should be honored

PR #3089, opened 2024-11-21
Enables e2e tests using k3d services export to external ips

0xavi0:
This PR changes some e2e environment files to be able to run e2e tests in systems in which we cannot access load balancers IPs from clusters.

This is specially intented to be able to execute the e2e tests locally in MAC OS systems in which docker IPs cannot be accessed.

* Get your IP address ( `localhost` or `127.0.0.1` is not valid. See [this](https://github.com/helm/helm/issues/11683#issuecomment-2183304555) )
* Export your IP with `export external_ip=YOUR_IP` or edit the dev/env.single-cluster-defaults file and add it there.
* Export the Git server URL with your ip (127.0.01 is also valid in this case).
 `export CI_GIT_REPO_URL=http://YOUR_IP:8080` either in the the dev/env.single-cluster-defaults file or exporting before running
* Run e2e tests normally

Note: when running tests with external_ip set ports passed to `k3d` are not random. This in intentional as we need to access the services in a well known port.

### DIFF
--- a/e2e/testenv/infra/cmd/setup.go
+++ b/e2e/testenv/infra/cmd/setup.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"bytes"
 	"context"
 	"crypto/tls"
 	"fmt"
@@ -13,6 +12,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/chartmuseum/helm-push/pkg/chartmuseum"
 	"github.com/rancher/fleet/e2e/testenv"
 	"github.com/rancher/fleet/e2e/testenv/kubectl"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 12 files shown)


manno (approved):

e2e/testenv/infra/cmd/setup.go
```diff
@@ -163,30 +166,14 @@ var setupCmd = &cobra.Command{
 				return "", nil
 			})
 
-			fmt.Println("determining Helm binary path...")
-			helmPath := os.Getenv("HELM_PATH")
-			if helmPath == "" {
-				helmPath = "/usr/bin/helm" // prevents eg. ~/.rd/bin/helm from being used, without support for skipping TLS
+			// TODO enable this when the Helm library supports `--insecure-skip-tls-verify`
```
weyfonk, line 169: Should we get rid of this comment then, or have I missed something? :)

0xavi0: as it is very possible that the `helmops` PR takes time to be merged... https://github.com/rancher/fleet/pull/3100

PR #3102, opened 2024-11-27
 Calculate Clusters resourceCount from BundleDeployments instead of GitRepos

aruiz14:
Refers to https://github.com/rancher/rancher/issues/46916

### Description

`status.resourceCounts` for `cluster.fleet.cattle.io` is calculated from GitRepos statuses, which include an equivalent `resourceCounts` field. This causes the count to be incorrect, since a GitRepo could deploy resources to different Clusters.

### How to reproduce the issue
1. Setup Rancher/Fleet with more than 1 downstream clusters
2. Add a GitRepo that deploys to all downstream clusters
3. Describe the status of any of the clusters and observe that `status.resourceCounts` is wrong, since it also includes resources deployed to other namespaces.

### Description of the fix
This can be fixed by populating `Clusters`' `status.resourceCounts` from their respective `BundleDeployments`, similarly to how it's calculated for `GitRepos`. In order to do this in a clean way, I had to move that code out of the `gitops` reconciler to its own package.

In addition, I've added some changes to make the existing code a bit more efficient:
 - Filter `bundleDeployment.Items` while it's being modified, removing the need to keep a separate map of deleted objects.
 - Create an index (map) of bundles to be deleted, instead of having a double `for` loop.
 - Removed duplicity of `repoKey` type, which was the same as `types.NamespacedKey`.


internal/cmd/controller/reconciler/cluster_controller.go
```diff
@@ -144,20 +151,25 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+			if err := r.Delete(ctx, &bd); err != nil {
+				logger.V(1).Info("deleting bundleDeployment returned an error", "error", err)
 			}
+			return true
 		}
-	}
+		// Omit fleet-agent bundle deployments from resource counts
+		if strings.HasPrefix(bd.Name, "fleet-agent") {
+			return true
+		}
```
aruiz14, line 170: This has an effect on the calculations performed later, since `fleet-agent` bundles (and its k8s resources) will be omitted from any calculations and display. This is also reflected in the e2e that started failing:
```
Checks status updates happen for a simple deployment when deployment is successful [It] correctly sets the status values for Clusters
...
  Expected
      <string>: '1/1'
  to equal
      <string>: '2/2'
```
However, the actual test expectation would not be obvious if you don't know that fleet-agents are deployed using bundles (e.g. why it expects 2 ready bundles if I only deployed 1?).
I thought users may have the same confusion, and I added this for that reason, but happy to revert if if you think otherwise.

weyfonk: I think this approach of making Fleet agent deployments transparent to users make sense, but am curious to know what others think. Either way, we should document this.

weyfonk (commented):
LGTM, thanks!
Approval pending discussion on whether or not to include Fleet agent bundles in cluster resource counts.

internal/cmd/controller/reconciler/cluster_controller.go
aruiz14: +1
For completeness, this only affects the Cluster resource counts. They will still appear under Bundles in the Advanced view

aruiz14: Consensus was to keep including fleet-agent for consistency.

weyfonk (approved):
LGTM, thanks for this! :)

PR #3114, opened 2024-12-03
[SURE-9137] Add template errors to bundle and gitrepo status

p-se:
When fleet.yaml is evaluated for targeting, errors can occur due to missing values, like for instance, if a cluster has a missing label that is used in the fleet.yaml file.

Those kind of errors would previously only be visible in logs. This PR introduces error messages in the bundle, more specifically in the Ready condition of the Bundle status, when errors occur on rendering the templates for targeting. Subsequently the error message is being copied from the Bundle to the GitRepo. It will not show up in Clusters or ClusterGroups.

It is new that Bundles will have their own errors. Errors are commonly created in BundleDeployments and are aggregated in the status of Bundle. Now, additionally, the bundle itself may have errors that result in a status change. Those errors will take precedence over any errors that may already exist and effectively overwrite the Ready status condition, should they occur.

Overwriting errors coming from BundleDeployments in Bundle will not happen if the Bundle never created a BundleDeployment that has an error, but will happen if BundleDeployments have successfully been created from the Bundle. If then a cluster is added that has a missing label or a cluster is edited so that a label value required for templating is removed, the error will be shown in the Bundle, not those errors that might be present in the BundleDeployment resources. The errors present in the BundleDeployment will not be changed at all, so that there is a possibility of retrieving both at the same time.

### DIFF
--- a/e2e/single-cluster/status_test.go
+++ b/e2e/single-cluster/status_test.go
@@ -1,13 +1,23 @@
 package singlecluster_test
 
 import (
+	"encoding/json"
 	"errors"
+	"fmt"
+	"math/rand"
+	"os"
+	"path"
 	"strings"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+
 	"github.com/rancher/fleet/e2e/testenv"
+	"github.com/rancher/fleet/e2e/testenv/githelper"
 	"github.com/rancher/fleet/e2e/testenv/kubectl"
+	fleet "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
+	"github.com/rancher/wrangler/v3/pkg/genericcondition"
+	corev1 "k8s.io/api/core/v1"
 )
 
 var _ = Describe("Checks status updates happen for a simple deployment", Ordered, func() {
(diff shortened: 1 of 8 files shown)


kkaempf:
Where are Bundle errors displayed in the UI ?

p-se:
> Where are Bundle errors displayed in the UI ?

Good point, the Bundle is not something we like to advertise to users. We're treating it more like an internal resource in the Rancher UI. It would be in `Continuous Delivery`, `Advanced`, `Bundles`.

That said, the same error will be in the GitRepo.

image

internal/cmd/controller/gitops/reconciler/status_controller.go
```diff
@@ -6,10 +6,13 @@ import (
 	"reflect"
 	"sort"
 
+	v1 "k8s.io/api/core/v1"
```
weyfonk, line 9: nit: This import should perhaps be grouped with other `k8s.io` imports below.

internal/cmd/controller/gitops/reconciler/status_controller.go
```diff
@@ -110,6 +113,28 @@ func (r *StatusReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+	//
+	// - If the bundle has no deployments, there is no status to be copied from the setStatus
+	// function, so that we won't overwrite anything.
+	//
+	// - If the bundle has rendering issues and there are deployments of which there is at least one
+	// in a failed state, the status of the bundle deployments would be overwritten by the bundle
+	// status.
+	//
+	// - If the bundle has no rendering issues and there are deployments in a failed state, the code
+	// will overwrite the gitreop status condition with the same status from the bundle. This is why
```
weyfonk, line 126:
```suggestion
	// will overwrite the gitrepo status condition with the same status from the bundle. This is why
```

I am not entirely sure of understanding this sentence, especially the `same status from the bundle` bit: this means to say that the bundle's status will be propagated to the gitrepo, right?

internal/cmd/controller/gitops/reconciler/status_controller.go
```diff
@@ -110,6 +113,28 @@ func (r *StatusReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		gitrepo.Status.Display.State = "GitUpdating"
 	}
 
+	// We're explicitly setting the ready status from a bundle here, but only if it isn't ready.
```
weyfonk, line 116: Thank you for this detailed write-up :)
I wonder if all of it needs to live here, instead of being part of `setReadyStatusFromBundle`'s method-level doc, but this provides much needed context :+1:

internal/cmd/controller/gitops/reconciler/status_controller.go
```diff
@@ -165,23 +190,80 @@ func setStatus(list *v1alpha1.BundleDeploymentList, gitrepo *v1alpha1.GitRepo) e
...
+	for _, bundle := range bList.Items {
+		if bundle.Status.Conditions == nil {
+			continue
+		}
+
+		for _, c := range bundle.Status.Conditions {
+			if c.Type == string(fleet.Ready) && c.Status == v1.ConditionFalse {
+				condition = c
+				found = true
+				break bundles
```
weyfonk, line 220: It looks like this will only propagate the _first_ bundle rendering error to the gitrepo status, if multiple such errors exist over multiple bundles.
Perhaps something to clarify in documentation.

internal/cmd/controller/gitops/reconciler/status_controller.go
```diff
@@ -165,23 +190,80 @@ func setStatus(list *v1alpha1.BundleDeploymentList, gitrepo *v1alpha1.GitRepo) e
...
+			newConditions = append(newConditions, condition)
+			found = true
+			continue
+		}
+		newConditions = append(newConditions, c)
+	}
+	if !found {
+		// Add the ready condition from the bundle to the gitrepo.
+		newConditions = append(newConditions, condition)
+	}
```
weyfonk, line 244: Just curious: would readiness of gitrepos be previously computed from bundle deployments only, then?

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -174,6 +178,18 @@ func (r *BundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
 	matchedTargets, err := r.Builder.Targets(ctx, bundle, manifestID)
 	if err != nil {
+		// When targeting fails, we don't want to continue and make the error message visible in the
+		// UI. For that we use a status condition of type Ready.
+		bundle.Status.Conditions = []genericcondition.GenericCondition{
+			{
+				Type:           string(fleet.Ready),
+				Status:         v1.ConditionFalse,
+				Message:        "Targeting error: " + err.Error(),
+				LastUpdateTime: metav1.Now().Format(time.RFC3339),
```
weyfonk, line 188:
```suggestion
				LastUpdateTime: metav1.Now().UTC().Format(time.RFC3339),
```

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -216,38 +232,25 @@ func (r *BundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+			// namespace
 			if err := r.createDeploymentOCISecret(ctx, bundle, bd); err != nil {
 				return ctrl.Result{}, err
 			}
 		}
 	}
 
 	updateDisplay(&bundle.Status)
-
-	err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
weyfonk, line 238: No more retries on conflicts? Is that deliberate?
Same question about other instances below.

e2e/assets/status/chart-with-template-vars/.helmignore
```diff
@@ -0,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
```
weyfonk, line 8: Do we really need all these patterns in this case, or even this file? :upside_down_face:

e2e/single-cluster/status_test.go
```diff
@@ -107,3 +117,175 @@ var _ = Describe("Checks status updates happen for a simple deployment", Ordered
...
+			"-c",
+			"fleet-controller",
+		)
+		Expect(err).ToNot(HaveOccurred())
+
+		// Errors about resources other than bundles or bundle deployments not being found at deletion time
+		// should be ignored, as they may result from other test suites.
+		Expect(out).ToNot(MatchRegexp(
+			`ERROR.*Reconciler error.*Bundle(Deployment)?.fleet.cattle.io \\".*\\" not found`,
+		))
```
weyfonk, line 201: Why do we need this?

e2e/single-cluster/status_test.go
```diff
@@ -107,3 +117,175 @@ var _ = Describe("Checks status updates happen for a simple deployment", Ordered
...
+			}).Should(Succeed())
+		})
+	})
+
+	When("a git repository is created that contains no template error", func() {
+		It("should have no error in the bundle", func() {
+			_, _ = ensureClusterHasLabelFoo()
+			Eventually(func(g Gomega) {
+				status := getBundleStatus(g, k, gitrepoName+"-examples")
+				expectNoError(g, status.Conditions)
```
weyfonk, line 275: Should we check the gitrepo status here as well, for the sake of completeness? (no strong opinions here)

e2e/single-cluster/status_test.go
```diff
@@ -107,3 +117,175 @@ var _ = Describe("Checks status updates happen for a simple deployment", Ordered
...
+			}).Should(Succeed())
+		})
+	})
+})
+
+// getGitRepoStatus retrieves the status of the gitrepo with the provided name.
```
weyfonk, line 281:
```suggestion
// getBundleStatus retrieves the status of the bundle with the provided name.
```

e2e/single-cluster/status_test.go
```diff
@@ -107,3 +117,175 @@ var _ = Describe("Checks status updates happen for a simple deployment", Ordered
 		})
 	})
 })
+
+var _ = Describe("Checks that template errors are shown in bundles and gitrepos", Ordered, Label("infra-setup"), func() {
```
weyfonk, line 121: What prevents these tests from being integration tests instead?

internal/cmd/controller/gitops/reconciler/status_controller.go
p-se: Yes. This is how it was:
```mermaid
flowchart LR
  GR[GitRepo]
  BD[Bundle Deployment]
  B[Bundle]

  GR -->|"`fetches the status of and
  aggregates the status of`"| BD
  GR -->|creates| B
  B -->|creates one or more and
  fetches the status of| BD
```
With this implementation it becomes:
```mermaid
flowchart LR
  GR[GitRepo]
  BD[Bundle Deployment]
  B[Bundle]

  GR -->|"`fetches the status of and
  aggregates the status of`"| BD
  GR -->|creates| B
  B -->|creates one or more and
  fetches the status of| BD

  GR ==>|needs to fetch
  the status of| B
```
And how I thought it perhaps could be, but I did not implement it:
```mermaid
flowchart LR
  GR[GitRepo]
  BD[Bundle Deployment]
  B[Bundle]

  GR -->|"`creates and fetches the
  (aggregated) status of`"| B
  B -->|creates one or more and
  fetches the status of| BD
```

internal/cmd/controller/gitops/reconciler/status_controller.go
p-se: I find it difficult to put everything in the documentation of setReadyStatusFromBundle, as I try to describe why we can use it there the way it is being used.

internal/cmd/controller/gitops/reconciler/status_controller.go
p-se: Yes, that's true. That is something that we might want to change at some point, but I'd need to think more about it, since there's [another possible optimization](https://github.com/rancher/fleet/pull/3114#discussion_r1878117893). In the meantime, I'm thinking of adding some documentation to our docs, too, since the errors we start showing in Bundle and GitRepo won't be found in Cluster or ClusterGroup.

internal/cmd/controller/reconciler/bundle_controller.go
p-se: Yes, it is deliberate! We've started migrating away from RetryOnConflict and let controller-runtime's back-off do more of the retries, just a bit further apart. Those were some of the few leftovers of RetryOnConflict and I was asked to take care of it while I'm already working on those controllers.

e2e/assets/status/chart-with-template-vars/.helmignore
p-se: You're right. I don't think we need that file at all.

e2e/single-cluster/status_test.go
p-se: Right, I don't think we need it.

internal/cmd/controller/gitops/reconciler/status_controller.go
p-se: I did adapt the text though, hoping that it might be clearer now.

e2e/single-cluster/status_test.go
p-se: I'm not actually sure that there is something that would prevent that, I was just worried about the complexity of having to work with several controllers in integration tests.

internal/cmd/controller/gitops/reconciler/status_controller.go
weyfonk: Nice, thanks for clarifying, cool diagrams too!

internal/cmd/controller/gitops/reconciler/status_controller.go
weyfonk: That's clearer, thanks!

internal/cmd/controller/gitops/reconciler/status_controller.go
weyfonk: No worries, let's leave this here for now, and thanks again for writing and polishing this.

internal/cmd/controller/reconciler/bundle_controller.go
weyfonk: Oh, that does ring a bell. Thanks for clarifying.

e2e/single-cluster/status_test.go
weyfonk: There were 2 suggested changes on this line ;)

e2e/single-cluster/status_test.go
weyfonk: We already [support](https://github.com/rancher/fleet/blob/main/integrationtests/gitjob/controller/suite_test.go#L107-L133) multiple reconcilers in integration tests ;)

weyfonk (approved):

PR #3134, opened 2024-12-10
Exports Metrics URLs to test with external IPs

0xavi0:
It also deletes the HELM_PATH env variable as it is no longer used.

Fixes errors in metrics tests when the test is trying to check for metrics when the service exporting them it's still not fully up.

Adds a custom file to add extra env variables (This was added so the `dev/env.single-cluster-defaults` is not changed and to avoid pushing changes to it by mistake)  

Adds an extra check in OCI e2 tests to verify that the `CI_OCI_USERNAME` and `CI_OCI_PASSWORD` are set.

Deletes the `CI_GIT_REPO_URL` env variable and uses the already existing `external_ip`.

### Example on how to run `e2e` tests with services exported to external IPs (useful for testing under MAC OS)

1. Create a file with your custom environment
 ```bash
## YOUR IP HERE
export external_ip=192.168.1.47

## PORTS FOR EXPORTING METRICS IN E2E TESTS
## IF YOU DON'T SET THESE PORTS THE DEFAULT IS TO 
## CREATE A RANDOM PORT FOR EACH EXECUTION AND YOU 
## WON'T BE ABLE TO ACCESS.
## WHEN ENABLING THESE VARIABLES K3D WILL EXPORT THESE PORTS
export METRICS_CONTROLLER_PORT=8083
export METRICS_GITJOB_PORT=8084
```
2. Setup the cluster
```bash
export CUSTOM_ENV_FILE=PATH_TO_YOUR_CUSTOM_ENV
source dev/setup-single-cluster
``` 

3. Run the tests
```bash
export CUSTOM_ENV_FILE=PATH_TO_YOUR_CUSTOM_ENV
source dev/env.single-cluster-defaults
ginkgo --github-output e2e/metrics
ginkgo --github-output e2e/single-cluster
```

### DIFF
--- a/dev/env.single-cluster-defaults
+++ b/dev/env.single-cluster-defaults
@@ -10,4 +10,7 @@ export CI_OCI_USERNAME=fleet-ci
 export CI_OCI_PASSWORD=foo
 export CI_OCI_CERTS_DIR=FleetCI-RootCA/
 
-export HELM_PATH=/usr/bin/helm
+# File with custom environment
+if [[ -f "$CUSTOM_ENV_FILE" ]]; then
+    source "$CUSTOM_ENV_FILE"
+fi
--- a/dev/env.single-cluster-defaults
+++ b/dev/env.single-cluster-defaults
@@ -10,7 +10,3 @@ export CI_OCI_USERNAME=fleet-ci
 export CI_OCI_PASSWORD=foo
 export CI_OCI_CERTS_DIR=FleetCI-RootCA/
 
-# File with custom environment
-if [[ -f "$CUSTOM_ENV_FILE" ]]; then
-    source "$CUSTOM_ENV_FILE"
-fi
--- a/dev/setup-k3d
+++ b/dev/setup-k3d
@@ -7,6 +7,14 @@ docker_mirror=${docker_mirror-}
 name=${1-upstream}
 i=${2-0}
 
+if [ ! -z METRICS_CONTROLLER_PORT ]; then
+    args="$args -p "${METRICS_CONTROLLER_PORT}:${METRICS_CONTROLLER_PORT}@server:0""
+fi
+
+if [ ! -z METRICS_GITJOB_PORT ]; then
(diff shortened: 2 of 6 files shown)


dev/env.single-cluster-defaults
```diff
@@ -10,4 +10,7 @@ export CI_OCI_USERNAME=fleet-ci
 export CI_OCI_PASSWORD=foo
 export CI_OCI_CERTS_DIR=FleetCI-RootCA/
 
-export HELM_PATH=/usr/bin/helm
+# File with custom environment
+if [[ -f "$CUSTOM_ENV_FILE" ]]; then
+    source "$CUSTOM_ENV_FILE"
+fi
```
manno, line 16: Not sure this is necessary.

"Do it all" scripts are meant to source https://github.com/rancher/fleet/blob/main/dev/setup-cluster-config which is used to switch between 
`DEFAULT_CONFIG="dev/env.single-cluster-defaults"` and  `CUSTOM_CONFIG_FILE="env.single-cluster"`. Where `env.single-cluster` is already in `.gitignore`.


Well, `setup-cluster-config` seems a bit too much me, it's only called from two scripts. Those scripts could just source their default/custom config, duplicating the if switches might even help with readability.

dev/setup-k3d
```diff
@@ -7,6 +7,14 @@ docker_mirror=${docker_mirror-}
 name=${1-upstream}
 i=${2-0}
 
+if [ ! -z METRICS_CONTROLLER_PORT ]; then
```
manno, line 10: I'm not too happy vs `setup-k3ds` vs. `setup-k3d`. In the beginning I used `setup-k3d` to set individual clusters, but both scripts diverged and now integrationtests only work with `setup-k3d` 🤷 

https://github.com/rancher/fleet/pull/3156/files#r1892583417

dev/env.single-cluster-defaults
manno, line 16:
```suggestion
```

PR #3136, opened 2024-12-10
Limit git clone to latest commit

weyfonk:
This should speed up git cloning.
A follow-up to this could implement filtering by path once [this go-git PR](https://github.com/go-git/go-git/pull/1108) is merged upstream.

### DIFF
--- a/internal/cmd/cli/gitcloner/cloner.go
+++ b/internal/cmd/cli/gitcloner/cloner.go
@@ -79,6 +79,7 @@ func cloneBranch(opts *GitCloner, auth transport.AuthMethod, caBundle []byte) er
 		CABundle:          caBundle,
 		SingleBranch:      true,
 		ReferenceName:     plumbing.ReferenceName(opts.Branch),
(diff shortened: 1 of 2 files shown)


internal/cmd/cli/gitcloner/cloner.go
```diff
@@ -79,6 +79,7 @@ func cloneBranch(opts *GitCloner, auth transport.AuthMethod, caBundle []byte) er
 		CABundle:          caBundle,
 		SingleBranch:      true,
 		ReferenceName:     plumbing.ReferenceName(opts.Branch),
+		Depth:             1,
```
manno, line 82: As discussed, it's unclear if this works with sub module cloning.

manno (changes requested):
We need to investigate submodules

manno:
Reminds me of sparse checkouts, which could also be interesting to reduce cloning time and bandwidth.

weyfonk:
Superseded by #4488.

PR #3137, opened 2024-12-10
Use UTC for timestamps

weyfonk:
This brings consistency to the use of `time.Now()` timestamps for resource creation, deletion, last sync, etc.
It should limit the risks of confusion and timestamp-related bugs.

### DIFF
--- a/internal/cmd/agent/deployer/summary/summarizers.go
+++ b/internal/cmd/agent/deployer/summary/summarizers.go
@@ -353,7 +353,7 @@ func checkInitializing(obj data.Object, conditions []Condition, summary fleetv1.
 	if summary.State == "" && hasConditions && len(conditions) == 0 && strings.Contains(apiVersion, "cattle.io") {
 		val := obj.String("metadata", "created")
 		if i, err := convert.ToTimestamp(val); err == nil {
-			if time.Unix(i/1000, 0).Add(5 * time.Second).After(time.Now()) {
+			if time.Unix(i/1000, 0).Add(5 * time.Second).After(time.Now().UTC()) {
 				summary.State = "initializing"
 				summary.Transitioning = true
 			}
@@ -393,7 +393,7 @@ func checkRemoving(obj data.Object, conditions []Condition, summary fleetv1.Summ
 
 	summary.Message = append(summary.Message, "waiting on "+f)
 	if i, err := convert.ToTimestamp(removed); err == nil {
-		if time.Unix(i/1000, 0).Add(5 * time.Minute).Before(time.Now()) {
+		if time.Unix(i/1000, 0).Add(5 * time.Minute).Before(time.Now().UTC()) {
 			summary.Error = true
 		}
 	}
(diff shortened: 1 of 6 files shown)


p-se (approved):

internal/cmd/agent/deployer/summary/summarizers.go
```diff
@@ -353,7 +353,7 @@ func checkInitializing(obj data.Object, conditions []Condition, summary fleetv1.
 	if summary.State == "" && hasConditions && len(conditions) == 0 && strings.Contains(apiVersion, "cattle.io") {
 		val := obj.String("metadata", "created")
 		if i, err := convert.ToTimestamp(val); err == nil {
-			if time.Unix(i/1000, 0).Add(5 * time.Second).After(time.Now()) {
+			if time.Unix(i/1000, 0).Add(5 * time.Second).After(time.Now().UTC()) {
```
manno, line 356: In Go time includes the time zone.
So, when comparing times,  `if time.Now().UTC().After(expire) {` is no different than `if time.Now().After(expire) {`?

internal/cmd/agent/deployer/summary/summarizers.go
```diff
@@ -393,7 +393,7 @@ func checkRemoving(obj data.Object, conditions []Condition, summary fleetv1.Summ
 
 	summary.Message = append(summary.Message, "waiting on "+f)
 	if i, err := convert.ToTimestamp(removed); err == nil {
-		if time.Unix(i/1000, 0).Add(5 * time.Minute).Before(time.Now()) {
+		if time.Unix(i/1000, 0).Add(5 * time.Minute).Before(time.Now().UTC()) {
```
manno, line 396: same

internal/cmd/controller/agentmanagement/controllers/clusterregistrationtoken/handler.go
```diff
@@ -271,7 +271,7 @@ func (h *handler) deleteExpired(token *fleet.ClusterRegistrationToken) (bool, er
 		return false, nil
 	}
 	expire := token.CreationTimestamp.Add(ttl.Duration)
-	if time.Now().After(expire) {
+	if time.Now().UTC().After(expire) {
```
manno, line 274: same

manno (changes requested):

internal/cmd/agent/deployer/summary/summarizers.go
weyfonk: Indeed, a few tests reveal that it doesn't make any difference, so we might as well save ourselves the additional method call.

internal/cmd/controller/imagescan/gitcommit_job.go
```diff
@@ -219,7 +219,7 @@ func (j *GitCommitJob) cloneAndReplace(ctx context.Context) {
 	if interval == nil || interval.Seconds() == 0.0 {
 		interval = &DefaultInterval
 	}
-	gitrepo.Status.LastSyncedImageScanTime = metav1.NewTime(time.Now())
+	gitrepo.Status.LastSyncedImageScanTime = metav1.NewTime(time.Now().UTC())
```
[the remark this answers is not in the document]
weyfonk: It sure does! Closing this PR as it is just plain noise then. Thanks for pointing this out :)

PR #3139, opened 2024-12-11
dev-scripts: Spawn arbitrary amount of downstream clusters

p-se:
Using the environment variable FLEET_E2E_DS_CLUSTER_COUNT, an arbitrary
amount of downstream clusters can be spawned, e.g.:

```
FLEET_E2E_DS_CLUSTER_COUNT=4 ./dev/setup-multi-cluster
```

This environment variable affects

- dev/setup-k3ds
- dev/import-images-k3d
- dev/setup-multi-cluster

### DIFF
--- a/dev/import-images-k3d
+++ b/dev/import-images-k3d
@@ -2,13 +2,26 @@
 
 set -euxo pipefail
 
+# The upstream cluster to import all the images to.
 upstream_ctx="${FLEET_E2E_CLUSTER-k3d-upstream}"
+
+# The single downstream cluster to import the agent image to.
 downstream_ctx="${FLEET_E2E_CLUSTER_DOWNSTREAM-k3d-downstream}"
+
+# If multi-cluster is enabled, import the agent image to all downstream clusters.
+FLEET_E2E_DS_CLUSTER_COUNT="${FLEET_E2E_DS_CLUSTER_COUNT:-1}"
 
 k3d image import rancher/fleet:dev rancher/fleet-agent:dev -m direct -c "${upstream_ctx#k3d-}"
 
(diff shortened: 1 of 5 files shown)


dev/import-images-k3d
```diff
@@ -2,13 +2,26 @@
...
+# If multi-cluster is enabled, import the agent image to all downstream clusters.
+FLEET_E2E_DS_CLUSTER_COUNT="${FLEET_E2E_DS_CLUSTER_COUNT:-1}"
+
 k3d image import rancher/fleet:dev rancher/fleet-agent:dev -m direct -c "${upstream_ctx#k3d-}"
 
 if [ "$upstream_ctx" != "$downstream_ctx" ]; then
-  k3d image import rancher/fleet-agent:dev -m direct -c "${downstream_ctx#k3d-}"
+  if [ "$FLEET_E2E_DS_CLUSTER_COUNT" -gt 1 ]; then
+    for cluster in $(k3d cluster list -o json | \
+        jq -r ".[].name | select(. | contains(\"${downstream_ctx#k3d-}\"))"); do
```
weyfonk, line 19: This will not work if `FLEET_E2E_CLUSTER_DOWNSTREAM` is set to a value other than `k3d-downstream`, right?
Unless `setup-k3ds` is updated to take that environment variable is taken into account for cluster naming.

weyfonk: Right, so if I'm understanding this well, the process towards adding more flexibility to this would start with editing `setup-k3ds` anyway, which is run earlier than this script anyway.
That makes sense to me. Resolving, thank you.

weyfonk (approved):

PR #3015, opened 2024-10-25
Limit debug error messages to debug log

manno:
Error() is printed at all verbosities

PR #3146, opened 2024-12-12
[v0.10] Calculate Clusters resourceCount from BundleDeployments instead of GitRepos

aruiz14:
Refers to https://github.com/rancher/rancher/issues/48397
Backports https://github.com/rancher/fleet/pull/3102


internal/cmd/controller/reconciler/cluster_controller.go
```diff
@@ -144,20 +150,21 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	toDeleteBundles := indexByNamespacedName(cleanup)
+
+	// Delete BundleDeployments for Bundles being removed while getting a filtered items list
+	bundleDeployments.Items = slices.DeleteFunc(bundleDeployments.Items, func(bd fleet.BundleDeployment) bool {
+		bundleNamespace := bd.Labels[fleet.BundleNamespaceLabel]
+		bundleName := bd.Labels[fleet.BundleLabel]
+		if _, ok := toDeleteBundles[types.NamespacedName{Namespace: bundleNamespace, Name: bundleName}]; ok {
+			logger.V(1).Info("cleaning up bundleDeployment not matching the cluster", "bundledeployment", bd)
+			if err := r.Delete(ctx, &bd); err != nil {
+				logger.V(1).Info("deleting bundleDeployment returned an error", "error", err)
```
aruiz14, line 162: @manno it seems like #3015 was not backported to 0.10 and this produced a conflict in this line.
Is it okay to keep the version that uses `Info` (like in 0.11) or should I change it back to use `Error`?:
```suggestion
				logger.V(1).Error(err, "deleting bundleDeployment returned an error")
```

weyfonk (approved):
LGTM, pending clarification on the above comment.

internal/cmd/controller/reconciler/cluster_controller.go
aruiz14: I've reconstructed the changes and used `logger.V(1).Error` to preserve the original behavior

manno: 👍

Issue #3138, opened 2024-12-11
[SURE-9460] Fleet not picking up gitrepo updates, no job created to update


kkaempf:
### Issue description
After upgrading Rancher to 2.9.3 / fleet to v0.10.4, some gitrepos are no longer receiving updates. Customer update the repository, but changes are not pushed to the clusters. No Job is created to pull in the changes that should be tracked by the gitRepo.

In fleet v0.10.4, there were changes made to how jobs are managed in fleet. Could these changes be the cause of the issue here? https://github.com/rancher/fleet/pull/2932 seems to change how jobs are managed.

### Business impact:
Unable to receive updates to applications using fleet for continuous delivery.

### Troubleshooting steps:
GitJob pod, does not show that jobs are completing for those gitRepos, We are also unable to find jobs for the

### Repro steps:

Upgrade to Rancher 2.9.3 from 2.9.2

### Workaround:
Is a workaround available and implemented? yes
What is the workaround:
Customer found that by editing a gitRepo in the Rancher UI, changing nothing, then saving, it will eventually cause the repo to pull the change and make the necessary updates.

When making those changes, a couple lines are changed within the gitRepo:
spec.correctDrift: {} is added
status.commit is updated
status.lastPollingTriggered time is updated (time changed by more than a day).

### Actual behavior:
repositories are not updated.

### Expected behavior:
Repositories are updated.

carneiroskeeled:
@kkaempf have you tried to enable this?
```
  #  correctDrift:
  #    enabled: false
  ```

manno:
We suspect that the GitRepo drops out of the new polling. The new polling is based on `RequeueAfter` and uses the reconcilers workqueue.

We suspect this could happen, because of 
* a condition prevents the resource from being queued, when it should
* resource drops out of requeue because of an unhandled error
* not enough reconcile workers, so that some reconciles get delayed indefinitely

Until we can reproduce this:
We are adding some jitter to the polling, so that 300 gitrepos don't reconcile at the same time.
We shorten the resync period to pick up dropped gitrepos.

rancherbot:
Not creating port issue, milestone releasev0.11 does not exist or is not an open milestone

mmartin24:
Verified as well `GITREPO_SYNC_PERIOD` default set to 2 hours in `v2.10-ee01a8e72c881512d9d2352229a46fcea7c9d8e8-head` with fleet `105.0.3+up0.11.3-rc.2`:

```yaml
   spec:
[...]

PR #3151, opened 2024-12-13
Add jitter and resync to polling

manno:
refers to https://github.com/rancher/fleet/issues/3138

* Add jitter to the pollingInterval of GitRepos, jitter is plus minus 1/10 the polling interval
* Gitops controller resyncs every hour, if gitrepos are lost from requeueAfter polling, resync should add them again.
* Since the 10h cache sync is an update event, change predicates like GenerationChanged, prevent the reconcile from running.

### DIFF
--- a/internal/cmd/controller/gitops/reconciler/gitjob_controller.go
+++ b/internal/cmd/controller/gitops/reconciler/gitjob_controller.go
@@ -103,8 +103,9 @@ func (r *GitJobReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&v1alpha1.GitRepo{},
 			builder.WithPredicates(
-				// do not trigger for GitRepo status changes (except for commit changes)
+				// do not trigger for GitRepo status changes (except for commit changes and cache sync)
 				predicate.Or(
+					TypedResourceVersionUnchangedPredicate[client.Object]{},
 					predicate.GenerationChangedPredicate{},
 					predicate.AnnotationChangedPredicate{},
 					predicate.LabelChangedPredicate{},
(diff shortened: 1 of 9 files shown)


manno:
Why, oh why would gosec think this is crypto:

```
 Error: internal/cmd/controller/gitops/reconciler/gitjob_controller.go:214:27: G404: Use of weak random number generator (math/rand or math/rand/v2 instead of crypto/rand) (gosec)
  	return d + time.Duration(rand.Int64N(int64(d)/5)-int64(d)/10)
  	                         ^
```

weyfonk (approved):

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -98,12 +99,42 @@ type GitJobReconciler struct {
...
+	return e.ObjectNew.GetResourceVersion() == e.ObjectOld.GetResourceVersion()
+}
+
 func (r *GitJobReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&v1alpha1.GitRepo{},
 			builder.WithPredicates(
 				// do not trigger for GitRepo status changes (except for commit changes)
 				predicate.Or(
+					TypedResourceVersionUnchangedPredicate[client.Object]{},
```
aruiz14, line 137: Could you please elaborate on why this is needed? Since this is used with `predicate.Or`, wouldn't any update to the resources (which make the resource version change) make it return `true`, making the rest of predicates redundant?

manno: I added more information to the commit, as discussed this relates to https://github.com/kubernetes-sigs/controller-runtime/issues/2355

PR #3156, opened 2024-12-18
dev/k3d-act-clean cleanup

p-se:
- Clean up downstream clusters if FLEET_E2E_DS_CLUSTER_COUNT is set
- Remove cleaning up any docker containers by name, including from
  nektos/act, therefore renaming k3d-act-clean to k3d-clean.

Follow-up for #3139


weyfonk (approved):
LGTM :)

dev/setup-k3ds
```diff
@@ -32,12 +32,16 @@ k3d cluster create upstream \
   $args
 
 for i in $(seq 1 "$FLEET_E2E_DS_CLUSTER_COUNT"); do
-  k3d cluster create "downstream$i" \
+  cluster="downstream"
```
manno, line 35: Should we just make it configurable, so we can get rid of `setup-k3d`, which was meant for more "customized" creation of clusters?

dev/setup-single-cluster
```diff
@@ -7,7 +7,7 @@ export CUSTOM_CONFIG_FILE="env.single-cluster"
 source ./dev/setup-cluster-config
 
 # Cleans with settings sourced, so it should be rather selective.
-./dev/k3d-act-clean
+./dev/k3d-clean
 
 RANDOM_PORT=0
 if [ -z ${external_ip} ];
```
manno, line 13: Always guard against globbing and word splitting.

```suggestion
if [ -z "$external_ip" ];
```

manno (approved):

PR #3165, opened 2024-12-20
Remove HELM_PATH, tests no longer use it

manno:
3f61ba5e removed the usage of the HELM_PATH environment variable.


weyfonk (approved):
Thanks for cleaning this up :+1:

PR #3177, opened 2025-01-03
Improve charts release script

thardeck:
The validate script seems to sort the release.yaml file so we have to do it too in our automation.

Followup to https://github.com/rancher/charts/actions/runs/12596195018/job/35106849382?pr=4918


kkaempf (approved):
looks ok

Issue #3191, opened 2025-01-09
[backport v2.9] [SURE-9137] ClusterValues dont apply changes if one of the clusters is missing the templateValues


rancherbot:
This is a backport issue for #2943, automatically created via [GitHub Actions workflow](https://github.com/rancher/fleet/actions/runs/12688685387) initiated by @p-se

Original issue body:

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

If a GitRepo is configured to target two or more clusters and the `fleet.yaml` file includes` ${ .ClusterValues}`, any missing templateValues in one of the cluster's spec will prevent updates or changes from being deployed to the clusters where `templateValues` are properly configured.

### Expected Behavior

- The changes should be applied in the cluster where the templatesValues are defined.
- UI should show the clear Error message 

### Steps To Reproduce

1. Install rancher 2.9.2 with fleet 0.10.3v
2. Register two downstream clusters, ensuring that one of them includes `templateValues`.

```
apiVersion: fleet.cattle.io/v1alpha1
kind: Cluster
metadata:
  annotations:
  labels:
    foo: bar
    management.cattle.io/cluster-display-name: rke2custom1
    management.cattle.io/cluster-name: c-m-qmc767s2
    objectset.rio.cattle.io/hash: 464bd091084175e4d5572051571f4dfb39bcf2fd
    provider.cattle.io: rke2
  name: rke2custom1
  namespace: fleet-default
spec:
  agentAffinity:
    nodeAffinity:
      preferredDuringSchedulingIgnoredDuringExecution:
        - preference:
            matchExpressions:
              - key: fleet.cattle.io/agent
                operator: In
                values:
                  - 'true'
          weight: 1
  clientID: pl882vs458n4lqqrj8jc58jvkvq4xgqdfv9l7q7spnrhh7s8wjgj8v
  kubeConfigSecret: rke2custom1-kubeconfig
  kubeConfigSecretNamespace: fleet-default
  templateValues:
    generated:
      cluster_metadata:
        fqdn: server-1.example.com
        name: server-1
```

3. create gitrepo from this [example](https://github.com/skanakal/fleet-examples/tree/templateValues)  path: `templateValues`
4. check the gitrepo dashboard for resourceReady

### Environment

```markdown
- Architecture: x86_64
- Fleet Version: fleet:104.0.3+up0.10.3
- Cluster:
  - Provider: custom
  - Options: 1
[...]

Issue #2943, opened 2024-10-08
[SURE-9137] ClusterValues dont apply changes if one of the clusters is missing the templateValues


skanakal:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

If a GitRepo is configured to target two or more clusters and the `fleet.yaml` file includes` ${ .ClusterValues}`, any missing templateValues in one of the cluster's spec will prevent updates or changes from being deployed to the clusters where `templateValues` are properly configured.

### Expected Behavior

- The changes should be applied in the cluster where the templatesValues are defined.
- UI should show the clear Error message 

### Steps To Reproduce

1. Install rancher 2.9.2 with fleet 0.10.3v
2. Register two downstream clusters, ensuring that one of them includes `templateValues`.

```
apiVersion: fleet.cattle.io/v1alpha1
kind: Cluster
metadata:
  annotations:
  labels:
    foo: bar
    management.cattle.io/cluster-display-name: rke2custom1
    management.cattle.io/cluster-name: c-m-qmc767s2
    objectset.rio.cattle.io/hash: 464bd091084175e4d5572051571f4dfb39bcf2fd
    provider.cattle.io: rke2
  name: rke2custom1
  namespace: fleet-default
spec:
  agentAffinity:
    nodeAffinity:
      preferredDuringSchedulingIgnoredDuringExecution:
        - preference:
            matchExpressions:
              - key: fleet.cattle.io/agent
                operator: In
                values:
                  - 'true'
          weight: 1
  clientID: pl882vs458n4lqqrj8jc58jvkvq4xgqdfv9l7q7spnrhh7s8wjgj8v
  kubeConfigSecret: rke2custom1-kubeconfig
  kubeConfigSecretNamespace: fleet-default
  templateValues:
    generated:
      cluster_metadata:
        fqdn: server-1.example.com
        name: server-1
```

3. create gitrepo from this [example](https://github.com/skanakal/fleet-examples/tree/templateValues)  path: `templateValues`
4. check the gitrepo dashboard for resourceReady

### Environment

```markdown
- Architecture: x86_64
- Fleet Version: fleet:104.0.3+up0.10.3
- Cluster:
  - Provider: custom
  - Options: 1
  - Kubernetes Version: v1.30.5+rke2r1
```

### Logs

```markdown
From the fleet-controller logs:

[...]

PR #3196, opened 2025-01-09
[backport v2.9] [SURE-9137] Add template errors to bundle and gitrepo status

p-se:
* Import v1alpha1 package as fleet

* Show bundle errors in Bundle and GitRepo

Refers to #2943

* Add E2E tests

Refers to #2943

(cherry picked from commit 235e8efa1e857fc482ddafc7b415ec760b343d24) (cherry picked from commit 34170715e2947a413c0d51fae3da6c525f2c0257)

Refers to #3191

### DIFF
--- a/internal/cmd/controller/reconciler/bundle_controller.go
+++ b/internal/cmd/controller/reconciler/bundle_controller.go
@@ -5,8 +5,10 @@ package reconciler
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/go-logr/logr"
+
 	"github.com/rancher/fleet/internal/cmd/controller/finalize"
 	"github.com/rancher/fleet/internal/cmd/controller/summary"
 	"github.com/rancher/fleet/internal/cmd/controller/target"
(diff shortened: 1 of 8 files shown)


internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -215,6 +220,18 @@ func (r *BundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 
 	matchedTargets, err := r.Builder.Targets(ctx, bundle, manifestID)
```
weyfonk, line 221: Do we want to take the opportunity to remove `RetryOnConflict` loops (around L150) for adding and removing finalizers here, as done for other release branches?

manno (approved):

internal/cmd/controller/reconciler/bundle_controller.go
manno: Is that like a one liner? Otherwise it sounds to me like a different issue/PR.

Issue #3186, opened 2025-01-08
Feature Request: extend GitRepoStatus for easier differentiation of per-cluster state


aruiz14:
### Is your feature request related to a problem?

While working on #3102 and https://github.com/rancher/dashboard/issues/12424, I realized that the dashboard has to do non-trivial computations in order to differentiate each cluster, which actually lead to a bug that made the resource list not to contain accurate information.

### Solution you'd like

The UI was changed to use BundleDeployments for calculating the list of resources, but Fleet could ease this task by exposing the necessary information in a way that can directly be consumed (as far as I could see, Fleet is already doing the computation part that would allow this, it only needs to extend its API to expose it).

manno:
Refering this comment which has some example json: https://github.com/rancher/dashboard/issues/11643#issuecomment-2442249231

PR #3209, opened 2025-01-10
Add PerClusterResourceCounts to GitRepo status

aruiz14:
Refers to #3186

This PR extends the current public API to add a new field to every GitRepoStatus:
 - `perClusterResourceCounts`, which is a map of related resource counts by cluster ID.
It uses the new status field added to BundleDeploymentStatus by #3203

In addition, it removes the dependency on the GitRepo's `status.summary` field, which is just a pre-calculated summary of BundleDeployment statuses, which was then aggregated and used to assign a resource state. Instead, I've changed it to use the actual per-BundleDeployment state since it's available.


kkaempf:
@aruiz14 - this is still flagged as draft 🤔

aruiz14:
> @aruiz14 - this is still flagged as draft 🤔

Yes. It depends on another PR being reviewed, but Github it's not nice showing stacked PRs

internal/resourcestatus/resourcekey.go
```diff
@@ -5,21 +5,29 @@ import (
...
+	status.ResourceCounts = sumResourceCounts(list)
+}
+
+func SetGitRepoResources(list *fleet.BundleDeploymentList, status *fleet.GitRepoStatus) {
+	setBaseStatusResources(list, &status.StatusBase)
+	status.PerClusterResourceCounts = resourceCountsPerCluster(list)
+}
+
+func SetHelmAppResources(list *fleet.BundleDeploymentList, status *fleet.HelmAppStatus) {
+	setBaseStatusResources(list, &status.StatusBase)
```
weyfonk, line 26: My understanding of this is that we don't yet want/need per-cluster resource counts on HelmApp because that resource is not yet visible in the Rancher UI. Is that correct, or am I missing an important aspect of this?

internal/resourcestatus/resourcekey_test.go
```diff
@@ -116,7 +116,7 @@ var _ = Describe("Resourcekey", func() {
 	})
 
 	It("returns a list", func() {
-		SetResources(list, &gitrepo.Status.StatusBase)
+		SetGitRepoResources(list, &gitrepo.Status)
```
weyfonk, line 119: I think it could make sense to validate the `gitrepo`'s `PerClusterResourceCounts` here. How do you see it?

weyfonk (commented):
LGTM, pending

weyfonk (approved):
LGTM, although I'd be more comfortable with an additional review from the team.

0xavi0 (approved):
lgtm

PR #3226, opened 2025-01-20
Fix ./internal/resourcestatus tests not being run

aruiz14:
While working on #3209, I realized these tests are being ignored, due to the absence of a Ginkgo suite file:
```
ok  	github.com/rancher/fleet/internal/resourcestatus	0.003s [no tests to run]
```
https://github.com/rancher/fleet/actions/runs/12823018691/job/35756872772

The existing test does not need a complex setup, so there is no need to use Ginkgo.

### DIFF
--- a/internal/resourcestatus/resourcekey_test.go
+++ b/internal/resourcestatus/resourcekey_test.go
@@ -1,8 +1,9 @@
 package resourcestatus
 
 import (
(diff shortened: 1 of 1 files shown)


weyfonk (approved):
Nice catch, thanks!

PR #3227, opened 2025-01-20
Adds new gitops poller controller

0xavi0:
Adds a new poller controller to avoid sharing events between the main `GitRepo` controller and the polling mechanism. New poller controller only does polling and resets its state upon generation change.

When a new commit is found by the poller controller it sets the `PollerCommit` field in the `GitRepo`'s status. The main controller watches for changes in that field and it is called to reflect the new commit in the `GitRepo`'s `Status.Commit` field.

A new `PollerGeneration` field is added so it does not interfere with the already existing `ObservedGeneration` used in the main `GitRepo` controller

### Changes in e2e tests
It now uses `BeforeAll` and `AfterAll` for OCI storage `e2e` tests and test are ordered. This is to avoid having to change the experimental env variable for almost each test. That was taking too much time and was being a source of errors. 

### Changes in integration tests
There is one integration test that was removed because it was passing because the poller inside the main controller was triggering an extra reconcile call. 
The test makes no sense as `gitjobs` are owned by the controller and we already look for the `succeded/failed` lifecycle

Refers to https://github.com/rancher/fleet/issues/3138

### DIFF
--- a/charts/fleet-crd/templates/crds.yaml
+++ b/charts/fleet-crd/templates/crds.yaml
@@ -6662,6 +6662,15 @@ spec:
                   description: PerClusterResourceCounts contains the number of resources
                     in each state over all bundles, per cluster.
                   type: object
+                pollerCommit:
+                  description: PollerCommit is the latest Git commit gathered by the
+                    GitRepo poller
+                  type: string
+                pollerGeneration:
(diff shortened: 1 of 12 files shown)


0xavi0:
Closing this one after https://github.com/rancher/fleet/pull/3307 was merged

Issue #2750, opened 2024-08-16
[SURE-8809] Fleet deployment fails when Helm chart repo uses custom CA / missing error


kkaempf:
### Issue description:

When
adding a git repo to fleet that has a fleet.yaml referencing an external helm chart.

and

the server serving the helm chart uses a custom CA

Then

the Git Repo is added and marked with State "active"

the "Clusters Ready" status remains on "0/0" indefinitely

no error is thrown or visible in the Fleet UI

### Expected behavior:
1) Fleet should honor the custom CA configured in the Rancher global settings or the custom CA configured in the GitRepo resource (spec.CaBundle) when downloading resources, such as Helm charts.

2) Fleet should display an error message in the UI, indicating that there was a problem downloading the Helm chart

### Business impact:
Customer can't use any helm charts from internet sources, because their corporate firewall performs TLS inception, replacing all SSL certs with their own CA.

manno:
We should add support to Fleet to fall back to a default value, unless overridden in the resource. 
This is most likely a Fleet install option, I think Fleet already gets re-installed when the Rancher CA changes.

However there are multiple clients in Fleet, which would need to support this. They all use a PEM block for CA in their spec instead. 
Specifying the CA in the resource directly has the advantage that we don't need to watch another resource for changes, e.g. to redeploy on certificate rotation. That's not possible, when we rely on a global setting.
Does re-installing Fleet re-render all bundles with the new CA, or do we need to implement this?

* git monitor (lsremote)
* git cloner
* chart downloader
* image scan (no custom CA yet)

marthydavid:
I've faced the same issue.

weyfonk:
## Additional QA
 
### Problem
Fleet fails to deploy a workload pointing to a Helm registry using a custom CA.
 
### Solution
Fleet takes Rancher-configured CA bundle [secrets](https://ranchermanager.docs.rancher.com/getting-started/installation-and-upgrade/installation-references/helm-chart-options#additional-trusted-cas) into account, enabling those to be used by default if a `GitRepo` does not provide a CA bundle. This happens in interactions with both git and Helm.

See new Fleet [docs](https://fleet.rancher.io/gitrepo-add#using-custom-ca-bundles).
 
### Testing
### Engineering Testing
See description of [this PR](https://github.com/rancher/fleet/pull/3233).

#### Manual Testing
N/A: manual testing ended up being automated.

#### Automated Testing
* Unit tests for:
[...]

PR #3233, opened 2025-01-22
Fall back to Rancher-configured CA bundles

weyfonk:
This enables Fleet to rely on Rancher-configured CA bundle [secrets](https://ranchermanager.docs.rancher.com/getting-started/installation-and-upgrade/installation-references/helm-chart-options#additional-trusted-cas) in the following clients:
* git remote lister and cloner: those CA bundles no longer need to be provided in a `GitRepo`'s spec. Fleet will extract CA bundle data from Rancher secrets if (and only if) no CA bundle is provided for a `GitRepo`.
* Helm client: Fleet will extract, populate and mount Rancher CA bundles in the following cases:
    * no `HelmSecretName` nor `HelmSecretNameForPaths` secret is provided on a `GitRepo`
    * one of those secrets is provided, but only specifies username(s) and password(s), without a CA bundle

This is tested through a combination of:
* unit tests for common Rancher CA bundle data extraction logic
* integration tests for mounting and use of that data by the gitOps controller
* end-to-end tests for successful deployment of workloads with git repositories and Helm charts hosted on infrastructure using a custom CA. Those tests required adapting configuration and deployment of our test git server to support HTTPS beside HTTP and make use of already generated custom certificates.

## Limitations

This does not cover imagescan (pending further discussion on the future of that feature), nor the HelmOps controller.

Falling back to Rancher CA bundles in HelmOps would require reading the HelmApp's `HelmSecretName` field, and populating a CA bundle if that field is empty or if the referenced secret does not contain any CA bundle.

How this should be done to support both [version handling](https://github.com/rancher/fleet/blob/main/internal/cmd/controller/helmops/reconciler/helmapp_controller.go#L294) in the controller and chart downloads from the agent, is unclear for now: should an empty `HelmSecretName` be overwritten by the HelmOps controller, or an incomplete secret edited to add a Rancher CA bundle? This could be handled in a follow-up patch.
    
Refers to #2750
Related docs PR: https://github.com/rancher/fleet-docs/pull/212

### DIFF
--- a/e2e/single-cluster/gitrepo_test.go
+++ b/e2e/single-cluster/gitrepo_test.go
@@ -26,8 +26,9 @@ import (
 )
 
 const (
-	port     = 8080
-	repoName = "repo"
+	port      = 8080
+	HTTPSPort = 4343
+	repoName  = "repo"
 )
 
 var _ = Describe("Monitoring Git repos via HTTP for change", Label("infra-setup"), func() {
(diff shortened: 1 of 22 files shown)


dev/setup-k3d
```diff
@@ -50,5 +50,6 @@ k3d cluster create "$name" \
   -p "$(( 8081 + offs )):8081@server:0" \
   -p "$(( 8082 + offs )):8082@server:0" \
   -p "$unique_tls_port:443@server:0" \
+  -p "8043:4343@server:0" \
```
0xavi0, line 53: Question: Why is not using the offset like the rest?

e2e/single-cluster/gitrepo_test.go
```diff
@@ -369,6 +370,184 @@ var _ = Describe("Monitoring Git repos via HTTP for change", Label("infra-setup"
...
+
+	JustBeforeEach(func() {
+
+		// Build git repo URL reachable _within_ the cluster, for the GitRepo
+		host, err := githelper.BuildGitHostname(env.Namespace)
+		Expect(err).ToNot(HaveOccurred())
+
+		addr, err := githelper.GetExternalRepoAddr(env, HTTPSPort, repoName)
+		Expect(err).ToNot(HaveOccurred())
+		addr = strings.Replace(addr, "http://", "https://", 1)
```
0xavi0, line 399: Question: 
Are these tests the same than the HTTP ones except for this line? 
(I haven't checked 100% but both blocks look very similar)
If so... maybe better to just add the `http/https` as a variable and reuse the code?

e2e/single-cluster/helm_auth_test.go
```diff
@@ -62,9 +64,38 @@ var _ = Describe("Single Cluster Examples", Label("infra-setup"), func() {
 				}).Should(ContainSubstring("sleeper-"))
 			})
 		})
+		Context("containing a private HTTP-based helm chart with repo path and no configured Helm secret name", Label("helm-registry"), func() {
```
0xavi0, line 67: It says "and no configured Helm secret"  but it is passing a secret? is it a typo?

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -651,28 +686,66 @@ func (r *GitJobReconciler) newJobSpec(ctx context.Context, gitrepo *v1alpha1.Git
...
 		volumeMounts = append(volumeMounts, volMnts...)
 	}
 
+	// In case no Helm secret volume has been created, because Helm secrets don't exist or don't contain a CA
+	// bundle, mount a volume with a Rancher CA bundle.
+	if !certVolCreated {
+		// Fall back to Rancher-configured secrets
+		// We need to copy secret data from Rancher, because Rancher secrets live in a different namespace and
+		// can therefore not be used as sources for a volume.
+		secretName := fmt.Sprintf("%s-rancher-cabundle", gitrepo.Name)
```
0xavi0, line 724: Nit: Maybe better to have these names in constants or a function? Similar to the `caBundleName` function.

0xavi0 (approved):
LGTM.

I wonder if we could get the Rancher secrets (names and values) from a configMap or something. (not sure if such configMap exists in Rancher or we could have that in fleet to centralise things)

And... in the same way....I would maybe set all the hardcoded strings for certs paths, secret values, keys, secret names, etc... as constants or functions. (So it is easier and faster to change them if needed)

e2e/single-cluster/helm_auth_test.go
weyfonk: Good catch! Fixed, thanks :)

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
weyfonk: Moved this one to a separate function. I think that moving others to separate variables/functions could be done as part of a separate effort, refactoring building of git jobs.

e2e/single-cluster/gitrepo_test.go
0xavi0: The loop stuff is cool! thanks!

PR #3238, opened 2025-01-22
Rework how resources are calculated

aruiz14:
Refers to #3186

This PR concludes the work started on #3203 and #3209, by refactoring how the `status.resources` (and especially `perClusterState`) are calculated.
The goal of this refactor is to keep all the information accessible, while still producing the same opinionated representation. This will allow to later change the format in an easier way, or adding new fields from the same information (e.g. `PerClusterResources`).

### DIFF
--- a/internal/resourcestatus/resourcekey.go
+++ b/internal/resourcestatus/resourcekey.go
@@ -44,10 +44,8 @@ type resourceStateEntry struct {
 
 type resourceStatesByResourceKey map[fleet.ResourceKey][]resourceStateEntry
 
-// fromResources inspects all bundledeployments for this GitRepo and returns a list of
-// Resources and error messages.
-//
-// It populates gitrepo status resources from bundleDeployments. BundleDeployment.Status.Resources is the list of deployed resources.
+// fromResources inspects a list of BundleDeployments and returns a list of per-cluster states per resource keys.
+// It also returns a list of errors messages produced that may have occurred during processing
 func fromResources(list *fleet.BundleDeploymentList) (resourceStatesByResourceKey, []string) {
 	var (
 		errors    []string
--- a/internal/resourcestatus/resourcekey.go
+++ b/internal/resourcestatus/resourcekey.go
@@ -10,9 +10,9 @@ import (
 )
 
 func SetResources(list *fleet.BundleDeploymentList, status *fleet.StatusBase) {
-	r, errors := fromResources(list)
+	byCluster, errors := fromResources(list)
 	status.ResourceErrors = errors
-	status.Resources = merge(r)
+	status.Resources = aggregateResourceStatesClustersMap(byCluster)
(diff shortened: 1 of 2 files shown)


internal/resourcestatus/resourcekey.go
```diff
@@ -189,54 +135,68 @@ func addState(bd fleet.BundleDeployment, resources map[fleet.ResourceKey][]fleet
...
-		if k.Name == key.Name &&
-			k.APIVersion == key.APIVersion &&
-			k.Kind == key.Kind {
-			delete(states, key)
-			k.Namespace = ""
-			states[k] = append(existing, state)
+			// "Ready" states are currently omitted
+			if entry.State == "Ready" {
+				continue
+			}
```
aruiz14, line 180: When I mentioned in the description about `perClusterState` being opinionated, this is one of the causes: `Ready` resources are omitted from this view (and that forces the UI to use `BundleDeployments` for displaying GitRepo resources BTW). If we wanted to change that, it would be as simple as removing these lines.

internal/resourcestatus/resourcekey.go
```diff
@@ -63,124 +37,96 @@ func resourceCountsPerCluster(list *fleet.BundleDeploymentList) map[string]*flee
...
+	fleet.ResourcePerClusterState
+	incomplete bool
+}
+
+type resourceStatesByResourceKey map[fleet.ResourceKey][]resourceStateEntry
+
 // fromResources inspects all bundledeployments for this GitRepo and returns a list of
 // Resources and error messages.
 //
 // It populates gitrepo status resources from bundleDeployments. BundleDeployment.Status.Resources is the list of deployed resources.
```
weyfonk, line 50: This is not entirely relevant anymore: it should probably mention that the first return value lists resources by cluster.

internal/resourcestatus/resourcekey_test.go
```diff
@@ -161,6 +161,14 @@ func TestSetResources(t *testing.T) {
 		Transitioning:   false,
 		Message:         "",
 		PerClusterState: []fleet.ResourcePerClusterState{
+			{
+				State:     "WaitApplied",
+				ClusterID: "c-ns1/cluster1",
+			},
+			{
+				State:     "NotReady",
```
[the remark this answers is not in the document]
aruiz14: Good catch! The `BundleDeploymentStatus` was not complete (missing `Ready` and `NonModified` fields), so the calculated state was incorrect.

weyfonk (approved):
Nice one, thanks! LGTM

kkaempf:
Would we need to adapt UI for this ?

aruiz14:
> Would we need to adapt UI for this ?

@kkaempf no, no change is needed. It only changes how the info is calculated internally, while still producing the same result

Issue #3163, opened 2024-12-19
[SURE-9487] Fleet takes long time to create a bundledeployment (180+ gitrepos)


mmartin24:
Transcribed steps:

## Issue description:
At a larger gitrepo scale a applying a new gitrepo using labels takes a long time to create the bundledeployment

## Repro steps:

- create a huge number of gitrepos. (k apply -f gitrepos.yaml) gitrepos.yaml is attached to the case
- create one extra git repo that has targets using labels
```
  paths:
    - /hello-world
  repo: https://github.com/rbreddy/bundledependency.git
  targets:
    - clusterSelector:
        matchLabels:
          foo: bar
```
- watch the resources :
```
kubectl get clusters -n fleet-default --output-watch-events --watch-only | ts '%Y-%m-%d %H:%M:%S' | tee  -a output.txt 
kubectl get bundles -n fleet-default --output-watch-events --watch-only | ts '%Y-%m-%d %H:%M:%S' | tee -a output.txt
```
- Label a cluster to get a gitrepo targeting it:
```
kubectl label cluster.management.cattle.io <CLUSTER-ID> -n fleet-default foo=bar --overwrite | ts '%Y-%m-%d %H:%M:%S' | tee -a output.txt 
```
*watch resources on the DS cluster: e.g.
```
kubectl get pods -n hello --watch | ts '%Y-%m-%d %H:%M:%S' | tee -a output.txt
```

- Actual behavior:
fleet takes time to create bundle deployments and start creating resources

- Expected behavior:
fleet should immediately create bundle deployments and start creating resources

- Additional notes:
With 300 gitrepos it takes 2 minutes
<details>
  <summary>Click to expand</summary>
```
2024-12-04 15:48:57 cluster.management.cattle.io/c-m-hz9bmvz5 labeled
2024-12-04 15:48:57 EVENT      NAME    CLUSTERCLASS   PHASE         AGE    VERSION
2024-12-04 15:48:57 MODIFIED   test1                  Provisioned   2d4h   
2024-12-04 15:48:57 MODIFIED   test1                  Provisioned   2d4h   
2024-12-04 15:48:57 MODIFIED   test1                  Provisioned   2d4h   
2024-12-04 15:48:57 MODIFIED   test1                  Provisioned   2d4h   
2024-12-04 15:50:29 EVENT      NAME                           BUNDLEDEPLOYMENTS-READY   STATUS
[...]

PR #3245, opened 2025-01-23
Prevent empty Patch updates

aruiz14:
Refers to #3163

By adding additional logging to the BundleReconciler, I realized that (in my experiment), every operation was taking ~2.5s, and most of it was on the `updateStatus` function. It seems that the client implementation will perform the call besides producing an empty JSON patch (which BTW is correct in my opinion).

In order to prevent those spurious requests, we need to skip calling `Patch()`. This can be done at the cost of generating the diff twice, but it's still worth it.


aruiz14:
@kkaempf this PR targets `main`, so it should have a v2.11 milestone, right? I've created a backport issue for 2.10 and set the v2.10.3 milestone there.

weyfonk (approved):

PR #3261, opened 2025-01-27
Upload Docker images to Prime registry and sign them

thardeck:
Refers to #3075

[Successful run on the test branch](https://github.com/rancher/fleet/actions/runs/12991154974/job/36227983118)
[Failed run while pushing manifests to the Prime registry](https://github.com/rancher/fleet/actions/runs/12991021004/job/36227556793)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -98,7 +98,7 @@ jobs:
 
       - name: Prepare Docker environment to support sbom/provenance parameters
         run: |
-          sudo bash -c 'echo "{\"features\": {\"containerd-snapshotter\": true}}" >> /etc/docker/daemon.json'
+          sudo bash -c 'echo "{\"features\": {\"containerd-snapshotter\": true}}" > /etc/docker/daemon.json'
           sudo systemctl restart docker
 
       - name: Log into Docker Container registry


.github/workflows/release.yml
```diff
@@ -89,6 +92,14 @@ jobs:
             secret/data/github/repo/${{ github.repository }}/dockerhub/rancher/credentials password | DOCKER_PASSWORD ;
             secret/data/github/repo/${{ github.repository }}/rancher-prime-stg-registry/credentials username | STAGE_REGISTRY_USERNAME ;
             secret/data/github/repo/${{ github.repository }}/rancher-prime-stg-registry/credentials password | STAGE_REGISTRY_PASSWORD ;
+            secret/data/github/repo/${{ github.repository}}/rancher-prime-registry/credentials username | PRIME_REGISTRY_USERNAME ;
+            secret/data/github/repo/${{ github.repository}}/rancher-prime-registry/credentials password | PRIME_REGISTRY_PASSWORD ;
+            secret/data/github/repo/${{ github.repository }}/rancher-prime-registry/credentials registry | PRIME_REGISTRY;
+
+      - name: Prepare Docker environment to support sbom/provenance parameters
+        run: |
+          sudo bash -c 'echo "{\"features\": {\"containerd-snapshotter\": true}}" >> /etc/docker/daemon.json'
```
weyfonk, line 101: Curiosity: why do we need to invoke Bash explicitly here?
Wouldn't this do the same, or is it eg. a matter of (another) shell vs Bash?
```
          echo '{"features": {"containerd-snapshotter": true}}' >> /etc/docker/daemon.json
```

.goreleaser.yaml
```diff
@@ -180,14 +184,183 @@ dockers:
     - "--build-arg=BUILD_ENV=goreleaser"
     - "--platform=linux/arm64/v8"
 
+  # fleet-controller images
```
weyfonk, line 187: nit: we could perhaps differentiate comments here, such as:
```suggestion
  # Prime fleet-controller images
```

.github/workflows/release.yml
aruiz14: It's because of using `sudo`.
Using `sudo echo ... > /path/to/file`, writing to that file will not use the elevated privileges provided by sudo (hence not being able to write in restricted directories).
By wrapping the redirection inside the `bash` execution, it ensures the whole operation is performed as `root`

aruiz14, line 101: small nit-pick: @thardeck I guess that this file is not meant to include multiple JSON documents, right? Since you are writing a complete object there, I suggest to override any existing contents:
```suggestion
          sudo bash -c 'echo "{\"features\": {\"containerd-snapshotter\": true}}" > /etc/docker/daemon.json'
```

weyfonk: Oh, of course, that makes perfect sense :) Thanks!

weyfonk (approved):

PR #3265, opened 2025-01-28
Skip `snap` dependency for `yq` download

weyfonk:
The `snap` binary is no longer available on self-hosted, non-containerised Github runners.
Instead, this relies on good old `curl` to identify and download the latest version of `yq`.

This fixes failures such as [this one](https://github.com/rancher/fleet/actions/runs/12989655888/attempts/1).
See [success on feature branch](https://github.com/rancher/fleet/actions/runs/13007860259/job/36278658939).


weyfonk:
> I would prefer to have snap installed by apt and then continue with the process or did the distribution change too? (The only reason we use snap is because Ubuntu moved `yq` from packages to snap).
> 
> Especially in our release process binaries should be verified, and this is the case for distribution packages.

Thanks for this feedback; I've updated this patch to install `snap` through `apt`, as suggested.
This fixes the test chart generation workflow, as can be seen [here](https://github.com/rancher/fleet/actions/runs/13012130168).

Issue #3266, opened 2025-01-28
HelmApp values does not support non-string templating


Danil-Grigorev:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

After attempting CAAPF integration with `HelmApp` resource - currently is not fully working, due to inability to specify template in a `yaml` format (not as a string with `|` or `|-`, as it does not produce correct value)

It works for simple templates without range values, also is possible to edit `BundleDeployment` values directly to apply the data.

```yaml
apiversion: fleet.cattle.io/v1alpha1
kind: HelmApp
metadata:
  name: sample1
spec:
  helm:
    releaseName: projectcalico
    repo: https://docs.tigera.io/calico/charts
    chart: tigera-operator
    values:
      installation:
        cni:
          type: Calico
          ipam:
            type: HostLocal
        calicoNetwork:
          bgp: Disabled
          mtu: 1350
          ipPools: # Works only with |- here, which is unfortunately incorrect, as it creates a string value
            ${- range $cidr := .ClusterValues.Cluster.spec.clusterNetwork.pods.cidrBlocks }
            - cidr: "${ $cidr }"
              encapsulation: None
              natOutgoing: Enabled
              nodeSelector: all()${- end}
  insecureSkipTLSVerify: true
  targets:
  - clusterName: test1
``` 

### Expected Behavior

Have an option for `values` field as a `map[string]string` which would allow to process template data before serializing it as yaml.

### Environment

```markdown
- Architecture:
- Fleet Version:
- Cluster:
  - Provider:
  - Options:
  - Kubernetes Version:
```

### Logs

```markdown

```

Danil-Grigorev:
Having the value specified as
```yaml
values:
  installation: |-
```
Results in error on HelmApp, which is only due to invalid type passed into `installation`, a string:
```yaml
status:
  conditions:
  - lastUpdateTime: "2025-01-31T15:55:58Z"
    message: 'ErrApplied(1) [Bundle sample1: template: tigera-operator/templates/tigera-operator/02-tigera-operator.yaml:81:17:
      executing "tigera-operator/templates/tigera-operator/02-tigera-operator.yaml"
      at <.Values.installation.kubernetesProvider>: can''t evaluate field kubernetesProvider
      in type interface {}]'
    status: "False"
    type: Ready
```

If I instead switch to just `installation:` in the template and apply again, it is invalid YAML:
```bash
[...]

PR #3267, opened 2025-01-28
Add `templateValues` to HelmApp, Bundle, etc

Danil-Grigorev:
Refers to #3266 

Implement new field inside `Helm` options: `templateValues`, which performs templating of each individual value before converting data to yaml. This allows to use more complex template expressions then currently available to template string data, like creation of maps and lists.

### DIFF
--- a/internal/cmd/controller/target/builder.go
+++ b/internal/cmd/controller/target/builder.go
@@ -2,8 +2,10 @@ package target
 
 import (
 	"bytes"
+	"cmp"
 	"context"
 	"fmt"
+	"maps"
 	"sort"
 	"strings"
 	"text/template"
@@ -182,10 +184,15 @@ func preprocessHelmValues(logger logr.Logger, opts *fleet.BundleDeploymentOption
 	}
 
 	opts.Helm = opts.Helm.DeepCopy()
-	if opts.Helm.Values == nil || opts.Helm.Values.Data == nil {
-		opts.Helm.Values = &fleet.GenericMap{
-			Data: map[string]interface{}{},
-		}
+	opts.Helm.Values = cmp.Or(opts.Helm.Values, &fleet.GenericMap{
+		Data: map[string]interface{}{},
+	})
+
+	if opts.Helm.Values.Data == nil {
+		opts.Helm.Values.Data = map[string]interface{}{}
+	}
+
+	if opts.Helm.TemplateValues == nil && len(opts.Helm.Values.Data) == 0 {
 		return nil
 	}
 
@@ -211,6 +218,14 @@ func preprocessHelmValues(logger logr.Logger, opts *fleet.BundleDeploymentOption
 		if err != nil {
 			return err
 		}
+
+		templatedData, err := processTemplateValuesData(opts.Helm.TemplateValues, values)
+		if err != nil {
+			return err
+		}
+
+		maps.Copy(opts.Helm.Values.Data, templatedData)
+
 		logger.V(4).Info("preProcess completed", "releaseName", opts.Helm.ReleaseName)
 	}
 
@@ -276,6 +291,38 @@ func tplFuncMap() template.FuncMap {
 	delete(f, "tpl")
 
 	return f
+}
+
+func processTemplateValuesData(helmTemplateData map[string]string, templateContext map[string]interface{}) (map[string]interface{}, error) {
+	renderedValues := make(map[string]interface{}, len(helmTemplateData))
+
+	for k, v := range helmTemplateData {
+		// fleet.yaml must be valid yaml, however '{}[]' are YAML control
+		// characters and will be interpreted as JSON data structures. This
+		// causes issues when parsing the fleet.yaml so we change the delims
+		// for templating to '${ }'
+		tmpl := template.New("values").Funcs(tplFuncMap()).Option("missingkey=error").Delims("${", "}")
+		tmpl, err := tmpl.Parse(v)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse helm values template: %w", err)
+		}
+
+		var b bytes.Buffer
+		err = tmpl.Execute(&b, templateContext)
+		if err != nil {
+			return nil, fmt.Errorf("failed to render helm values template: %w", err)
+		}
+
+		var value interface{}
+		err = kyaml.Unmarshal(b.Bytes(), &value)
+		if err != nil {
+			return nil, fmt.Errorf("failed to interpret rendered template as helm values: %s, %v", b.String(), err)
+		}
+
+		renderedValues[k] = value
+	}
+
+	return renderedValues, nil
 }
 
 func processTemplateValues(helmValues map[string]interface{}, templateContext map[string]interface{}) (map[string]interface{}, error) {
--- a/internal/cmd/controller/target/target_test.go
+++ b/internal/cmd/controller/target/target_test.go
@@ -112,6 +112,17 @@ func TestProcessLabelValues(t *testing.T) {
 const bundleYamlWithTemplate = `namespace: default
 helm:
   releaseName: labels
+  templateValues:
+    mapData: |
+      ${- range $key := .ClusterValues.items }
+      "${ $key }":
+        nested: "true"
+      ${- end}
+    listData: |
+      ${- range $key := .ClusterValues.items }
+      - "${ $key }":
+        nested: "true"
+      ${- end}
   values:
     clusterName: "${ .ClusterLabels.name }"
     fromAnnotation: "${ .ClusterAnnotations.testAnnotation }"
@@ -154,6 +165,10 @@ func TestProcessTemplateValues(t *testing.T) {
 				"thirdTier": "bar",
 			},
 		},
+		"items": []string{
+			"one",
+			"two",
+		},
 		"list": []string{
 			"alpha",
 			"beta",
@@ -187,7 +202,7 @@ func TestProcessTemplateValues(t *testing.T) {
 
 	templatedValues, err := processTemplateValues(bundle.Helm.Values.Data, values)
 	if err != nil {
-		t.Fatalf("error during label processing %v", err)
+		t.Fatalf("error during templated values processing %v", err)
 	}
 
 	clusterName, ok := templatedValues["clusterName"]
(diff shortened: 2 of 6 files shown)


Danil-Grigorev:
With this change I was able to deploy helm chart with proper modifications:
```yaml

apiversion: fleet.cattle.io/v1alpha1
kind: HelmApp
metadata:
  finalizers:
  - fleet.cattle.io/helmapp-finalizer
  generation: 3
  name: sample1
  namespace: default
  resourceVersion: "23694"
  uid: 511580b5-733a-4304-9cbb-229448989a3a
spec:
  helm:
    chart: tigera-operator
    releaseName: projectcalico
    repo: https://docs.tigera.io/calico/charts
    templateValues:
      installation: |-
        cni:
          type: Calico
          ipam:
            type: HostLocal
        calicoNetwork:
          bgp: Disabled
          mtu: 1350
          ipPools:
            ${- range $cidr := .ClusterValues.Cluster.spec.clusterNetwork.pods.cidrBlocks }
            - cidr: "${ $cidr }"
              encapsulation: None
              natOutgoing: Enabled
              nodeSelector: all()${- end}
  ignore: {}
  insecureSkipTLSVerify: true
  targets:
  - clusterName: docker-demo
```
created correct `BundleDeployment`:
```yaml

apiversion: fleet.cattle.io/v1alpha1
kind: BundleDeployment
metadata:
  creationTimestamp: "2025-01-28T15:00:20Z"
  finalizers:
  - fleet.cattle.io/bundle-deployment-finalizer
  generation: 1
  labels:
    fleet.cattle.io/bundle-name: sample1
    fleet.cattle.io/bundle-namespace: default
    fleet.cattle.io/cluster: docker-demo
    fleet.cattle.io/cluster-namespace: default
    fleet.cattle.io/fleet-helm-name: sample1
    fleet.cattle.io/managed: "true"
  name: sample1
  namespace: cluster-default-docker-demo-9a1d57cc2ad8
  resourceVersion: "22820"
  uid: 37600a86-c9e4-4445-a757-5611c5d27ace
spec:
  deploymentID: :fb788a089566f35d87c7d4b2ed4a499a2d3e36315b44cfa9446f66996b6a7c82
  helmChartOptions:
    helmAppInsecureSkipTLSVerify: true
  options:
    helm:
      chart: tigera-operator
      releaseName: projectcalico
      repo: https://docs.tigera.io/calico/charts
      templateValues:
        installation: |-
          cni:
            type: Calico
            ipam:
              type: HostLocal
          calicoNetwork:
            bgp: Disabled
            mtu: 1350
            ipPools:
              ${- range $cidr := .ClusterValues.Cluster.spec.clusterNetwork.pods.cidrBlocks }
              - cidr: "${ $cidr }"
                encapsulation: None
                natOutgoing: Enabled
                nodeSelector: all()${- end}
      values:
        installation:
          calicoNetwork:
            bgp: Disabled
            ipPools:
            - cidr: 10.1.0.0/16
              encapsulation: None
              natOutgoing: Enabled
              nodeSelector: all()
            mtu: 1350
          cni:
            ipam:
              type: HostLocal
            type: Calico
      version: v3.29.1
    ignore: {}
  stagedDeploymentID: :fb788a089566f35d87c7d4b2ed4a499a2d3e36315b44cfa9446f66996b6a7c82
…
```

internal/cmd/controller/target/builder.go
```diff
@@ -182,10 +184,15 @@ func preprocessHelmValues(logger logr.Logger, opts *fleet.BundleDeploymentOption
...
-		opts.Helm.Values = &fleet.GenericMap{
-			Data: map[string]interface{}{},
-		}
+	opts.Helm.Values = cmp.Or(opts.Helm.Values, &fleet.GenericMap{
+		Data: map[string]interface{}{},
+	})
+
+	if opts.Helm.Values.Data == nil {
+		opts.Helm.Values.Data = map[string]interface{}{}
+	}
```
weyfonk, line 193: Why do we still need this block? My understanding is that `cmp.Or` would populate the `Data` field if needed :thinking:

internal/cmd/controller/target/builder.go
```diff
@@ -182,10 +184,15 @@ func preprocessHelmValues(logger logr.Logger, opts *fleet.BundleDeploymentOption
...
+
+	if opts.Helm.TemplateValues == nil && (opts.Helm.Values == nil || opts.Helm.Values.Data == nil) {
```
weyfonk, line 195: Since we've ensured that `opts.Helm.Values.Data` exists, wouldn't the following be enough?
```suggestion
	if opts.Helm.TemplateValues == nil && opts.Helm.Values.Data == nil {
```

internal/cmd/controller/target/builder.go
```diff
@@ -211,6 +218,14 @@ func preprocessHelmValues(logger logr.Logger, opts *fleet.BundleDeploymentOption
 		if err != nil {
 			return err
 		}
+
+		templatedData, err := processTemplateValuesData(opts.Helm.TemplateValues, values)
+		if err != nil {
+			return err
+		}
+
+		maps.Copy(opts.Helm.Values.Data, templatedData)
```
weyfonk, line 227: This effectively means that, if both a value and a template value exist with the same key, the latter will overwrite the former. Not a problem, but something we'll want to document, and ideally test :)

internal/cmd/controller/target/builder.go
```diff
@@ -278,6 +293,38 @@ func tplFuncMap() template.FuncMap {
...
+		var b bytes.Buffer
+		err = tmpl.Execute(&b, templateContext)
+		if err != nil {
+			return nil, fmt.Errorf("failed to render helm values template: %w", err)
+		}
+
+		var value interface{}
+		err = kyaml.Unmarshal(b.Bytes(), &value)
+		if err != nil {
+			return nil, fmt.Errorf("failed to interpret rendered template as helm values: %#v, %v", renderedValues, err)
```
weyfonk, line 319: Why output the whole set of rendered values here? Is the rationale to enable the user to understand where successful processing stopped? Otherwise, how about outputting `b.String()` (or rather a string conversion of the previous `b.Bytes()`)?

internal/cmd/controller/target/target_test.go
```diff
@@ -314,6 +329,70 @@ func TestProcessTemplateValues(t *testing.T) {
 		t.Fatal("join func was not right")
 	}
 
+	templatedValuesData, err := processTemplateValuesData(bundle.Helm.TemplateValues, values)
+	if err != nil {
+		t.Fatalf("error during label processing %v", err)
```
weyfonk, line 334: :raised_eyebrow: Not sure why that error message is phrased this way (same for `processTemplateValues`)
```suggestion
		t.Fatalf("error during templated values processing %v", err)
```

weyfonk (commented):
I like the idea, but I wonder if we _really_ need both _values_ and `templateValues`; perhaps we could template each member of the existing `value` field before serialising it to YAML, WDYT?

Danil-Grigorev:
> I like the idea, but I wonder if we _really_ need both _values_ and `templateValues`; perhaps we could template each member of the existing `value` field before serialising it to YAML, WDYT?

I also thought about it, but that has the same problem as previous solution. As the value in the map is interface{}, what type it actually contains is unknown. Template processing requires string data, which is only achievable after serialization - so it has to go through yaml decoder. These template expressions do not render as valid yaml, so we can't do anything.

Even if the interface is initially checked to be string, and we can skip yaml pre-processing, it is unclear if the templated result needs to stay a sting, or it has to be a map/list. `templatedValues` makes an assumption that the string inside will contain raw data (map/list). So I made a field addition to make it distinct in order to avoid breaking changes in API.

internal/cmd/controller/target/builder.go
Danil-Grigorev: I updated the doc message on the field. Not sure we need to test this, since the method is a part of standard lib, but added a scenario

manno:
> ```yaml
> installation: |-
> ```

I think this is supposed to be `installation:`, like in the issue. Otherwise, within a string value, the old approach should work.

Danil-Grigorev:
Having the value specified as
```yaml
values:
  installation: |-
```
Results in error on HelmApp, which is only due to invalid type passed into `installation`, a string:
```yaml
status:
  conditions:
  - lastUpdateTime: "2025-01-31T15:55:58Z"
    message: 'ErrApplied(1) [Bundle sample1: template: tigera-operator/templates/tigera-operator/02-tigera-operator.yaml:81:17:
      executing "tigera-operator/templates/tigera-operator/02-tigera-operator.yaml"
      at <.Values.installation.kubernetesProvider>: can''t evaluate field kubernetesProvider
      in type interface {}]'
    status: "False"
    type: Ready
```

If I instead switch to just `installation:` in the template and apply again, it is invalid YAML:
```bash
dgrigorev@localhost:~/Documents/work/cluster-api-addon-provider-fleet> kubectl apply -f helm.yaml 
error: error parsing helm.yaml: error converting YAML to JSON: yaml: line 21: mapping values are not allowed in this context
```

Then if I go to generated `BundleDeployment` and edit it so it contains no `|-` in it, the rollout succeeds. It also succeeds with correct configuration (note the `cidr` field):
```yaml
kind: Installation
metadata:
  annotations:
    meta.helm.sh/release-name: projectcalico
    meta.helm.sh/release-namespace: default
    objectset.rio.cattle.io/id: default-sample1-fleet-addon-agent
  creationTimestamp: "2025-01-31T16:04:01Z"
  finalizers:
  - operator.tigera.io/installation-controller
  - tigera.io/operator-cleanup
  - operator.tigera.io/apiserver-controller
  generation: 4
  labels:
    app.kubernetes.io/managed-by: Helm
    objectset.rio.cattle.io/hash: cf1aceee4a248c0997592ec3509a41cb53a2fece
  name: default
  resourceVersion: "2589"
  uid: 9022fd0d-ea88-4d04-ac1a-ebc322adfc98
spec:
  calicoNetwork:
    bgp: Disabled
    hostPorts: Enabled
    ipPools:
    - allowedUses:
      - Workload
      - Tunnel
      blockSize: 26
      cidr: 10.1.0.0/16
```

None of this is needed in updated `templateValues` implementation.

manno (approved):
Merging this early for an alpha release

Issue #2898, opened 2024-09-25
OCI Experimental Storage: job service account missing RBAC?


puffitos:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

We’ve been trying to use the OCI experimental feature to allow bigger bundles containing multiple versions of the same helm chart (to enable canary deployments via modifying the fleet.yaml). This seems to address https://github.com/rancher/fleet/issues/2442, which is absolutely great. 

After going through some authentication problems with our quay registry, we’ve reached a point, where this _seems_ to be a fleet problem.

As the logs of the job created by the `GitRepo` with the experimental OCI feature show below, some secret needs to be created first by the `serviceaccount` of the pod. Unfortunately, the default role created by fleet for those `serviceaccounts` doesn’t provide such RBAC (see log section for the `Role` bound to the `serviceaccount` in question).

```shell
time="2024-09-25T08:41:28Z" level=fatal msg="secrets is forbidden: User \"system:serviceaccount:fleet-default:git-caas-cluster-monitoring\" cannot create resource \"secrets\" in API group \"\" in the namespace \"fleet-default\""
Stream closed EOF for fleet-default/caas-cluster-monitoring-46bd3-ftfx7 (fleet)
```

After changing the ability to create secrets, I was briefly disappointed to see the old error message:

```shell
_=/usr/bin/env
time="2024-09-25T08:54:45Z" level=fatal msg="rpc error: code = ResourceExhausted desc = trying to send message larger than max (2489846 vs. 2097152)"
Stream closed EOF for fleet-default/caas-cluster-monitoring-46bd3-6tppb (fleet)
```

After that, I just deleted the created Bundle and tried again, and it seems to have worked. I’m baffled as why the error would appear at first but not the second time, but that's a different issue / fleet tried to update the bundle and it already contained too much info?

### Expected Behavior

The feature works out of the box without having to manually change the Role.

### Steps To Reproduce

1. (optional) Deploy a GitRepo that deploys a big chart in multiple clusters (like kube-prometheus-stack)
2. Use the OCI experimental feature for this gitrepo and set up a repository and credentials for the OCI backend
3. When the pod runs to  create the bundle, the error message should appear, because the serviceaccount has no RBAC to create secrets

### Environment

```markdown
- Architecture: arm64
- Fleet Version: 0.10.1
- Cluster:
  - Provider: rke
  - Options: - 
  - Kubernetes Version: 1.29
```

### Logs

```markdown
<details>
[...]

PR #3305, opened 2025-02-06
Migrate GitJob RBAC rules

p-se:
Part of #2898

### DIFF
--- a/charts/fleet/templates/rbac_gitjob.yaml
+++ b/charts/fleet/templates/rbac_gitjob.yaml
@@ -85,12 +85,20 @@ rules:
       - escalate
       - create
       - bind
+      - get
+      - list
+      - watch
+      - update
   - apiGroups:
       - rbac.authorization.k8s.io
     resources:
       - rolebindings
     verbs:
       - create
+      - get
+      - list
+      - watch
+      - update
 
 ---
 apiVersion: rbac.authorization.k8s.io/v1
--- a/integrationtests/gitjob/controller/controller_test.go
+++ b/integrationtests/gitjob/controller/controller_test.go
@@ -8,6 +8,7 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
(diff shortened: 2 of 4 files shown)


manno (commented):
I think this needs more permissions, because `Get` is cached. Maybe using `client.Reader` can be used?

manno:
> I think this needs more permissions, because `Get` is cached. Maybe using `client.Reader` can be used?

That service account already has lots of permissions, so no need to optimize these.

manno (approved):

p-se:
## Additional QA

### Problem

Fleet creates a role for each GitRepo which is named after the GitRepo but prefixed with `git-` and is created in the same namespace as the GitRepo resource.

Before Fleet supported OCI storage, this role did not have the permission to create secrets. Since Fleet supports OCI storage, the role needs to have the permission to create secrets.

Using a Fleet version that supports OCI storage, the role is create with the permission to create secrets, however, if a user updates from an older Fleet version (without OCI storage support) to a newer Fleet version (with OCI storage support), the permission of the role is not automatically updated. The user would need to remove and recreate the GitRepo for the role to be created with the permission to create secrets.

Since this is undesired behavior, Fleet is supposed to update the permission if updated to a version with OCI storage, so that the GitRepo does not need to be recreated.

### Solution

Fleet updates the permissions of the role when the GitRepo is updated or a force resync is triggered.

### Testing

### Engineering Testing

#### Manual Testing

- Use a Fleet version prior to the one that supports OCI storage
- Create a GitRepo resource
- Check that the role created for the GitRepo does not have the permission to create secrets
- Update Fleet to a version that supports OCI storage
- Update the GitRepo resource, either
  - force a resync, or
  - set OCI storage credentials, or
  - push a commit to the git repository monitored by the GitRepo.

  It is not necessary to use OCI storage, the permission (rule) is supposed to be added to the role independently of the used storage type.
- Check that the role created for the GitRepo now has the permission to create secrets, either by
  - checking the role, or
  - by checking that the Bundle has been created, in case an OCI registry has been configured as storage.

#### Automated Testing

Integration tests have been added. They first remove the permission (rule) to create secrets from the role, then trigger a forced resync and finally it checks that Fleet (the GitJob controller) has updated the role so that it contains the permission (rule) to create secrets.

### QA Testing Considerations

Suggestion: Follow the steps in Manual Testing.

N/A

kkaempf:
Won't backport to 2.10.x since a workaround exists (and will be documented)

Issue #3313, opened 2025-02-07
[SURE-8883] Add the options for the user to add custom or automatic tolerations to tolerations for Helm operation pods


kkaempf:
### Motivation
Whenever a user does any operation on a helm chart, a helm-operation-xxx pod is created to execute the operation. Customers that have all nodes tainted will have a problem with this.

Changes were made to the backend SURE-8593 so that a customer may provide tolerations to helm operation pods or ask that it automatically adds the needed tolerations to run on control plane nodes. The changes were made to the install/upgrade/uninstall charts endpoints.

See JIRA for an example of payload

### Acceptance criteria
The user can manually specify tolerations to be added to a helm operation pod when installing/upgrading/uninstalling a chart
The user can specify that the helm operation pods should add tolerations to all taints from control plane nodes.

kkaempf:
Tentatively adding to v2.11 so engineering can triage it and give a rough estimate.
The final schedule will be determined based on the estimate (and other backlog items 😉 )

manno:
For Fleet:

- [x] make sure tolerations are passed to the fleet helm chart
- [ ] make sure fleetcluster [passes tolerations to local cluster](https://github.com/rancher/rancher/blob/release/v2.9/pkg/controllers/provisioningv2/fleetcluster/fleetcluster.go#L207)
- [x] fleet chart uses tolerations for cleanup/etc. jobs 
- [ ] fleet controller uses tolerations for apply/clone job pods

gaktive:
UI received something similar and has a draft to help out, but we seem to be blocked by either this or https://github.com/rancher/rancher/issues/49101

Dashboard: https://github.com/rancher/dashboard/issues/13125
Internal reference: SURE-9823 (blocked by SURE-8883 mentioned at the top)

@kkaempf as I help out @momesgin with his ticket, does his ticket look connected?

0xavi0:
https://github.com/rancher/rancher/pull/49355 was pushed to `rancher/rancher` to add tains in control plane as tolerations to the the Fleet's local cluster.

kkaempf:
Done.

PR #3323, opened 2025-02-10
Adds tolerations to GitRepo

0xavi0:
When applying a `GitRepo` in a cluster where all the nodes are tainted the pod created to call `fleet apply` remans waiting as Pending.

This PR adds tolerations to the `GitRepo` spec. Those tolerations will be added to the job spec when running `fleet apply` so the pod is scheduled.

Refers to: https://github.com/rancher/fleet/issues/3313


0xavi0:
Closing to copy toleration values from fleet-controller

PR #3328, opened 2025-02-10
Fix integrationtest for drift correction

manno:
After applying ginkgolinter fixes, several tests failed. Some tests where not effective, as they were missing `Should(Succeed())`.

This fixes test expectations and relaxes some tests.

It seems `UpdateStatus` sometimes overwrites the Ready error message or resets the conditions. The  error appears in either Ready or Deployed as "modified" or "cannot patch".


weyfonk (approved):
LGTM, thanks for looking into this :bow:

PR #3329, opened 2025-02-11
Fix default state for GitRepo resources

aruiz14:
Follow up to #3209
Found while testing https://github.com/rancher/dashboard/pull/13244

#3209 changed the default state for resources from the Bundle summary state to use the BundleDeployment, but I didn't take into account that this is sometimes calculated from the resource states (when it's not `WaitApplied` or `ErrApplied`).
This change prevents that, and provides resources with correct default state assumptions.

### DIFF
--- a/internal/resourcestatus/resourcekey.go
+++ b/internal/resourcestatus/resourcekey.go
@@ -81,9 +81,11 @@ func toType(apiVersion, kind string) string {
 	return group + strings.ToLower(kind)
 }
 
+// resourceDefaultState calculates the state for items in the status.Resources list.
+// This default state may be replaced individually for each resource with the information from NonReadyStatus and ModifiedStatus fields.
 func resourcesDefaultState(bd *fleet.BundleDeployment) string {
 	switch bdState := summary.GetDeploymentState(bd); bdState {
-	// NotReady and Modified BD states are inferred from resource statuses, so it's incorrect to used that to calculate resource states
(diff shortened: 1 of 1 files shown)


internal/resourcestatus/resourcekey.go
```diff
@@ -81,10 +81,24 @@ func toType(apiVersion, kind string) string {
 	return group + strings.ToLower(kind)
 }
 
+func resourcesDefaultState(bd *fleet.BundleDeployment) string {
+	switch bdState := summary.GetDeploymentState(bd); bdState {
+	// NotReady and Modified BD states are inferred from resource statuses, so it's incorrect to used that to calculate resource states
```
weyfonk, line 86: nit:
```suggestion
	// NotReady and Modified BD states are inferred from resource statuses, so it's incorrect to use that to calculate resource states
```

internal/resourcestatus/resourcekey.go
```diff
@@ -81,10 +81,24 @@ func toType(apiVersion, kind string) string {
...
+	case fleet.NotReady, fleet.Modified:
+		if bd.Status.IncompleteState {
+			return "Unknown"
+		} else {
+			return string(fleet.Ready)
```
weyfonk, line 91: A comment here could make sense, explaining that this does not mean that the bundle deployment itself is considered ready, but that this is a default state which may well be overwritten by individual resources' `Modified` or `NonReady` statuses. This could be function-level documentation too.

weyfonk (approved):

PR #3344, opened 2025-02-13
Add charts branch creation to release action

thardeck:
to have a repo with a versioned chart which can be used for automation.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -141,27 +141,26 @@ jobs:
       - name: Create Docker manifest for Prime and sign it
         shell: bash
         run: |
-            for IMAGE in fleet fleet-agent; do
-              URL="${{ env.PRIME_REGISTRY }}/rancher/${IMAGE}:${{ github.ref_name }}"
-              docker buildx imagetools create -t "${URL}" \
-                "${URL}-linux-amd64" \
-                "${URL}-linux-arm64"
+          for IMAGE in fleet fleet-agent; do
+            URL="${{ env.PRIME_REGISTRY }}/rancher/${IMAGE}:${{ github.ref_name }}"
+            docker buildx imagetools create -t "${URL}" \
+              "${URL}-linux-amd64" \
+              "${URL}-linux-arm64"
 
-              cosign sign --oidc-provider=github-actions --yes "${URL}"
-            done
+            cosign sign --oidc-provider=github-actions --yes "${URL}"
+          done
 
       - name: Attest provenance
         shell: bash
         run: |
           for IMG_NAME in $(yq e '.dockers[].image_templates[0]' .goreleaser.yaml | grep PRIME_REGISTRY | sed "s/{{ .Env.PRIME_REGISTRY }}/${{ env.PRIME_REGISTRY }}/g" | sed "s/{{ .Tag }}/${{ github.ref_name }}/g"); do
             # Extract Docker image reference plus digest from local image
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -203,3 +203,41 @@ jobs:
           for f in $(find dist/ -name '*.tgz'); do
             gh release upload $tag $f
           done
+
+      - name: Add charts to branch
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          version=$(jq -r '.version' <<< "${{ steps.goreleaser.outputs.metadata }}")
+          branch_version=$(cut -d'.' -f1,2 <<< "$version")
```
weyfonk, line 212: Why not use the full version number? On a given minor charts release branch, will charts for a later patch release overwrite existing ones?

.github/workflows/release.yml
```diff
@@ -203,3 +203,41 @@ jobs:
...
+            git checkout --orphan "$charts_branch"
+            git rm -rf .
+
+            echo "# Fleet Helm Charts for $branch_version versions" > README.md
+            echo "The documentation is centralized in a unique place, checkout https://fleet.rancher.io/." >> README.md
+
+            git checkout main -- .gitignore
+
+            git add README.md .gitignore
+            git commit -m "Initial commit for $charts_branch"
```
weyfonk, line 231: This initial commit will remove all contents from the index.
Without being knowledgeable enough in git internals, I wonder how repeating this process for multiple release branches, in addition to adding charts, might increase this repository's `.git` directory :thinking:

.github/workflows/release.yml
```diff
@@ -203,3 +203,41 @@ jobs:
...
+          else
+            git checkout "$charts_branch"
+          fi
+
+          mkdir charts
+          find dist/ -name '*.tgz' -exec tar -xf {} -C charts/ \;
```
weyfonk, line 237: Why do we need to do this, since we've edited values files in-place during step ` Upload charts to release`?
Couldn't we simply reuse the `charts` dir in its current state? Or what am I missing?

weyfonk (commented):
Still a bit confused about this approach, and interested to better understand its rationale.

.github/workflows/release.yml
weyfonk: Oh, I see, thanks for clarifying :+1:

.github/workflows/release.yml
weyfonk: Yes, what I meant to say was not so much that we'd need to keep all those unneeded files around, but rather that using a different repository for this could make sense, similarly to what we have with [`rancher/fleet-helm-charts`](https://github.com/rancher/fleet-helm-charts/).

.github/workflows/release.yml
weyfonk: Makes sense, thanks.

.github/workflows/release.yml
```diff
@@ -203,3 +203,41 @@ jobs:
           for f in $(find dist/ -name '*.tgz'); do
             gh release upload $tag $f
           done
+
+      - name: Add charts to branch
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          version=$(jq -r '.version' <<< "${{ steps.goreleaser.outputs.metadata }}")
```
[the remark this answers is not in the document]
weyfonk: I don't think that adapting L184 to this style is critical, but that's up to you ;)
Thanks for the explanation!

weyfonk (approved):

Issue #3360, opened 2025-02-15
Reconcile Failed Update to Git Repo Status


freexploit:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

The reconcile process for the `GitRepo` status failed with an error. The error message shows that `status.lastPollingTriggered` and `status.lastSyncedImageScanTime` are invalid, as they should be of type string but are currently null.

### Expected Behavior

The reconcile process should successfully update the GitRepo status without errors.

### Steps To Reproduce

1. Trigger a reconcile process for the `random-repo` GitRepo status.
2. Observe that the reconcile process fails with an error.

### Environment

```markdown
- Architecture:  x86_64
- Fleet Version: 0.11.3
- Cluster:
  - Provider: k3s, k3d
  - Options: k3s on cloud, with 5 nodes, k3d local development.
  - Kubernetes Version: 1.30
```

### Logs

```markdown
{
  "level": "error",
  "ts": "2025-02-10T19:33:11Z",
  "logger": "gitops-status",
  "msg": "Reconcile failed update to git repo status",
  "controller": "GitRepoStatus",
  "controllerGroup": "fleet.cattle.io",
  "controllerKind": "GitRepo",
  "GitRepo": {
    "name": "random-repo",
    "namespace": "clusters"
  },
  "namespace": "clusters",
  "name": "webhook-citadel",
  "reconcileID": "02417797-4e45-4cde-93a2-6828ee9cac7b",
  "generation": 1,
  "commit": "63a7d81282c57e31341046acd1054988a4fa6a23",
  "conditions": [
    {
      "type": "Ready",
      "status": "True",
      "lastUpdateTime": "2025-02-10T19:01:31Z"
    },
    {
      "type": "Accepted",
      "status": "True",
      "lastUpdateTime": "2025-02-10T19:12:41Z"
    },
    {
      "type": "ImageSynced",
      "status": "True",
      "lastUpdateTime": "2025-02-06T13:01:38Z"
    },
    {
      "type": "Reconciling",
      "status": "False",
      "lastUpdateTime": "2025-02-10T19:01:25Z"
    },
    {
      "type": "Stalled",
      "status": "False",
      "lastUpdateTime": "2025-02-10T19:01:25Z"
    },
    {
      "type": "Synced",
      "status": "True",
      "lastUpdateTime": "2025-02-10T19:12:41Z"
    }
  ],
  "status": {
    "observedGeneration": 1,
[...]

PR #3361, opened 2025-02-15
Nullable gitrepo status

cvalerio-va:
Refers to #3360

I added annotations to `lastPollingTriggered` and `lastSyncedImageScanTime` to be nullable and stop the error.


cvalerio-va:
Oh, sorry I will fix this. Thanks!!!!

0xavi0:
Hey @cvalerio-va
Could you please elaborate a bit when are you observing that error?
Also a test `GitRepo` so we're able to recreate would be really appreciated. 

I've tried to recreate but I didn't see any error message related to those 2 fields having an empty value

PR #3362, opened 2025-02-17
Adds tolerations from fleet-controller Deployment

0xavi0:
Adding the toleration needed to the helm chart in Fleet is not enough when running the agent and the fleet apply job.

This PR adds the tolerations found in the `fleet-controller` deployment to the agent and to the fleet apply job.

Refers to: https://github.com/rancher/fleet/issues/3313

### DIFF
--- a/internal/cmd/controller/gitops/reconciler/gitjob_test.go
+++ b/internal/cmd/controller/gitops/reconciler/gitjob_test.go
@@ -14,12 +14,14 @@ import (
 	"go.uber.org/mock/gomock"
 
 	"github.com/rancher/fleet/internal/cmd/controller/finalize"
+	"github.com/rancher/fleet/internal/config"
 	"github.com/rancher/fleet/internal/mocks"
 	fleetv1 "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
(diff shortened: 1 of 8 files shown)


internal/cmd/controller/gitops/reconciler/gitjob_test.go
```diff
@@ -1169,9 +1259,10 @@ func TestGenerateJob_EnvVars(t *testing.T) {
 	for name, test := range tests {
 		t.Run(name, func(t *testing.T) {
 			r := GitJobReconciler{
-				Client: fake.NewFakeClient(),
-				Image:  "test",
-				Clock:  RealClock{},
+				Client:          getTestClient(testClientSimple, []corev1.Toleration{}),
```
weyfonk, line 1262: nit: Is updating this line really necessary, if no checks are run on tolerations?

weyfonk (approved):

internal/cmd/controller/gitops/reconciler/gitjob_test.go
0xavi0: Yes, it is necessary, because the test client now adds the Deployment to the schema. 
That is needed when creating the gitjob Spec

Issue #3160, opened 2024-12-19
[SURE-9542] fleet apply logs from fleet job Pod do not indicate source of "no chart name found" errors in GitRepo containing multiple bundles


kkaempf:
### Issue description:
fleet apply logs from fleet job Pod do not indicate source of "no chart name found" errors in GitRepo containing multiple bundles. If one (or more) charts defined by a fleet.yaml bundle in the GitRepo do no exist in the index.yaml for their configured repository, only a generic `level=fatal msg="no chart name found"` is logged in the Job Pod, with no indication of which chart/bundle the error originates from.

### Business impact:
Makes troubleshooting of the issue a heavy lift manual task

### Repro steps:

Provision a Rancher v2.9.4 instance with a single all-role node custom RKE cluster (I used github.com/axeal/tf-do-rancher2)
In fleet-default workspace of Fleet, add a GitRepo with the reposistory https://github.com/axeal/fleet-test.git and branch 01563984
Observe successful deployment of the two charts/bundles
Update the GitRepo branch to 01563984-invalid, in which the rancher-monitoring-crd chart name is updated to the invalid rancher-monitoring-crds (use of the the rancher-logging-crd and rancher-monitoring-crd charts was arbitrary for the purpose of reproduction, in customer environment they are deploying their internal applications).
Observe after a short time that the GitRepo goes into an error state with `Job Failed. failed: 3/1time="2024-12-18T13:24:13Z" level=fatal msg="no chart name found"`

Observe the fleet Job pods for the GitRepo contain only the following logs:
~~~
HOSTNAME=test-ec8c6-wmczp
KUBERNETES_PORT_443_TCP_PROTO=tcp
COMMIT=8d6c3473d9d1b4afe6b1d479466226c247c6783b
KUBERNETES_PORT_443_TCP_ADDR=10.43.0.1
KUBERNETES_PORT=tcp://10.43.0.1:443
PWD=/workspace/source
HOME=/fleet-home
KUBERNETES_SERVICE_PORT_HTTPS=443
KUBERNETES_PORT_443_TCP_PORT=443
KUBERNETES_PORT_443_TCP=tcp://10.43.0.1:443
SHLVL=1
KUBERNETES_SERVICE_PORT=443
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
KUBERNETES_SERVICE_HOST=10.43.0.1
_=/usr/bin/env
[...]

Issue #3234, opened 2025-01-22
[SURE-9488]  Fleet UI errors will simplify troubleshooting and reduce escalations


kkaempf:
### Request Description: Descriptive Error Messages in the Fleet UI

Expected behaviour: UI should provide more descriptive error messages where you can pinpoint where the issue is since Hosted Rancher logs are not accessible. Or allow viewing the Rancher logs in the UI

*Actual behaviour:* The message show like 'failed: 3/1time="2024-11-28T09:04:55Z" level=fatal msg="context canceled"' where it doesn't provide useful context where to check the issue.

*Actual issue:* fleet.yaml contains duplicate key entry and validation fails or sometimes the helm chart doesn't exist.

## TODO

Let's check all the error conditions and add a "context", like "error in gitjob: ".
- [ ] Maybe strip or format the timestamp, too so they look better.
- [ ] Replace all "context canceled" error messages, because users don't understand "context canceled". This probably translates into "timeout waiting for gitjob to complete"
- [ ] Revisit `Failure` and `Readiness` Conditions (see SURE-9488)

manno:
Let's check all the error conditions and add a "context", like "error in gitjob: ". 
Maybe strip or format the timestamp, too.
Replace "context canceled with timeout?".

This probably translates into "timeout waiting for gitjob to complete"

kkaempf:
I closed JIRA today since we've made good progress in `0.12.0`

sbulage:
## System Information

### Before Upgrade:

| Rancher Version | Fleet Version |
|--------|--------|
| v2.11.0 | 106.0.0+up0.12.0 | 

### Steps followed
- Created a GitRepo using this [fleet.yaml](https://github.com/Martin-Weiss/rancher-fleet/blob/main/rke-monitoring/app/fleet.yaml)
- Wait for the Job to be created.
- GitRepo is in Error state (see below screenshot for more details)
  ```bash
 	Job Failed. failed: 1/1time="2025-05-19T18:27:31Z" level=fatal msg="failed to process bundle: context canceled"
  ```

<details><summary>Screenshot showing Error message.</summary>

![Image](https://github.com/user-attachments/assets/dc1fc5ba-534b-4c46-9570-a8d975e55d9e)

</details> 

---

### After Upgrade

| Rancher Version | Fleet Version |
|--------|--------|
| v2.12-49289cc9c6590b361d64950977dd20b1214908d7-head | 107.0.0+up0.13.0-alpha.3 |

[...]

PR #3368, opened 2025-02-18
Add extra error info when downloading remote charts

0xavi0:
Error example (using the provided git repository)
```bash
time="2025-02-18T10:37:59+01:00" level=fatal msg="repo=https://charts.rancher.io chart=rancher-monitoring-crds version=104.1.2+up57.0.3: no chart name found"
```

Refers to: https://github.com/rancher/fleet/issues/3160

### DIFF
--- a/integrationtests/cli/apply/helm_test.go
+++ b/integrationtests/cli/apply/helm_test.go
@@ -1,7 +1,9 @@
 package apply
 
 import (
+	"fmt"
 	"io/fs"
+	"os"
 	"path/filepath"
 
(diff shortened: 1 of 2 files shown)


weyfonk (approved):
LGTM; I wonder if Fleet produces other related errors, e.g. when downloading the chart itself, which could be clarified or better exposed as well.

0xavi0:
Yeah, we have also https://github.com/rancher/fleet/issues/3234 in which I hope to add context to more errors.
I kept this one simple so we can merge is fast and will go more in deep in the other one ;)

Issue #3359, opened 2025-02-15
OCIStorage: bundles still created as k8s-resources ?


strowi:
Hi,

i started testing the OCIStorage experimental feature. I added this to the gitRepo:

```yaml
  ociRegistry:
    authSecretName: oci-secret
    basicHTTP: true
    reference: core.harbor.x.y/fleet-oci-storage
```

And on deleting all bundles in the gitRepo i can see blobs popping up in the registry.  So far so good, but parallel i still see the bundles being generated inside the `fleet-default`-namespace (using rancher-2.10.2 with fleet-0.11.3 ). Eg the rancher-monitoring bundle still takes ~1,2MB in etcd and the resources are listed:

```yaml
  resources:
  - content: bas64content
    encoding: base64+gz
    name: chart/CHANGELOG.md
  - content: bas64content
    encoding: base64+gz
    name: chart/CONTRIBUTING.md
```

It was my understanding that this is either OCIStorage OR k8s-resource? Am i wrong or is this just not implemented yet?

0xavi0:
Hi, thanks for reporting this.

Resources are still listed as a reference for the UI, but should no include the real content. 
Fleet creates a `Content` resource (not namespaced) with the real contents of the Bundle. 
The name of the resource is the same name you see in the OCI storage blobs, (it's basically a hash with a `s-` prefix). 

That `Content` resource  should **not** be created when using OCI storage.

0xavi0:
If it's in the spec then sounds like a bug :) 

We will triage and schedule for being fixed.

Thanks!!

manno:
Should be fixed in 0.12

PR #3389, opened 2025-02-24
Bundlereader needs to update OCI storage bundles, too

manno:
Refers to https://github.com/rancher/fleet/issues/3359

### DIFF
--- a/internal/cmd/cli/apply/apply.go
+++ b/internal/cmd/cli/apply/apply.go
@@ -343,7 +343,7 @@ func saveOCIBundle(ctx context.Context, c *client.Client, bundle *fleet.Bundle, 
 	if err != nil {
 		return err
 	}
-	logrus.Infof("OCI artifact stored successful: %s %s", bundle.Name, manifestID)
+	logrus.Infof("OCI artifact stored successfully: %s %s", bundle.Name, manifestID)
 
 	obj, err := c.Fleet.Bundle().Get(bundle.Namespace, bundle.Name, metav1.GetOptions{})
 	if err != nil && !apierrors.IsNotFound(err) {
--- a/internal/cmd/cli/apply/apply.go
+++ b/internal/cmd/cli/apply/apply.go
@@ -223,7 +223,6 @@ func Dir(ctx context.Context, client Getter, name, baseDir string, opts *Options
 		return err
 	}
 
-	bundle = bundle.DeepCopy()
 	bundle.Namespace = client.GetNamespace()
 
 	if len(bundle.Spec.Resources) == 0 {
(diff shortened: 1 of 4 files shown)


0xavi0 (approved):

internal/cmd/cli/apply/apply.go
```diff
@@ -218,20 +218,20 @@ func Dir(ctx context.Context, client Getter, name, baseDir string, opts *Options
...
 
-	bundle, scans, err := readBundle(ctx, bundleID, baseDir, opts)
+	bundle, scans, err := newBundle(ctx, bundleID, baseDir, opts)
 	if err != nil {
 		return err
 	}
 
-	def := bundle.DeepCopy()
-	def.Namespace = client.GetNamespace()
+	bundle = bundle.DeepCopy()
```
weyfonk, line 226: Why do a deep copy if we're not keeping the original?

internal/cmd/cli/apply/apply.go
```diff
@@ -331,30 +344,62 @@ func saveOCIBundle(ctx context.Context, c *client.Client, bundle *fleet.Bundle,
 	if err != nil {
 		return err
 	}
-	// OCI artifact stored successfully
-	// We don't store the resources in the bundle. Just keep the manifestID for
-	// being able to access the bundle's contents later.
-	bundle.Spec.Resources = nil
-	bundle.Spec.ContentsID = manifestID
+	logrus.Infof("OCI artifact stored successful: %s %s", bundle.Name, manifestID)
```
weyfonk, line 347:
```suggestion
	logrus.Infof("OCI artifact stored successfully: %s %s", bundle.Name, manifestID)
```

internal/cmd/cli/apply/apply.go
```diff
@@ -331,30 +344,62 @@ func saveOCIBundle(ctx context.Context, c *client.Client, bundle *fleet.Bundle,
...
+		// We don't store the resources in the bundle. Just keep the manifestID for
+		// being able to access the bundle's contents later.
+		bundle.Spec.Resources = nil
+		bundle.Spec.ContentsID = manifestID
+		obj, err = c.Fleet.Bundle().Create(bundle)
+		if err != nil {
+			return err
+		}
+
+		if err := createOrUpdate(c, newOCISecret(manifestID, obj, opts)); err != nil {
```
weyfonk, line 364: I wonder if at some point we might want to make this transactional: if creating the secret fails, the bundle will still have been created, but would be useless. We could (eventually) try to roll the process back by deleting the bundle before returning an error.

internal/cmd/cli/apply/apply.go
manno: Right, I wanted to test if it can be removed. Since this is wrangler's k8s client, I think the returned resources are pointers to the cache. Modifying their content might modify the cached content,  too.

internal/cmd/cli/apply/apply.go
manno: Not sure, since we do not have locking. If an error occurs the reconcile will be retried and should pick up where the previous one failed. Ideally the reconciles are idempotent.

PR #3407, opened 2025-02-28
Retries in fleet apply and single patch in webhook

0xavi0:
Adds a retry mechanism to `fleet apply` after observing several errors due to conflicts under high loads. As the job created is not retrying by itself the `GitRepo` gets hung.

Also changes from 1 Status Update and Patch to one single patch in the webhook handling to mitigate conflicts and possible race conditions.

Refers to: https://github.com/rancher/fleet/issues/3067

### DIFF
--- a/internal/cmd/cli/apply.go
+++ b/internal/cmd/cli/apply.go
@@ -8,6 +8,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
 	"github.com/rancher/fleet/internal/bundlereader"
@@ -78,7 +79,11 @@ func (a *Apply) Run(cmd *cobra.Command, args []string) error {
 	// Apply retries on conflict errors.
 	// We could have race conditions updating the Bundle in high load situations
 	var err error
-	for range GetOnConflictRetries() {
+	retries, err := GetOnConflictRetries()
+	if err != nil {
+		logrus.Errorf("failed parsing env variable %s, using defaults, err: %v", FleetApplyConflictRetriesEnv, err)
+	}
+	for range retries {
 		err = a.run(cmd, args)
 		if !errors.IsConflict(err) {
 			break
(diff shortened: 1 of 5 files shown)


internal/cmd/cli/apply.go
```diff
@@ -210,3 +231,15 @@ func currentCommit() string {
 	}
 	return ""
 }
+
+func GetOnConflictRetries() int {
+	s := os.Getenv(FleetApplyConflictRetriesEnv)
+
+	retries := defaultApplyConflictRetries
```
manno, line 238: The default will only be returned when Atoi fails? 
I would prefer something like `if v, ok := os.LookupEnv("FLEET_VERSION"); ok {`, instead of relying on the Atoi` failure for empty strings. If `Atoi` fails the user likely passed a string and we could return an error to let them know.

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -881,6 +882,10 @@ func argsAndEnvs(gitrepo *v1alpha1.GitRepo, debug bool, CACertsPathOverride stri
 			Name:  "HOME",
 			Value: fleetHomeDir,
 		},
+		{
+			Name:  fleetcli.FleetApplyConflictRetriesEnv,
+			Value: strconv.Itoa(fleetcli.GetOnConflictRetries()),
```
manno, line 887: Now we are converting it back?

internal/cmd/cli/apply.go
```diff
@@ -210,3 +231,15 @@ func currentCommit() string {
 	}
 	return ""
 }
+
+func GetOnConflictRetries() int {
+	s := os.Getenv(FleetApplyConflictRetriesEnv)
```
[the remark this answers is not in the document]
0xavi0: I added that extra function because we need to get the env variable in 2 different places and we also need to ensure the value is a valid integer.

internal/cmd/cli/apply.go
0xavi0: [os.LookupEnv](https://cs.opensource.google/go/go/+/go1.24.0:src/os/env.go;l=112) may also return "" which is a valid value for an environment variable that is set. 

We still need a valid integer, that's why it calls `Atoi`, to parse it and return an error if the value is either empty or not an `int`.

I changed the function to return the error, so we can also log it and make the user aware, and not use the default value silently.

I added that extra env variable to avoid hardcoding the number of retries and make it a bit more flexible, but I guess that it should not be a common user variable to play with.

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
0xavi0: corev1.EnvVar expects a `string` and logically the number of retries is an `int` so we need to convert from the env value (`string`) to `int` when using the value in the retries loop or here.

manno (approved):

PR #3408, opened 2025-03-03
Fix BundleDeployment resource counts when a resource is missing

aruiz14:
Refers to #3409

Fixes a regression introduced by #3203, where I introduced a new `resourceCounts` field to `BundleDeployment` statuses. I misinterpreted the result of `plan.Objects`, ignoring that it only includes information of live objects in the cluster:
```
type Plan struct{
    ...
	// Objects contains objects, already existing in the cluster, that have
	// valid metadata
	Objects []runtime.Object
```
Hence the `DesiredReady` field does not take into account missing objects, as well as many test expectations are incorrect and have been fixed too.


internal/cmd/agent/deployer/monitor/updatestatus_test.go
```diff
@@ -104,7 +97,7 @@ func Test_updateFromResources(t *testing.T) {
 				assert.Equal(t, status.ResourceCounts, fleet.ResourceCounts{DesiredReady: 1, Missing: 1})
 				assert.Truef(t, status.Ready, "unexpected ready status")
 				assert.Falsef(t, status.NonModified, "unexpected non-modified status")
-				assert.Lenf(t, status.Resources, 1, "unexpected resources length")
+				assert.Emptyf(t, status.Resources, "unexpected resources to be empty")
```
weyfonk, line 100:
```suggestion
				assert.Emptyf(t, status.Resources, "expected resources to be empty")
```

internal/cmd/agent/deployer/monitor/updatestatus.go
```diff
@@ -258,6 +258,7 @@ func calculateResourceCounts(all []fleet.BundleDeploymentResource, nonReady []fl
 	for _, r := range modified {
 		if r.Create {
 			counts.Missing++
+			counts.DesiredReady++ // plan.Objects do not include missing resources
```
weyfonk, line 261: I'd be tempted to expand on this comment a bit, to make it easier to understand for future readers:
```suggestion
                         // counts are determined from `all`, itself computed from `plan.Objects` (see `toBundleDeploymentResources` which do not include missing resources
			counts.DesiredReady++
```

... which makes me wonder: would it not make more sense for this correction to happen in `toBundleDeploymentResources`, which could take `modifiedResources` as an additional argument to add resources to be created to the total?

aruiz14: I've refactored it to a separate function, hopefully making it clearer

weyfonk (approved):

PR #3410, opened 2025-03-03
Skip chart release for hotfixes

aruiz14:
Forward port of #3405 and #3406

It also simplifies the hotfix creation process: since now the CI can push directly to the prime registry, we can remove the staging repo handling/duality and skip those steps instead.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -88,7 +88,6 @@ dockers:
   -
     use: buildx
     id: fleet-amd64-public
-    skip_build: "{{ .Env.IS_HOTFIX }}"
     skip_push: "{{ .Env.IS_HOTFIX }}"
 
     # GOOS of the built binaries/packages that should be used.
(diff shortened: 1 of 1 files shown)


.goreleaser.yaml
```diff
@@ -298,17 +305,19 @@ dockers:
 
 docker_manifests:
   # https://goreleaser.com/customization/docker_manifest/
-  - name_template: "{{ .Env.REGISTRY }}/rancher/fleet:{{ .Tag }}"
+  - name_template: "docker.io/rancher/fleet:{{ .Tag }}"
     id: fleet-manifest-public
+    skip_push: "{{ .Env.IS_HOTFIX }}"
```
aruiz14, line 310: There is no [`skip_build` support for this type](https://goreleaser.com/customization/docker_manifest/#customization), so I don't know if it will still run with `dry-run`. If that's the case, it may fail due to the images not. being present locally? Should I remove `skip_build` for the images just in case?

Issue #3377, opened 2025-02-20
Improve Fleet Agent Failover


manno:

p-se:
Needs QA for #3463 and #3457.

p-se:
## QA Template

### Problem

Failover is supposed to be improved.

### Solution

For faster (and possibly more reliable) failover, Fleet controllers support running replicas of controllers, but these have so far not been configurable. Now they can be configured by using helm template values.

### Testing

Set different values for the desired amount of replicas for various controllers, deploy the helm chart and observe that the replica size for those controllers has been set correctly by verifying that the specified amount of pods have been deployed. You may also check the logs of those pods and verify that only one controller is actively processing while the others are waiting for the leader lease, which they will only get if the active controllers stops renewing it by, for instance, dying.

mmartin24:
Verified in `v2.12-62362dfd905a75f77865a47aff683a195cc6ac66-head` with Fleet `107.0.0+up0.13.0-alpha.3`

---

Replicas for `fleet-agent`, `helmops` as well as `gitjob` are configurable (set to 3 replicas as can be seen in screenshot:

![Image](https://github.com/user-attachments/assets/30215d9f-15e5-46f3-9b8b-27908cbd43e5)

Confirmed when deploying to 3 downstream clusters, only 1 gitjob pod gets the leader lease:

```
> k get pod -n cattle-fleet-system 
NAME                               READY   STATUS    RESTARTS   AGE
fleet-controller-95fb94b4c-tbhfb   3/3     Running   0          11m
gitjob-57f6fd9664-4xlpl            1/1     Running   0          11m
gitjob-57f6fd9664-n79ff            1/1     Running   0          11m
gitjob-57f6fd9664-rwp7m            1/1     Running   0          11m
helmops-69d9c9f697-fg5xq           1/1     Running   0          11m
helmops-69d9c9f697-jtq65           1/1     Running   0          11m
helmops-69d9c9f697-xs79l           1/1     Running   0          11m
```


```
> k logs -n cattle-fleet-system gitjob-57f6fd9664-4xlpl  | grep "lease"
I0602 11:43:56.522344       7 leaderelection.go:257] attempting to acquire leader lease cattle-fleet-system/fleet-gitops-leader-election-shard...
I0602 11:44:29.260412       7 leaderelection.go:271] successfully acquired lease cattle-fleet-system/fleet-gitops-leader-election-shard
[...]

PR #3416, opened 2025-03-04
[SURE-9419] Move agent registration into controller container

p-se:
Since we have changed the agent from using a StatefulSet to a Deployment, we also needed and have implemented leader election for every container of the agent pod. This ensures that
- only  one cluster registration is created at a time from one downstream cluster,
- one agent is running on a downstream cluster at a time and
- one container is updating the status at at time,

even if the replicas are temporarily (as in a rollout) or permanently (as in replica size) above one.

Now, if replicas are to be set to anything above one, this will currently lead to the agent of the second pod to perform the cluster registration, even though the agent cannot run after it completed the registration. The cluster registration will complete, but the controller will wait for the lease. This state potentially leads to confusion, as there are cluster registrations in the upstream cluster that have never and may never be used. Apparently this state is also considered inconsistent with the UI. Lastly, we use more leases than necessary.

While theoretically possible that the name for the lease is the same for the cluster registration container as well as for the agent controller, this may lead to extended waiting times, as the leases need to be acquired. Also, it is not guaranteed that the agent controller of the same pod will get the lease after the cluster registration container has let go of the lease , potentially leading to the same problem with unused cluster registrations (imagine replica size 3 from 1).

For those reasons the agent registration is to be moved into the controller container.

Part of #3377

### DIFF
--- a/internal/cmd/agent/root.go
+++ b/internal/cmd/agent/root.go
@@ -43,7 +43,6 @@ type AgentReconcilerWorkers struct {
 const (
 	leaseLockName              = "fleet-agent"
 	leaseLockNameClusterStatus = "fleet-agent-clusterstatus"
-	leaseLockNameRegister      = "fleet-agent-register"
 )
 
 var (
@@ -125,6 +124,18 @@ func (a *FleetAgent) Run(cmd *cobra.Command, args []string) error {
 		RenewDeadline: *leaderOpts.RenewDeadline,
 		Callbacks: leaderelection.LeaderCallbacks{
 			OnStartedLeading: func(ctx context.Context) {
+				// Handle agent registration.
+				r := &Register{
+					UpstreamOptions: UpstreamOptions{
+						Namespace: a.Namespace,
+					},
+				}
+				err := r.RegisterAgent(ctx)
+				if err != nil {
+					setupLog.Error(err, "failed to register with upstream cluster")
+					return
+				}
+
 				workersOpts := AgentReconcilerWorkers{}
 
 				if d := os.Getenv("BUNDLEDEPLOYMENT_RECONCILER_WORKERS"); d != "" {
@@ -178,7 +189,6 @@ func App() *cobra.Command {
 
 	root.AddCommand(
 		NewClusterStatus(),
-		NewRegister(),
 	)
 	return root
 }
--- a/charts/fleet-agent/templates/deployment.yaml
+++ b/charts/fleet-agent/templates/deployment.yaml
@@ -11,43 +11,6 @@ spec:
       labels:
         app: fleet-agent
     spec:
-      initContainers:
-      - env:
-        - name: NAMESPACE
-          valueFrom:
-            fieldRef:
-              fieldPath: metadata.namespace
-        {{- if $.Values.leaderElection.leaseDuration }}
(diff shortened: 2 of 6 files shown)


internal/cmd/agent/root.go
```diff
@@ -125,6 +124,15 @@ func (a *FleetAgent) Run(cmd *cobra.Command, args []string) error {
 		RenewDeadline: *leaderOpts.RenewDeadline,
 		Callbacks: leaderelection.LeaderCallbacks{
 			OnStartedLeading: func(ctx context.Context) {
+				// Handle agent registration.
+				r := &Register{}
+				r.Namespace = a.Namespace
+				err := r.RegisterAgent(ctx)
+				if err != nil {
```
weyfonk, line 131: nit: this could perhaps be simplified a bit, e.g.:
```suggestion
				r := Register{Namespace: a.Namespace}
				if err := r.RegisterAgent(ctx); err != nil {
```

internal/cmd/agent/root.go
```diff
@@ -125,6 +124,15 @@ func (a *FleetAgent) Run(cmd *cobra.Command, args []string) error {
 		RenewDeadline: *leaderOpts.RenewDeadline,
 		Callbacks: leaderelection.LeaderCallbacks{
 			OnStartedLeading: func(ctx context.Context) {
```
weyfonk, line 126: Right, so if I properly understand the new process, the (single) agent container will seek to acquire a leader election lease and, once it has it, will register the cluster _before_ proceeding to running the agent reconcilers. :+1: 

In my understanding (just checking if it is aligned with yours), the next step would involve launching a goroutine, in parallel to running the reconcilers, for reporting cluster status. Happy to discuss if I've misunderstood any of this ;)

weyfonk (approved):
Nice one, this looks cleaner to me than the previous approach.

internal/cmd/agent/root.go
p-se: AFAIK, that wouldn't work, since the `Namespace` attribute is part of the `UpstreamOptions` that is embedded into `Register`, which would rather make it look like this:

```diff
-               r := &Register{}
-               r.Namespace = a.Namespace
+               r := &Register{
+                   UpstreamOptions: UpstreamOptions{
+                       Namespace:  a.Namespace,
+                   },
+               }
```

I do not mind making it look like that. Perhaps it's clearer that way.

weyfonk: Oh I see. Nevermind then, the current version is more concise.

internal/cmd/agent/root.go
p-se: > Right, so if I properly understand the new process, the (single) agent container will seek to acquire a leader election lease and, once it has it, will register the cluster before proceeding to running the agent reconcilers. 👍

Correct.

> In my understanding (just checking if it is aligned with yours), the next step would involve launching a goroutine, in parallel to running the reconcilers, for reporting cluster status. Happy to discuss if I've misunderstood any of this ;)

As we've discussed, my initial idea was to make it an ordinary reconciler and use requeue to immitate a ticker, but now I'm going to put the functionality into a (controller-runtime) Runnable and have the manager execute it. Then I'll take care of migration the other parts from wrangler/lasso to controller-runtime.

Issue #3437, opened 2025-03-12
Editing targets for a GitRepo takes some time to delete orphaned BundleDeployments


aruiz14:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When modifying an existing `GitRepo` that is deployed to multiple target clusters, if one of the clusters no longer matches the conditions, Fleet "undeploys" those manifests (`BundleDeployments`) included in the `GitRepo` for that cluster.
However, while the `Bundle` status is updated immediately, the corresponding `BundleDeployments` are not, which indirectly causes the `GitRepo` status to contain mismatching information, since `BundleDeployments` are used to construct it.

After some time, the offending `BundleDeployments` will be garbage-collected, causing every resource to reconcile to the desired state. However, in some cases it can take up to 10 minutes, which is undesirable as it could lead to confusion (e.g. Force Updating the `GitRepo` does not have any effect on this).

### Expected Behavior

Garbage-collection of unneeded `BundleDeployments` should happen on a timely fashion.

### Steps To Reproduce

1. Setup Rancher and at least 2 downstream clusters within the same cluster group.
2. In the `fleet-default` namespace (or where the downstream clusters were placed), add a new `GitRepo` that will deploy the same workload to all downstream clusters
   - Just anything should suffice, e.g. https://github.com/rancher/fleet-test-data/tree/master/simple
3. Wait for the `GitRepo` to become `Ready`, then observe the value that appears for `Resources` count and "ready clusters".
4. Edit the `GItRepo`'s targets configuration to only deploy to one downstream cluster.
5. Wait for `GitRepo` to become `Ready` again, then observe the `Resources` count and "ready clusters" didn't change.
6. After some time, this will eventually decrease to the desired value, meaning that `BundleDeployments` were effectively deleted.

### Environment

```markdown
- Fleet Version: 0.12
```

### Logs

```markdown

```

### Anything else?

I was uncertain of what process was actually deleting the `BundleDeployments` after a few minutes, thinking that it could be some scheduled or background job.
However, after some investigation I found that is only done from the `Cluster` reconciler, which not only triggers on their own events but also based on certain changes on `BundleDeployment` objects. This was causing the difference in behavior I observed.

sbulage:
## System Information

### Before Upgrade:

| Rancher Version | Fleet Version |
|--------|--------|
| v2.11.0 | 106.1.0+up0.12.0 | 

### Steps followed
[...]

PR #3438, opened 2025-03-12
Prune BundleDeployments when GitRepo/Bundle targets change

aruiz14:
Also changes/improves how the finalizers are handled, both for Bundles and BundleDeployments.

Refers to #3437

`BundleDeployments` are only being deleted on the `Cluster` reconciler, which can take considerable time to run and cause confusion in the status of resources.
Note: this PR was superseded by #3509 while still on review. However, I believe this implementation is cleaner, so some of the changes there are being replaced.

### DIFF
--- a/internal/cmd/controller/reconciler/bundle_controller.go
+++ b/internal/cmd/controller/reconciler/bundle_controller.go
@@ -525,7 +525,7 @@ func (r *BundleReconciler) cleanupOrphanedBundleDeployments(ctx context.Context,
 		return err
 	}
 	toDelete := slices.DeleteFunc(list, func(bd fleet.BundleDeployment) bool {
-		return uidsToKeep.Has(bd.ObjectMeta.UID)
+		return uidsToKeep.Has(bd.UID)
 	})
 	return batchDeleteBundleDeployments(ctx, r.Client, toDelete)
 }
--- a/internal/cmd/controller/reconciler/bundle_controller.go
+++ b/internal/cmd/controller/reconciler/bundle_controller.go
@@ -37,6 +37,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
 
+const (
+	// period after which the Bundle reconciler is re-scheduled, in order to wait for the BundleDeploymentReconciler cleanup to finish
+	requeueAfterBundleDeploymentCleanup = 2 * time.Second
+)
+
 type BundleQuery interface {
 	// BundlesForCluster is used to map from a cluster to bundles
 	BundlesForCluster(context.Context, *fleet.Cluster) ([]*fleet.Bundle, []*fleet.Bundle, error)
@@ -133,7 +138,6 @@ func (r *BundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	if err := r.Get(ctx, req.NamespacedName, bundle); err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
-	bundleOrig := bundle.DeepCopy()
 
 	if bundle.Labels[fleet.RepoLabel] != "" {
 		logger = logger.WithValues(
@@ -142,9 +146,15 @@ func (r *BundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		)
 	}
 
-	if res, err := r.addOrRemoveFinalizer(ctx, logger, req, bundle); res {
+	if !bundle.DeletionTimestamp.IsZero() {
+		return r.handleDelete(ctx, logger, req, bundle)
+	}
+
+	if err := r.ensureFinalizer(ctx, bundle); err != nil {
 		return ctrl.Result{}, err
 	}
+
+	bundleOrig := bundle.DeepCopy()
 
 	logger.V(1).Info(
 		"Reconciling bundle, checking targets, calculating changes, building objects",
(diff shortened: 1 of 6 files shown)


internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -487,3 +497,32 @@ func (r *BundleReconciler) updateStatus(ctx context.Context, orig *fleet.Bundle,
 	metrics.BundleCollector.Collect(ctx, bundle)
 	return nil
 }
+
+// cleanupOrphanedBundleDeployments will delete all existing BundleDeployments corresponding that do not match a provided list
+func (r *BundleReconciler) cleanupOrphanedBundleDeployments(ctx context.Context, bundle *fleet.Bundle, created []*fleet.BundleDeployment) error {
+	uidsToKeep := make(sets.Set[types.UID], len(created))
+	for _, bd := range created {
+		uidsToKeep.Insert(bd.UID)
+	}
```
weyfonk, line 506: If we only use `created` for this purpose, wouldn't it make more sense to simply feed the equivalent of `uidsToKeep` directly to `cleanupOrphanedBundleDeployments`?
L282 would then append the `bd.UID` to a slice of `types.UID` (aka `string`), and we'd save ourselves some complexity.

aruiz14: Thanks, good point. I applied this suggestion but kept using a `Set`, as it works better for lookups.

weyfonk (approved):

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -487,3 +497,27 @@ func (r *BundleReconciler) updateStatus(ctx context.Context, orig *fleet.Bundle,
...
+
+	var errs []error
+	for _, bd := range list.Items {
+		if uidsToKeep.Has(bd.UID) {
+			continue
+		}
+		if err := r.Delete(ctx, &bd); client.IgnoreNotFound(err) != nil {
+			errs = append(errs, err)
+		}
```
weyfonk, line 520:
```suggestion
		if err := r.Delete(ctx, &bd); client.IgnoreNotFound(err) != nil {
			errs = append(errs, err)
		}
		
		if err := finalize.PurgeContent(ctx, r.Client, bd.Name, bd.Spec.DeploymentID); err != nil {
		        errs = append(errs, err)
		}
```

aruiz14: I've added 39ac379 with that I think is the proper way to handle this (let the reconciliation from adding a `deletionTimestamp` to the `BundleDeployment` handle the Content management).
Please let me know if you feel strongly about this!

weyfonk: Oh of course, that's much better! :trophy: 
The only amendment I would make to it is removing calls to `PurgeContent` from [here](https://github.com/rancher/fleet/blob/main/internal/cmd/controller/finalize/finalize.go#L103), and we'd be good to go.

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -499,33 +496,26 @@ func (r *BundleReconciler) updateStatus(ctx context.Context, orig *fleet.Bundle,
 	return nil
 }
 
-func findNotTargetedBundleDeployments(ctx context.Context, c client.Client, bundleId types.NamespacedName, targets []*target.Target) ([]types.NamespacedName, error) {
```
aruiz14, line 502: @0xavi0 @manno I'm afraid that this implementation added at https://github.com/rancher/fleet/pull/3509/files#diff-38f599d4b6f84abae0fd9c6582e85b27f3f8a52774a565742e207ef52302274e completely overlaps with the changes I was proposing.
Nonetheless, I believe my proposal is a bit cleaner? (since it does not need to find again which BundleDeployments should be kept, but instead keeps them from the "creation phase").

Let me know if you think otherwise and I can trim this PR to only the improvements in finalizers.

0xavi0 (approved):
lgtm!

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -499,33 +505,43 @@ func (r *BundleReconciler) updateStatus(ctx context.Context, orig *fleet.Bundle,
...
+
+// cleanupOrphanedBundleDeployments will delete all existing BundleDeployments which do not have a match in a provided list of UIDs
+func (r *BundleReconciler) cleanupOrphanedBundleDeployments(ctx context.Context, bundle *fleet.Bundle, uidsToKeep sets.Set[types.UID]) error {
+	list, err := r.listBundleDeploymentsForBundle(ctx, bundle)
 	if err != nil {
-		return []types.NamespacedName{}, err
+		return err
 	}
+	toDelete := slices.DeleteFunc(list, func(bd fleet.BundleDeployment) bool {
+		return uidsToKeep.Has(bd.ObjectMeta.UID)
```
aruiz14, line 528:
```suggestion
		return uidsToKeep.Has(bd.UID)
```

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -496,26 +505,43 @@ func (r *BundleReconciler) updateStatus(ctx context.Context, orig *fleet.Bundle,
...
+	})
+	return batchDeleteBundleDeployments(ctx, r.Client, toDelete)
+}
 
+func batchDeleteBundleDeployments(ctx context.Context, c client.Client, list []fleet.BundleDeployment) error {
 	var errs []error
-	for _, bd := range list.Items {
-		if uidsToKeep.Has(bd.UID) {
+	for _, bd := range list {
+		if bd.DeletionTimestamp != nil {
```
weyfonk, line 536: nit: why check the `DeletionTimestamp` against `nil` here instead of using `IsZero()`?

weyfonk (approved):

internal/cmd/controller/reconciler/bundle_controller.go
aruiz14: I think they are equivalent. `IsZero` uses a pointer receiver so it will work even if `DeletionTimestamp` is `nil`.
I guess it's a matter of style; personally I find it more idiomatic to check against nil, than a negated `IsZero`, slightly easier thought process 😅 

BTW, I searched in the `kubernetes/kubernetes` codebase, and found [78 occurrences for `!= nil`](https://github.com/search?q=repo%3Akubernetes%2Fkubernetes%20%22DeletionTimestamp%20!%3D%20nil%22&type=code) vs. [only 14 for `IsZero`](https://github.com/search?q=repo%3Akubernetes%2Fkubernetes+%22DeletionTimestamp.IsZero%22&type=code) 😄

PR #3457, opened 2025-03-18
[SURE-9419] Configurable replica count

p-se:
Refers to #3377

### Checklist

- [x]  I have updated the documentation via pull request in [website](https://github.com/updatecli/website) repository: see [this PR](https://github.com/rancher/fleet-docs/pull/229)


tp-drosssmith:
thanks for this (from a Fleet end user)

charts/fleet/tests/controller-replica-count.yaml
```diff
@@ -0,0 +1,20 @@
+suite: replica values test for fleet controller
+templates:
+  - deployment.yaml
+  - deployment_gitjob.yaml
+  - deployment_helmops.yaml
+tests:
+  - it: should set spec.replicas to the number of replicas specified in agent.replicas
```
weyfonk, line 7: Is this the result of a copy/paste operation?
How about something like this, or ideally a bit shorter with a similar idea:
```suggestion
  - it: should set spec.replicas for each controller to the number of replicas specified in the controller's replicas field
```

Also, should we set the same value for all deployments in this test? This wouldn't detect issues such as, say, the gitOps controller deployment using the replica count of helmOps.

internal/cmd/options.go
```diff
@@ -48,3 +49,22 @@ func NewLeaderElectionOptions() (LeaderElectionOptions, error) {
...
+func ParseEnvAgentReplicaCount() (int32, error) {
+	replicas, err := parseEnvInt32("FLEET_AGENT_REPLICA_COUNT")
+	if err != nil {
+		return 0, err
+	}
+	return replicas, nil
+}
+
+func parseEnvInt32(envVar string) (int32, error) {
+	if d := os.Getenv(envVar); d != "" {
```
weyfonk, line 62: nit: this could make use of [os.Lookupenv](https://pkg.go.dev/os#LookupEnv) and check its 2nd return parameter which indicates whether the variable is set.

.github/workflows/ci.yml
```diff
@@ -36,6 +36,11 @@ jobs:
       -
         name: unit-test
         run: go test -shuffle=on $(go list ./... | grep -v -e /e2e -e /integrationtests -e /benchmarks)
+      -
+        name: helm-unittest
+        run: |
+          helm plugin install https://github.com/helm-unittest/helm-unittest.git
+          helm unittest -f 'charts/fleet/tests/*.yaml' charts/fleet
```
0xavi0, line 43: I think we don't need to use `-f 'charts/fleet/tests/*.yaml'` if the test files have the `_test.yaml` suffix.

charts/fleet/tests/controller-replica-count.yaml
p-se: I have expanded the tests into three files so that each controller gets only the value set that it should respect, thereby testing specifically for values that are only related to the deployment manifest being tested.

Note that it is not possible to have those three tests in a single file if they have different values to be tested. This was only possible as long they have had the same values passed to it.

charts/fleet/tests/fleet_controller_replica_count_test.yaml
```diff
@@ -0,0 +1,16 @@
+suite: replica values test for fleet controller
+templates:
+  - deployment.yaml
+tests:
+  - it: should set spec.replicas to the number of replicas specified in the controller's replicas field
+    set:
+      controller.replicas: 3
+      extraEnv:
+        - name: 'EXPERIMENTAL_HELM_OPS'
+          value: 'true'
```
weyfonk, line 10: We only need this for the HelmOps controller, don't we?
```suggestion
```

charts/fleet/tests/gitjob_controller_replica_count_test.yaml
```diff
@@ -0,0 +1,16 @@
+suite: replica values test for fleet controller
+templates:
+  - deployment_gitjob.yaml
+tests:
+  - it: should set spec.replicas to the number of replicas specified in the controller's replicas field
+    set:
+      gitjob.replicas: 3
+      extraEnv:
+        - name: 'EXPERIMENTAL_HELM_OPS'
+          value: 'true'
```
weyfonk, line 10:
```suggestion
```

charts/fleet/tests/helmops_controller_replica_count_test.yaml
```diff
@@ -0,0 +1,16 @@
+suite: replica values test for fleet controller
+templates:
+  - deployment_helmops.yaml
+tests:
+  - it: should set spec.replicas to the number of replicas specified in the controller's replicas field
+    set:
+      helmops.replicas: 3
+      extraEnv:
+        - name: 'EXPERIMENTAL_HELM_OPS'
+          value: 'true'
```
weyfonk, line 10: Do we want to add a test case checking that no replicas are deployed if `EXPERIMENTAL_HELM_OPS` is `false`?

p-se: That's difficult. If `EXPERIMENTAL_HELM_OPS` isn't set to true, we don't get a manifest at all.

```yaml
 FAIL  replica values test for fleet controller charts/fleet/tests/helmops_controller_replica_count_test.yaml
    - should not set spec.replicas to the number of replicas specified in the controller's replicas field

        - asserts[0] `isKind` fail
            Template:   fleet/templates/deployment_helmops.yaml
            Expected to be kind:
                Deployment
            Actual:
                no manifest found

        - asserts[1] `equal` fail
            Template:   fleet/templates/deployment_helmops.yaml
            Path:   spec.replicas
            Expected to equal:
                1
            Actual:
                no manifest found
            Diff:
                --- Expected
                +++ Actual
                @@ -1,2 +1,2 @@
                -1
                +no manifest found


Charts:      1 failed, 0 passed, 1 total
Test Suites: 1 failed, 3 passed, 4 total
Tests:       1 failed, 4 passed, 5 total
Snapshot:    0 passed, 0 total
Time:        18.542171ms

Error: plugin "unittest" exited with error
```

weyfonk (approved):

PR #3463, opened 2025-03-20
[SURE-9419] Configurable agent leader election

p-se:
Refers to #3377   

### Checklist

- [ ]  I have updated the documentation via pull request in [website](https://github.com/updatecli/website) repository.

### DIFF
--- a/internal/cmd/options.go
+++ b/internal/cmd/options.go
@@ -13,15 +13,15 @@ type LeaderElectionOptions struct {
 	// LeaseDuration is the duration that non-leader candidates will
 	// wait to force acquire leadership. This is measured against time of
 	// last observed ack. Default is 15 seconds.
-	LeaseDuration *time.Duration
+	LeaseDuration time.Duration
 
 	// RenewDeadline is the duration that the acting controlplane will retry
 	// refreshing leadership before giving up. Default is 10 seconds.
-	RenewDeadline *time.Duration
+	RenewDeadline time.Duration
 
 	// RetryPeriod is the duration the LeaderElector clients should wait
 	// between tries of actions. Default is 2 seconds.
-	RetryPeriod *time.Duration
+	RetryPeriod time.Duration
 }
 
 // NewLeaderElectionOptions returns a new LeaderElectionOptions struct with the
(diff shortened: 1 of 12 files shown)


internal/cmd/options.go
```diff
@@ -24,27 +24,52 @@ type LeaderElectionOptions struct {
 // NewLeaderElectionOptions returns a new LeaderElectionOptions struct with the
 // values parsed from environment variables.
 func NewLeaderElectionOptions() (LeaderElectionOptions, error) {
-	leaderOpts := LeaderElectionOptions{}
-	if d := os.Getenv("CATTLE_ELECTION_LEASE_DURATION"); d != "" {
-		v, err := time.ParseDuration(d)
-		if err != nil {
-			return leaderOpts, fmt.Errorf("failed to parse CATTLE_ELECTION_LEASE_DURATION with duration %s: %w", d, err)
-		}
-		leaderOpts.LeaseDuration = &v
```
aruiz14, line 33: Is there any reason to use pointers here? I believe this method should return a safe `LeaderElectionOptions` object, including default values for every parameter. (e.g. running the agent without setting this environment variables will panic with a nil pointer dereference). So please consider both changing the fields of `LeaderElectionOptions` to not be pointer and populate default values here.

p-se: I did not question the implementation I adopted, but it looks like it would be safe to change it to take a non-pointer value.

However, I'm not sure if we want the agent to silently ignore that a (so far) necessary option has not been set, or possibly has been set incorrectly. Like, if there's a typo in the environment variable, I'd like to know that by having the agent refuse to start rather than to assume the correct value is being used by having the agent starting up anyway.

And since I'm reusing the LeaderElectionOptions struct, any change would also affect the behavior of all the other controllers that take environment variables for configuring leader election.

So, I'm a bit unsure whether we want that (at least in this PR).

manno: If the value is specified it needs to be correct. If it is omitted we pass nil to use the leader election client-go defaults. There should not be a panic in the agent.

aruiz14: That's not completely accurate. `client-go` uses non-pointer `time.Duration` values, so it will only apply its own defaults when `0` is provided, not `nil`. The panic is actually happening here:
https://github.com/rancher/fleet/blob/main/internal/cmd/agent/root.go#L123-L125
That causes a nil-pointer dereference if If the leader election env. vars are not set

Issue #3484, opened 2025-03-27
Unstable gitrepo ready condition for multiple non-ready bundles


rbreddy:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

when there is more than one bundle with "ready: false" state, gitrepos objects will be updated frequently (creating a new resource version) with a "new" message from a different bundle
This causes unnecessary activity at the control plane, eventually slowing down Rancher Continuous Delivery UI

### Expected Behavior

the same gitrepo status is produced if there is no changes on the bundle states

### Steps To Reproduce

1. Installing Rancher v2.10.3, running Fleet 0.11.4
2. Create a GitRepo with more than one bundle in error state
`kind: GitRepo
apiVersion: fleet.cattle.io/v1alpha1
metadata:
  name: bundledependency
  namespace: fleet-default
spec:
  repo: https://github.com/rbreddy/bundledependency 
  branch: main
  targets:
    - clusterSelector:
        matchExpressions:
          - key: provider.cattle.io
            operator: NotIn
            values:
              - harvester`
 3. To analyze it: leave it running for a while and then watch gitrepos, e.g.:
`for kind in gitrepos bundles bundledeployments; do
  {
    kubectl get -A --show-managed-fields --chunk-size=0 --watch-only --output-watch-events -o yaml $kind >$kind-watch-only-events.yaml &
    pid=$!
    sleep 180
    kill $pid
  } &
done
wait`
4. Parse the gitrepos-watch-only-events.yaml with: https://gist.github.com/aruiz14/b58fcc96fde894cbf85562e888d8e1bd

### Environment

```markdown
- Architecture: amd64
- Fleet Version:v0.11.4
```

### Logs

```markdown
Example of frequent updates on the message
@@ -100,8 +100,8 @@
 status:
   commit: 124109ec64e6c2ef5de39cd7a704bda6e2d4b49e
   conditions:
-    - lastUpdateTime: "2025-03-27T09:47:24Z"
-      message: 'ErrApplied(1) [Cluster fleet-default/downstream-0-0: list bundledeployments: no bundles matching labels fleet.cattle.io/bundle-name=logging-logging-crd,fleet.cattle.io/bundle-namespace=fleet-default in namespace fleet-default]'
+    - lastUpdateTime: "2025-03-27T09:46:20Z"
[...]

PR #3485, opened 2025-03-27
Ensure same bundle is selected for non-ready status

rbreddy:
Refers to #3484

Adding a sort by UID for the bundle list, so the same bundle is selected to update the non ready state. 

## Additional Information
This only makes sure that the same bundle is always selected. But it still picks one (as long as it is the first and in non ready state). 
A follow-up could aggregate all the non-ready bundles on the message.
And also avoid overiding the status condition previously set at: https://github.com/rancher/fleet/blob/main/internal/cmd/controller/gitops/reconciler/status_controller.go#L159

### DIFF
--- a/internal/cmd/controller/gitops/reconciler/status_controller.go
+++ b/internal/cmd/controller/gitops/reconciler/status_controller.go
@@ -180,6 +180,7 @@ func (r StatusReconciler) setReadyStatusFromBundle(ctx context.Context, gitrepo 
 	}
 
 	// Make sure the bundles are always iterated in the same order
+	// The code below will pick the first element matching the condition, so successive executions should produce the same result.
 	sort.Slice(bList.Items, func(i, j int) bool {
 		return bList.Items[i].UID < bList.Items[j].UID
 	})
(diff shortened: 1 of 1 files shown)


rbreddy:
tagging @aruiz14 as we have been working together on this

aruiz14 (approved):

manno (approved):

0xavi0:
Can we add a brief comment explaining why we are sorting?
I remember we saw sorting before and we thought it was useless...

internal/cmd/controller/gitops/reconciler/status_controller.go
```diff
@@ -179,6 +179,11 @@ func (r StatusReconciler) setReadyStatusFromBundle(ctx context.Context, gitrepo
 		return err
 	}
 
+	// Make sure the bundles are always iterated in the same order
```
aruiz14, line 182:
```suggestion
	// Make sure the bundles are always iterated in the same order
	// The code below will pick the first element matching the condition, so successive executions should produce the same result.
```

0xavi0: Thanks, lgtm. I can't review the PR (I don't know why).

Issue #2776, opened 2024-08-26
Support Webhook Secret per GitRepo


0xavi0:
Webhook secret is unique per cluster, and it's shared for all gitrepos and all webhooks (its name has to be **gitjob-webhook**).

This is not flexible as it forces users to use/share the same secret when configuring their `gitrepos`. 

One possibility could be to add a new field in the `gitrepo` resource (something like `webhookSecret`) and, when we receive a webhook payload and we have identified which `gitrepo` is referencing, check for the secret in the `gitrepo`

We're instantiating different 3rd party webhook handlers and we need to pass the secret.

- [x] remove the watch on secrets, process in request. Instantiation is cheap. Request is already operating on all gitrepo resources.
- [x] keep old secret as a fallback, log a warning if old secret is missing, but authentication info is in header. If possible.
- [x] error if gitrepo references a secret, but secret is missing

PR #3490, opened 2025-03-31
New webhook secret handling

0xavi0:
This PR changes the secret handling for webhooks in order to add the possibility to define a secret per `GitRepo`. The actual `main` code has a single global secret shared by all the possible `GitRepos` in the system, which forces the users to share the same secret for all their webhooks in the cluster.

The secret is also cached, which means that it listens to the informer cache for all secrets in the cluster and, if any secret is updated it checks if it's the global secret for webhooks. In that case it reinstantiates all the webhook parsers for all the supported providers with the new secret value (or none if the secret was deleted).

It also needs to instantiate the webhook parsers for all supported providers at the beginning.

Also... a bug was found in that secret caching. If all the `gitjob` pods are restarted, the secret will be completely ignored, accepting any webhook received (which might mean a security problem).

The new implementation does not cache any secret and does not create all the different providers in advance. It just checks for the secret right before updating the `WebhookCommit` and, depending on the payload, it will instantiate the needed parser for that provider.

The price we pay for this is that we need to double parse the webhook request. This is because when we receive the payload we don't know yet the name of the `GitRepo` it is trying to update and we cannot check if it has a secret defined for it.

The first round of parsing will be used to determine the name of the `GitRepo` and, if a secret is defined (global or per `GitRepo`) it will parse again setting that secret.

The library used to parse the webhook payload doesn't offer the possibility to check just the secret given a payload.

The new field added in the `GitRepo` CRD is called `webhookSecret` and references the name of the secret, which should be located at the same namespace as the `GitRepo`

If a `Gitrepo` secret is defined it will take preference over the global secret, that is still supported.

This is an example of a `GitRepo` with the new field.

```yaml
kind: GitRepo
metadata:
  name: simple
  namespace: fleet-local
spec:
  repo: "https://github.com/rancher/fleet-examples"
  paths:
  - simple
  disablePolling: true
  webhookSecret: webhook-secret-name
```

Refers to: https://github.com/rancher/fleet/issues/2776

### Checklist

### DIFF
--- a/pkg/webhook/parser.go
+++ b/pkg/webhook/parser.go
@@ -1,0 +1,159 @@
+package webhook
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/go-playground/webhooks/v6/azuredevops"
+	"gopkg.in/go-playground/webhooks.v5/bitbucket"
+	bitbucketserver "gopkg.in/go-playground/webhooks.v5/bitbucket-server"
+	"gopkg.in/go-playground/webhooks.v5/github"
+	"gopkg.in/go-playground/webhooks.v5/gitlab"
+	"gopkg.in/go-playground/webhooks.v5/gogs"
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	githubKey          = "github"
+	gitlabKey          = "gitlab"
+	bitbucketKey       = "bitbucket"
+	bitbucketServerKey = "bitbucket-server"
+	gogsKey            = "gogs"
+	azureUsername      = "azure-username"
+	azurePassword      = "azure-password"
+)
+
+func parseWebook(r *http.Request, secret *corev1.Secret) (interface{}, error) {
+	switch {
+	//Gogs needs to be checked before Github since it carries both Gogs and (incompatible) Github headers
+	case r.Header.Get("X-Gogs-Event") != "":
+		return parseGogs(r, secret)
+	case r.Header.Get("X-GitHub-Event") != "":
+		return parseGithub(r, secret)
+	case r.Header.Get("X-Gitlab-Event") != "":
+		return parseGitlab(r, secret)
+	case r.Header.Get("X-Hook-UUID") != "":
+		return parseBitbucket(r, secret)
+	case r.Header.Get("X-Event-Key") != "":
+		return parseBitbucketServer(r, secret)
+	case r.Header.Get("X-Vss-Activityid") != "" || r.Header.Get("X-Vss-Subscriptionid") != "":
+		return parseAzureDevops(r, secret)
+	}
+
+	return nil, nil
+}
+
+func getValue(secret corev1.Secret, key string) (string, error) {
+	value, ok := secret.Data[key]
+	if !ok {
+		return "", fmt.Errorf("secret key %q not found in secret %q", key, secret.Name)
+	}
+
+	return string(value), nil
+}
+
+func parseGogs(r *http.Request, secret *corev1.Secret) (interface{}, error) {
+	opts := []gogs.Option{}
+	if secret != nil {
+		value, error := getValue(*secret, gogsKey)
+		if error != nil {
+			return nil, error
+		}
+		opts = append(opts, gogs.Options.Secret(value))
+	}
+	p, err := gogs.New(opts...)
+	if err != nil {
+		return nil, err
+	}
+
+	return p.Parse(r, gogs.PushEvent)
+}
+
+func parseGithub(r *http.Request, secret *corev1.Secret) (interface{}, error) {
+	opts := []github.Option{}
+	if secret != nil {
+		value, error := getValue(*secret, githubKey)
+		if error != nil {
+			return nil, error
+		}
+		opts = append(opts, github.Options.Secret(value))
+	}
+	p, err := github.New(opts...)
+	if err != nil {
+		return nil, err
+	}
+
+	return p.Parse(r, github.PushEvent)
+}
+
+func parseGitlab(r *http.Request, secret *corev1.Secret) (interface{}, error) {
+	opts := []gitlab.Option{}
+	if secret != nil {
+		value, error := getValue(*secret, gitlabKey)
+		if error != nil {
+			return nil, error
+		}
+		opts = append(opts, gitlab.Options.Secret(value))
+	}
+	p, err := gitlab.New(opts...)
+	if err != nil {
+		return nil, err
+	}
+
+	return p.Parse(r, gitlab.PushEvents, gitlab.TagEvents)
+}
+
+func parseBitbucket(r *http.Request, secret *corev1.Secret) (interface{}, error) {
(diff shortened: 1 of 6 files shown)


pkg/webhook/webhook_test.go
```diff
@@ -356,83 +373,308 @@ func TestGitHubWrongSecret(t *testing.T) {
...
+					secret.Data = map[string][]byte{
+						tt.gitrepoSecretKey: []byte(tt.gitrepoSecretValue),
+					}
+					return nil
+				} else if tt.globalSecret {
+					secret.Data = map[string][]byte{
+						tt.globalSecretKey: []byte(tt.globalSecretValue),
+					}
+					return nil
+				}
```
weyfonk, line 591: nit: why check those conditions again when we already have an if/else block over them above?

pkg/webhook/webhook_test.go
```diff
@@ -356,83 +373,308 @@ func TestGitHubWrongSecret(t *testing.T) {
...
-
-	ctlr := gomock.NewController(t)
-	mockClient := mocks.NewMockClient(ctlr)
-
-	gitRepo := &v1alpha1.GitRepo{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: "test",
+	gitrepoSecretName := "gitrepoSecret"
+	tests := map[string]struct {
+		secretValueInWebhook string
```
weyfonk, line 380: This field doesn't seem to be used :thinking: 
Should it be used instead of a hard-coded string on L641 maybe? (and ideally renamed to `secretValueInRequest`, or something similar to clarify that the secret is to be found in the request received from the webhook provider)

pkg/webhook/webhook_test.go
```diff
@@ -356,83 +373,308 @@ func TestGitHubWrongSecret(t *testing.T) {
...
+			setSecretInGitrepo:   false,
+			expectedResCode:      http.StatusOK,
+			expectedCommitUpdate: true,
 		},
-	).Times(1)
 
-	w := &Webhook{
-		client:    mockClient,
-		namespace: "default",
+		// when no secret is defined we accept the payload
```
weyfonk, line 519:
```suggestion
		// when a referenced secret does not exist, we throw an error
```

pkg/webhook/parser_test.go
```diff
@@ -0,0 +1,1275 @@
...
+			body: []byte(`{
+				"ref": "refs/heads/main",
+				"after": "af69d162de5a276abc86e0686b2b44033cd3f442",
+				"repository": {
+					"html_url": "https://gogs.example.com/example/repo"
+				}
+			}`),
+			headers: map[string]string{
+				"X-Gogs-Event":     "push",
+				"X-Gogs-Signature": "sha1=57968570459352679905f2372454008f9015101a",
```
weyfonk, line 107: This is not very relevant to this test case, but why is the signature format different here?
Also, why `sha1`? I thought that algorithm was deprecated.

pkg/webhook/parser_test.go
```diff
@@ -0,0 +1,1275 @@
...
+					t.Fatalf("parseBitbucketServer() got nil, want not nil")
+				}
+				gotBB, ok := got.(bitbucketserver.RepositoryReferenceChangedPayload)
+				if !ok {
+					t.Fatalf("parseBitbucketServer() got %T, want bitbucketserver.RepositoryReferenceChangedPayload", got)
+				}
+
+				_, okCloneLinks := gotBB.Repository.Links["clone"]
+				if okCloneLinks {
+					// if len(gotBB.Repository.Links["clone"]) > 0 {}
```
weyfonk, line 1041:
```suggestion
```

weyfonk (commented):
Thanks for this, _especially_ for its thorough testing! :bow:

pkg/webhook/webhook.go
```diff
@@ -19,13 +21,14 @@ import (
 	"gopkg.in/go-playground/webhooks.v5/gitlab"
 	"gopkg.in/go-playground/webhooks.v5/gogs"
 
-	v1alpha1 "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
+	"github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
+	fleet "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
```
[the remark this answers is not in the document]
0xavi0: looks like I forgot to change 1 `v1aplha1` to `fleet` and VSC did the rest :) 
Changed

pkg/webhook/parser.go
```diff
@@ -0,0 +1,159 @@
...
+	githubKey          = "github"
+	gitlabKey          = "gitlab"
+	bitbucketKey       = "bitbucket"
+	bitbucketServerKey = "bitbucket-server"
+	gogsKey            = "gogs"
+	azureUsername      = "azure-username"
+	azurePassword      = "azure-password"
+)
+
+func parseWebook(r *http.Request, secret *corev1.Secret) (interface{}, error) {
```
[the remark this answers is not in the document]
0xavi0: changed

pkg/webhook/webhook_test.go
0xavi0: The idea was to do the error checking first and the return of secret data later.
And, as this is a test file.. I thought it was better to do it that way to separate functionalities (the code is confusing enough if you are not familiar with it)

No strong opinion on any on these things, tbh.

pkg/webhook/webhook_test.go
0xavi0: changed as suggested and using it instead of the hardcoded one

pkg/webhook/webhook_test.go
0xavi0: changed as suggested

pkg/webhook/parser_test.go
0xavi0: It's simpler... :P I did copy/paste that from a Github test.
[Github still supports HMAC-SHA1 for legacy purposes](https://docs.github.com/en/webhooks/using-webhooks/validating-webhook-deliveries#troubleshooting) and the [library](https://github.com/go-playground/webhooks/blob/59fb74817c3d5c702553964f4d2faff80f23d350/github/github.go#L157:L160) we use for parsing the webhook uses the legacy header (which is HMAC-SHA1).
The latest version of the library (v6) still uses the HMAC-SHA1 header, btw.

Anyway... at this particular test the signature is irrelevant (as you noticed) because we're not even reading it, but I changed it so it does not say sha1=

pkg/webhook/webhook_test.go
weyfonk: Understood. As you've noted, it's a test file, no worries.

pkg/webhook/parser_test.go
weyfonk: Nice, thanks for clarifying :bow:

weyfonk (approved):

Issue #3483, opened 2025-03-27
Multiple fleet agents created when deploying Rancher 2.12-head


mmartin24:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Issue

Multiple fleet agents are deployed when deploying Rancher `v2.12-b55fb8c0a326d73fbb3dec30d39d9842901c4db4-head` and Fleet `fleet:106.0.0+up0.12.0`

Easily reproduceable installing latest Rancher version with Fleet

### Steps to reproduce

Deploy k3d and latest Rancher. I used this

<details closed>
  <summary>Click to expand installation steps</summary>

  ```
# Go to main branc on fleet repo
# k3d prep
./dev/build-fleet 
./dev/setup-k3d
k3d image import rancher/fleet-agent:dev rancher/fleet:dev -m direct -c upstream

# Cert Manager installation
helm repo add jetstack https://charts.jetstack.io
helm repo update
helm upgrade --install cert-manager --namespace cert-manager jetstack/cert-manager \
    --create-namespace \
    --set installCRDs=true \
    --set "extraArgs[0]=--enable-certificate-owner-ref=true" \
    --wait
sleep 8

# Rancher install
export MY_IP='' # Add local IP here
export SYSTEM_DOMAIN="${MY_IP}.nip.io"
export RANCHER_USER=admin RANCHER_PASSWORD=password
export RANCHER_URL=https://${MY_IP}.nip.io/dashboard

helm repo add rancher-latest https://releases.rancher.com/server-charts/latest
helm repo update

helm upgrade --install rancher rancher-latest/rancher \
  --devel \
  --set "rancherImageTag=head" \
  --namespace cattle-system --create-namespace \
  --set "extraEnv[1].name=CATTLE_AGENT_IMAGE" \
  --set "extraEnv[0].name=CATTLE_SERVER_URL" \
  --set hostname=$SYSTEM_DOMAIN \
  --set bootstrapPassword=password \
  --set replicas=1 \
  --set agentTLSMode=system-store \
  \
  --set "extraEnv[1].value=rancher/rancher-agent:head" \
  \
  --wait
```
</details>

Observe how multiple agents are created from start. 

```
> k get pods -A
NAMESPACE                         NAME                                        READY   STATUS      RESTARTS   AGE
cattle-fleet-local-system         fleet-agent-57c4fd678d-scqqj                1/1     Running     0          111m
[...]

PR #3492, opened 2025-04-01
Set agent deployment strategy and history

manno:
Refers to #3483 and #3377

Even if this turns out to be a garbage collector issue, I would still update the deployment options. 
Not sure if we need to bring back the explicit deletion of pods: https://github.com/rancher/fleet/pull/1905/files#diff-be5264e16d1776adf438294ce1bcdb18d6c5df2c06d03e3469acaff79ba310a8

### DIFF
--- a/charts/fleet-agent/templates/deployment.yaml
+++ b/charts/fleet-agent/templates/deployment.yaml
@@ -6,6 +6,10 @@ spec:
   selector:
     matchLabels:
       app: fleet-agent
+  replicas: 1
+  revisionHistoryLimit: 0
+  strategy:
+    type: Recreate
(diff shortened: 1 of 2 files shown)


charts/fleet-agent/templates/deployment.yaml
```diff
@@ -6,6 +6,10 @@ spec:
   selector:
     matchLabels:
       app: fleet-agent
+  replicas: 1
```
weyfonk, line 9: This would conflict with #3457.

manno: Only if it was merged...

Issue #2773, opened 2024-08-22
Testing: use bundles instead of `GitRepo`s


weyfonk:
Fleet's end-to-end tests spend too much time running logic which is tested repeatedly across its suites, by creating `GitRepo`s everywhere.
Let's migrate end-to-end tests to directly create bundles instead wherever possible, which will speed up those tests.

PR #3493, opened 2025-04-01
Create bundles directly in release cleanup end-to-end tests

weyfonk:
This simplifies end-to-end tests for release cleanup by directly creating bundles, instead of dealing with `GitRepo`s.
Brief tests on my machine suggest a 5x improvement in execution time (from ~50s to ~10s) of the set of 3 test cases affected by this change, which no longer need test infrastructure.

Refers to #2773


0xavi0 (approved):
lgtm

Issue #3038, opened 2024-10-31
Review/Update/Cleanup dev and .github/scripts scripts 


0xavi0:
There are a few scripts under the [dev](https://github.com/rancher/fleet/tree/main/dev) folder and the [.github/scripts](https://github.com/rancher/fleet/tree/main/.github/scripts) that are outdated or possibly old and are no longer used. 

Ideally, when using one of those script for deploying a cluster with the dev version of Fleet, or when updating Fleet to the dev version in a Rancher setup we should not edit those scripts. 

If the script is no longer valid, we should delete it to avoid possible confusions.

PR #3499, opened 2025-04-02
Clean up dev and CI scripts

manno:
Refers to #3038


0xavi0:
Other possible scripts to check if they are necessary or not:

* dev/airgap-rancher.sh 
* dev/delve-debug 
* dev/run-integration-tests.sh --> seems to do the same than the already used in CI .github/scripts/run-integration-tests.sh
* dev/setup-fleet-downstream

PR #3509, opened 2025-04-04
Delete no longer targeted bundle deployments

0xavi0:
Delete Bundle deployments that exist in the Bundle, that are also deployed, but that are no longer targeted.

When changing either the target customisations or targets in the `GitRepo` definition, Fleet does not delete bundle deployments that are no longer targeted using the new target rules.

Refers to: https://github.com/rancher/fleet/issues/2995


manno (approved):

aruiz14:
@0xavi0 @manno I think there is quite some overlap between this and https://github.com/rancher/fleet/pull/3438? Note @weyfonk and I have already been discussing a bit about how finalizers are handled.

manno:
@aruiz14 👍 looks like 3438 is an improvement to handle finalizers smarter? It is on the 2.12 milestone, do you think it should be backported to 2.11? We could target 2.11.2, 2.11.1 bug freeze already passed.

Issue #3295, opened 2025-02-03
Feature Request: Add annotation for fleet-controller and gitjob deployement.


amolkharche13:
### Is your feature request related to a problem? Yes

Both the fleet-controller and gitjob pods have metrics available that can potentially be sent to our SUSE Observability environment, and this would benefit all Rancher customers that use Fleet and SUSE Observability.

When we are trying to configure through terraform we dont see option to add annotation.

### Solution you'd like

The following annotation we need to add in deployment spec:

```
...
metadata:
  annotations:
        ad.stackstate.com/fleet-controller.check_names: '["openmetrics"]'
        ad.stackstate.com/fleet-controller.init_configs: '[{}]'
        ad.stackstate.com/fleet-controller.instances: |
          [ 
            {
              "prometheus_url": "http://%%host%%:8080/metrics",
              "namespace": "cattle_fleet_system", 
              "metrics": ["*"]
            }
          ]
      creationTimestamp: null
...
# This already exists in the pod spec, the container name needs to match the container that is exposing the openmetrics endpoint
spec:
  containers:
   - name: fleet-controller
...
 ```
```
 ...
metadata:
  annotations:
        ad.stackstate.com/ gitjob.check_names: '["openmetrics"]'
        ad.stackstate.com/ gitjob .init_configs: '[{}]'
        ad.stackstate.com/ gitjob .instances: |
          [ 
            {
              "prometheus_url": "http://%%host%%:8081/metrics",
              "namespace": "cattle_fleet_system", 
              "metrics": ["*"]
            }
          ]
      creationTimestamp: null
...
# This already exists in the pod spec, the container name needs to match the container that is exposing the openmetrics endpoint
spec:
  containers:
   - name:  gitjob  
  ```

### Alternatives you've considered

It would be nice to add in helm chart flag that will add the stackstate annotations automatically. Maybe something like "enable_observability_metrics: true"

manno:
We can add `extraLabels` and `extraAnnotations` to the helm chart.

manno:
@amolkharche13 we just published a 0.13 release candidate, which can be configured with these annotations.

dnoland1:
@amolkharche13 can you check if https://github.com/rancher/fleet/pull/3531 looks correct? I believe the stackstate annotations need to ultimately land on the pod spec and not directly on the deployment spec. Can you try out the RC to confirm it works?

amolkharche13:
[...]

PR #3531, opened 2025-04-09
Adds extra labels and annotations in charts

0xavi0:
Adds the `extraLabels` and `extraAnnotations` values to the helm charts. Those values are propagated to the `fleet-controller`, `gitjob` and `helmops` deployments.

It also introduces helm charts unittests testing those 2 new values.

Refers to https://github.com/rancher/fleet/issues/3295

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,8 +11,7 @@ on:
 env:
   GOARCH: amd64
   CGO_ENABLED: 0
-  HELM_VERSION: 3.17.2
-  HELM_UNITTEST_VERSION: 0.8.1
+
 jobs:
   unit-test:
     runs-on: runs-on,runner=16cpu-linux-x64,mem=16,run-id=${{ github.run_id }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,8 @@ on:
 env:
   GOARCH: amd64
   CGO_ENABLED: 0
-
+  HELM_VERSION: 3.17.2
+  HELM_UNITTEST_VERSION: 0.8.1
 jobs:
   unit-test:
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -28,6 +29,13 @@ jobs:
       -
         name: Run chart-testing (lint)
         run: ct lint --all --validate-maintainers=false charts/
+      - 
+        name: Run helm charts unittests
```
weyfonk, line 33: I think Helm is already available on the runner, see also [this](https://github.com/rancher/fleet/pull/3457/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f) which only installs the `helm-unittest` plugin.

0xavi0: Oh, great! I didn’t see that PR was also introducing unit tests for the Helm charts, thanks!

I’ve updated it to match that PR, with the exception of the `-f` flag. (I mentioned in the PR that if the files have the `_test.yaml` suffix, the plugin loads them automatically.)

.github/workflows/ci.yml
```diff
@@ -11,7 +11,8 @@ on:
 env:
   GOARCH: amd64
   CGO_ENABLED: 0
-
+  HELM_VERSION: 3.17.2
+  HELM_UNITTEST_VERSION: 0.8.1
```
weyfonk, line 15:
```suggestion
```

.github/workflows/ci.yml
weyfonk: Nice, thanks :)
It seems there's a couple of superfluous env vars now; once removed, this should be ready to merge.

.github/workflows/ci.yml
0xavi0: Deleted

weyfonk (approved):

Issue #3532, opened 2025-04-10
[v0.11] Too many updates for BundleDeployments with more than 10 non-ready or modified resources


aruiz14:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

For `Bundles` deploying a considerable amount of resources, when many of those (more than 10) are in Modified or NonReady state, the `fleet-agent` will update the `BundleDeployment` status too often, producing a different set of `modifiedStatus` and/or `nonReadyStatus` resources on every update. This also causes a cascade effect, since this information bubbles up to `Bundle` and `GitRepo` statuses, causing unnecessary noise in the system (CPU usage due to reconciliation, network and k8s API calls, UI refreshes, etc.).

### Expected Behavior

`status.modifiedStatus` and `status.nonReadyStatus` fields have a stable list based on the the same resources state, hence avoiding too-frequent updates.

### Steps To Reproduce

1. Deploy a `Bundle` with more than 10 resources, e.g. this will create N ConfigMaps in the `default` namespace:
```
apiVersion: fleet.cattle.io/v1alpha1
kind: Bundle
metadata:
  name: many-configmaps
  namespace: fleet-local
spec:
  correctDrift: {}
  helm:
    chart: many-configmaps
    values:
      count: 20 # configure the amount of configmaps here
  ignore: {}
  namespace: default
  resources:
  - content: |
      apiVersion: v2
      name: many-configmaps
      type: application
      version: 0.1.0
      appVersion: "1.16.0"
    name: many-configmaps/Chart.yaml
  - content: |
      {{- $top := . -}}
      {{- $count := .Values.count | add 1 }}
      {{- range $i := untilStep 1 (int $count) 1 }}
      ---
      apiVersion: v1
      kind: ConfigMap
      metadata:
        name: many-cm-{{.}}
      data:
        key: value
      {{end -}}
    name: many-configmaps/templates/configmap.yaml
  targetRestrictions:
    - clusterSelector:
        matchExpressions:
          - key: provider.cattle.io
            operator: NotIn
            values:
              - harvester
  targets:
    - clusterSelector:
        matchExpressions:
          - key: provider.cattle.io
            operator: NotIn
            values:
              - harvester
      ignore: {}
```
2. Watch the `BundleDeployment`:
```
kubectl get bundledeployments -A -l fleet.cattle.io/bundle-name=many-configmaps -w --output-watch-events -o yaml
```
[...]

PR #3533, opened 2025-04-10
[v0.11] Stable BundleDeployment modified/nonReady status

aruiz14:
Refers to #3532

While we are sorting the result of `modified` and `nonReady`, since we early aborting after reaching the items limit, that subset of items could be different on every execution, ending up with an ordered subset but not idempotent results.

I'm introducing the new Go iterator library to make sure we are iterating over the map inputs in order directly.


internal/cmd/agent/deployer/monitor/updatestatus.go
```diff
@@ -263,7 +285,11 @@ func modified(ctx context.Context, c client.Client, logger logr.Logger, plan des
 			return sortKey(result[i]) < sortKey(result[j])
```
aruiz14, line 285: We need to keep sorting result, since the keys in `plan`'s `Create`, `Delete`, `Modified` could be different.

manno (approved):

Issue #3501, opened 2025-04-03
fleet-agent-register crashes maybe with missing leases


marthydavid:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

0.12.0 fleet-agent-0 Init:crashloopbackoff within fleet-agent-register
```
$ kubectl get pods -n cattle-fleet-local-system
NAME            READY   STATUS                  RESTARTS        AGE
fleet-agent-0   0/2     Init:CrashLoopBackOff   5 (2m29s ago)   5m33s
```

### Steps To Reproduce

Upgrade rancher 2.11.0 from 2.10.3

### Environment

```markdown
- Architecture: arm64
- Fleet Version: 0.12.0
- Cluster:
  - Provider: k3s
  - Options: Deployed by rancher-2.11.0
  - Kubernetes Version: v1.31.3+k3s1
```

### Logs

```markdown
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x2042b7c]
goroutine 1 [running]:
github.com/rancher/fleet/internal/cmd/agent.(*FleetAgent).Run(0x4000b2fa40, 0x4000b42008, {0x0?, 0x0?, 0x0?})
	/home/runner/_work/fleet/fleet/internal/cmd/agent/root.go:123 +0x33c
github.com/rancher/fleet/internal/cmd.Command.bind.func4(0x4000b42008, {0x40003c18b0, 0x1, 0x1})
	/home/runner/_work/fleet/fleet/internal/cmd/builder.go:272 +0xf8
github.com/spf13/cobra.(*Command).execute(0x4000b42008, {0x400004c0b0, 0x1, 0x1})
	/home/runner/go/pkg/mod/github.com/spf13/cobra@v1.9.1/command.go:1015 +0x82c
github.com/spf13/cobra.(*Command).ExecuteC(0x4000b42008)
	/home/runner/go/pkg/mod/github.com/spf13/cobra@v1.9.1/command.go:1148 +0x350
github.com/spf13/cobra.(*Command).Execute(...)
	/home/runner/go/pkg/mod/github.com/spf13/cobra@v1.9.1/command.go:1071
github.com/spf13/cobra.(*Command).ExecuteContext(0x4000078738?, {0x2d804e0?, 0x4000b2f9f0?})
[...]

PR #3534, opened 2025-04-10
Avoid panic if leader election opts are not set for the client

manno:
client-go uses non-pointer time.Duration values, so it will only apply its own defaults when 0 is provided, not nil.

Refers to: https://github.com/rancher/fleet/issues/3501

See also: https://github.com/rancher/fleet/pull/3463#discussion_r2028648698

### DIFF
--- a/internal/cmd/options.go
+++ b/internal/cmd/options.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"os"
 	"time"
+
+	"k8s.io/utils/ptr"
 )
 
 type LeaderElectionOptions struct {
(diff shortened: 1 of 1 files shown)


weyfonk (approved):

internal/cmd/options.go
```diff
@@ -32,6 +34,10 @@ func NewLeaderElectionOptions() (LeaderElectionOptions, error) {
 		}
 		leaderOpts.LeaseDuration = &v
 	}
+	if leaderOpts.LeaseDuration == nil {
```
weyfonk, line 37: Do we need something similar with the renew deadline and retry period? :thinking:

manno: Not according to https://github.com/rancher/fleet/pull/3463#discussion_r2028648698

manno: Or maybe we do. I misread that comment as LeaseDuration, when it says Duration

aruiz14:
https://github.com/rancher/fleet/pull/3536

manno:
closed for https://github.com/rancher/fleet/pull/3536

Issue #3535, opened 2025-04-10
Fleet 0.12 changes GitRepo.Resource.PerClusterState From List To Map


manno:
> We had a similar problem.
> 
> In our case, the gitjob container also didn't start (or crashloop).
> The logs hinted us in the Direction, that some Gitrepo resources were faulty after the update.
> 
> In the Status of some there were 
> ```
> perClusterState: 
> - {}
> ```
> 
> I've changed them all to 
> ```
> perClusterState: {}
> ```
> 
> important: to update the status, you need to use the --subresource=status flag.
> 
> After that, everything worked fine again for us.
> Might be a completely other Problem at your end, but maybe that helps. 

 _Originally posted by @BiggA94 in [#3501](https://github.com/rancher/fleet/issues/3501#issuecomment-2775801039)_

aruiz14:
This could be tested together with https://github.com/rancher/fleet/issues/3548, starting from 0.11, create some GitRepo resources with some non-ready resources, upgrading to 0.12, then back to 0.11.

aruiz14:
# QA Template

## Solution
Use custom deserializer to allow ignoring incompatible data formats.

## Testing
### Case 1
 - Run Fleet 0.11.
 - Apply a `GitRepo` with some sample data. E.g. `https://github.com/rancher/fleet-test-data`, with path `simple`.
 - Wait for Fleet to deploy it, then modify or delete any of the deployed resources.
 - Observe how the GItRepo status is Modified/Missing, and includes `perClusterState`:
```
perClusterState:
- clusterId:
   ...
```
- Upgrade Fleet to 0.12 including this patch.
- The gitjob controller should run correctly and show no warnings about this payload.
- The GitRepo status yaml is updated to either omit this field, or include a new value (using a map instead of an array):
```
perClusterState:
  ready:
  - fleet-default/downstream-0-0
```

PR #3542, opened 2025-04-10
[v0.12] Fix resource.PerClusterState incompatibility

aruiz14:
Refers to #3535

Deserializing existing data with the previous format (an array instead of a map) will make the strict YAML (actually JSON) deserializer in Go to complain about data being incorrect. Adding a custom `UnmarshalJSON` function for this field should allow to process that existing data correctly, ignoring the offending field, which should be populated on the following resource reconciliation.


manno (approved):

sbulage:
There is issue present to verify this PR, moving to Done.

Issue #3548, opened 2025-04-11
[v0.11] Prevent breaking change in gitrepos resource.PerClusterState when rolling back from Fleet 0.12


aruiz14:
Complementary to #3542, we'll a similar approach in 0.11 in case there is a rollback from upgrading to 0.12.

aruiz14:
# QA Template

## Solution
Use custom deserializer to allow ignoring incompatible data formats.

## Testing
### Case 1
 - Run Fleet 0.12.
 - Apply a `GitRepo` with some sample data. E.g. `https://github.com/rancher/fleet-test-data`, with path `simple`.
 - Wait for Fleet to deploy it, then check the status YAML: every item in the `resources` field should contain `perClusterState` like:
```
perClusterState:
  ready:
  - fleet-default/downstream-0-0
```
 - Downgrade Fleet to 0.11 including this patch.
 - The `gitjob` controller should run correctly and show no warnings about this payload.
 - The `GitRepo` status yaml is updated to omit this field.
 - Modify or deleting any of the deployed resources.
 - Observe how the GItRepo status is Modified/Missing, and includes a `perClusterState` field in the entry for the modified/resource (it must be an array, not a map):
```
perClusterState:
- clusterId:
...
```

sbulage:
## Steps Performed
- I have followed steps from: https://github.com/rancher/fleet/issues/3535#issuecomment-2804665522
- Thanks @aruiz14 for providin above steps.
- See below video for more details.

### Video of steps performed

<details><summary><s>Upgrade</s> Downgrade of Rancher from 2.11.1-alpha2 to 2.10.5-alpha3 </summary>

https://github.com/user-attachments/assets/14c55acb-d87e-4353-a967-00cb91082175

</details> 

_**Note: Please play video in 1.5x speed.**_

### Before ~Upgrade~ Downgrade

| Rancher Version | Fleet Version |
|--------|--------|
| v2.11.1-alpha2 | 106.0.1+up0.12.1-rc.1 | 


<details><summary>Screenshots of GitRepo YAML missing resource on 2.11.1-alpha2 for perClusterState</summary>

![Image](https://github.com/user-attachments/assets/4b547680-3e6a-4ff1-b42d-f159b9a8d7ab)

</details> 

### After ~Upgrade~ Downgrade

| Rancher Version | Fleet Version |
|--------|--------|
| v2.10.5-alpha3 | 105.0.6+up0.11.6-rc.2 | 

[...]

PR #3549, opened 2025-04-11
[v0.11] Prevent breaking change in PerClusterState on rollback

aruiz14:
Refers to #3548

Complementary to https://github.com/rancher/fleet/pull/3542, we'll a similar approach in 0.11 in case there is a rollback from upgrading to 0.12.


manno (approved):

sbulage:
There is issue present to verify this PR, moving to Done.

Issue #3061, opened 2024-11-08
The hasConfigChanged has flaws and causes the unnecessary re-deployment of fleet-agent on local cluster


w13915984028:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

https://github.com/rancher/fleet/blob/1cddbbff1c2cd71c9b9011c3738754e5b4c8fa89/internal/cmd/controller/agentmanagement/controllers/cluster/import.go#L98

fleet-controller assumes the config is changed if following conditions are met

```
		hasConfigChanged := config.APIServerURL != cluster.Status.APIServerURL ||
			hashStatusField(config.APIServerCA) != cluster.Status.APIServerCAHash ||
			config.AgentTLSMode != cluster.Status.AgentTLSMode ||
			hasGarbageCollectionIntervalChanged(config, cluster)
```

however, the status values are fetched from an related secret, and if they are empty, then fallback to a configmap

https://github.com/rancher/fleet/blob/1cddbbff1c2cd71c9b9011c3738754e5b4c8fa89/internal/cmd/controller/agentmanagement/controllers/cluster/import.go#L232

```
...
	logrus.Debugf("Cluster import for '%s/%s'. Setting up agent with kubeconfig from secret '%s/%s'", cluster.Namespace, cluster.Name, kubeConfigSecretNamespace, cluster.Spec.KubeConfigSecret)
	var (
		cfg          = config.Get()
		apiServerURL = string(secret.Data[config.APIServerURLKey])
		apiServerCA  = secret.Data[config.APIServerCAKey]
	)

	if apiServerURL == "" {
		if len(cfg.APIServerURL) == 0 {
			return status, fmt.Errorf("missing apiServerURL in fleet config for cluster auto registration")
		}
		logrus.Debugf("Cluster import for '%s/%s'. Using apiServerURL from fleet-controller config", cluster.Namespace, cluster.Name)
		apiServerURL = cfg.APIServerURL
	}

	if len(apiServerCA) == 0 {
		apiServerCA = cfg.APIServerCA
	}


the cluster.fleet status is updated from:

	status.AgentDeployedGeneration = &cluster.Spec.RedeployAgentGeneration
	status.AgentMigrated = true
	status.CattleNamespaceMigrated = true
	status.Agent = fleet.AgentStatus{
		Namespace: cluster.Spec.AgentNamespace,
	}
	status.AgentNamespaceMigrated = true
	status.AgentConfigChanged = false
[...]

PR #3551, opened 2025-04-14
Trigger cluster imports only from relevant config updates

weyfonk:
Fleet's agent management logic re-imports clusters based on configuration updates, which includes API server URL and CA.

However, those config fields are only used by clusters which do not source API server URL or CA from secrets.
Clusters which do use secrets instead must not be affected by changes to API server URL and CA in config.

This patch provides unit tests validating this new behaviour.

Refers to #3061


w13915984028 (approved):
LGTM, thanks for the fix.

manno (approved):
This matches https://fleet.rancher.io/ref-registration#secrets-during-agent-deployment

PR #3553, opened 2025-04-14
Removes the random string separator calculation.

0xavi0:
For being too complex for an issue that will possibly never occur and because the implementation fails when the strings in the bundle description have a suffix that matches the beginning of the random string generated.

It tries now all the possible characters, and if all of them are found in the bundle definitions, it returns an error.

I'll also add that to the documentation.

### DIFF
--- a/internal/cmd/controller/gitops/reconciler/gitjob_test.go
+++ b/internal/cmd/controller/gitops/reconciler/gitjob_test.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"os"
 	"slices"
-	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


internal/cmd/controller/gitops/reconciler/gitjob_test.go
```diff
@@ -2426,61 +2423,21 @@ func TestDrivenScanSeparator(t *testing.T) {
...
-								Base:    fmt.Sprintf("test?one:%stwo", separator),
-								Options: "options.yaml",
-							},
-							{
-								Base:    "test,one<>",
-								Options: "options2|.yaml",
-							},
-						},
-					},
+				expectedErrorMessage := fmt.Sprintf("bundle base and/or options paths contain all possible characters from %q, please update those paths to remedy this", bundleOptionsSeparatorChars)
```
weyfonk, line 2435: nit: I'm not sure we _need_ to check against the full error message, as that could make our tests brittle. Checking that an error occurs could be enough.

weyfonk (approved):

internal/cmd/controller/gitops/reconciler/gitjob_test.go
0xavi0: Oh, that was because as we're giving extra info to the user I thought it was good to check for the full error we return instead to just check that an error happened.

PR #3558, opened 2025-04-16
Delay GitRepo status reconciliations from bundle changes

aruiz14:
Refers to #XXX

### Checklist

### DIFF
--- a/internal/cmd/controller/gitops/reconciler/status_controller.go
+++ b/internal/cmd/controller/gitops/reconciler/status_controller.go
@@ -35,10 +35,11 @@ type StatusReconciler struct {
 func (r *StatusReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&fleet.GitRepo{}).
-		Watches(
+		WatchesRawSource(source.TypedKind(
 			// Fan out from bundle to gitrepo
+			mgr.GetCache(),
 			&fleet.Bundle{},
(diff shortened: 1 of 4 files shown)


manno (approved):

internal/cmd/controller/gitops/reconciler/status_controller.go
```diff
@@ -12,19 +13,23 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
+// enqueueDelay is used as an artificial delay for enqueuing GitRepo status reconciliation requests
+// This allows aggregating multiple consecutive Bundle update events, reducing the number of GitRepo status changes at the cost of introducing a delay in the notification
+const enqueueDelay = 3 * time.Second
```
weyfonk, line 31: Where does this value come from? I'm interested in the process behind determining it.
(not talking about making it configurable... yet ;) )

aruiz14: I must admit I didn't follow a very scientific approach to pick this value, which should be big enough to defeat its purpose but not so much to affect user experience.
I considered adding some kind of adaptive mechanism but that increased the complexity for not much benefit (IMO).

PR #3559, opened 2025-04-16
fleet cli json output

0xavi0:
Uses a JSONFormatter in fleet cli.
It also filters out the output received from fleet cli so the error set in the conditions is only the one sent by the fleet cli code and not from any used libraries.

In order to use the json formatter we need to set the `FLEET_JSON_OUTPUT` env var to `true`. This way the command line user won't get the json output unless it's explicitly specified.

Difference in the UI

Before the change:
Screenshot 2025-04-15 at 16 10 54

After the change:
Screenshot 2025-04-15 at 15 37 49

 
Refers to: https://github.com/rancher/fleet/issues/3234

### DIFF
--- a/cmd/fleetcli/main.go
+++ b/cmd/fleetcli/main.go
@@ -6,18 +6,17 @@ import (
 	"strings"
 
 	// Ensure GVKs are registered
-
 	_ "github.com/rancher/fleet/pkg/generated/controllers/fleet.cattle.io"
 	_ "github.com/rancher/wrangler/v3/pkg/generated/controllers/apiextensions.k8s.io"
 	_ "github.com/rancher/wrangler/v3/pkg/generated/controllers/apps"
 	_ "github.com/rancher/wrangler/v3/pkg/generated/controllers/core"
 	_ "github.com/rancher/wrangler/v3/pkg/generated/controllers/rbac"
-	"github.com/sirupsen/logrus"
 
 	// Add non-default auth providers
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
 	"github.com/rancher/wrangler/v3/pkg/signals"
+	"github.com/sirupsen/logrus"
 
 	cmds "github.com/rancher/fleet/internal/cmd/cli"
 )
--- a/cmd/fleetcli/main.go
+++ b/cmd/fleetcli/main.go
@@ -2,18 +2,22 @@
 package main
 
 import (
+	"os"
+	"strings"
+
 	// Ensure GVKs are registered
+
 	_ "github.com/rancher/fleet/pkg/generated/controllers/fleet.cattle.io"
 	_ "github.com/rancher/wrangler/v3/pkg/generated/controllers/apiextensions.k8s.io"
 	_ "github.com/rancher/wrangler/v3/pkg/generated/controllers/apps"
 	_ "github.com/rancher/wrangler/v3/pkg/generated/controllers/core"
 	_ "github.com/rancher/wrangler/v3/pkg/generated/controllers/rbac"
+	"github.com/sirupsen/logrus"
 
 	// Add non-default auth providers
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
 	"github.com/rancher/wrangler/v3/pkg/signals"
-	"github.com/sirupsen/logrus"
 
 	cmds "github.com/rancher/fleet/internal/cmd/cli"
 )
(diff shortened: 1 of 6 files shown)


cmd/fleetcli/main.go
```diff
@@ -3,6 +3,10 @@ package main
 
 import (
 	// Ensure GVKs are registered
```
aruiz14, line 5: Please keep this comment together with the _blank imports_.

integrationtests/gitjob/controller/controller_test.go
```diff
@@ -542,10 +542,15 @@ var _ = Describe("GitJob controller", func() {
...
+							// using Stalled because the Compute function uses Stalled
+							// for returning the condition message and it's simpler.
+							// For testing it in a different way we would need to setup a more complex
+							// scenario defining the job pods
+							// We are simulating job failures.
+							Type:    "Stalled",
 							Status:  "True",
 							Reason:  "BackoffLimitExceeded",
-							Message: "Job has reached the specified backoff limit",
+							Message: `{"fleetErrorMessage":"fleet error message","level":"fatal","msg":"Fleet cli failed","time":"2025-04-15T14:53:15+02:00"}`,
```
aruiz14, line 553: Beware that `message` in status conditions is by definition meant to be human-readable:
```
❯ kubectl explain job.status.conditions.message
GROUP:      batch
KIND:       Job
VERSION:    v1

FIELD: message <string>


DESCRIPTION:
    Human readable message indicating details about last transition.
```

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -1632,3 +1642,57 @@ func separatorInBundleDefinitions(gitrepo v1alpha1.GitRepo, sep rune) bool {
...
+}
+
+func getFleetCLIErrorsFromLine(l string) string {
+	type LogEntry struct {
+		Level         string `json:"level"`
+		FleetErrorMsg string `json:"fleetErrorMessage"`
+		Time          string `json:"time"`
+		Msg           string `json:"msg"`
+	}
+	s := ""
```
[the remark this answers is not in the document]
manno: I am not too worried about the performance and it's kind of a short func... but `s = s + getFleetCLIErrorsFromLine(l)` could be changed to `s += getFleetCLIErrorsFromLine(l)` if no builder is used.

integrationtests/gitjob/controller/controller_test.go
manno: I think for jobs it's okay and it allows much better errors in the UI? 

Additionally:
 
> JSON is a text-based, human-readable format for representing ...

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -1632,3 +1642,57 @@ func separatorInBundleDefinitions(gitrepo v1alpha1.GitRepo, sep rune) bool {
...
+			// ignore everything up to that closing bracked and try again
+			return getFleetCLIErrorsFromLine(l[close+1:])
+		}
+	} else if close == -1 {
+		// line does not contain a valid json string
+		return ""
+	}
+	var entry LogEntry
+	if err := json.Unmarshal([]byte(l[open:close+1]), &entry); err == nil {
+		s = s + entry.FleetErrorMsg + "\n"
```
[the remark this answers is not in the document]
0xavi0: `fleetErrorMessage` is set in here:
```go
if err := cmd.ExecuteContext(ctx); err != nil {
  if strings.ToLower(os.Getenv(cmds.JSONOutputEnvVar)) == "true" {
	log := logrus.New()
	log.SetFormatter(&logrus.JSONFormatter{})
	// use a fleet specific field name so we are sure logs from other libraries
	// are not considered.
	log.WithFields(logrus.Fields{
		"fleetErrorMessage": err.Error(),
	}).Fatal("Fleet cli failed")
```

So the only possible way for it to be an empty string is if the error was different from `nil` and the message was an empty string. 

In that case, I think it would be more of a problem with the code that generated the error.
Why would anyone want to return a non-nil error with an empty error message? :thinking:

Since it was something very quick to do, I added a check to cover that case (it will return `Unknown error`)

cmd/fleetcli/main.go
```diff
@@ -2,18 +2,22 @@
 package main
 
 import (
+	"os"
+	"strings"
+
 	// Ensure GVKs are registered
+
```
manno, line 9:
```suggestion
```

cmd/fleetcli/main.go
```diff
@@ -2,18 +2,22 @@
...
 	_ "github.com/rancher/fleet/pkg/generated/controllers/fleet.cattle.io"
 	_ "github.com/rancher/wrangler/v3/pkg/generated/controllers/apiextensions.k8s.io"
 	_ "github.com/rancher/wrangler/v3/pkg/generated/controllers/apps"
 	_ "github.com/rancher/wrangler/v3/pkg/generated/controllers/core"
 	_ "github.com/rancher/wrangler/v3/pkg/generated/controllers/rbac"
+	"github.com/sirupsen/logrus"
```
manno, line 15:
```suggestion
```

cmd/fleetcli/main.go
0xavi0: should be fixed now

cmd/fleetcli/main.go
0xavi0: should be fixed now

weyfonk (approved):

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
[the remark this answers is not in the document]
weyfonk: Thanks for checking!

PR #3566, opened 2025-04-22
Avoids assigning values to a nil helm values map when merging

0xavi0:


manno (approved):

0xavi0:
Closing after merging https://github.com/rancher/fleet/pull/3567

Issue #3161, opened 2024-12-19
Add more logging/metrics to GitOps


manno:
The introduction of events seems to have removed some log lines. We also need more metrics to troubleshoot large fleet deployments with a huge number of gitrepos.

- [ ] add INFO level logging for all events, like "new commit" found
- [ ] add metrics like
  * number of jobs created

sbulage:
## System Information

| Rancher Version | Fleet Version |
|--------|--------|
| v2.12-90de58cccdcc8254854a5517d73010cf638d1655-head | 107.0.0+up0.13.0-alpha.3 | 

### Metrics to be checked
- List of metrics found here: https://github.com/rancher/fleet/issues/3668#issue-3055783493
   ```bash
   gitjobs_created_success_total
   gitjobs_created_failure_total
   gitjob_duration_seconds
   gitrepo_fetch_latest_commit_success_total
   gitrepo_fetch_latest_commit_failure_total
   # bucket with _bucket _sum _count suffixes
   gitrepo_fetch_latest_commit_duration_seconds
   ```
### Below steps performed.
- Create 2-3 `GitRepo's`
- Wait for them to be ready
- Create another `GitRepo` with repository doesn't exists (for _failure_ total metrics)
- Go to `Services` and find
  ```
  - monitoring-gitjob
  - monitoring-fleet-controller
  ```
- In order to check above metrics, we have to `PORT-FORWARD` `monitoring-gitjob` `service` to `localhost` by using below commands.
- First download `kubconfig` file from cluster (where above services are available, in my case `local` cluster) to local machine.
  ```bash
  kubectl --kubeconfig <kubeconfig> -n cattle-fleet-system port-forward svc/monitoring-gitjob 8081:8081
  ```
- On local machine, using `curl` command to get the metrics and store in file `metrics_to_be_checked.txt`
  ```bash
  curl http://localhost:8081/metrics >metrics_to_be_checked.txt
  ```
- Check above metrics are available in `metrics_to_be_checked.txt` file

### See below video for metrics available in gitjob.

<details><summary>Part: 1 Setup local machine to get Metrics</summary>

https://github.com/user-attachments/assets/1d42543b-c5fd-4205-987e-de376eced64e

</details> 

<details><summary>Part: 2 Check fetched metrics</summary>

https://github.com/user-attachments/assets/e25f92f3-6237-4fb9-a9da-46eddb524647

</details>

PR #3649, opened 2025-05-06
Metrics for gitjob

p-se:
*Counters*

fleet_gitrepo_fetch_latest_commit_success_total
fleet_gitrepo_fetch_latest_commit_failure_total
fleet_gitjobs_created_success_total
fleet_gitjobs_created_failure_total

*Histograms*

fleet_gitjob_duration_seconds
fleet_gitrepo_fetch_latest_commit_duration_seconds

*Gauges*

fleet_gitjob_duration_seconds_gauge

Refers to #3161

### Checklist

### DIFF
--- a/internal/cmd/controller/gitops/reconciler/gitjob.go
+++ b/internal/cmd/controller/gitops/reconciler/gitjob.go
@@ -48,6 +48,7 @@ const (
 
 func (r *GitJobReconciler) createJobAndResources(ctx context.Context, gitrepo *v1alpha1.GitRepo, logger logr.Logger) error {
 	logger.V(1).Info("Creating Git job resources")
+
 	if err := r.createJobRBAC(ctx, gitrepo); err != nil {
 		return fmt.Errorf("failed to create RBAC resources for git job: %w", err)
 	}
@@ -60,6 +61,7 @@ func (r *GitJobReconciler) createJobAndResources(ctx context.Context, gitrepo *v
 	if err := r.createJob(ctx, gitrepo); err != nil {
 		return fmt.Errorf("error creating git job: %w", err)
 	}
+
 	r.Recorder.Event(gitrepo, fleetevent.Normal, "Created", "GitJob was created")
 	return nil
 }
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
+
 	fleet "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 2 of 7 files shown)


internal/metrics/metrics.go
```diff
@@ -209,3 +214,95 @@ func getStatusMetrics(subsystem string, labels []string) map[string]prometheus.C
 		),
 	}
 }
+
+// ObjCounter creates and registers a new CounterVec metric with the given name and help
+// text. The returned CounterVec embeds the CounterVec from the prometheus package and adds a method
+// to increment the counter for a given object. The labels of the metric are determined from the
+// name and the namespace of the given object.
+func ObjCounter(name, help string) (c ObjCounterVec) {
+	labels := []string{"name", "namespace"}
```
aruiz14, line 223: I'm worried about tracking objects individidually could turn this into a "high-cardinality" metrics on big scale systems.

pkg/git/fetch.go
```diff
@@ -15,6 +17,28 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+var (
+	fetchLatestCommitSuccess metrics.ObjCounterVec
+	fetchLatestCommitFailure metrics.ObjCounterVec
+	timeToFetchLatestCommit  metrics.ObjHistogramVec
+)
+
+func init() {
```
aruiz14, line 26: Why using `init()` instead of the variable initialization directly?

aruiz14 (changes requested):
I miss some test describing the requirements or the rationale behind these new metrics.
Could you please elaborate there a bit more?

internal/metrics/metrics.go
p-se: I don't see a problem with those two labels. It's basically "one" metric per GitRepo. Even if a customer had a thousand GitRepos, I don't think that this would cause any problems for Prometheus. I said "one" because the combination of both name and namespace is the identifier. The label values aren't arbitrary and do not change for the whole lifetime of the GitRepo. The metric itself denotes the type, e.g. `gitrepo`. I wouldn't consider that high cardinality. Also, this allows for insights as to which GitRepo might be causing problems, or at least how long it takes to download and process them, as opposed to having that information without the ability to connect that with any particular GitRepo.

Where do your concern come from?

p-se:
> I miss some test describing the requirements or the rationale behind these new metrics. Could you please elaborate there a bit more?

The purpose of adding those metrics is to have them, since we do have metrics for each fleet resource, in particular the status of those resources, but none about what the gitjob controller does. They should help in visualizing how long it takes to process a git repository to create a bundle and how often that happens, succeeds and fails. It also records the time it takes to query for new commits.

The resulting metrics can, for instance, be visualized like this:

Screenshot 2025-05-07 at 16-30-16 Fleet GitOps - Development Dashboard - Dashboards - Grafana

That said, I'm not sure what you mean by missing a test in the context of explaining the purpose. Can you please clarify?

internal/metrics/metrics.go
aruiz14: Whether its high-cardinality or not is actually very subjective and depend on a case-by-case basis. I'm just pointing out this could create an unexpected situation for users with a bigger amount of GitRepos.
I understand there is value being able to get per-object insights, but we could argue that Prometheus may not be the best choice for showing those stats. Part of this information may belong to the logs, while would offer an aggregated view of it?
At the very least, I think we should provide a way to configure this behavior (not sure what the default value should be).

aruiz14:
@p-se
> > I miss some test describing the requirements or the rationale behind these new metrics. Could you please elaborate there a bit more?

> That said, I'm not sure what you mean by missing a test in the context of explaining the purpose. Can you please clarify?

Oops, sorry for the typo, I meant "text" 😅

internal/cmd/controller/gitops/reconciler/gitjob.go
```diff
@@ -48,18 +48,25 @@ const (
 
 func (r *GitJobReconciler) createJobAndResources(ctx context.Context, gitrepo *v1alpha1.GitRepo, logger logr.Logger) error {
 	logger.V(1).Info("Creating Git job resources")
+
 	if err := r.createJobRBAC(ctx, gitrepo); err != nil {
+		gitjobsCreatedFailure.Inc(gitrepo)
```
aruiz14, line 53: Wouldn't be easier to record metrics in the caller instead? Same for the other case below.

internal/metrics/metrics.go
aruiz14: FTR, after talking about this offline, it turns out that these changes are consistent with the existent metrics, but we agreed it's not ideal. @p-se will create a separate issue to look into this from a broader perspective.

aruiz14 (commented):
@p-se what about GitRepos being deleted? Shouldn't we delete the metrics when receiving the `DELETED` event?

p-se:
> @p-se what about GitRepos being deleted? Shouldn't we delete the metrics when receiving the `DELETED` event?

Yes, but the metrics need to be treated differently on deletes. I've already started with the implementation but wanted to do that separate from this PR.

manno (approved):

PR #3671, opened 2025-05-12
Update GitHub Actions runners to use 'ubuntu-latest' 

thardeck:
because they seem at least similarly reliable as our own hosted runners at the moment.

| Task          | ubuntu-latest | self hosted |
|---------------|---------------|-------------|
| CI            | 15 minutes    | 13 minutes  |
| e2e k3s new   | 8 minutes     | 6 minutes   |
| e2e mc        | 9 minutes     | 6 minutes   |
| lint          | 3 minutes     | 3 minutes   |

So at this time ubuntu-latest was a bit slower but more reliable. In my opinion it is worth trying.

### DIFF
--- a/.github/workflows/check-changes.yml
+++ b/.github/workflows/check-changes.yml
@@ -10,7 +10,7 @@ env:
 
 jobs:
   check-changes:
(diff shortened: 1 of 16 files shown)


weyfonk (approved):
Let's try this :rocket:

PR #3677, opened 2025-05-12
Allow integration tests to be run in parallel

thardeck:
the second group automatically tests newly added directories to make sure all tests are executed.

But it might make sense to adapt the groups from time to time to have a similar run time.

### DIFF
--- a/.github/scripts/run-integration-tests-group2.sh
+++ b/.github/scripts/run-integration-tests-group2.sh
@@ -11,7 +11,7 @@ KUBEBUILDER_ASSETS=$(setup-envtest use --use-env -p path "$ENVTEST_K8S_VERSION")
 export KUBEBUILDER_ASSETS
 
 # Group 2: Run everything else - this will automatically include newly added directories
-find ./integrationtests -type d -not -path './integrationtests' -not -path '*/.git*' \
+find ./integrationtests -type d -not -path '*/.git*' \
   -not -path './integrationtests/agent*' \
   -not -path './integrationtests/bundlereader*' \
   -not -path './integrationtests/cli*' \


.github/scripts/run-integration-tests-group1.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+SETUP_ENVTEST_VER=${SETUP_ENVTEST_VER-v0.0.0-20221214170741-69f093833822}
+ENVTEST_K8S_VERSION=${ENVTEST_K8S_VERSION-1.25}
```
weyfonk, line 6: nit: could we perhaps take the opportunity to default to more recent versions of these tools?

.github/scripts/run-integration-tests-group2.sh
```diff
@@ -0,0 +1,19 @@
...
+SETUP_ENVTEST_VER=${SETUP_ENVTEST_VER-v0.0.0-20221214170741-69f093833822}
+ENVTEST_K8S_VERSION=${ENVTEST_K8S_VERSION-1.25}
+
+# install and prepare setup-envtest
+go install sigs.k8s.io/controller-runtime/tools/setup-envtest@"$SETUP_ENVTEST_VER"
+KUBEBUILDER_ASSETS=$(setup-envtest use --use-env -p path "$ENVTEST_K8S_VERSION")
+export KUBEBUILDER_ASSETS
+
+# Group 2: Run everything else - this will automatically include newly added directories
+find ./integrationtests -type d -not -path './integrationtests' -not -path '*/.git*' \
```
weyfonk, line 14: Why do we need ` -not -path './integrationtests'`? :thinking:

weyfonk (approved):
Nice one, I like the idea of being able to reshuffle test suites between groups.

Issue #3482, opened 2025-03-27
Bundle status too big for etcd


webD97:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

I have updated a fleet controller managing ~ 72 clusters to Fleet 0.11.5 and noticed a lot of reconciler errors in the fleet-controller pod logs that look like this:

```json
{
    "level": "error",
    "ts": "2025-03-27T10:43:27Z",
    "msg": "Reconciler error",
    "controller": "bundle",
    "controllerGroup": "fleet.cattle.io",
    "controllerKind": "Bundle",
    "Bundle": {
        "name": "xyz-kube-prometheus-stack",
        "namespace": "xyz"
    },
    "namespace": "xyz",
    "name": "xyz-kube-prometheus-stack",
    "reconcileID": "9eacefc5-fb1d-45cb-b5fe-2fc7de6bdb00",
    "error": "etcdserver: request is too large",
    "stacktrace": "sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.0/pkg/internal/controller/controller.go:316\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.0/pkg/internal/controller/controller.go:263\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.2\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.0/pkg/internal/controller/controller.go:224"
}
```

I then enabled debug logging and looked at the request body that fleet was trying to send to etcd and noticed that the new bundle status yaml has more than 30k lines because `.status.resourceKey` now contains an array with every single resource being deployed which in my case means "72 clusters * 100 resources = 7200 entries". This status alone has a size of 1MB. Combined with the rest of the resource, this exceeds etcd's limits. See this attachment: [status.resourceKey.yaml.txt](https://github.com/user-attachments/files/19484796/status.resourceKey.yaml.txt)

These are pretty rough numbers and I know that I might have even missed quite some lines from the log while trying to extract the request.

Previously, we were running a 0.9.x version which only had 100 resources in the status. 

[...]

PR #3681, opened 2025-05-13
Remove bundle.ResourceKey computation

manno:
Refers to #3482

### DIFF
--- a/internal/cmd/controller/gitops/reconciler/status_controller.go
+++ b/internal/cmd/controller/gitops/reconciler/status_controller.go
@@ -161,6 +161,7 @@ func (r *StatusReconciler) updateStatus(ctx context.Context, orig *fleet.GitRepo
 }
 
 func setStatus(list *fleet.BundleDeploymentList, gitrepo *fleet.GitRepo) error {
+	// sort bundledeployments so lists in status are always in the same order
 	sort.Slice(list.Items, func(i, j int) bool {
 		return list.Items[i].UID < list.Items[j].UID
 	})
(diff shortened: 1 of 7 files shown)


internal/cmd/controller/gitops/reconciler/status_controller.go
```diff
@@ -161,7 +161,6 @@ func (r *StatusReconciler) updateStatus(ctx context.Context, orig *fleet.GitRepo
 }
 
 func setStatus(list *fleet.BundleDeploymentList, gitrepo *fleet.GitRepo) error {
-	// sort for resourceKey?
```
aruiz14, line 164: I think these comments actually meant `namespace/name` rather than _those_ `ResourceKey`s
(e.g. if they would make it a more comprehensive list for users)

pkg/apis/fleet.cattle.io/v1alpha1/bundle_types.go
```diff
@@ -369,7 +369,7 @@ type BundleStatus struct {
 	Display BundleDisplay `json:"display,omitempty"`
 	// ResourceKey lists resources, which will likely be deployed. The
 	// actual list of resources on a cluster might differ, depending on the
-	// helm chart, value templating, etc..
+	// helm chart, value templating, etc.. (deprecated to reduce bundle size)
 	// +nullable
 	ResourceKey []ResourceKey `json:"resourceKey,omitempty"`
```
aruiz14, line 374: Well we keep it in 0.13, then drop it in 0.14?

manno: Good idea!

internal/cmd/controller/gitops/reconciler/status_controller.go
manno: Replacing with a comment that explains why we sort.

aruiz14 (approved):

PR #3682, opened 2025-05-13
Simplify GitRepo end-to-end tests

weyfonk:
This reduces the risk of flakiness in single-cluster end-to-end tests for `GitRepo`, by:
* refactoring test cases to use separate `When` blocks from Ginkgo, instead of a loop with `Skip` statements, to ease tracking of which test cases run and which do not
* removing the check on logs from `AfterEach`, which could cause failures. That check was added back when our reconcilers were not as mature as they are now. Such a check would anyway not belong there, but rather in integration tests covering bundle deployment cleanup.

### DIFF
--- a/e2e/single-cluster/gitrepo_test.go
+++ b/e2e/single-cluster/gitrepo_test.go
@@ -32,7 +32,7 @@ const (
 	repoName  = "repo"
 )
 
-var _ = FDescribe("Monitoring Git repos via HTTP for change", Label("infra-setup"), func() {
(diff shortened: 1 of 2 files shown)


manno (changes requested):
Looks like `repoName` exists as a variable and as a constant.

weyfonk:
> Looks like `repoName` exists as a variable and as a constant.

Fixed by moving the constant to suite level, as it is used by multiple test drivers, and by renaming the variable to make the override explicit.

manno (approved):

PR #3706, opened 2025-05-19
[v0.10] Add known-hosts update script

weyfonk:
The script is required to enable updatecli to compute the latest entries for the `known-hosts` config map after checking out this branch.

Backport of the script from #3680.


weyfonk:
Not needed anymore as per [this comment](https://github.com/rancher/fleet/pull/3709#discussion_r2095524498).

PR #3709, opened 2025-05-19
Update `known-hosts` config map in all release branches

weyfonk:
Follow-up to #3680, expanded to cover all currently maintained release branches.

Any new release branches created from `main` would already contain that script; expanding updates to a new release branch would simply require adding a reference to that branch to `updatecli/values.d/scm.yaml`.

Script added to currently maintained release branches through #3706, #3707 and #3708.


updatecli/values.d/scm.yaml
```diff
@@ -1,8 +1,29 @@
...
+    owner: rancher
+    repository: fleet
+    branch: 'release/v0.11'
+  v10:
+    enabled: true
+    user: 'fleet-bot'
+    email: 'fleet@suse.de'
+    owner: rancher
+    repository: fleet
+    branch: 'release/v0.10'
```
aruiz14, line 29: You could use a YAML anchors to avoid repeating these (provided that updatecli's YAML parser supports it):
```suggestion
scms:
  main: &main
    enabled: true
    user: 'fleet-bot'
    email: 'fleet@suse.de'
    owner: rancher
    repository: fleet
    branch: main
  v12:
    <<: *main
    branch: 'release/v0.12'
  v11:
    <<: *main
    branch: 'release/v0.11'
  v10:
    <<: *main
    branch: 'release/v0.10'
```

updatecli/updatecli.d/known-hosts.yaml
```diff
@@ -28,15 +28,18 @@ targets:
         spec:
           files:
             - charts/fleet/templates/configmap_known_hosts.yaml
+      # TODO ensure this script is available on each branch
       command: bash .github/scripts/update_known_hosts_configmap.sh
```
aruiz14, line 32: Since this is a very simple use case, and only one script is used, you could make this run the script from the `main` branch by using a small git trick: `git show <branch>:</path/to/file>` will retrieve the output of the file as it's in the provided branch.
So maybe something like this could work?
```suggestion
      command: bash <(git show main:.github/scripts/update_known_hosts_configmap.sh)
```

updatecli/values.d/scm.yaml
weyfonk: Nice! I've tried this, and a local `updatecli diff` run suggests that updatecli supports it :rocket:

updatecli/updatecli.d/known-hosts.yaml
weyfonk: That's much simpler than pushing the script to separate branches indeed. Thanks, applied.

aruiz14 (approved):

PR #3721, opened 2025-05-26
Refactor fleet apply and enable concurrent bundle scanning

aruiz14:
Refers to #3742

The goal of this PR is to detach the processing of every individual `Bundle` from the directory scanning.
Using [errgroup](https://pkg.go.dev/golang.org/x/sync/errgroup#Group) provides synchronization, error propagation, and Context cancelation over just goroutines.

I took the chance to refactor some part of the code (basically splitting `Dir` between `Bundle` "creation" and "writing"), so it was easier to implement. 

## Additional Information

Some tests relied on a specific order, which is no longer guaranteed when processing bundles concurrently, so they have been adapted (I placed those commits at the bottom and pushed them early to ensure they continue to pass without my code changes).

### Checklist

- N/A -  I have updated the documentation via a pull request in the

### DIFF
--- a/internal/cmd/cli/apply/apply.go
+++ b/internal/cmd/cli/apply/apply.go
@@ -106,7 +106,7 @@ func CreateBundles(pctx context.Context, client client.Client, repoName string, 
 	// Using an errgroup to manage concurrency
 	// 1. Goroutines will be launched, honouring the concurrency limit, and eventually block trying to write to `bundlesChan`.
 	// 2. The main function will read from `bundlesChan`, hence unblocking the goroutines. This will continue to read from `bundlesChan` until it is closed.
-	// 3. We use another goroutine to wait for all goroutines to finish, then close `bundleChan`, finally unblocking the main function.
+	// 3. We use another goroutine to wait for all goroutines to finish, then close `bundlesChan`, finally unblocking the main function.
 	bundlesChan := make(chan *fleet.Bundle)
 	eg, ctx := errgroup.WithContext(pctx)
 	eg.SetLimit(bundleCreationMaxConcurrency + 1) // extra goroutine for WalkDir loop
@@ -208,7 +208,7 @@ func CreateBundlesDriven(pctx context.Context, client client.Client, repoName st
 	// Using an errgroup to manage concurrency
 	// 1. Goroutines will be launched, honouring the concurrency limit, and eventually block trying to write to `bundlesChan`.
 	// 2. The main function will read from `bundlesChan`, hence unblocking the goroutines. This will continue to read from `bundlesChan` until it is closed.
-	// 3. We use another goroutine to wait for all goroutines to finish, then close `bundleChan`, finally unblocking the main function.
+	// 3. We use another goroutine to wait for all goroutines to finish, then close `bundlesChan`, finally unblocking the main function.
 	bundlesChan := make(chan *fleet.Bundle)
 	eg, ctx := errgroup.WithContext(pctx)
 	eg.SetLimit(bundleCreationMaxConcurrency + 1) // extra goroutine for WalkDir loop


internal/cmd/cli/apply/apply.go
```diff
@@ -177,32 +197,55 @@ func CreateBundles(ctx context.Context, client client.Client, repoName string, b
...
+			case <-ctx.Done():
+				return ctx.Err()
+			case bundlesChan <- bundle:
+			}
+			return writeBundle(ctx, client, bundle, scans, opts)
+		})
 	}
+	go func() {
+		_ = eg.Wait()
+		close(bundlesChan)
```
0xavi0, line 248: Couldn't we just return in here if there's any error? 
The second `eg.Wait()` would not be required, then. 

Does the second `Wait` really wait? (reaching that code depends on closing the channel, which is done when the first `Wait` finished)

internal/cmd/cli/apply/apply.go
```diff
@@ -177,32 +197,55 @@ func CreateBundles(ctx context.Context, client client.Client, repoName string, b
...
+func CreateBundlesDriven(pctx context.Context, client client.Client, repoName string, baseDirs []string, opts Options) error {
 	if len(baseDirs) == 0 {
 		baseDirs = []string{"."}
 	}
 
-	foundBundle := false
-	gitRepoBundlesMap := make(map[string]bool)
+	eg, ctx := errgroup.WithContext(pctx)
+	eg.SetLimit(bundleCreationMaxConcurrency)
+	bundlesChan := make(chan *fleet.Bundle)
```
0xavi0: Why don't we have the outer routine?
Without it... we could have deadlocks, no?

The `CreateBundles` function has the outer routine that would avoid that.

aruiz14: Good catch, it wasn't necessary when there was no concurrency limit, but after I added it could indeed create a deadlock.

internal/cmd/cli/apply/apply.go
aruiz14: Returning from a goroutine will not make the main function to return as well.

I added a comment to `CreateBundles` trying to explain how concurrency is managed here, but I acknowledge it's not straightforward:
1. Many goroutines will be launched, which will all block trying to write into `bundlesChan` until it's being read.
2. The main function will start reading from `bundlesChan`, hence unblocking the goroutines. This will block until `bundlesChan` is closed.
3. This goroutine will wait for all goroutines to be finished, then close `bundleChan`, hence unblocking the main function.

The second `eg.Wait()` recovers any error that could happen in the ErrorGroup, without necessarily waiting.

0xavi0: yeah, but in case of error... the first `eq.Wait()` will have exactly the same error as the second would get, no?
If so... why can't we just return the error after closing the channel?

Calling `Wait` 2 times makes the second call get any different result?

aruiz14: Yes, the error will be the same, but the goroutine is in a different scope from the main function, so it's not possible to `return` from there. That's why the `eg.Wait` is needed in the main thread, despite knowing it will not actually wait.

0xavi0: oh, yeah :). Thanks!

internal/cmd/cli/apply/apply.go
```diff
@@ -253,43 +308,46 @@ func pruneBundlesNotFoundInRepo(ctx context.Context, c client.Client, repoName,
...
+		}
+	}
+	bundle.Namespace = opts.Namespace
+	return bundle, scans, nil
 }
 
 // Dir reads a bundle and image scans from a directory and writes runtime objects to the selected output.
 //
 // name: the gitrepo name, passed to 'fleet apply' on the cli
 // basedir: the path from the walk func in Dir, []baseDirs
```
0xavi0: We need to change the documentation for the `Dir` function as it no longer "writes" any resources.
(It fact it is now a wrapper around `newBundle` that does not much else) 
We could, maybe, move the little it does to `newBundle` that also has a more meaningful name than `Dir` (that name always confused me)
What do you think?

integrationtests/cli/apply/apply_test.go
```diff
@@ -581,12 +607,22 @@ var _ = Describe("Fleet apply with helm charts with dependencies", Ordered, func
...
-			Expect(bundle).To(HaveLen(3))
-			remoteDepl := bundle[0]
-			simpleDepl := bundle[1]
-			noDepsDepl := bundle[2]
+			Expect(bundles).To(HaveLen(3))
+			remoteDepl, _, ok := sliceFind(bundles, func(bundle *v1alpha1.Bundle) bool {
+				return bundle.Spec.Helm.ReleaseName == "remote-chart-with-deps"
+			})
+			Expect(ok).To(BeTrue())
+			//remoteDepl := bundles[0]
```
0xavi0: nit: Forgotten commented line

integrationtests/cli/apply/helm_test.go
```diff
@@ -55,12 +55,11 @@ var _ = Describe("Fleet apply helm release", Serial, func() {
 			})
 		})
 
-		When("path credentials are provided for both root and sub folder", func() {
+		When("path credentials are provided for sub folder", func() {
 			It("fleet apply works fine", func() {
 				Eventually(func() error {
 					return fleetApply("helm", []string{cli.AssetsPath + "helm_path_credentials"}, apply.Options{
 						AuthByPath: map[string]bundlereader.Auth{
-							cli.AssetsPath + "helm_path_credentials":           {Username: username, Password: password},
```
0xavi0: Question: why do we need to delete this? Was the test incorrect?

internal/cmd/cli/apply/apply.go
aruiz14: I was intentionally avoiding to rename, since it was an exported function. But since it's not really public (it's under `internal/`) and I'm changing the signature anyways, I don't see why not changing it.
From my POV, it still provides some value over `newBundle`, such as bundleID generation and treating an empty bundle as an error.

integrationtests/cli/apply/apply_test.go
aruiz14: Dead code shall not pass, thanks!

integrationtests/cli/apply/helm_test.go
aruiz14: I can't remember well... but looking at it now indeed doesn't seem related, let me try reverting it.

aruiz14: It worked! 😅 so it must be a leftover of trying to fix some tests, sorry for the noise!

0xavi0 (approved):

internal/cmd/cli/apply/apply.go
```diff
@@ -95,64 +94,89 @@ func globDirs(baseDir string) (result []string, err error) {
...
-		}
-		for _, baseDir := range matches {
-			if i > 0 && opts.Output != nil {
-				if _, err := opts.Output.Write([]byte("\n---\n")); err != nil {
-					return fmt.Errorf("writing to bundle output: %w", err)
-				}
+	// Using an errgroup to manage concurrency
+	// 1. Goroutines will be launched, honouring the concurrency limit, and eventually block trying to write to `bundlesChan`.
+	// 2. The main function will read from `bundlesChan`, hence unblocking the goroutines. This will continue to read from `bundlesChan` until it is closed.
+	// 3. We use another goroutine to wait for all goroutines to finish, then close `bundleChan`, finally unblocking the main function.
```
weyfonk: nit (found it by chance):
```suggestion
	// 3. We use another goroutine to wait for all goroutines to finish, then close `bundlesChan`, finally unblocking the main function.
```
Same remark about L212.

internal/cmd/cli/apply/apply.go
```diff
@@ -95,64 +94,89 @@ func globDirs(baseDir string) (result []string, err error) {
...
-		if err != nil {
-			return fmt.Errorf("invalid path glob %s: %w", baseDir, err)
-		}
-		for _, baseDir := range matches {
-			if i > 0 && opts.Output != nil {
-				if _, err := opts.Output.Write([]byte("\n---\n")); err != nil {
-					return fmt.Errorf("writing to bundle output: %w", err)
-				}
+	// Using an errgroup to manage concurrency
+	// 1. Goroutines will be launched, honouring the concurrency limit, and eventually block trying to write to `bundlesChan`.
```
weyfonk, line 107: Does this blocking _need_ to happen? I mean, we could use a buffered channel to prevent writes from blocking until the buffer is full. Since we know how many bundles (at most) we'd like to create (`len(baseDirs)`, if I'm not mistaken), we could dimension that buffer accordingly. Or am I missing something? Is there a concern on memory usage for large bundles?
Anyway, this could be an enhancement idea for a future iteration.

aruiz14: Please don't be confused by the wording in the explanation, which I intended to be as detailed as possible.
In practice, every goroutine will only block during a negligible period of time, as the main function starts consuming the channel immediately (and every item does not incur into any heavy processing either).
Using a buffered channel won't provide any substantial improvement (if any), while still incurring into the "complexity" of managing a channel.

internal/cmd/cli/apply/apply.go
```diff
@@ -94,64 +93,89 @@ func globDirs(baseDir string) (result []string, err error) {
...
+	// 2. The main function will read from `bundlesChan`, hence unblocking the goroutines. This will continue to read from `bundlesChan` until it is closed.
+	// 3. We use another goroutine to wait for all goroutines to finish, then close `bundleChans`, finally unblocking the main function.
```
weyfonk: nit: it is still `bundlesChan` :D
```suggestion
	// 3. We use another goroutine to wait for all goroutines to finish, then close `bundlesChan`, finally unblocking the main function.
```

weyfonk (approved):

internal/cmd/cli/apply/apply.go
aruiz14: Oh man 🤦 
https://github.com/rancher/fleet/commit/5722fa8b8701e89788ac0a802faeb5a469f7fa15

Issue #3394, opened 2025-02-26
Discussion on GitOps poller factory vs requeueAfter


manno:

weyfonk:
Related to https://github.com/rancher/fleet/issues/3746.

PR #3734, opened 2025-05-28
Support k8s 1.33

thardeck:
Refers to https://github.com/rancher/rancher/issues/50450

Obsoletes #3729

### DIFF
--- a/.github/workflows/e2e-ci.yml
+++ b/.github/workflows/e2e-ci.yml
@@ -23,7 +23,7 @@ jobs:
           # k3d version list k3s | sed 's/+/-/' | sort -h
           # https://hub.docker.com/r/rancher/k3s/tags
           - name: k3s-new
-            version: v1.32.1-k3s1
+            version: v1.33.1-k3s1
           - name: k3s-old
             version: v1.30.9-k3s1
         test_type:
--- a/.github/workflows/e2e-fleet-upgrade-ci.yml
+++ b/.github/workflows/e2e-fleet-upgrade-ci.yml
@@ -19,7 +19,7 @@ jobs:
       matrix:
         k3s:
           - name: k3s-new
-            version: v1.32.1-k3s1
+            version: v1.33.1-k3s1
     name: fleet-upgrade-test-${{ matrix.k3s.name }}
 
     steps:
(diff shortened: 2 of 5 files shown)


go.mod
```diff
@@ -52,20 +52,20 @@ require (
...
-	k8s.io/kubectl v0.32.5
-	k8s.io/kubernetes v1.32.2
+	k8s.io/kubectl v0.33.1
+	k8s.io/kubernetes v1.33.1
 	k8s.io/utils v0.0.0-20250502105355-0f33e8f1c979
 	oras.land/oras-go/v2 v2.6.0
 	sigs.k8s.io/cli-utils v0.37.2
-	sigs.k8s.io/controller-runtime v0.20.4
-	sigs.k8s.io/controller-tools v0.17.3
+	sigs.k8s.io/controller-runtime v0.21.0
```
weyfonk, line 67: A few words or warning on this:
* v0.21.0 [deprecates `Result.Requeue`](https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.21.0), which we use extensively
* When quickly scanning changelogs between these 2 versions, I have not found anything suggesting that the implementation detail on which we rely (namely events with a higher timestamp being discarded from the queue if events with a lower timestamp are added for the same object) might have been altered, but this may need a closer look.
@0xavi0, do you know if our tests would detect such a change?

Also, [v0.20.0](https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.20.0) introduces a priority queue :eyes:

thardeck: Should I create an issue for fixing the deprecations?

weyfonk: Yes, that would be a good idea; it could target 2.13.0 for instance.

thardeck: https://github.com/rancher/fleet/issues/3746

weyfonk: Resolving following offline discussion. We've concluded that, should the `controller-runtime` implementation detail on which we rely be undone, this would result in more reconcile cycles, but not break anything.
We're still keeping an eye on this though, and are planning to rewrite the `GitRepo` polling mechanism to use more explicit polling; see #3394.

PR #3744, opened 2025-05-30
Add agent management label to skip clusters

manno:
This allows experimenting with different agent management models.

### DIFF
--- a/pkg/apis/fleet.cattle.io/v1alpha1/cluster_types.go
+++ b/pkg/apis/fleet.cattle.io/v1alpha1/cluster_types.go
@@ -42,6 +42,9 @@ var (
 	// ClusterLabel is used on a bundledeployment to refer to the targeted
 	// cluster
 	ClusterLabel = "fleet.cattle.io/cluster"
+
+	// ClusterManagementLabel can be used to specify a custom cluster manager
+	ClusterManagementLabel = "fleet.cattle.io/cluster-management"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/apis/fleet.cattle.io/v1alpha1/cluster_types.go
```diff
@@ -42,6 +42,9 @@ var (
 	// ClusterLabel is used on a bundledeployment to refer to the targeted
 	// cluster
 	ClusterLabel = "fleet.cattle.io/cluster"
+
+	// ClusterManagementLabel can be used to specify a custom cluster manager
```
weyfonk, line 46: Should we clarify that any non-empty value will lead to the current agent management logic skipping the cluster?
e.g.:
```suggestion
	// ClusterManagementLabel can be used to specify a custom cluster manager.
	// A non-empty value on a cluster means that agent management should skip this cluster.
	// This is useful for testing alternative agent management models.
```

manno: I think that is too specific and we would need to update the comment for each manager we add?

weyfonk: I see; we may want to later implement managers which behave differently based on different (sets of) values for this label. Nevermind then.

weyfonk (approved):

PR #3745, opened 2025-05-30
Implements initial OCI storage garbage collection

0xavi0:
This PR implements basic OCI storage garbage collection.

It deletes OCI artifacts either when the Bundle is deleted, when the contentID for the Bundle changes (deleting the previous one) and also when a Bundle was deployed using OCI before and it's now deployed using the default etcd storage.

Moves (and renames) the secret name's constant to the config package and changes `types.NamespacedName` to `client.ObjectKey` in the ocistorage secret functions to be inline with the client's operations.

It also introduces changes to the infra, adding a new OCI registry user with read-only permissions. This allows to test the agentSecret and also to write an extra test to test errors when using a user with not enough permissions.

Follow up: (not for 2.12)- Add a CronJob that checks for orphan OCI artifacts and purges them.

Refers to: https://github.com/rancher/fleet/issues/3749

### Checklist

### DIFF
--- a/e2e/single-cluster/oci_registry_test.go
+++ b/e2e/single-cluster/oci_registry_test.go
@@ -389,8 +389,7 @@ var _ = Describe("Single Cluster Deployments using OCI registry", Label("oci-reg
 					verifySecretsAreEqual(config.DefaultOCIStorageSecretName, env.Namespace, contentsID, downstreamNamespace)
 				})
 				By("not creating a contents resource for this chart", func() {
-					var content fleet.Content
-					k8sclient.ObjectShouldNotExist(clientUpstream, contentsID, "", &content, false)
+					k8sclient.ObjectShouldNotExist(clientUpstream, contentsID, "", &fleet.Content{}, false)
 				})
 				By("deploying the helm chart", func() {
 					var cm corev1.ConfigMap
@@ -443,8 +442,7 @@ var _ = Describe("Single Cluster Deployments using OCI registry", Label("oci-reg
 					verifySecretsAreEqual(deploySpecificSecretName, env.Namespace, contentsID, downstreamNamespace)
 				})
 				By("not creating a contents resource for this chart", func() {
-					var content fleet.Content
-					k8sclient.ObjectShouldNotExist(clientUpstream, contentsID, "", &content, false)
+					k8sclient.ObjectShouldNotExist(clientUpstream, contentsID, "", &fleet.Content{}, false)
 				})
 				By("deploying the helm chart", func() {
 					var cm corev1.ConfigMap
@@ -491,12 +489,10 @@ var _ = Describe("Single Cluster Deployments using OCI registry", Label("oci-reg
 					k8sclient.GetObjectShouldSucceed(clientUpstream, contentsID, "", &content)
 				})
 				By("not creating a bundle secret", func() {
-					var secret corev1.Secret
-					k8sclient.ObjectShouldNotExist(clientUpstream, contentsID, env.Namespace, &secret, false)
+					k8sclient.ObjectShouldNotExist(clientUpstream, contentsID, env.Namespace, &corev1.Secret{}, false)
 				})
 				By("not creating a bundle deployment secret", func() {
-					var secret corev1.Secret
-					k8sclient.ObjectShouldNotExist(clientUpstream, contentsID, downstreamNamespace, &secret, false)
+					k8sclient.ObjectShouldNotExist(clientUpstream, contentsID, downstreamNamespace, &corev1.Secret{}, false)
 				})
 				By("deploying the helm chart", func() {
 					var cm corev1.ConfigMap
@@ -537,8 +533,7 @@ var _ = Describe("Single Cluster Deployments using OCI registry", Label("oci-reg
 					}).Should(Succeed())
 				})
 				By("not creating the bundle", func() {
-					var bundle fleet.Bundle
-					k8sclient.ObjectShouldNotExist(clientUpstream, "sample-simple-chart-oci", env.Namespace, &bundle, true)
+					k8sclient.ObjectShouldNotExist(clientUpstream, "sample-simple-chart-oci", env.Namespace, &fleet.Bundle{}, true)
 				})
 				By("not creating the bundle secret", func() {
 					var secrets corev1.SecretList
@@ -583,8 +578,7 @@ var _ = Describe("Single Cluster Deployments using OCI registry", Label("oci-reg
 					}).Should(Succeed())
 				})
 				By("not creating the bundle", func() {
-					var bundle fleet.Bundle
-					k8sclient.ObjectShouldNotExist(clientUpstream, "sample-simple-chart-oci", env.Namespace, &bundle, true)
+					k8sclient.ObjectShouldNotExist(clientUpstream, "sample-simple-chart-oci", env.Namespace, &fleet.Bundle{}, true)
 				})
 				By("not creating the bundle secret", func() {
 					var secrets corev1.SecretList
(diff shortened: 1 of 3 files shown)


internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -432,6 +437,11 @@ func (r *BundleReconciler) createBundleDeployment(
 			}
 		}
 
+		// check if there's any OCI secret that can be purged
+		if err := maybePurgeOCIReferenceSecret(ctx, r.Client, bd, updated); err != nil {
+			logger.Error(err, "Reconcile failed to purge old oci reference secret")
```
```suggestion
			logger.Error(err, "Reconcile failed to purge old OCI reference secret")
```

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -556,6 +566,29 @@ func (r *BundleReconciler) cloneSecret(
 	return nil
 }
 
+func maybePurgeOCIReferenceSecret(ctx context.Context, c client.Client, old, new *fleet.BundleDeployment) error {
+	if !old.Spec.OCIContents || old.Spec.DeploymentID == "" {
+		return nil
+	}
+
+	if !new.Spec.OCIContents || (old.Spec.DeploymentID != new.Spec.DeploymentID) {
```
weyfonk, line 574: Why would we need to delete the secret every time the deployment ID changes?
Wouldn't credentials to the OCI registry potentially remain unchanged then?
Or is this an approximation to delete and recreate the same secret rather than risk keeping outdated secrets in namespaces of downstream clusters, which would lead to failing deployments when the upstream/parent OCI secret, referenced by the user, does change?

internal/cmd/cli/apply/apply.go
```diff
@@ -469,6 +480,17 @@ func saveOCIBundle(ctx context.Context, c client.Client, bundle *fleet.Bundle, o
 			return fmt.Errorf("a helmOps bundle with name %q already exists", bundle.Name)
 		}
 
+		// If the actual manifestID is different from the previous one,
```
weyfonk: nit: did you mean this?
```suggestion
		// If the current manifestID is different from the previous one,
```

internal/cmd/cli/apply/apply.go
```diff
@@ -424,6 +423,18 @@ func save(ctx context.Context, c client.Client, bundle *fleet.Bundle) (*fleet.Bu
 			return fmt.Errorf("a helmOps bundle with name %q already exists", bundle.Name)
 		}
 
+		if bundle.Spec.ContentsID != "" {
+			// this bundle was previously deployed to an OCI registry.
+			// Delete the OCI artifact as it's no longer required.
+			if err := deleteOCIManifest(ctx, c, bundle, ocistorage.OCIOpts{}); err != nil {
+				// we log the error and continue, since the OCI registry is an entity to the the cluster
```
weyfonk: What's the intended message here? I don't quite understand what "entity to the cluster" expresses. Missing word maybe?
```suggestion
				// we log the error and continue, since the OCI registry is an entity to the cluster
```

internal/cmd/cli/apply/apply.go
```diff
@@ -469,6 +480,17 @@ func saveOCIBundle(ctx context.Context, c client.Client, bundle *fleet.Bundle, o
 			return fmt.Errorf("a helmOps bundle with name %q already exists", bundle.Name)
 		}
 
+		// If the actual manifestID is different from the previous one,
+		// delete the previous OCI artifact
+		if bundle.Spec.ContentsID != "" && bundle.Spec.ContentsID != manifestID {
+			if err := deleteOCIManifest(ctx, c, bundle, opts); err != nil {
+				// we log the error and continue, since the OCI registry is an entity to the the cluster
```
weyfonk: Same nitpick + question as on L430.

internal/cmd/cli/apply/apply.go
```diff
@@ -497,6 +519,35 @@ func saveOCIBundle(ctx context.Context, c client.Client, bundle *fleet.Bundle, o
...
+			return err
+		}
+	}
+	if err := ocistorage.NewOCIWrapper().DeleteManifest(ctx, opts, bundle.Spec.ContentsID); err != nil {
+		return err
+	}
+
+	// also delete the bundle secret as it's no longer needed
+	var secret corev1.Secret
+	if err := c.Get(ctx, secretID, &secret); err != nil {
```
weyfonk: Out of curiosity: since we already have the secret ID, do we need to get it again to be able to delete it? Or could we just build a secret in memory with the same metadata (name and namespace), and use that to delete the one in the cluster?

e2e/single-cluster/oci_registry_test.go
```diff
@@ -654,4 +669,142 @@ var _ = Describe("Single Cluster Deployments using OCI registry", Label("oci-reg
...
+					Eventually(func(g Gomega) {
+						k8sclient.GetObjectShouldSucceed(clientUpstream, "sample-simple-chart-oci", env.Namespace, &bundle2)
+						g.Expect(bundle2.Spec.ContentsID).To(ContainSubstring("s-"))
+						g.Expect(previousContentsID).ToNot(Equal((bundle2.Spec.ContentsID)))
+						contentsID = bundle2.Spec.ContentsID
+					}).Should(Succeed())
+				})
+				By("checking that the previous oci bundle key was deleted", func() {
+					var secret corev1.Secret
+					k8sclient.ObjectShouldNotExist(clientUpstream, previousContentsID, env.Namespace, &secret, false)
```
weyfonk: nit: since `secret` does not need to be used, this could probably be shortened as:
```suggestion
					k8sclient.ObjectShouldNotExist(clientUpstream, previousContentsID, env.Namespace, &corev1.Secret{}, false)
```

weyfonk (commented):
Nice one! Leaving one doubt about the conditions for deleting old OCI access secrets, the rest is just nitpicks.

internal/cmd/controller/reconciler/bundle_controller.go
0xavi0: The secret name at Bundle level depends on the DeploymentID. 
If the deployment id changed it means the OCI artifact has changed (the deployment ID is also a hash of the contents).
The OCI artifact is also named depending on the DeploymentID.

If the OCI artifact has changed the previous secret will no longer be used, as the OCI artifact won't be used anymore.

internal/cmd/cli/apply/apply.go
0xavi0: Missing "external". Fixed.

internal/cmd/cli/apply/apply.go
0xavi0: Fixed (missing word)

internal/cmd/cli/apply/apply.go
0xavi0: No, technically we don't need to `Get` it.
Using a secret in memory.

e2e/single-cluster/oci_registry_test.go
0xavi0: Changed for all `ObjectShouldNotExist` calls in that file.

internal/cmd/cli/apply/apply.go
weyfonk: Thanks! My understanding is that the OCI registry _may_ live in the cluster, but I see what you mean: these failures should not block further processing.

0xavi0: we could also use an external one (dockerHub, for example).
But yes... it is something that Fleet has no control...

internal/cmd/controller/reconciler/bundle_controller.go
weyfonk: Got it, thanks for clarifying :)

weyfonk (approved):

PR #3775, opened 2025-06-06
Oci storage label and event

0xavi0:
PLEASE DON'T REVIEW AS THIS PR DEPENDS ON ANOTHER ONE STILL UNDER REVIEW

### Checklist


internal/cmd/cli/apply/apply.go
```diff
@@ -679,3 +683,18 @@ func deleteSecretIfExists(ctx context.Context, c client.Client, name, ns string)
 
 	return nil
 }
+
+func sendWarningEvent(r record.EventRecorder, namespace, artifactID string, errorToLog error) {
+	jobName := os.Getenv("JOB_NAME")
```
weyfonk: Why not use the constant with that value?
```suggestion
	jobName := os.Getenv(JobNameEnvVar)
```

internal/cmd/cli/apply/apply.go
```diff
@@ -736,3 +741,18 @@ func deleteSecretIfExists(ctx context.Context, c client.Client, name, ns string)
...
+		logrus.Warn("JOB_NAME environment variable not set")
+		return
+	}
+	job := &batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      jobName,
+			Namespace: namespace,
+		},
+	}
+	r.Event(job, fleetevent.Warning, "Failed", fmt.Sprintf("deleting OCI artifact: %q, error: %v", artifactID, errorToLog.Error()))
```
weyfonk: Should we be a bit more explicit about the reason for the event?
e.g.:
```suggestion
	r.Event(job, fleetevent.Warning, "FailedToDeleteOCIArtifact", fmt.Sprintf("deleting OCI artifact %q: %v", artifactID, errorToLog.Error()))
```

pkg/apis/fleet.cattle.io/v1alpha1/bundle_types.go
```diff
@@ -39,6 +39,10 @@ const (
 	// SecretTypeOCIStorage is the secret type used internally to store bundle resources in an OCI registry instead
 	// of etcd.
 	SecretTypeOCIStorage = "fleet.cattle.io/bundle-oci-storage/v1alpha1"
+
+	// InternalSecretLabel is a label added to any secret created by Fleet to propagate Bundle or
+	// BundleDeployment secrets storing credential details for OCI storage or HelmOps
```
weyfonk: nit:
```suggestion
	// BundleDeployment secrets storing credential details for OCI storage or HelmOps.
```

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -664,6 +668,10 @@ func (r *BundleReconciler) maybeDeleteOCIArtifact(ctx context.Context, bundle *f
 		return err
 	}
 	err = ocistorage.NewOCIWrapper().DeleteManifest(ctx, opts, bundle.Spec.ContentsID)
+	if err != nil {
+		r.Recorder.Event(bundle, fleetevent.Warning, "Failed", fmt.Sprintf("deleting OCI artifact: %q, error: %v", bundle.Spec.ContentsID, err.Error()))
```
weyfonk: Same remark/question as in `apply.go` on the event reason.

weyfonk (commented):
Thanks! With a question on the event reason, the rest LGTM :)

internal/cmd/cli/apply/apply.go
0xavi0: It was to avoid a circular dependency at some point (which points that the package was maybe not the best place for the constant)
The constant was declared in the `internal/cmd/cli` package but needed from the `internal/cmd/cli/apply` package and the first already imports the later. 

That variable was only used for the `apply` command.
I've moved it to the `internal/cmd/cli/apply` package along with other constants and 1 function that were also only related to `apply`.

internal/cmd/cli/apply/apply.go
0xavi0: Yup, thanks!

internal/cmd/controller/reconciler/bundle_controller.go
0xavi0: changed to the same suggestion

weyfonk (approved):

internal/cmd/cli/apply/apply.go
weyfonk: I see, thanks for clarifying and making this change :)

Issue #3585, opened 2025-04-25
Document and test basic rollout use cases


weyfonk:
The goal here is to be able to answer the following questions, with tests and documentation, about basic aspects of rollout:
- [x] How does auto-partitioning work? Does e.g. 10% of 4 clusters result in 1 partition?
- [x] Example use case from https://github.com/rancher/fleet/issues/2891
- [x] [rollout.go](https://github.com/rancher/fleet/blob/main/internal/cmd/controller/target/rollout.go,): does manual partitioning really distribute the remaining partitions as stated in the [docs](https://fleet.rancher.io/ref-fleet-yaml)?

Documentation must include user-facing [docs](https://fleet.rancher.io/ref-fleet-yaml), and optionally code comments/refactoring.

p-se:
> How does auto-partitioning work? Does e.g. 10% of 4 clusters result in 1 partition?

No. We have an undocumented immutable limit of 200 clusters, which disables auto-partitioning if less than 200 clusters are selected or available for targeting. However, if that limitation was removed or made configurable and then be disabled, the `limit` function in `rollout.go` would make sure to return at least `1`, no matter how small the given percentage.

> [rollout.go](https://github.com/rancher/fleet/blob/main/internal/cmd/controller/target/rollout.go?rgh-link-date=2025-04-25T12%3A54%3A39.000Z): does manual partitioning really distribute the remaining partitions as stated in the [docs](https://fleet.rancher.io/ref-fleet-yaml)?

No, every target that is not successfully selected using manual partitioning is not being added to any partition.

PR #3780, opened 2025-06-06
Unit tests for rollout

p-se:
Add rollout and status unit tests

- Add status tests
- Add rollout tests

Refers to #3585

### Checklist

- [ ]  I have updated the documentation via a pull request
  in the [fleet-docs](https://github.com/rancher/fleet-docs) repository.

### DIFF
--- a/internal/cmd/controller/target/partition.go
+++ b/internal/cmd/controller/target/partition.go
@@ -53,7 +53,7 @@ func UpdatePartitions(status *fleet.BundleStatus, allTargets []*Target) (err err
 			updateTarget(currentTarget, status, &partition.Status)
 		}
 
-		if updateStatusUnavailable(&partition.Status, partition.Targets) {
+		if updateStatusAndCheckUnavailable(&partition.Status, partition.Targets) {
 			status.UnavailablePartitions++
 		}
 
--- a/internal/cmd/controller/target/rollout.go
+++ b/internal/cmd/controller/target/rollout.go
@@ -18,7 +18,13 @@ func partitions(targets []*Target) ([]partition, error) {
 	return manualPartition(rollout, targets)
 }
 
-// getRollout returns the rollout strategy for the specified targets (pure function)
+// getRollout returns the rollout strategy for the specified targets (pure
+// function).
+//
+// If targets contains several elements, the rollout strategy of the first
+// element is used. If no rollout strategy is found, an empty is created and
+// returned. This function therefore assumes that all bundles in targets have
+// the same rollout strategy.
 func getRollout(targets []*Target) *fleet.RolloutStrategy {
 	var rollout *fleet.RolloutStrategy
 	if len(targets) > 0 {
@@ -50,6 +56,10 @@ func manualPartition(rollout *fleet.RolloutStrategy, targets []*Target) ([]parti
 					partitionTargets = append(partitionTargets, target)
 					continue targetLoop
 				}
+			}
+			if len(target.ClusterGroups) == 0 && matcher.Match(target.Cluster.Name, "", nil, target.Cluster.Labels) {
+				partitionTargets = append(partitionTargets, target)
+				continue targetLoop
 			}
 		}
 
@@ -89,10 +99,7 @@ func autoPartition(rollout *fleet.RolloutStrategy, targets []*Target) ([]partiti
 		if len(targets) == 0 {
 			return partitions, nil
 		}
-		end := maxSize
-		if len(targets) < maxSize {
-			end = len(targets)
-		}
+		end := min(len(targets), maxSize)
 
 		partitionTargets := targets[:end]
 		name := fmt.Sprintf("Partition %d - %d", offset, offset+end)
--- a/internal/cmd/controller/target/rollout_test.go
+++ b/internal/cmd/controller/target/rollout_test.go
@@ -1,0 +1,321 @@
+package target
+
+import (
+	"fmt"
+	"strconv"
+	"testing"
+
+	fleet "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+// createTargets creates a slice of targets with sequentially numbered clusters
+// and bundles. Both values start and stop are inclusive, meaning the targets
+// will be created from start to stop.
+func createTargets(start, stop int) []*Target {
+	targets := make([]*Target, stop-start+1)
+	for i := range stop - start + 1 {
+		targets[i] = &Target{
+			Cluster: &fleet.Cluster{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: "default",
+					Name:      "cluster-" + strconv.Itoa(start),
+				},
+			},
+			Bundle: &fleet.Bundle{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "bundle-" + strconv.Itoa(start),
+				},
+			},
+			Deployment: &fleet.BundleDeployment{
+				ObjectMeta: metav1.ObjectMeta{},
+				Spec:       fleet.BundleDeploymentSpec{},
+				Status:     fleet.BundleDeploymentStatus{},
+			},
+			DeploymentID: "deployment-" + strconv.Itoa(start),
+		}
+		start++
+	}
+	if start != stop+1 {
+		panic(fmt.Sprintf("createTargets: start and stop values are not equal: start: %d, stop: %d", start, stop))
+	}
+	return targets
+}
+
+func Test_createTargets(t *testing.T) {
+	tests := []struct {
+		name        string
+		start, stop int
+		want        []*Target
+	}{
+		{
+			name:  "start and stop should be inclusive",
+			start: 1,
+			stop:  5,
+			want: []*Target{
+				{DeploymentID: "deployment-1"},
+				{DeploymentID: "deployment-2"},
+				{DeploymentID: "deployment-3"},
+				{DeploymentID: "deployment-4"},
+				{DeploymentID: "deployment-5"},
+			},
+		},
+	}
+	for _, tt := range tests {
+		got := createTargets(tt.start, tt.stop)
+		if err := targetsEqual(got, tt.want); err != nil {
+			t.Errorf("createTargets(%d, %d): %v", tt.start, tt.stop, err)
+		}
+	}
+}
+
+func withCluster(targets []*Target, cluster *fleet.Cluster) {
+	for _, target := range targets {
+		target.Cluster = cluster
+	}
+}
+
+func withClusterGroup(targets []*Target, clusterGroup *fleet.ClusterGroup) {
+	for _, target := range targets {
+		target.ClusterGroups = append(target.ClusterGroups, clusterGroup)
+	}
+}
+
+func Test_withClusterGroup(t *testing.T) {
+	clusterGroup := &fleet.ClusterGroup{}
+	target1 := &Target{}
+	target2 := &Target{}
+	targets := []*Target{target1, target2}
+
+	withClusterGroup(targets, clusterGroup)
+
+	for _, target := range targets {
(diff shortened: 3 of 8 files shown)


internal/cmd/controller/target/rollout.go
```diff
@@ -51,6 +57,10 @@ func manualPartition(rollout *fleet.RolloutStrategy, targets []*Target) ([]parti
 					continue targetLoop
 				}
 			}
+			if len(target.ClusterGroups) == 0 && matcher.Match(target.Cluster.Name, "", nil, target.Cluster.Labels) {
+				partitionTargets = append(partitionTargets, target)
+				continue targetLoop
+			}
```
p-se, line 63: This addition fixes an issue where partitions name a single cluster, which would not result in a partition with the named cluster.

p-se: Just noticed that we don't even have it documented. But if we decide to document it, it'll actually work.

internal/cmd/controller/target/partition.go
```diff
@@ -14,24 +14,25 @@ type partition struct {
 // UpdatePartitions recomputes status, including partitions, from data in allTargets.
 // It creates Deployments in allTargets if they are missing.
 // It updates Deployments in allTargets if they are out of sync (DeploymentID != StagedDeploymentID).
-func UpdatePartitions(status *fleet.BundleStatus, allTargets []*Target) (err error) {
+func UpdatePartitions(bdStatus *fleet.BundleStatus, allTargets []*Target) (err error) {
```
weyfonk: `bdStatus` suggests that this might be a _bundle deployment_ status, yet at least in the [bundle reconciler](https://github.com/rancher/fleet/blob/main/internal/cmd/controller/reconciler/bundle_controller.go#L309), this function is called with a _bundle status_.

internal/cmd/controller/target/partition.go
```diff
@@ -75,9 +76,9 @@ func maxUnavailablePartitions(partitions []partition, targets []*Target) (int, e
 	return limit(len(partitions), rollout.MaxUnavailablePartitions, &defMaxUnavailablePartitions)
 }
 
-// updateTarget will update DeploymentID and Options for the target to the
+// tryUpdateDeploymentFromStaged will update DeploymentID and Options for the target to the
 // staging values, if it's in a deployable state
-func updateTarget(t *Target, status *fleet.BundleStatus, partitionStatus *fleet.PartitionStatus) {
+func tryUpdateDeploymentFromStaged(t *Target, bundleStatus *fleet.BundleStatus, partitionStatus *fleet.PartitionStatus) {
```
weyfonk: Just curious: why `try`? I mean, sure, there are conditions under which no update would happen, but this is true of quite a few different functions throughout Fleet ;)

internal/cmd/controller/target/rollout.go
```diff
@@ -18,7 +18,13 @@ func partitions(targets []*Target) ([]partition, error) {
 	return manualPartition(rollout, targets)
 }
 
-// getRollout returns the rollout strategy for the specified targets (pure function)
+// getRollout returns the rollout strategy for the specified targets (pure
+// function).
+//
+// If targets contains several elements, the rollout strategy of the first
+// element is used. If no rollout strategy is found, an empty is created and
```
weyfonk:
```suggestion
// element is used. If no rollout strategy is found, an empty one is created and
```

internal/cmd/controller/target/status.go
```diff
@@ -70,7 +71,8 @@ func upToDate(target *Target) bool {
 	return true
 }
 
-// isUnavailable checks if target is not available (pure function)
+// isUnavailable checks if target is available (pure function). If no target is
+// provided, it returns true, assuming that a nil target is always available.
```
weyfonk:
```suggestion
// provided, it returns false, assuming that a nil target is always available.
```

internal/cmd/controller/target/status.go
```diff
@@ -70,7 +71,8 @@ func upToDate(target *Target) bool {
 	return true
 }
 
-// isUnavailable checks if target is not available (pure function)
+// isUnavailable checks if target is available (pure function). If no target is
```
weyfonk:
```suggestion
// isUnavailable checks if target is unavailable (pure function). If no target is
```

internal/cmd/controller/target/rollout_test.go
```diff
@@ -0,0 +1,456 @@
...
+			Deployment: &fleet.BundleDeployment{
+				ObjectMeta: metav1.ObjectMeta{},
+				Spec:       fleet.BundleDeploymentSpec{},
+				Status:     fleet.BundleDeploymentStatus{},
+			},
+			DeploymentID: "deployment-" + strconv.Itoa(start),
+		}
+		start++
+	}
+	if start != stop+1 {
```
weyfonk: By construction of the loop, this check should be superfluous :thinking: 
`start` would be incremented by (stop-start + 1), which verifies `start == stop + 1` when the loop exits.

internal/cmd/controller/target/rollout_test.go
```diff
@@ -0,0 +1,456 @@
...
+	targets := []*Target{target1, target2}
+
+	withClusterGroup(targets, clusterGroup)
+
+	for _, target := range targets {
+		if len(target.ClusterGroups) != 1 || target.ClusterGroups[0] != clusterGroup {
+			t.Errorf("expected cluster group to be appended to target, got %+v", target.ClusterGroups)
+		}
+	}
+}
```
weyfonk: This code testing test code feels a bit much to me. Isn't it equivalent to testing that when passing a pointer variable to a function, mutations of that variable are applied beyond the function's scope?

internal/cmd/controller/target/rollout_test.go
```diff
@@ -0,0 +1,456 @@
...
+		min, max string
+	}
+	stats := func() []*targetStats {
+		stats := make([]*targetStats, len(got))
+		for i, p := range got {
+			stats[i] = &targetStats{
+				length: len(p.Targets),
+				// targets are sorted by Name
+				min: p.Targets[0].DeploymentID,
+				max: p.Targets[len(p.Targets)-1].DeploymentID,
```
weyfonk: Why call these fields `min` and `max`?

internal/cmd/controller/target/rollout_test.go
```diff
@@ -0,0 +1,456 @@
...
+			name:    "rounding with 230 clusters",
+			rollout: &fleet.RolloutStrategy{},
+			targets: createTargets(1, 230),
+			want: []partition{
+				{Targets: createTargets(1, 57), Status: fleet.PartitionStatus{MaxUnavailable: 57}},
+				{Targets: createTargets(58, 114), Status: fleet.PartitionStatus{MaxUnavailable: 57}},
+				{Targets: createTargets(115, 171), Status: fleet.PartitionStatus{MaxUnavailable: 57}},
+				{Targets: createTargets(172, 228), Status: fleet.PartitionStatus{MaxUnavailable: 57}},
+				{Targets: createTargets(229, 230), Status: fleet.PartitionStatus{MaxUnavailable: 2}},
+			},
```
weyfonk, line 222: Nice, I really like how easy these test cases are to understand :clap:

internal/cmd/controller/target/rollout_test.go
```diff
@@ -0,0 +1,456 @@
...
+						Name:        "Partition 1",
+						ClusterName: "cluster-1",
+					},
+					{
+						Name:        "Partition 2",
+						ClusterName: "cluster-2",
+					},
+				},
+			},
+			// targets: createTargets(1, 4),
```
weyfonk: Is this a leftover?

internal/cmd/controller/target/rollout_test.go
```diff
@@ -0,0 +1,456 @@
...
+							"group": "a",
+						},
+					},
+				})
+				return targets
+			},
+			want: []partition{
+				{Targets: createTargets(1, 60), Status: fleet.PartitionStatus{MaxUnavailable: 60}},
+				{Targets: createTargets(41, 100), Status: fleet.PartitionStatus{MaxUnavailable: 60}},
+				// {Targets: createTargets(61, 100), Status: fleet.PartitionStatus{MaxUnavailable: 40}},
```
weyfonk:
```suggestion
```

internal/cmd/controller/target/status_test.go
```diff
@@ -0,0 +1,504 @@
...
+			},
+			want: 5,
+		},
+		{
+			name:  "fixed value above count",
+			count: 10,
+			val: []*intstr.IntOrString{
+				{IntVal: 15},
+			},
+			want: 15,
```
weyfonk: In effect, wouldn't this be equivalent in behaviour to having the result be 10?
As in, specifying a maximum of 15 unavailable clusters for a partition containing 10 of them will never result in more than 10 unavailable clusters for that partition.

Just a thought on the implementation, not on the test case itself.

internal/cmd/controller/target/status_test.go
```diff
@@ -0,0 +1,504 @@
...
+			want: 1,
+		},
+		{
+			name:  "no value should match count",
+			count: 50,
+			val:   []*intstr.IntOrString{},
+			want:  50,
+		},
+		{
+			name:  "percentage below 1 should return 1",
```
weyfonk: Isn't this the same test case as `percent value 10 with count 5`?

internal/cmd/controller/target/status_test.go
```diff
@@ -0,0 +1,504 @@
...
+	}
+}
+
+func TestUnavailable(t *testing.T) {
+	tests := []struct {
+		name    string
+		targets []*Target
+		want    int
+	}{
+		{
```
weyfonk, line 404: nit: this could be expanded with a couple of simple test cases: one with all targets available, and another one with all targets unavailable; that would prevent this test suite from still passing if we update `Unavailable` to return a hard-coded value of `2`, as unlikely as it seems ;)

internal/cmd/controller/target/rollout.go
weyfonk: Is this covered by test case `should match cluster names` in `rollout_test.go`? That's the impression I get, but just to be sure.

p-se: Yeah, that's how I found it in the first place. The field was in the struct, so I tested it.

internal/cmd/controller/target/rollout_test.go
p-se: :bow:

internal/cmd/controller/target/status_test.go
p-se: I think it would behave more like we'd expect it to, so I changed it. Also for `150%` which resulted in a value of `150%` of the count.

internal/cmd/controller/target/status_test.go
p-se: Yes! Removed it.

internal/cmd/controller/target/partition.go
p-se: :shrug: Just trying to be be precise, but I've removed the prefix as it doesn't seem to be consistent with the other code. Thanks!

internal/cmd/controller/target/partition.go
p-se: Yes, you're right.

internal/cmd/controller/target/status.go
```diff
@@ -107,6 +108,13 @@ func limit(count int, val ...*intstr.IntOrString) (int, error) {
...
 
 	if maxUnavailable.Type == intstr.Int {
+		i := maxUnavailable.IntValue()
+		if i > count {
+			return count, nil
+		}
+		if i < 0 {
+			return 0, nil
+		}
 		return maxUnavailable.IntValue(), nil
```
weyfonk, line 118: nit: this could be rewritten as (or even compressed further, at the expense of readability):
```suggestion
		i := maxUnavailable.IntValue()
		i = max(0, i)
		i = min(i, count)

		return i, nil
```
Same remark about L139-145 below.

Issue #3764, opened 2025-06-04
Poll Helm for new available versions


weyfonk:
Similarly to what GitOps supports with polling git repositories, and following semantic versioning support, Fleet should be able to poll Helm repositories in search of new versions matching a HelmOp's `version` field, whether it be:
* a version constraint, meaning that the latest available version satisfying that constraint should be installed
* an empty or wildcard (`*`) value, in which case the latest available version should be installed

Polling must be configurable using a new `pollingInterval` field, consistently with what GitOps supports.

weyfonk:
## Additional QA
 
### Problem
When installing a Helm chart using a Fleet HelmOp resource with a wildcard or empty version field, Fleet would initially resolve the latest available version for the chart. Subsequent updates of the HelmOp would repeat this process, but the user did not have any way of ensuring that a newer chart version would be installed as soon as it would be available.

### Solution
Fleet is now able to poll a Helm repository referenced by a HelmOp , similarly to what GitOps supports for git repositories, as soon as the HelmOp meets both of the following conditions:
* its `version` field may resolve to more than one version over time, e.g. `*` (wildcard), `< 1.0.0`, `2.x.x`. Referencing a static version, such as `1.2.3`, disables polling
* its new `pollingInterval` field, which value defaults to 0, is set to a non-zero duration

### Testing
### Engineering Testing
#### Manual Testing
N/A

#### Automated Testing
##### End-to end tests
1. Creating a HelmOp, meeting the above conditions for polling, referencing a chart hosted on a test ChartMuseum instance, where version `0.1.0` of the chart is the latest version matching the version constraint
2. Checking that version `0.1.0` of the chart is installed
3. Uploading version `0.2.0` of the chart to ChartMuseum, with version `0.2.0` also matching the HelmOp's version constraint
4. Checking that version `0.2.0` of the chart is installed

##### Integration tests
* if a HelmOp bundle exists with a version constraint (instead of a static version, which should have been set by the HelmOps controller when checking Helm for the latest version matching the HelmOp's constraint), then no bundle is created
* if a polling job fails, it updates the HelmOp status' new `Polled` condition to `False`, with an error message explaining the failure

##### Unit tests
- polling job scheduling: 
    - Does not happen if one of the above conditions for polling is not met
    - Is triggered if the HelmOp's `version`, `chart`, `repo` or polling interval field changes
    - Ensures that jobs replace one another for a given key which is unique to the HelmOp
[...]

PR #3784, opened 2025-06-10
Poll Helm repository from HelmOps reconciler

weyfonk:
This patch enhances the HelmOps reconciler with polling, enabling it to periodically check a Helm repository for new versions matching the constraints set in the `version` field of the HelmOp.

The polling job:
* patches the bundle with the resolved version upon success, allowing Fleet to install the corresponding chart
* records Kubernetes events, whether it succeeds (info events) or fails (warnings, with reason of error)
* updates a new `Polled` condition in the HelmOp status, with `True` if it succeeds, and with the error if it fails
* sets a new `Last Polling Time` field in the HelmOp status if the job could get to the point where it would try to poll Helm, and whether it succeeds or not.

Refers to #3764

### Checklist


charts/fleet-crd/templates/crds.yaml
```diff
@@ -8612,6 +8617,11 @@ spec:
                         BundleState according to StateRank.'
                       type: string
                   type: object
+                lastPollingTriggered:
+                  description: LastPollingTime is the last time the polling check
```
0xavi0, line 8621: The description does not match the name. 
I would make this and the `GitRepo` one be the same.

e2e/assets/gitrepo/sleeper-chart2/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
```
0xavi0: Should this go in the helmops directory? This chart is only used in helmops tests, right?

charts/fleet/templates/rbac_helmops.yaml
```diff
@@ -11,6 +11,7 @@ rules:
     verbs:
       - "create"
       - "list"
+      - "watch"
```
0xavi0, line 14: Out of curiosity: why do we need to `watch` secrets now? 
Is it becase of the cache?

e2e/single-cluster/helmop_test.go
```diff
@@ -178,3 +270,9 @@ var _ = Describe("HelmOp resource tests with oci registry", Label("infra-setup",
 		})
 	})
 })
+
+// getExternalHelmAddr retrieves the external URL where our local Helm registry can be reached, based on the provided
+// port.
+func getExternalHelmAddr(k kubectl.Command) (string, error) {
+	return k.Namespace(cmd.InfraNamespace).Get("service", "chartmuseum-service", "-o", "jsonpath={.status.loadBalancer.ingress[0].ip}")
+}
```
0xavi0: I think this won't work in Macs.
If the `external_ip` env variable is set... could you consider returning that value?

internal/cmd/controller/helmops/reconciler/helmop_controller.go
```diff
@@ -203,7 +214,9 @@ func (r *HelmOpReconciler) calculateBundle(helmop *fleet.HelmOp) *fleet.Bundle {
 			Namespace: helmop.Namespace,
 			Name:      helmop.Name,
 		},
-		Spec: spec,
+		// We ensure the bundle and HelmOp spec are independent. This prevents versions constraints from being overwritten
+		// in the HelmOp spec with actual versions resolved from Helm when the bundle version is updated.
```
0xavi0, line 215:
```suggestion
		// We ensure the bundle and HelmOp spec are independent. This prevents versions constraints from being overwritten
		// in the HelmOp spec with actual versions resolved from the Helm registry when the bundle version is updated.
```

Could it also work not passing `helmOp` as a pointer to this function?

internal/cmd/controller/helmops/reconciler/helmop_controller.go
```diff
@@ -236,6 +249,96 @@ func (r *HelmOpReconciler) calculateBundle(helmop *fleet.HelmOp) *fleet.Bundle {
...
+		if err != nil {
+			return err
+		}
+	}
+	auth.InsecureSkipVerify = helmop.Spec.InsecureSkipTLSverify
+
+	version, err := bundlereader.ChartVersion(*helmop.Spec.Helm, auth)
+	if err != nil {
+		return err
+	}
```
0xavi0: Most of this code is also present in the `pollHelm` function from the `polling_job.go` file. 
Maybe we could create a function that gathers the version and then use it in both places... so if we need to change something we don't need to updated it twice.
what do you think?

internal/cmd/controller/helmops/reconciler/helmop_controller.go
```diff
@@ -236,6 +249,96 @@ func (r *HelmOpReconciler) calculateBundle(helmop *fleet.HelmOp) *fleet.Bundle {
...
+			logger.V(1).Info("Scheduled new polling job")
+		} else if err != nil {
+			// This should never happen
+			return fmt.Errorf("an unknown error occurred when looking for a polling job: %w", err)
+		}
+	} else if !errors.Is(err, quartz.ErrJobNotFound) {
+		// A job still exists, but is no longer needed; delete it.
+		if err = r.Scheduler.DeleteJob(jobKey); err != nil {
+			return fmt.Errorf("failed to delete polling job: %w", err)
+		}
```
0xavi0: we also delete if the error is something else (different to `ErrJobNotFound`)? 
I think `GetScheduledJob` returns other error codes (and it might returns more in future versions)

internal/cmd/controller/helmops/reconciler/helmop_controller.go
```diff
@@ -236,6 +249,96 @@ func (r *HelmOpReconciler) calculateBundle(helmop *fleet.HelmOp) *fleet.Bundle {
...
+// This is calculated in the upstream cluster so all downstream bundle deployments have the same
+// version. (Potentially we could be gathering the version at the very moment it is being updated, for example)
+func (r *HelmOpReconciler) handleVersion(ctx context.Context, oldBundle *fleet.Bundle, bundle *fleet.Bundle, helmop fleet.HelmOp) error {
+	if _, err := semver.StrictNewVersion(helmop.Spec.Helm.Version); err == nil {
+		bundle.Spec.Helm.Version = helmop.Spec.Helm.Version
+		return nil
+	}
+
+	// Prevent version computation from interfering with polling but, if polling applies, set the version when first
+	// creating the bundle, to avoid having to wait a full polling interval.
```
0xavi0: What if the semver constraint (or the the Repo, or any other value) changed? Should we wait the whole polling interval? 
I think we're applying Spec changes immediately to `GitRepos` and don't wait for the polling interval.

internal/cmd/controller/helmops/reconciler/helmop_controller_test.go
```diff
@@ -306,3 +309,410 @@ func TestReconcile_CreatesBundleAndUpdatesStatus(t *testing.T) {
...
+				},
+			},
+			expectedSchedulerCalls: func(ctrl *gomock.Controller, scheduler *mocks.MockScheduler, helmop fleet.HelmOp) {
+				existingTrigger := quartz.NewSimpleTrigger(2 * helmop.Spec.PollingInterval.Duration)
+
+				existingJob := mocks.NewMockScheduledJob(ctrl)
+				existingJob.EXPECT().Trigger().Return(existingTrigger)
+
+				scheduler.EXPECT().GetScheduledJob(gomock.Any()).Return(existingJob, nil)
+				scheduler.EXPECT().ScheduleJob(gomock.Any(), gomock.Any()).Return(nil)
```
0xavi0: We could maybe check if `ScheduleJob` is receiving the expected parameter to replace the job.
`quartz.JobDetailOptions.Replace` is set to false by default, so I think it's good to check that we're passing it set to true and we're not adding extra jobs.

internal/cmd/controller/helmops/reconciler/polling_job.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	auth := bundlereader.Auth{}
+	if h.Spec.HelmSecretName != "" {
+		req := types.NamespacedName{Namespace: h.Namespace, Name: h.Spec.HelmSecretName}
+
+		var err error
+		auth, err = bundlereader.ReadHelmAuthFromSecret(ctx, j.client, req)
+		if err != nil {
+			return fail(fmt.Errorf("could not read Helm auth from secret: %w", err), "FailedToReadHelmAuth")
```
0xavi0: This call to `fail` will have a zero `pollingTimestamp`, right? 
We should maybe move
```go
pollingTimestamp = time.Now().UTC()
```
to be executed before?

internal/cmd/controller/helmops/reconciler/polling_job.go
```diff
@@ -0,0 +1,215 @@
...
+	b := &fleet.Bundle{}
+
+	if err := j.client.Get(ctx, nsName, b); err != nil {
+		return fail(fmt.Errorf("could not get bundle before patching its version: %w", err), "FailedToGetBundle")
+	}
+
+	orig := b.DeepCopy()
+	b.Spec.Helm.Version = version
+
+	j.recorder.Event(h, fleetevent.Normal, "GotNewChartVersion", version)
```
0xavi0: What if the version gathered was the same?
Say the constraint was `1.*.*` and the gathered version for the moment this was created was `1.1.1` and... when the first polling is executed is still `1.1.1`... 
The patch later will be empty (which is cool) but... would we also create an event?

internal/cmd/controller/helmops/reconciler/polling_job.go
```diff
@@ -0,0 +1,215 @@
...
+	if patchData, err := patch.Data(b); err == nil && string(patchData) == "{}" {
+		// skip update if patch is empty
+		return nil
+	}
+
+	if err := j.client.Patch(ctx, b, patch); err != nil {
+		return fail(fmt.Errorf("could not patch bundle to set the resolved version: %w", err), "FailedToPatchBundle")
+	}
+
+	merr := []error{}
```
0xavi0: Question: why do we need a slice of errors in here?

pkg/apis/fleet.cattle.io/v1alpha1/helmop_types.go
```diff
@@ -51,19 +52,30 @@ type HelmOpList struct {
...
+
 	// InsecureSkipTLSverify will use insecure HTTPS to clone the helm app resource.
 	InsecureSkipTLSverify bool `json:"insecureSkipTLSVerify,omitempty"`
 }
 
 type HelmOpStatus struct {
 	StatusBase `json:",inline"`
+
+	// LastPollingTime is the last time the polling check was triggered
+	LastPollingTime metav1.Time `json:"lastPollingTriggered,omitempty"`
```
0xavi0, line 77: Same remark as before. (LastPollingTime vs lastPollingTriggered)

0xavi0 (commented):
A few questions and nitpicks.
I wonder if it's still possible to have race conditions when updating the `Version` from the reconciler and the job (as they run in different routines). 

I think we created the field `WebhookCommit` so it is only updated from the webhook handler and then the reconciler is the one updating the final `Commit` when `WebhookCommit` changed to avoid conflicts. 

It's totally a corner case and I guess the race condition is only possible for very short polling intervals mixed with conflicts when updating the Version from the reconciler or similar. 
Maybe it is not even possible to incur in a race condition, I'm not sure.

charts/fleet/templates/rbac_helmops.yaml
weyfonk: It seems so yes: removing it leads to errors such as:
```
2025-06-11T14:53:15Z    ERROR   controller-runtime.cache.UnhandledError Failed to watch {"reflector": "pkg/mod/k8s.io/client-go@v0.33.1/tools/cache/reflector.go:285", "type": "*v1.Secret", "error": "secrets is forbidden: User \"system:serviceaccount:cattle-fleet-system:hel
mops\" cannot watch resource \"secrets\" in API group \"\" at the cluster scope"}                                                                                                                                                                                                
k8s.io/apimachinery/pkg/util/runtime.logError                                                                                                                                                                                                                                    
        /home/tan/go/pkg/mod/k8s.io/apimachinery@v0.33.1/pkg/util/runtime/runtime.go:226                                   
k8s.io/apimachinery/pkg/util/runtime.handleError                                                                                                                                                                                                                                 
        /home/tan/go/pkg/mod/k8s.io/apimachinery@v0.33.1/pkg/util/runtime/runtime.go:217                                                                                                                                                                                         
k8s.io/apimachinery/pkg/util/runtime.HandleErrorWithContext                                                                                                                                                                                                                      
        /home/tan/go/pkg/mod/k8s.io/apimachinery@v0.33.1/pkg/util/runtime/runtime.go:203                                                                                                                                                                                         
k8s.io/client-go/tools/cache.DefaultWatchErrorHandler                                                                                                                                                                                                                            
        /home/tan/go/pkg/mod/k8s.io/client-go@v0.33.1/tools/cache/reflector.go:200                                                                                                                                                                                               
k8s.io/client-go/tools/cache.(*Reflector).RunWithContext.func1                                                                                                                                                                                                                   
        /home/tan/go/pkg/mod/k8s.io/client-go@v0.33.1/tools/cache/reflector.go:360                                                                                                                                                                                               
k8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1                                                                                                                                                                                                                             
        /home/tan/go/pkg/mod/k8s.io/apimachinery@v0.33.1/pkg/util/wait/backoff.go:233
k8s.io/apimachinery/pkg/util/wait.BackoffUntilWithContext.func1    
        /home/tan/go/pkg/mod/k8s.io/apimachinery@v0.33.1/pkg/util/wait/backoff.go:255
k8s.io/apimachinery/pkg/util/wait.BackoffUntilWithContext                                                                               
        /home/tan/go/pkg/mod/k8s.io/apimachinery@v0.33.1/pkg/util/wait/backoff.go:256                                                                                                                                                                                            
k8s.io/apimachinery/pkg/util/wait.BackoffUntil                                                                                          
        /home/tan/go/pkg/mod/k8s.io/apimachinery@v0.33.1/pkg/util/wait/backoff.go:233                                                                                                                                                                                            
k8s.io/client-go/tools/cache.(*Reflector).RunWithContext                                                                                                                                                                                                                         
        /home/tan/go/pkg/mod/k8s.io/client-go@v0.33.1/tools/cache/reflector.go:358                                                 
k8s.io/apimachinery/pkg/util/wait.(*Group).StartWithContext.func1                                                                                                                                                                                                                
        /home/tan/go/pkg/mod/k8s.io/apimachinery@v0.33.1/pkg/util/wait/wait.go:63                                                                                                                                                                                                
k8s.io/apimachinery/pkg/util/wait.(*Group).Start.func1                                                                                                                                                                                                                           
        /home/tan/go/pkg/mod/k8s.io/apimachinery@v0.33.1/pkg/util/wait/wait.go:72
```

charts/fleet-crd/templates/crds.yaml
weyfonk: As discussed offline: this has been copied over from the `GitRepo` field.

e2e/assets/gitrepo/sleeper-chart2/Chart.yaml
weyfonk: Yes! Thanks for pointing this out, moving it to `e2e/assets/helmop`.

internal/cmd/controller/helmops/reconciler/helmop_controller.go
weyfonk: I tried it, thinking that it would work, but it did not.

e2e/single-cluster/helmop_test.go
weyfonk: Done, thanks for raising this :)

internal/cmd/controller/helmops/reconciler/helmop_controller.go
weyfonk: If I understand your comment properly, then we should remove this check and have reconciles always update the version from Helm, to prevent that resolution from having to wait until the next polling interval. This would be more consistent with what we do on the GitOps side. Is that correct?
We could do that, at an increased risk of conflict, but we have retries in place.

My other concern about this is that it may make Helm chart updates less predictable as, apart from HelmOp creation time, they may happen from any reconcile. But I'm not sure how much of a concern that might be to users.

internal/cmd/controller/helmops/reconciler/polling_job.go
weyfonk: Yes, this is where the definition of polling may get a bit philosophical. Do we consider polling to have happened if auth information could not be extracted from the Helm access secret? I've decided that we should not, because in that case, no actual polling request could be sent to the Helm registry.
However, this could be viewed differently, as in: polling happens as soon as the job is started, in which case yes, I'd agree with your comment.

Which approach do you think is best?

internal/cmd/controller/helmops/reconciler/polling_job.go
weyfonk: Good catch, thanks! This was a leftover from `updateErrorStatus`, and multi-errors are indeed not needed here.

internal/cmd/controller/helmops/reconciler/polling_job.go
weyfonk: Good point, thanks for raising this; I've added a condition on versions, so that a new event will only be created if the resolved version is newer than the one already set in the bundle.

internal/cmd/controller/helmops/reconciler/helmop_controller_test.go
weyfonk: Yes! Added a matcher to check for the replace parameter, and using it in this test case. Thanks for suggesting this.

internal/cmd/controller/helmops/reconciler/helmop_controller.go
weyfonk: Yes, other errors may appear in future Quartz versions. I've refactored this to treat unknown errors separately, and only delete the job if it could be fetched without any error.

internal/cmd/controller/helmops/reconciler/polling_job.go
weyfonk: On the other hand, following your suggestion would also simplify the refactoring effort suggested [here](https://github.com/rancher/fleet/pull/3784#discussion_r2139668323), so I'm tempted to simply agree with your original comment ;)

internal/cmd/controller/helmops/reconciler/helmop_controller.go
0xavi0: Oh, right... The `Helm` field of the `Spec` is a pointer.

internal/cmd/controller/helmops/reconciler/helmop_controller.go
0xavi0: I don't mean to reconcile always.
Only when the spec of what defines a Helm chart changes. (Fields `Repo`, `Chart` or `Version` inside `spec.helm`)
For example, say we have a polling interval of 5 hours.
And the users changes the chart version constraint because version 2 is available and now he/she wants to deploy with the `2.*.*` constraint. 
He/she must wait 5 hours in the worst case scenario (if polling just happened right before the change) in order to get the update.

I think `GitOps` re-schedules the polling job when the spec changes, that way we don't have conflicts.

Also `GitRepo` have that Force Update feature that `HelmOps` don't.

internal/cmd/controller/helmops/reconciler/polling_job.go
0xavi0: Oh, I think it's okay to consider polling has happened even if there was an error.
My comment was because at that particular call to `fail` `pollingTimestamp` has been declared but not assigned yet.
If the error happens and that `fail` is called it will log timestamp zero

weyfonk: > My comment was because at that particular call to `fail` `pollingTimestamp` has been declared but not assigned yet.

I understand, but this had been accounted for: `updateErrorStatus` only includes the polling timestamp in the patch if it is non-zero ;)
Anyway, this is no longer an issue: I'm moving this variable assignment further up as suggested.
Thanks!

internal/cmd/controller/helmops/reconciler/helmop_controller.go
weyfonk: Thanks, this is clearer now. As discussed offline, I've refactored scheduling logic with a custom trigger, so that a job is scheduled to be run right away when the chart ref (name, repo or version) changes.

internal/cmd/controller/helmops/reconciler/helmop_controller.go
weyfonk: Done, thanks for suggesting this :+1:

pkg/apis/fleet.cattle.io/v1alpha1/helmop_types.go
0xavi0: Are we still going to keep `LastPollingTime` and `lastPollingTriggered`? 
I thought we said to fix it for `HelmOps` and keep it for now for `GitRepos` (but maybe I got confused)

internal/cmd/controller/helmops/reconciler/helmop_controller_test.go
```diff
@@ -310,3 +319,635 @@ func TestReconcile_CreatesBundleAndUpdatesStatus(t *testing.T) {
...
+			},
+			expectedSchedulerCalls: func(ctrl *gomock.Controller, scheduler *mocks.MockScheduler, helmop fleet.HelmOp) {
+				trigger := quartz.NewSimpleTrigger(2 * helmop.Spec.PollingInterval.Duration)
+
+				job := mocks.NewMockScheduledJob(ctrl)
+				job.EXPECT().Trigger().Return(trigger)
+				job.EXPECT().JobDetail().Return(nil)
+
+				scheduler.EXPECT().GetScheduledJob(gomock.Any()).Return(job, nil)
+				scheduler.EXPECT().ScheduleJob(gomock.Any(), gomock.Any()).Return(errors.New("something happened!"))
```
0xavi0: Nit: as it is checked in other calls
```suggestion
				scheduler.EXPECT().ScheduleJob(matchesJobDetailReplace(true), gomock.Any()).Return(errors.New("something happened!"))
```

internal/cmd/controller/helmops/reconciler/helmop_controller_test.go
```diff
@@ -310,3 +319,635 @@ func TestReconcile_CreatesBundleAndUpdatesStatus(t *testing.T) {
...
+								Chart:   "chart",
+								Version: "1.x.x",
+							},
+						},
+					},
+				},
+			},
+			expectedSchedulerCalls: func(_ *gomock.Controller, scheduler *mocks.MockScheduler, helmop fleet.HelmOp) {
+				scheduler.EXPECT().GetScheduledJob(gomock.Any()).Return(nil, quartz.ErrJobNotFound)
+				scheduler.EXPECT().ScheduleJob(gomock.Any(), gomock.Any()).Return(nil)
```
0xavi0: same as above
```suggestion
				scheduler.EXPECT().ScheduleJob(matchesJobDetailReplace(true), gomock.Any()).Return(nil)
```

internal/cmd/controller/helmops/reconciler/helmop_controller_test.go
```diff
@@ -310,3 +319,635 @@ func TestReconcile_CreatesBundleAndUpdatesStatus(t *testing.T) {
...
+								Chart:   "chart",
+								Version: "1.x.x",
+							},
+						},
+					},
+				},
+			},
+			expectedSchedulerCalls: func(ctrl *gomock.Controller, scheduler *mocks.MockScheduler, helmop fleet.HelmOp) {
+				scheduler.EXPECT().GetScheduledJob(gomock.Any()).Return(nil, quartz.ErrJobNotFound)
+				scheduler.EXPECT().ScheduleJob(gomock.Any(), gomock.Any()).Return(errors.New("something happened!"))
```
0xavi0: same as above:

```suggestion
				scheduler.EXPECT().ScheduleJob(matchesJobDetailReplace(true), gomock.Any()).Return(errors.New("something happened!"))
```

PR #3786, opened 2025-06-10
Fix panic found in OCI test

p-se:
Refers to #3792

### Checklist

### DIFF
--- a/internal/bundlereader/loaddirectory.go
+++ b/internal/bundlereader/loaddirectory.go
@@ -422,26 +422,25 @@ func newHttpGetter(auth Auth) *getter.HttpGetter {
 		header.Add("Authorization", "Basic "+basicAuth(auth.Username, auth.Password))
 		httpGetter.Header = header
 	}
+
+	transport := http.DefaultTransport.(*http.Transport).Clone()
 	if auth.CABundle != nil {
 		pool, err := x509.SystemCertPool()
 		if err != nil {
 			pool = x509.NewCertPool()
 		}
 		pool.AppendCertsFromPEM(auth.CABundle)
-		transport := http.DefaultTransport.(*http.Transport).Clone()
 		transport.TLSClientConfig = &tls.Config{
 			RootCAs:            pool,
 			MinVersion:         tls.VersionTLS12,
 			InsecureSkipVerify: auth.InsecureSkipVerify, // nolint:gosec
 		}
-		httpGetter.Client.Transport = transport
 	} else if auth.InsecureSkipVerify {
-		transport := http.DefaultTransport.(*http.Transport).Clone()
 		transport.TLSClientConfig = &tls.Config{
 			InsecureSkipVerify: auth.InsecureSkipVerify, // nolint:gosec
 		}
-		httpGetter.Client.Transport = transport
 	}
+	httpGetter.Client.Transport = transport
 
 	return httpGetter
 }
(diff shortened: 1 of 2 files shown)


internal/bundlereader/loaddirectory.go
```diff
@@ -422,26 +422,25 @@ func newHttpGetter(auth Auth) *getter.HttpGetter {
...
 		}
-		httpGetter.Client.Transport = transport
 	} else if auth.InsecureSkipVerify {
-		transport := http.DefaultTransport.(*http.Transport).Clone()
 		transport.TLSClientConfig = &tls.Config{
 			InsecureSkipVerify: auth.InsecureSkipVerify, // nolint:gosec
 		}
-		httpGetter.Client.Transport = transport
 	}
+	httpGetter.Client.Transport = transport
```
weyfonk, line 443: Good call :+1:

internal/bundlereader/loaddirectory_test.go
```diff
@@ -29,7 +29,29 @@ func TestGetContent(t *testing.T) {
...
+		   chart: "oci://ghcr.io/fleetqa/fleet-qa-examples/fleet-test-configmap-chart"
+		   version: "0.1.0"
+		   values:
+		     replicas: 2`,
+			},
+			source: "oci://ghcr.io/fleetqa/fleet-qa-examples/fleet-test-configmap-chart",
+			auth: bundlereader.Auth{
+				Username: "foo",
+				Password: "bar",
+				//InsecureSkipVerify: true, // When commented out, the panic disappears
```
weyfonk:
```suggestion
				// InsecureSkipVerify is false by default
```

internal/bundlereader/loaddirectory_test.go
```diff
@@ -29,7 +29,29 @@ func TestGetContent(t *testing.T) {
...
+			directoryStructure: fsNode{
+				name: "fleet.yaml",
+				contents: `namespace: fleet-helm-oci-with-auth-example
+		 helm:
+		   chart: "oci://ghcr.io/fleetqa/fleet-qa-examples/fleet-test-configmap-chart"
+		   version: "0.1.0"
+		   values:
+		     replicas: 2`,
+			},
+			source: "oci://ghcr.io/fleetqa/fleet-qa-examples/fleet-test-configmap-chart",
```
weyfonk: This may not need actual OCI registry URLs; we could use dummy ones (e.g. `oci://foo/bar/baz`) to clarify that this fits in a unit test case and that exposing/fixing the issue does not involve actually reaching the registry.

internal/bundlereader/loaddirectory_test.go
```diff
@@ -29,7 +29,29 @@ func TestGetContent(t *testing.T) {
...
+			auth: bundlereader.Auth{
+				Username: "foo",
+				Password: "bar",
+				//InsecureSkipVerify: true, // When commented out, the panic disappears
+			},
+			expectedErr: "response status code 403: denied",
```
weyfonk: Following my comment above on OCI URLs, this could become something like:
```suggestion
			expectedErr: "no such host",
```

internal/bundlereader/loaddirectory_test.go
p-se: I do get inconsistent results. On my machine I could check for "no such host", on [Github](https://github.com/rancher/fleet/actions/runs/15607929114/job/43966719236) I get

```
2025/06/12 11:41:14 http: TLS handshake error from 127.0.0.1:35048: remote error: tls: bad certificate
```

weyfonk (approved):

PR #3796, opened 2025-06-16
Filter cluster events to trigger bundle deployment creation less often

manno:
Fleet controller logs a lot of "unchanged bundleployment" messages.

### Old behavior: 

Every time a cluster’s resource version changes, e.g. because `.status.agent.lastSeen` is updated, we enqueue all bundles targeting the cluster.
* At 1000 clusters, each cluster has one agent bundle. This will do 1000 reconciles in roughly one DefaultClusterCheckInterval (15min). Each reconcile will calculate and update **one bundledeployment**. Fine, though lots of spam in the logs.
* Additionally, when one bundle targets 1000 clusters, each change to a cluster (at least 1000 in DefaultClusterCheckInterval) will reconcile the bundle, each reconcile will compute **1000 bundledeployments**. De-duplication happens, but this is still a lot of requests.

### Why re-compute bundledeployments for a cluster change?

* cluster was deleted
* cluster name/labels changed, used in templating and targeting
* cluster spec.templateValues (templating)
* cluster annotations (templating)

### DIFF
--- a/internal/cmd/controller/reconciler/bundle_controller.go
+++ b/internal/cmd/controller/reconciler/bundle_controller.go
@@ -6,6 +6,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
+	"reflect"
 	"slices"
 	"time"
 
@@ -36,6 +38,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
@@ -79,7 +82,8 @@ func (r *BundleReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		For(&fleet.Bundle{}).
 		// Note: Maybe improve with WatchesMetadata, does it have access to labels?
 		Watches(
-			// Fan out from bundledeployment to bundle
+			// Fan out from bundledeployment to bundle, this is useful to update the
+			// bundle's status fields.
 			&fleet.BundleDeployment{},
 			handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, a client.Object) []ctrl.Request {
 				bd := a.(*fleet.BundleDeployment)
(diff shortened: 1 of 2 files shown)


internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -125,13 +127,50 @@ func (r *BundleReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+func clusterChangedPredicate() predicate.Funcs {
+	return predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return true
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			n := e.ObjectNew.(*fleet.Cluster)
+			o := e.ObjectOld.(*fleet.Cluster)
+			// handle cluster deletion
+			if n == nil || n.DeletionTimestamp.IsZero() {
```
aruiz14: Should this be
```suggestion
			if n == nil || !n.DeletionTimestamp.IsZero() {
```
?

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -125,13 +127,50 @@ func (r *BundleReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+				return true
+			}
+			// labels and annotations are used for templating and targeting
+			if !reflect.DeepEqual(n.Labels, o.Labels) {
+				return true
+			}
+			if !reflect.DeepEqual(n.Annotations, o.Annotations) {
+				return true
+			}
```
aruiz14: Consider using [`maps.Equal`](https://pkg.go.dev/maps#Equal) for simple map comparisons instead of reflection. Same about `.Status.Namespace` below (assuming they are just strings).

internal/cmd/controller/reconciler/bundle_controller.go
manno: I wanted to be careful and rather trigger more than less. But I agree, if we skip nil, we can also skip a deletion timestamp. The Delete event will trigger a reconcile.

aruiz14: Hmm I think we are misunderstanding something here. I commented based on the comment, since `n.DeletionTimestamp.IsZero()` will be `true` when the object is **not** being deleted.

manno:
This PR could interfere with the bundle's status updates. The bundle reconcile not only updates the bundledeployments, it also computes the bundle status.

Q: A change to a cluster might change bundle status fields, like:
A: No, the cluster status is not used in the bundle's status.

Q: A reconcile on cluster change might still be needed just as a trigger, because existing events are not enough?
A: A change to the bundledeployment, always triggers a reconcile, there are a lot of these.

weyfonk (approved):

PR #3805, opened 2025-06-18
Incomplete Agent Registration Stops BundleDeployment Creation

manno:
Refers to https://github.com/rancher/fleet/issues/3830

Try to continue if a namespace for a bundledeployment is missing.


manno:
It's probably less invasive to catch the `"error": "namespaces \"cluster-fleet-default-downstream-1002-ab4388254ea0\" not found",` error and continue. Checking the namespace is an extra API request and breaks tests.

manno:
See https://github.com/rancher/fleet/pull/4074

Issue #3818, opened 2025-06-19
Enable OCI registry by default


0xavi0:
Change the name of the env variable from `EXPERIMENTAL_OCI_STORAGE` to `OCI_STORAGE`.
The feature will be enabled by default. And disabled when `OCI_STORAGE=false`

subtask of: https://github.com/rancher/fleet/issues/3396

PR #3819, opened 2025-06-19
Enables OCI storage by default

0xavi0:
Changes the env variable from `EXPERIMENTAL_OCI_STORAGE` to `OCI_STORAGE`

The feature is enabled by default and in order to disable it users should pass `OCI_STORAGE=false` in `extraEnv`

Refers to: https://github.com/rancher/fleet/issues/3818

### Checklist


internal/ocistorage/ociwrapper.go
```diff
@@ -280,9 +280,19 @@ func (o *OCIWrapper) DeleteManifest(ctx context.Context, opts OCIOpts, id string
 	return repo.Delete(ctx, desc)
 }
 
-// ExperimentalOCIIsEnabled returns true if the EXPERIMENTAL_OCI_STORAGE env variable is set to true
+// OCIIsEnabled returns true if the OCI_STORAGE env variable is not set to true
```
weyfonk: Unless I've missed an important aspect of this:
```suggestion
// OCIIsEnabled returns true if the OCI_STORAGE env variable is not set to false
```

0xavi0: I changed it to reflect that it's `true` also when it's not set

weyfonk (approved):

PR #3831, opened 2025-06-23
Skip polling for OCI and tarball charts

weyfonk:
OCI and direct referencing through tarballs does not involve an `index.yaml` file which Helm can search for new versions over time. Therefore, Fleet now skips Helm polling in such cases.

Refers to #3764
Follow-up to #3784.

### Checklist

[fleet-docs](https://github.com/rancher/fleet-docs) repository: see https://github.com/rancher/fleet-docs/pull/272

### DIFF
--- a/internal/cmd/controller/helmops/reconciler/helmop_controller.go
+++ b/internal/cmd/controller/helmops/reconciler/helmop_controller.go
@@ -409,10 +409,6 @@ func usesPolling(helmop fleet.HelmOp) bool {
 
 	// Polling does not apply to OCI and tarball charts, where no index.yaml file is available to check for new
 	// chart versions.
-	if helmop.Spec.Helm.Repo == "" {
-		return false
-	}
-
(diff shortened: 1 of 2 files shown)


internal/cmd/controller/helmops/reconciler/helmop_controller.go
```diff
@@ -407,6 +407,20 @@ func usesPolling(helmop fleet.HelmOp) bool {
 		return false
 	}
 
+	// Polling does not apply to OCI and tarball charts, where no index.yaml file is available to check for new
+	// chart versions.
+	if helmop.Spec.Helm.Repo == "" {
+		return false
+	}
```
0xavi0: Question: 
After merging https://github.com/rancher/fleet/pull/3795 do we really need to check for this here?
If it's empty... the only possible option is that it is a tarball (which is checked at the line below)

The lack of a test that checks for empty `Repo` explicitly made me wonder if it's really needed.

weyfonk: Correct, this is superfluous. Removed it, thanks.

PR #3837, opened 2025-06-25
Avoid trying to import clusters if there is no valid apiServerURL available

aruiz14:
Refers to #XXX

### Checklist


internal/cmd/controller/agentmanagement/controllers/cluster/import.go
```diff
@@ -291,7 +305,10 @@ func (i *importHandler) importCluster(cluster *fleet.Cluster, status fleet.Clust
 
 	if apiServerURL == "" {
 		if len(cfg.APIServerURL) == 0 {
-			return status, fmt.Errorf("missing apiServerURL in fleet config for cluster auto registration")
+			// Current config cannot be deployed, so remove the "config changed" mark
+			logrus.Warnf("cannot import cluster '%s/%s', missing apiServerURL in fleet config for cluster auto registration", cluster.Namespace, cluster.Name)
+			status.AgentConfigChanged = false
```
aruiz14, line 313: The problem with this is that the KubeconfigSecret could eventually be updated to contain a valid apiServerURL key, but the code that checks if changes are needed is `onConfig`, which won't be triggered.

aruiz14: Solved by adding a new handler on Secrets that will retrieve all Clusters referencing a Secret (efficiently, using an indexer), then trigger the config check for them.

internal/cmd/controller/agentmanagement/controllers/cluster/import.go
```diff
@@ -77,47 +87,60 @@ func RegisterImport(
 	clusters.OnChange(ctx, "import-cluster", h.OnChange)
 	fleetcontrollers.RegisterClusterStatusHandler(ctx, clusters, "Imported", "import-cluster", h.importCluster)
 	config.OnChange(ctx, h.onConfig)
+
+	clustersCache := clusters.Cache()
```
weyfonk, line 91: I wonder how expensive the `Cache()` operation might be; would it make sense to do this?
```suggestion
	clustersCache := h.clustersCache
```

internal/cmd/controller/agentmanagement/controllers/cluster/import.go
```diff
@@ -77,47 +87,60 @@ func RegisterImport(
...
+	clustersCache.AddIndexer(clusterForKubeconfigSecretIndexer, func(cluster *fleet.Cluster) ([]string, error) {
+		if cluster == nil || len(cluster.Spec.KubeConfigSecret) == 0 {
+			return []string{}, nil
+		}
+		secretKey := getKubeConfigSecretNS(cluster) + "/" + cluster.Spec.KubeConfigSecret
+		return []string{secretKey}, nil
+	})
+	secrets.OnChange(ctx, "kubeconfig-secrets-watch", func(key string, secret *corev1.Secret) (*corev1.Secret, error) {
+		clusters, err := clustersCache.GetByIndex(clusterForKubeconfigSecretIndexer, key)
```
weyfonk, line 100: Does this mean that more than one clusters can share the same key, therefore the same `KubeConfigSecret`? :thinking:

internal/cmd/controller/agentmanagement/controllers/cluster/import_test.go
```diff
@@ -89,39 +89,79 @@ func TestOnConfig(t *testing.T) {
...
+					clusters:      clustersController,
+					clustersCache: clustersCache,
+					secretsCache:  secretsCache,
+				}
+			},
+		},
+		"non-ready config and no URL or CA in secret, do not trigger import when CA changes": {
+			cfg: config.Config{
+				APIServerURL:              "",
+				AgentTLSMode:              "system-store",
```
weyfonk: nit: this test case could even exhibit a TLS mode update, to demonstrate that other changes would not be taken into account anyway if no API server URL is available:
```suggestion
				AgentTLSMode:              "strict",
```

weyfonk (approved):
Nice one! Leaving a couple of questions and a nitpick, but LGTM.

internal/cmd/controller/agentmanagement/controllers/cluster/import.go
aruiz14: Good question.
It's actually not expensive, since using `.Cache()` on the same `Controller` will provide the same underlying storage:
https://github.com/rancher/wrangler/blob/main/pkg/generic/controller.go#L271-L277
`Cache` is just a wrapper around the indexer from the sharedInformer configured by the controller.

Having said that, this is only "cheap" because we already had handlers/controllers registered for the `Cluster` kind. If we only had a Client, converting it into a Cache will indeed incur into the costs of populating that cache.

internal/cmd/controller/agentmanagement/controllers/cluster/import.go
aruiz14: It's technically possible, yes, because we don't see any restriction on the unicity of `Spec.KubeConfigSecret` across Clusters, although it may not make sense at all in the practice. Most of the times, `clusters` will only have a length of 0 or 1.

internal/cmd/controller/agentmanagement/controllers/cluster/import_test.go
aruiz14: Great point, thanks!

internal/cmd/controller/agentmanagement/controllers/cluster/import.go
weyfonk: Thanks for clarifying :bow: .

internal/cmd/controller/agentmanagement/controllers/cluster/import.go
weyfonk: Indeed, my understanding is that a length of more than 1 would, given the way the index key is built, imply multiple clusters sharing a namespace, which would cause other issues for Fleet.

PR #3840, opened 2025-06-26
Separate E2E sharding and metrics tests

p-se:
Refers to #XXX

### Checklist

### DIFF
--- a/.github/workflows/e2e-ci.yml
+++ b/.github/workflows/e2e-ci.yml
@@ -30,6 +30,9 @@ jobs:
           - name: default
           - name: sharding
             shards: '[{"id":"shard0"},{"id":"shard1"},{"id":"shard2"}]'
+          - name: sharded-metrics
+            shards: '[{"id":"shard0"}]'
+          - name: metrics
           - name: infra-setup
     name: e2e-fleet-test-${{ matrix.k3s.name }}-${{ matrix.test_type.name }}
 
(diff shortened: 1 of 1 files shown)


p-se:
I've separate sharding from metrics tests, since running the sharding tests apparently took the longest (closely followed by infra tests, which still do run a tiny part of metrics tests). Now the sharding tests are the first E2E tests that finish. We have more jobs and each jobs takes about 3 minutes for the setup. But we will see where the problem occurs more directly. I might do the same to the infra tests and extract metrics from them.

By a large margin, currently running the longest:

- metrics (~11m)
- sharded-metrics (~11m)
- infra-setup (untouched so far, 17m)

So, sharding tests never were the culprit and we got away from one test that took up to 17 minutes down to three, of which two take now up to roughly 11 minutes.

p-se:
The tiny metrics test in infra runs on my local machine without any setup in 22 seconds, so I guess we can rule that one out as a culprit (and I'd refrain from separating it from the infra tests).

p-se:
I have further split up the long running infra-test into
- git & helm
- OCI

Now the longest running tests in order (test seem to be slow currently) are:
- infra-setup-git-helm (16m & 18m)
- infra-setup-oci (12m & 13m)
- sharded-metrics (2x 11m)
- metrics (2x 10m)

p-se:
I could not find a reason for why `e2e-fleet-test-k3s-new-infra-setup` would be displayed, except as a remnant of Githubs caching.

PR #3850, opened 2025-07-02
Fix release against rancher script

thardeck:
Rancher 2.12 needs controller-gen.

While fixing the script I also updated the defaults.

### DIFF
--- a/.github/workflows/release-against-rancher.yml
+++ b/.github/workflows/release-against-rancher.yml
@@ -3,15 +3,15 @@ on:
   workflow_dispatch:
     inputs:
       rancher_ref:
-        description: "Submit PR against the following rancher/rancher branch (e.g. release/v2.7)"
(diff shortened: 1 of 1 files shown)


.github/workflows/release-against-rancher.yml
```diff
@@ -3,15 +3,15 @@ on:
   workflow_dispatch:
     inputs:
       rancher_ref:
-        description: "Submit PR against the following rancher/rancher branch (e.g. release/v2.7)"
+        description: "Submit PR against the following rancher/rancher branch (e.g. release/v2.12)"
         required: true
-        default: "release/v2.10"
+        default: "release/v2.11"
```
weyfonk, line 8: Since the description mentions v2.12 as an example, is v2.11 here deliberate? Or is it a typo?

thardeck: It is deliberate because atm rancher/rancher still uses `main` for Rancher v2.12.

Issue #3763, opened 2025-06-04
Support semantic versioning for HelmOps


weyfonk:
The `version` field of a HelmOp resource currently only supports static versions, such as `1.2.3`, and wildcard or empty values, leading Fleet to fetch the latest available chart.

However, users may need more flexibility in choosing the chart version Fleet will install. This could be achieved by means of [semantic versioning constraints](semver.org/), such as `1.x.x`, `< 2.0`, etc.

Acceptance criteria:
- [ ] Fleet validates semantic versioning constraints specified in a HelmOp's `version` field, updating the HelmOp's status with an error if a constraint is invalid
- [ ] Fleet updates the HelmOp's status with an error if a constraint does not match any available version for the chart
- [ ] Fleet installs the latest available chart version matching the constraint, if that constraint is valid

weyfonk:
## Additional QA
### Problem
Fleet only supported static or wildcard versions in HelmOps' `version` field. More flexibility is needed.
 
### Solution
Fleet now supports semantic versioning constraints, which may resolve to multiple versions: e.g. `1.x.x`, `< 2.0`, etc.
 
### Testing
### Engineering Testing
#### Manual Testing
N/A

#### Automated Testing
Integration tests for:
* invalid constraints
* constraints not matching any available chart version
* valid constraints

### QA Testing Considerations
Errors related to semantic versioning constraints should be visible in the `Accepted` condition of the HelmOp resource.
 
#### Regressions Considerations
Static, wildcard (`*`) and empty versions should still work as expected.

mmartin24:
Sharing here test results on `V2.12.0-alpha10` with Fleet `107.0.0+up0.13.0-alpha.5`

Testing environment: 1 simple cluster + 1 downstream cluster.
Testing sample file
```yaml
apiVersion: fleet.cattle.io/v1alpha1
kind: HelmOp
metadata:
  name: test-collision-simple
  namespace: fleet-default
spec:
  helm:
    releaseName: collide-demo
    repo: https://charts.bitnami.com/bitnami
    chart: nginx
    version: 13.2.0
    namespace: default
```
--- 

- Scenario 1: Simple correct version number should allow bundle deployment and be displayed -> OK

![Image](https://github.com/user-attachments/assets/645bb850-6cfd-498b-bc6f-597346a38e1c)
- Scenario 2: Invalid versions should trigger error (Setting version to `v.2.0`  -> OK

[...]

PR #3858, opened 2025-07-03
Check Helm for static version

weyfonk:
When a HelmOp resource references a static chart version, the HelmOps controller checks Helm for existence of that version.

This prevents errors about the version not existing in the Helm index from going undetected until a bundle deployment is created and an agent attempts to deploy it.
Consequently, this also ensures that errors about the version not existing are produced even if no target cluster is found for the HelmOp.

Refers to #3763
Follow-up to #3747.

### Checklist

### DIFF
--- a/internal/cmd/controller/helmops/reconciler/helmop_controller.go
+++ b/internal/cmd/controller/helmops/reconciler/helmop_controller.go
@@ -262,7 +262,7 @@ func (r *HelmOpReconciler) handleVersion(ctx context.Context, oldBundle *fleet.B
 
 	version, err := getChartVersion(ctx, r.Client, *helmop)
 	if err != nil {
-		return fmt.Errorf("could not get chart version: %w", err)
+		return err
 	}
 
 	if usesPolling(*helmop) {
--- a/internal/cmd/controller/helmops/reconciler/helmop_controller.go
+++ b/internal/cmd/controller/helmops/reconciler/helmop_controller.go
@@ -254,11 +254,6 @@ func (r *HelmOpReconciler) handleVersion(ctx context.Context, oldBundle *fleet.B
 		return fmt.Errorf("the provided HelmOp is nil; this should not happen")
 	}
 
-	if _, err := semver.StrictNewVersion(helmop.Spec.Helm.Version); err == nil {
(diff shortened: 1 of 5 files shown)


internal/cmd/controller/helmops/reconciler/helmop_controller.go
```diff
@@ -524,7 +519,7 @@ func helmChartSpecChanged(o *fleet.HelmOptions, n *fleet.HelmOptions, statusVers
 	}
 	// check also against statusVersion in case that Reconcile is called
 	// before the status subresource has been fully updated in the cluster (and the cache)
-	if o.Version != n.Version && statusVersion != o.Version {
+	if o.Version != n.Version && statusVersion == o.Version {
```
0xavi0, line 522: I was not sure if that extra check against `statusVersion` was still required, and tried running a test.
I think I've found a bug, because I don't see `helmop.Status.Version` being updated at all.

I've used the following yamls:
```yaml
kind: HelmOp
metadata:
  name: helmapp-test
  namespace: fleet-local
spec:
  namespace: test-helmops2
  helm:
    releaseName: mosquitto
    chart: mosquitto
    repo: https://storage.googleapis.com/t3n-helm-charts
    version: 2.4.1
    values:
      service:
        type: LoadBalancer
```

and 

```yaml
kind: HelmOp
metadata:
  name: helmapp-test
  namespace: fleet-local
spec:
  namespace: test-helmops2
  helm:
    releaseName: mosquitto
    chart: mosquitto
    repo: https://storage.googleapis.com/t3n-helm-charts
    version: 2.4.x
    values:
      service:
        type: LoadBalancer
```

(The only difference is the version field)

I've added a few logs and I can confirm the version is correct up to the call to 
```go
if err := c.Status().Patch(ctx, t, statusPatch); err != nil {
			return err
		}
```

but I don't see the `Status` value being set (don't know if it's overwritten by the Status controller)

```bash
$ kubectl get helmop -n fleet-local helmapp-test -ojsonpath={.status.version}


```

0xavi0:
I've created https://github.com/rancher/fleet/pull/3884 to continue with the `Status.Version` issue

Issue #3435, opened 2025-03-11
Helm Ops Part 2


manno:
Follow up for  https://github.com/rancher/fleet/issues/2962
Updated from the RFC

* [x] make sure GitOps bundle is not overwritten by helmapp of the same name, collision (see #3762)
* [x] support three chart sources: tarball, helm oci registry, helm http repo (includes all options, see #3589)
* [x] add polling for registry, repo, allow semver constraints `>= 1.0`
* [x] document under https://fleet.rancher.io/enableexperimental
* [x] rename structs, crd to `HelmOp`
* fields in struct (notes from discussion with francesco):
  * [ ] `spec.YAML` unusable?
  * [ ] `HelmApp.spec.helm.valuesFiles`  is a list of strings pointing to file paths. In Helm CLI context, the files are in the local system, I'm not sure in Fleet context where they are
* [ ] In `HelmAppSpec`, `HelmSecretName` and `InsecureSkipTLSverify` appear unused. There is `BundleSpec.HelmAppOptions.*` which is copied into them for the bundle options.
* [x] copy api fields instead of embedding? 
    * Still embedding for now to avoid having 2 distinct bundle types, which would then require 2 very similar bundle reconcilers
* [x] use a new secret type for the helmapp secret
* [ ] ~what about pull image secrets, same namespace, different registries?~ See #3705 

Ideas:
* optional: cleaner deployment options than GitRepo? Don't duplicate code comments?
  *  less targetCustomization options (no url changes?)
* UI extension for marketplace?

PR #3865, opened 2025-07-07
Add HelmOps to cluster status and metrics

weyfonk:
HelmOp resources were previously missing from cluster statuses and metrics.
They are now reflected there next to GitRepos, with 2 new status fields:
* `DesiredReadyHelmOps`
* `ReadyHelmOps`

Refers to #3435
Related to https://github.com/rancher/dashboard/issues/14713

### Checklist


p-se (approved):

p-se:
I wonder if we should try to put those additional two metrics into the Grafana dashboards.

Issue #3855, opened 2025-07-02
insecure-registries no longer respected in fleet-cli


rscharer:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

In fleet-cli version 0.12.3, if I refer to a chart in a local registry (e.g. oci://localhost:5001/test-chart), fleet correctly pulls the chart via HTTP rather than HTTPs. As of 0.12.4 and 0.13-beta2, however, the apply fails with a message like

``` helm chart download: failed to perform "FetchReference" on source: Get "https://localhost:5001/v2/test-chart/manifests/0.1.0": http: server gave HTTP response to HTTPS client```

Is there a new switch we need to flip to enable the old behavior, or was this an accidental regression?

### Expected Behavior

The chart applies correctly

### Steps To Reproduce

1. Install fleet-cli 0.12.4 via homebrew or by compiling source
2. Ensure that localhost:5001 is in the list of insecure registries on your local docker
3. Start up a registry:2 container
4. Push a local chart to registry:2
5. Attempt to apply a local bundle that references the chart

### Environment

```markdown
- Architecture: arm64
- Fleet Version: 0.12.4
- Cluster: N/A
```

### Logs

```markdown

```

p-se:
Can you please explain how exactly you have referenced the local chart in the bundle?

rscharer:
Yes, we're using a targetCustomization to override the remote registry value for local testing:
```
targetCustomizations:
  - name: '0.1.0'
    helm:
      version: '0.1.0'
      chart: 'oci://localhost:5001/test-chart'
    clusterSelector:
      matchLabels:
        ourcompany.com/test-chart: '0.1.0'
```

p-se:
There is an issue with how v0.12.4 handles it compared to previous releases. v0.13.0-beta2 might or might not have it, since there's no `--oci-insecure` and `--oci-basic-http` CLI args anymore, but only the option to provide a `--oci-registry-secret`, which I have not tested.

### CLI Options

```
==> v0.12.0
fleet-linux-amd64-v0.12.0 apply -h | grep oci
      --oci-basic-http                         Use HTTP to access the OCI regustry
      --oci-insecure                           Allow connections to OCI registry without certs
      --oci-password-file string               Path of file containing basic auth password for OCI registry
      --oci-reference string                   OCI registry reference
      --ociusername string                     Basic auth username for OCI registry
==> v0.12.1
[...]

PR #3872, opened 2025-07-09
Adds InsercureTLS and BasicHTTP to helm secret

0xavi0:
After a recent Helm update the code used to download a helm chart from an OCI registry no longer supports basic HTTP. This PR adds 2 new values read from the helm secret and passes the values to fleet apply.

The new values are:
* `insecureSkipTLS` : skips TLS checks when set to true
* `basicHTTP` : uses basic HTTP when set to true

The values that are not a correct boolean values set to true are ignored as those values are security related.

Refers to https://github.com/rancher/fleet/issues/3855

### Checklist


internal/cmd/cli/apply.go
```diff
@@ -58,6 +58,8 @@ type Apply struct {
 	KeepResources               bool              `usage:"Keep resources created after the GitRepo or Bundle is deleted" name:"keep-resources"`
 	DeleteNamespace             bool              `usage:"Delete GitRepo target namespace after the GitRepo or Bundle is deleted" name:"delete-namespace"`
 	HelmCredentialsByPathFile   string            `usage:"Path of file containing helm credentials for paths" name:"helm-credentials-by-path-file"`
+	HelmBasicHTTP               bool              `usage:"Uses plai HTTP connections when downloading from helm repositories" name:"helm-basic-http"`
```
weyfonk:
```suggestion
	HelmBasicHTTP               bool              `usage:"Uses plain HTTP connections when downloading from helm repositories" name:"helm-basic-http"`
```

integrationtests/cli/apply/helm_test.go
```diff
@@ -229,6 +237,66 @@ func testHelmRepo(path, port string) {
...
+		Expect(err).ToNot(HaveOccurred())
+
+		pwd, err := os.Getwd()
+		Expect(err).NotTo(HaveOccurred())
+		relTmpDir, err = filepath.Rel(pwd, tmpDir)
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	AfterEach(func() {
+		Expect(container.Terminate(context.Background())).NotTo(HaveOccurred())
```
weyfonk: Does the container need to be started and stopped after each test case? Or wouldn't it rather make sense to use an `Ordered` tag and make setting up and tearing down the environment `BeforeAll` and `AfterAll` blocks, respectively?

internal/cmd/controller/gitops/reconciler/gitjob.go
```diff
@@ -777,12 +792,13 @@ func volumes(targetsConfigName string) ([]corev1.Volume, []corev1.VolumeMount) {
...
 // volumesFromSecret generates volumes and volume mounts from a Helm secret, assuming that that secret exists.
 // If the secret has a cacerts key, it will be mounted into /etc/ssl/certs, too.
+// It also returns the boolean values of the InsecureSkipTLS and BasicHTTP keys.
 func volumesFromSecret(
 	ctx context.Context,
 	c client.Client,
 	namespace string,
 	secretName, volumeName, mountPath string,
-) ([]corev1.Volume, []corev1.VolumeMount, bool) {
+) ([]corev1.Volume, []corev1.VolumeMount, bool, bool, bool) {
```
weyfonk: I am a bit concerned that this function is becoming increasingly harder to maintain, now with 5 return values...
Refactoring ideas:
- return a `struct` instead
- or move the logic reading the secret and extracting its values into a separate function, distinct from this function mounting the secret

internal/cmd/controller/gitops/reconciler/gitjob.go
```diff
@@ -327,9 +327,11 @@ func (r *GitJobReconciler) newJobSpec(ctx context.Context, gitrepo *v1alpha1.Git
 
 	volumes, volumeMounts := volumes(configMap.Name)
 	var certVolCreated bool
+	var helmInsecure bool
+	var helmBasicHTTP bool
 
 	if gitrepo.Spec.HelmSecretNameForPaths != "" {
-		vols, volMnts, hasCertVol := volumesFromSecret(ctx, r.Client,
+		vols, volMnts, hasCertVol, _, _ := volumesFromSecret(ctx, r.Client,
```
weyfonk: If `HelmSecretNameForPaths` is specified, then `insecureSkipTLS` and `basicHTTP` will not be read, but each path may have its own `insecureSkipVerify` value.
Is the new `insecureSkipTLS` option deliberately named that way, or should both be named `insecureSkipVerify` for consistency? Would we want path-specific credentials to support `basicHTTP` as well?

... which leads me to a broader question: how much sense does it make to specify a secret with credentials for accessing a Helm or OCI registry, and to force use of basic HTTP? Wouldn't that introduce a risk of credentials leak?
In other words, should `basicHTTP` only be supported when no `username`/`password` combination is provided in the secret, in which case none of the other fields (e.g. `caBundle`, and of course `sshPrivateKey`) would make sense?

Taking this one step further, shouldn't `basicHTTP` live outside of a secret instead, for instance as a Helm option, mutually exclusive with a `helmSecretName`/`helmSecretNameForPaths`, where the presence of both would either lead to an error, or to a warning with `basicHTTP` being ignored for security reasons?

Happy to discuss. Am I missing an important aspect?

internal/cmd/cli/apply.go
0xavi0: Typo fixed

integrationtests/cli/apply/helm_test.go
0xavi0: Changed as suggested

internal/cmd/controller/gitops/reconciler/gitjob.go
0xavi0: Changed to return a `struct`.
It would be, maybe, better to refactor this whole function as it does too many things, and I think there are bits that should be revisited. For example: it checks for `CACerts` also for `HelmSecretNameForPaths` in the same way it does it for a single `HelmSecretName`.
When using `HelmSecretNameForPaths` the CA Bundles are defined for each path, not globally.

I've chosen the option that involves changing less code for that reason.

internal/cmd/controller/gitops/reconciler/gitjob.go
0xavi0: I’ve changed the option to `insecureSkipVerify` so that it matches the behavior used with `HelmSecretNameForPaths`.

When `HelmSecretNameForPaths` is used, `basicHTTP` will need to be defined for each path, since each path has its own options.

As for whether `basicHTTP` should only be allowed depending on whether credentials are used or not, IMHO I believe it (along with insecureSkipVerify) is an option that should be recommended only for testing environments.

I wouldn’t recommend using either of these options in a client-facing production setup.

In fact, the client used by GoGetter could already perform that security check itself, but it allows bypassing it.
I think we should clearly document what these options imply, but I wouldn’t go any further—since a client might want to test their registry in a test environment first (with credentials and all) before moving to production.

Regarding the relationship between `HelmSecretNameForPaths` and `HelmSecretName`, the code gives precedence to the former, but in no case are both considered at the same time.

integrationtests/cli/apply/helm_test.go
```diff
@@ -73,6 +77,66 @@ var _ = Describe("Fleet apply helm release", Serial, func() {
...
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("fails when calling fleet apply not passing helm-basic-http", func() {
+		err := fleetApply("helm", []string{relTmpDir}, apply.Options{})
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("http: server gave HTTP response to HTTPS client"))
+	})
+
+	It("fails when calling fleet apply not passing helm-basic-http=false", func() {
```
weyfonk:
```suggestion
	It("fails when calling fleet apply passing helm-basic-http=false", func() {
```

integrationtests/cli/apply/helm_test.go
```diff
@@ -73,6 +77,66 @@ var _ = Describe("Fleet apply helm release", Serial, func() {
...
+		Expect(err.Error()).To(ContainSubstring("http: server gave HTTP response to HTTPS client"))
+	})
+
+	It("works fine when calling fleet apply passing helm-basic-http=true", func() {
+		err := fleetApply("helm", []string{relTmpDir}, apply.Options{Auth: bundlereader.Auth{BasicHTTP: true}})
+		Expect(err).ToNot(HaveOccurred())
+	})
+
+	AfterAll(func() {
+		Expect(container.Terminate(context.Background())).NotTo(HaveOccurred())
```
weyfonk, line 136: nit:
```suggestion
		Expect(container.Terminate(context.Background())).NotTo(HaveOccurred())
		os.RemoveAll(tmpDir)
```

e2e/single-cluster/helm_auth_test.go
```diff
@@ -47,14 +47,96 @@ var _ = Describe("GitRepo using Helm chart with auth", Label("infra-setup"), fun
...
+				)
+				Expect(err).ToNot(HaveOccurred(), out)
+
+				// create the gitrepo content on the fly, to set the external IP
+				// of the load balancer so it complains about the certificate.
+				repoBaseDir := createFleetYAMLWithExternalIP(k, gitRepoPath)
+				tmpdir, gh = setupGitRepo(repoBaseDir, gitRepoPath, repoName, port)
+
+				DeferCleanup(func() {
+					_, _ = k.Delete("secret", helmSecretName)
```
weyfonk, line 71:
```suggestion
					_, _ = k.Delete("secret", helmSecretName)
					os.RemoveAll(repoBaseDir)
					os.RemoveAll(tmpdir)
```
(also applies to the next test case)

internal/cmd/controller/gitops/reconciler/gitjob.go
```diff
@@ -777,12 +798,13 @@ func volumes(targetsConfigName string) ([]corev1.Volume, []corev1.VolumeMount) {
 
 // volumesFromSecret generates volumes and volume mounts from a Helm secret, assuming that that secret exists.
 // If the secret has a cacerts key, it will be mounted into /etc/ssl/certs, too.
+// It also returns the boolean values of the InsecureSkipTLS and BasicHTTP keys.
```
weyfonk:
```suggestion
// It also returns a struct containing boolean values indicating if a volume has
// been created for CA bundles, along with values (defaulting to false) of the
// `insecureSkipVerify` and `basicHTTP` keys of the secret.
```

internal/cmd/controller/gitops/reconciler/gitjob.go
weyfonk: Thanks, that's a good first step :)

internal/cmd/controller/gitops/reconciler/gitjob.go
weyfonk: Right, we should indeed document that neither `basicHTTP` not `insecureSkipVerify` are recommended for production use, with a very visible warning :+1:

integrationtests/cli/apply/helm_test.go
0xavi0: Fixed

integrationtests/cli/apply/helm_test.go
0xavi0: Ginkgo takes care of deleting the folder.
From their [docs](https://onsi.github.io/ginkgo/):

_(here we're using GinkgoT().TempDir() to access Ginkgo's implementation of t.TempDir() which cleans up after itself - there's no magic here. You could have simply called os.MkdirTemp and cleaned up afterwards yourself.)_

e2e/single-cluster/helm_auth_test.go
0xavi0: `repoBaseDir` is also created with `GinkgoT().TempDir()`
I changed `tmpdir`,which was created that way before this PR, to also use `GinkgoT().TempDir()`

internal/cmd/controller/gitops/reconciler/gitjob.go
0xavi0: Changed as suggested

integrationtests/cli/apply/helm_test.go
weyfonk: Oh you're right, nothing to see here then :)

Issue #3883, opened 2025-07-10
HelmOps main controller does not update Status.Version


0xavi0:
The call to `Patch` in the main `helmops` controllers does not really update the Version when we are not running polling, because the Version found in the original object and the Version found in the new object are identical.

When calcultating the patch Version is ignored because it is the same in both objects.

0xavi0:
# QA Template

1. Apply a `HelmOp` with NO polling, using a fixed helm chart version. 
2. Check that `.Status.Version` has the version value in the `HelmOp`

You can also check with a non-fixed version (something like 1.0.x), the important bit is to not use polling.

mmartin24:
Verified in Rancher `v2.12-122ba407047ed762689d3778c1bdafb4fbb91004-head` with Fleet `fleet:107.0.0+up0.13.0-beta.4`

---

Used for test this file which expects to deploy latest grafana chart (`2.9.10` at the time of this test):

```yaml
apiVersion: fleet.cattle.io/v1alpha1
kind: HelmOp
metadata:
  name: grafana-test
  namespace: fleet-local
spec:
  correctDrift:
    enabled: false
  helm:
    chart: grafana
    releaseName: grafana
    repo: https://grafana.github.io/helm-charts
    version: 9.2.x
  pollingInterval: 0s
  targets:
    - clusterSelector:
        matchExpressions:
          - key: provider.cattle.io
            operator: NotIn
            values:
              - harvester
```

Status version shown: 
<details close>
  <summary>grafana_result.yaml</summary>
  
  ```yaml
  apiVersion: fleet.cattle.io/v1alpha1
kind: HelmOp
metadata:
  creationTimestamp: '2025-07-15T12:29:04Z'
  finalizers:
    - fleet.cattle.io/helmop-finalizer
  generation: 2
  managedFields:
    - apiVersion: fleet.cattle.io/v1alpha1
      fieldsType: FieldsV1
      fieldsV1:
        f:metadata:
          f:finalizers:
            .: {}
            v:"fleet.cattle.io/helmop-finalizer": {}
      manager: fleetcontroller
      operation: Update
      time: '2025-07-15T12:29:04Z'
    - apiVersion: fleet.cattle.io/v1alpha1
      fieldsType: FieldsV1
      fieldsV1:
        f:spec:
          .: {}
          f:correctDrift: {}
          f:helm:
            .: {}
            f:chart: {}
            f:releaseName: {}
            f:repo: {}
            f:version: {}
[...]

PR #3884, opened 2025-07-10
Fix version not being set in non-polling Helmops

0xavi0:
This fixes the problem of Version not being set in the Status of HelmOps resources that does not use polling.

The fix could be temporary until a more elegant way is implemented.

Refers to: https://github.com/rancher/fleet/issues/3883

### Checklist

### DIFF
--- a/internal/cmd/controller/helmops/reconciler/helmop_controller.go
+++ b/internal/cmd/controller/helmops/reconciler/helmop_controller.go
@@ -438,12 +438,12 @@ func usesPolling(helmop fleet.HelmOp) bool {
 // updateStatus updates the status for the HelmOp resource. It retries on
 // conflict. If the status was updated successfully, it also collects (as in
 // updates) metrics for the HelmOp resource.
-func updateStatus(ctx context.Context, c client.Client, req types.NamespacedName, orig *fleet.HelmOp, orgErr error) error {
-	if orig == nil {
+func updateStatus(ctx context.Context, c client.Client, req types.NamespacedName, helmop *fleet.HelmOp, orgErr error) error {
+	if helmop == nil {
 		return fmt.Errorf("the HelmOp provided for a status update is nil; this should not happen")
 	}
 
-	objToPatchFrom := orig.DeepCopy()
+	objToPatchFrom := helmop.DeepCopy()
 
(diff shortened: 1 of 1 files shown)


internal/cmd/controller/helmops/reconciler/helmop_controller.go
```diff
@@ -443,14 +443,26 @@ func updateStatus(ctx context.Context, c client.Client, req types.NamespacedName
 		return fmt.Errorf("the HelmOp provided for a status update is nil; this should not happen")
 	}
 
+	objToPatchFrom := orig.DeepCopy()
```
manno: This variable is not really like the others we name `orig`:

```suggestion
	objToPatchFrom := helmOp.DeepCopy()
```

internal/cmd/controller/helmops/reconciler/helmop_controller.go
```diff
@@ -470,7 +482,7 @@ func updateStatus(ctx context.Context, c client.Client, req types.NamespacedName
 
 		setAcceptedConditionHelm(&t.Status, orgErr)
 
-		statusPatch := client.MergeFrom(orig)
+		statusPatch := client.MergeFrom(objToPatchFrom)
```
manno, line 485: We could also DeepCopy the helmOp in the reconciler and pass it to `updateStatus`, like in internal/cmd/agent/controller/bundledeployment_controller.go
However, this is the smaller change?

0xavi0: yeah, I didn't want to do that because this change covers a smaller part of the code and we're close to bug freeze.
(I'm not sure if we could be breaking something else by passing a DeepCopy calculated earlier)
Also... the DeepCopy would not have the version value we calculated in this reconcile call.

We should definitely revisit this...

internal/cmd/controller/helmops/reconciler/helmop_controller.go
0xavi0: I've changed the parameter name to `helmop`

manno:
Is it necessary to run Patch in Retry? We should compare all our `updateStatus` funcs and make sure they are all good.

PR #3889, opened 2025-07-14
Fix race condition creating options secret

aruiz14:
Follow up of #3527

I ran into a problem where BundleDeployments weren't properly initialized because their secret didn't exist, but the step that `CreatesOrUpdate` this secret depends on this `builder.Targets` success.

### Checklist

### DIFF
--- a/internal/cmd/controller/target/builder.go
+++ b/internal/cmd/controller/target/builder.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/Masterminds/sprig/v3"
 	"github.com/go-logr/logr"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 
 	"github.com/rancher/fleet/internal/cmd/controller/options"
 	"github.com/rancher/fleet/internal/cmd/controller/target/matcher"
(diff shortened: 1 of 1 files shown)


internal/cmd/controller/target/builder.go
```diff
@@ -122,24 +123,30 @@ func (m *Manager) Targets(ctx context.Context, bundle *fleet.Bundle, manifestID
 
 	byNamespace := map[string]*fleet.BundleDeployment{}
 	for _, bd := range bundleDeployments.Items {
+		bd := bd.DeepCopy()
```
aruiz14, line 126: `bundleDeployments.Items` may be the result from reading from the cache. Since we perform changes in this object, it's safer to work with a copy of it.

manno: I expect it to be safe unless we disable the deep copy explicitly:
https://github.com/kubernetes-sigs/controller-runtime/blob/v0.22.0/pkg/cache/internal/cache_reader.go#L170-L177

kkaempf:
What's the status of this PR ?

weyfonk:
Superseded by #4284, which takes new error reporting into account and somewhat simplifies bundle controller unit tests.
Thanks for the idea :)

Issue #3866, opened 2025-07-08
GitRepo has incorrect `status.resources`


torchiaf:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Steps To Reproduce

0. Prepare an environment with local cluster + 2 downstream cluster
1. Create a new GitRepo
2. URL: `https://github.com/rancher/fleet-examples` , path: `simple`
3. Deployment `redis-slave` has to be in `updating` status in 1 cluster and `ready` in 1 the other cluster

cluster: `fleet-default/rc-ec2` - deployment is not ready
<img width="931" height="404" alt="Image" src="https://github.com/user-attachments/assets/89b77d06-3ffa-4fbc-9045-8e48eca4b8d8" />

cluster: `fleet-default/c-rkrnf` - deployment is ready
<img width="946" height="359" alt="Image" src="https://github.com/user-attachments/assets/04e85f2e-b3ba-4b4a-8803-4942fa30325a" />

`GitRepo.status.resources[index]` is incorrect:

```
    - apiVersion: apps/v1
      id: default/redis-slave
      kind: Deployment
      name: redis-slave
      namespace: default
      perClusterState:
        ready:
          - fleet-default/c-rkrnf
      state: updating
      type: apps.deployment
```

It should be:

```
    - apiVersion: apps/v1
      id: default/redis-slave
      kind: Deployment
      name: redis-slave
      namespace: default
      perClusterState:
        ready:
          - fleet-default/c-rkrnf
        updating:
          - fleet-default/rc-ec2
      state: updating
      type: apps.deployment
```

### Environment

```markdown
- Architecture:
- Fleet Version: `rancher/fleet-agent:v0.13.0-beta.2`
- Cluster:
  - Provider:
  - Options:
  - Kubernetes Version:
```

### Logs

```markdown

```

p-se:
Can you explain how to reproduce this issue? I've tried with four downstream clusters (as I had that running locally already) and then removed a resource on downstream1. The `status.resources` field reports properly:

```yaml
  resources:
  - apiVersion: apps/v1
    id: default/frontend
    kind: Deployment
    name: frontend
    namespace: default
    perClusterState:
      missing:
      - fleet-default/downstream1
      ready:
      - fleet-default/downstream2
      - fleet-default/downstream3
      - fleet-default/downstream4
    state: Missing
    type: apps.deployment
```

I did use the suggested https://github.com/rancher/fleet-examples repo with path `simple`.

[...]

PR #3890, opened 2025-07-15
Treat updating status as notReady for perClusterState

p-se:
Converts
```yaml
  resources:
  - apiVersion: apps/v1
    id: default/foo
    kind: Deployment
    name: foo
    namespace: default
    perClusterState: {}  # <--
    state: updating
    type: apps.deployment
```
to
```yaml
  resources:
  - apiVersion: apps/v1
    id: default/foo
    kind: Deployment
    name: foo
    namespace: default
    perClusterState:       # <--
      notReady:            # <--
      - fleet-local/local  # <--
    state: updating
    type: apps.deployment
```

Refers to #3866

### Checklist

### DIFF
--- a/internal/resourcestatus/resourcekey_test.go
+++ b/internal/resourcestatus/resourcekey_test.go
@@ -214,3 +214,94 @@ func TestSetResources(t *testing.T) {
 	}, status.PerClusterResourceCounts)
 
 }
+
+func TestPerClusterState(t *testing.T) {
+	bundleDeploymentWithState := func(state string) fleet.BundleDeployment {
+		return fleet.BundleDeployment{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "bd1",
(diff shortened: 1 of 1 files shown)


internal/resourcestatus/resourcekey_test.go
```diff
@@ -214,3 +214,94 @@ func TestSetResources(t *testing.T) {
...
+						Name:      "web",
+						PerClusterState: fleet.PerClusterState{
+							NotReady: []string{"namespace/cluster"},
+						},
+					},
+				},
+			},
+		},
+		{
+			name:              "if the state of the NonReadyStatus resource is unknown, then it should report ignore the state",
```
weyfonk: nit:
```suggestion
			name:              "if the state of the NonReadyStatus resource is unknown, then it should ignore the state",
```

weyfonk (approved):

PR #3891, opened 2025-07-15
CLI arguments to disable leader election

p-se:
CLI arguments to disable leader election for
- fleet agent and
- fleet controller.

Consistent with how they are implemented for [GitJob](https://github.com/rancher/fleet/blob/578682ca34b91daab509dd76849ccf911f7c86d6/internal/cmd/controller/gitops/operator.go#L54).

### Checklist

### DIFF
--- a/internal/cmd/agent/root.go
+++ b/internal/cmd/agent/root.go
@@ -33,7 +33,8 @@ type FleetAgent struct {
 	command.DebugConfig
 	AgentScope string `usage:"An identifier used to scope the agent bundleID names, typically the same as namespace" env:"AGENT_SCOPE"`
 	UpstreamOptions
-	CheckinInterval string `usage:"How often to post cluster status" env:"CHECKIN_INTERVAL"`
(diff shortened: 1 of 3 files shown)


internal/cmd/agent/root.go
```diff
@@ -33,7 +33,8 @@ type FleetAgent struct {
 	command.DebugConfig
 	AgentScope string `usage:"An identifier used to scope the agent bundleID names, typically the same as namespace" env:"AGENT_SCOPE"`
 	UpstreamOptions
-	CheckinInterval string `usage:"How often to post cluster status" env:"CHECKIN_INTERVAL"`
+	CheckinInterval      string `usage:"How often to post cluster status" env:"CHECKIN_INTERVAL"`
+	EnableLeaderElection bool   `name:"leader-elect" default:"true" usage:"Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager."`
```
weyfonk:
```suggestion
	EnableLeaderElection bool   `name:"leader-elect" default:"true" usage:"Enable leader election for the agent. Enabling this will ensure there is only one active agent."`
```

Issue #1962, opened 2023-11-23
[RFE] add schemastore integration


olblak:
### Is your feature request related to a problem?

It is not always easy to identify available configuration for a `fleet.yaml`

### Solution you'd like

I would like to generate a jsonschema for Fleet using https://github.com/invopop/jsonschema
This schema could automatically be published on https://github.com/rancher/fleet-docs
And then referenced from https://github.com/schemastore/schemastore 

The final goal is to provide information to any IDE that support jsonschema store like VScode or Neovim that if we open a file named "fleet.yaml" with a parent directory name "fleet.d" then we could get documentation, validation, and auto-completion without additional configuration.

d3flex:
I have created a PR against fleet-product-docs and prepared another commit one for schemastore which is still on my local branch. 

those should complement two PR in the fleet side: one with the generation script and one for the CI validation.

However, the approach concentrates on the elimination of the constant burden of the maintenance and consistency of the schema on various places. fleet, docs and schemastore have one source of true and we never have to worry about docs and schemastore as they both points to the schema stored in the fleet which is updated and validated for each pull request.

d3flex:
https://github.com/SchemaStore/schemastore/pull/6209

which had already entry for "Rancher Fleet" with schema for 0.5 and 0.8 schemas. Both have removed

olblak:
Ideally the jsonschema should be generated automatically so we could automate the process au updating it

d3flex:
> Ideally the jsonschema should be generated automatically so we could automate the process au updating it

What I can tell is that any PR with inconsistent schema will fail, forcing the committer to commit the up to date one.

d3flex:
https://github.com/SchemaStore/schemastore/pull/6209 was merged. searching at https://www.schemastore.org/  will redirect you to https://raw.githubusercontent.com/rancher/fleet/main/schemas/fleet.yaml.json

sbulage:
## QA Verification — PASS

| Rancher | Fleet |
|---------|-------|
| v2.16-head | v0.17.0-alpha.3 |

**Verified on v0.17.0-alpha.3** (Rancher 2.16-head, k3s v1.36.1).

### Result

| # | Scenario | Result |
|---|----------|--------|
| 1 | Fresh install — annotation present | **PASS** |
| 2 | Matching version — hot-reload works | **PASS** |
| 3 | Mismatched version — hot-reload skipped | **PASS** |
| 4 | No annotation — hot-reload skipped | **PASS** |
[...]

PR #3897, opened 2025-07-16
Generating JSON Schema when CRDSs change

kakabisht:
Creating a new JSON Schema whenever the CRDs are modified. For more info, refer to https://github.com/rancher/fleet/issues/1962


.github/workflows/check-crds.yml
```diff
@@ -0,0 +1,59 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout this repo
+        uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.21'
```
olblak, line 20: This version of go is a bit outdated

.github/workflows/check-crds.yml
```diff
@@ -0,0 +1,59 @@
+name: Fleet CRD for JSON Schema
```
olblak, line 1: If interested, I could help contribute to a Updatecli manifest.
Here is an example of the manifest that I am using on Updatecli 
https://github.com/updatecli/website/blob/master/updatecli/updatecli.d/jsonschema.yaml

It opens PR like https://github.com/updatecli/website/

In this case it would be a bit different as we would need to use a second scm to execute the target from 

https://github.com/rancher/fleet

p-se:
I would like to propose a different process for updating. Instead of checking the changes made to CRDs when a PR has been merged, we can generate the JSON schema alongside the re-generation/update of the CRDs.

When a developer does changes to the Go structs we use for CRDs, the CRDs need to be regenerated (locally) and accompany the other changes in the PR of the developer. When could regenerate the JSON schemas alongside the regeneration of the CRDs, so that they will always be in sync with each other and also part of the PR that introduces the change. For that the generation of the schemas would need to happen when we call `go generate`. What happens when we call `go generate` is defined in [generate.go](https://github.com/rancher/fleet/blob/main/generate.go). 

That way all the changes would be part of the PR of the developer and CRDs and JSON schema would always be in sync.



That said, it might make sense to extract the JSON schema from the CRDs. Apparently they are contained in the CRDs. I'm worried that some comments that are used for the generation of the CRDs (and the JSON schema that are part of it) might not be picked up if we generate the JSON schemas separately from the same structs. This would also reduce our dependence on another library.


The other thing I noticed is that we might want to have separate JSON schemas for each CRD. JSON schemas are apparently most often mapped to file names. At least that's the case for how they are submitted to https://www.schemastore.org/.

Example from the front page: 

```json
{
  "name": "bower.json",
  "description": "Bower package description file",
  "fileMatch": [ "bower.json", ".bower.json" ],
  "url": "https://www.schemastore.org/bower.json"
}
```

Issue #3807, opened 2025-06-18
[SURE-10335] GitRepo Bundle shows hash mismatch error after upgrading rancher to 2.10.6 from 2.10.4


skanakal:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

After upgrading rancher from v2.10.4 to v2.10.6, one of the bundle  "'k8s005-uat2-deployments-bi-reports"of a GitRepo shows error:

Content does not match hash. Got x, expected y
And its corresponding deployment is not deployed in the cluster.

### Expected Behavior

The bundle should continue to work even after Rancher upgrade. 

### Environment

```markdown
- Architecture: 
- Fleet Version: 0.11.8
- Cluster:
  - Provider: RKE2
  - Options:
  - Kubernetes Version: 1.31
```

### Logs

```markdown
msg":"Reconciler error","controller":"bundle","controllerGroup":"fleet.cattle.io","controllerKind":"Bundle","Bundle":{"name":"deployments-xxx-xxx","namespace":"fleet-default"},"namespace":"fleet-default","name":"deployments-xxx-xxx","reconcileID":"cfcf9eeb-38e1-4192-b82f-2ba30bce1b3a","error":"Operation cannot be fulfilled on bundles.fleet.cattle.io \"deployments-xxx-xxx\": StorageError: invalid object, Code: 4, Key: /registry/fleet.cattle.io/bundles/fleet-default/deployments-xxx-xxx, ResourceVersion: 0, AdditionalErrorMsg: Precondition failed: UID in precondition: xxxxxxxxxx, UID in object meta: ","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.0/pkg/internal/controller/controller.go:316\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.0/pkg/internal/controller/controller.go:263\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.2\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.0/pkg/internal/controller/controller.go:224"}
```

### Anything else?

[...]

PR #3917, opened 2025-07-28
Calculate hash with all data, not only processed

0xavi0:
Please refer to the issue for more information.

Refers to https://github.com/rancher/fleet/issues/3807

### Checklist

### DIFF
--- a/internal/manifest/manifest.go
+++ b/internal/manifest/manifest.go
@@ -40,7 +40,7 @@ func FromJSON(data []byte, expectedSHAsum string) (*Manifest, error) {
 		return nil, err
 	}
 	m.raw = data
-	// #3807 Writing all the data to the hasher to avoid unprocessed data issues.
+	// Writing all the data to the hasher to avoid unprocessed data issues.
 	// See full details in #3807
 	h := sha256.New()
 	h.Write(data)


internal/manifest/manifest.go
```diff
@@ -35,14 +35,15 @@ func FromBundle(bundle *fleet.Bundle) *Manifest {
...
-	r := io.TeeReader(bytes.NewReader(data), h)
-
 	var m Manifest
-	if err := json.NewDecoder(r).Decode(&m); err != nil {
+	if err := json.NewDecoder(bytes.NewBuffer(data)).Decode(&m); err != nil {
 		return nil, err
 	}
 	m.raw = data
+	// #3807 Writing all the data to the hasher to avoid unprocessed data issues.
+	// See full details in #3807
```
weyfonk: nit: mentioning the issue number twice may not be needed.
```suggestion
	// Writing all the data to the hasher to avoid unprocessed data issues.
	// See full details in #3807
```

0xavi0:
I was thinking on adding a test, but all this depends on an implementation detail of the json Decoder.
Right now, the first block is 512 bytes long, and I thought to add a test with a json of 512 bytes (plus some extra bytes that would be discarded) but maybe in future versions they change the initial block size and that example will be useless (won't fail but will be testing something that adds no value)

The processed vs read block size is not fixed, it increases and also depends on the remaining buffer to be processed (not easy to track)

So... I didn't add any test because it is something very deep into the implementation of the Decoder.

Issue #3851, opened 2025-07-02
[SURE-5492] Fleet-agent: Wrong Signal Handling


kkaempf:
### Issue description:
With Kubernetes Version v1.23, 'Graceful Node Shutdown' Feature becomes GA. This feature requires all containers to handle at least SIGTERM signal for an graceful shutdown. This means a container stopped with SIGTERM have to return 0. Otherwise, those pods will be listed as 'error' when host has rebooted

### Business impact:
Medium

### Troubleshooting steps:
Create a single node k3s cluster with kubelet config ShutdownGracePeriod and ShutdownGracePeriodCriticalPods. Try rebooting the node with "systemctl reboot". Cluster-agent and fleet-agent status immediately changes into "error" with Exit code as 1 . Even though new cluster agent comes online , the old one still remains in an error state, which causes confusion while monitoring the agents

### Repro steps:

1) k3s config file(/etc/rancher/k3s/config.yaml ) with below content
~~~
kubelet-arg: "config=/etc/rancher/k3s/kubelet.config" 
~~~
2) kubelet config (/etc/rancher/k3s/kubelet.config) with below configuration
~~~
apiVersion: kubelet.config.k8s.io/v1beta1
kind: KubeletConfiguration
shutdownGracePeriod: 30s
shutdownGracePeriodCriticalPods: 10s 
~~~
3) Install k3s 
~~~
curl -sfL https://get.k3s.io |INSTALL_K3S_VERSION="v1.23.8+k3s2" K3S_TOKEN=SECRET sh -s - server --cluster-init 
~~~
4) Reboot the node:
~~~
systemctl reboot
~~~

### Workaround:
Is workararound available and implemented? no

### Actual behavior:

Cluster-agent and Fleet agent get * killed * with exit code 1
~~~
    State:          Running
      Started:      Fri, 28 Oct 2022 12:06:19 +0000
    Last State:     Terminated
      Reason:       Error
      Exit Code:    1
      Started:      Fri, 28 Oct 2022 12:02:57 +0000
      Finished:     Fri, 28 Oct 2022 12:06:06 +0000 
~~~
### Expected behavior:
Cluster-agent and Fleet agent should get gracefully terminated with a reason as "Completed" , possibly with exit code as 0
~~~
    Last State:     Terminated
      Reason:       Completed
      Exit Code:    0
      Started:      Fri, 28 Oct 2022 11:59:55 +0000
      Finished:     Fri, 28 Oct 2022 12:00:21 +0000 
[...]

PR #3918, opened 2025-07-28
Support graceful agent shutdown

weyfonk:
The Fleet agent must be able to properly handle `SIGTERM`, which amounts to exiting with error code 0 (instead of 1) when its context is canceled.

Refers to #3851.

### Checklist


weyfonk:
> Should we add a test for that, since the issue was a regression?

Good shout, I've added end-to-end tests in a separate commit (see `e2e/single-cluster/signals_test.go`).

Issue #3953, opened 2025-08-05
HelmOps does not support repositories with version having the "v" prefix in versions


0xavi0:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Some helm repositories out there list their versions like "v1.3.4" instead to the recommended semver compliant "1.3.4".
A couple of examples are:
* https://charts.jetstack.io (default repository for cert-manager)
* https://docs.tigera.io/calico/charts

Those repositories don't work with `HelmOps` because the `Bundle` reconciler checks for strict semver versions and we get the following error: "chart version cannot be deployed; check HelmOp status for more details: invalid characters in version".

Steps to recreate:
Deploy the following `HelmOp`:
```yaml
apiVersion: fleet.cattle.io/v1alpha1
kind: HelmOp
metadata:
  name: cert-manager
  namespace: fleet-local
spec:
  helm:
    releaseName: cert-manager
    chart: cert-manager
    repo: https://charts.jetstack.io
    version: 1.18.2
    values:
      crds:
        enabled: true
```

### Environment

```markdown
- Architecture:
- Fleet Version:
- Cluster:
  - Provider:
  - Options:
  - Kubernetes Version:
```

### Logs

```markdown

```

kkaempf:
Fixed in v2.12.1

PR #3955, opened 2025-08-05
Trims the v prefix in versions before checking if it's a valid semver

0xavi0:
As decribed in the related issue, some helm repositories identify versions using the "v" prefix (like v1.5.6). Although that's not recommended by Helm, those repositories are still valid.

This PR trims the "v" prefix (if exists) before checking if the version stored in the Bundle spec is semver compliant.

Refers to: https://github.com/rancher/fleet/issues/3953

### Checklist

### DIFF
--- a/internal/cmd/controller/reconciler/bundle_controller.go
+++ b/internal/cmd/controller/reconciler/bundle_controller.go
@@ -9,6 +9,7 @@ import (
 	"maps"
 	"reflect"
 	"slices"
+	"strings"
 	"time"
 
 	"github.com/Masterminds/semver/v3"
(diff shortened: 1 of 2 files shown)


internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -231,7 +232,13 @@ func (r *BundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	// version constraint. That constraint should be resolved into a strict version by the HelmOps reconciler before bundle
 	// deployments can be created.
 	if contentsInHelmChart && bundle.Spec.Helm != nil && len(bundle.Spec.Helm.Version) > 0 {
-		if _, err := semver.StrictNewVersion(bundle.Spec.Helm.Version); err != nil {
+		// #3953
+		// There are helm repositories that list their version with the "v" prefix.
```
```suggestion
		// There are helm repositories that list chart versions with the "v" prefix.
```

furkatgofurov7:
@0xavi0 thanks for the quick fix 👍🏼 Do we have approximate dates for this change to be shipped in the new Rancher version (I assume it will be v2.12.1, right)?

0xavi0:
> Tanks for the quick fix 👍🏼 Do we have approximate dates for this change to be shipped in the new Rancher version (I assume it will be v2.12.1, right)?

Yup, 2.12.1.
Once this is merged I will push a backport for 2.12 (this PR is going to `main`)

Issue #3907, opened 2025-07-22
Issue #21


weyfonk:
Just a reminder... (backport of https://github.com/rancher/fleet/issues/3832 to v0.11/Rancher v2.10)

mmartin24:
Please see the verification done [here](https://github.com/rancher/fleet-security/issues/19#issuecomment-3164971500).
If the extend of the verification is ok, please feel free to close this ticket.
Thanks @0xavi0 for the guidance here. 

CC: @kkaempf

kkaempf:
Discussed in weekly, closing.

PR #3962, opened 2025-08-06
[v0.11] Harden values files exclusion from bundle resources

weyfonk:
Values files may be referenced by file name alone, or with a directory prefix. Both variants now lead to values files being excluded from bundles resources, regardless of where a `fleet.yaml` file referencing them lives (inside or outside of a chart directory).

Refers to #3907
Backport of #3956 to `release/v0.11`.

### Checklist


thardeck:
E2E upgrade test fails in [general](https://github.com/rancher/fleet/pull/3966), will investigate this later.

PR #3990, opened 2025-08-13
Adds basicHTTP to HelmOps when using OCI registries

0xavi0:
BasicHTTP should be also supporter when downloading OCI stored helm charts for `HelmOps`

follow-up to: https://github.com/rancher/fleet/pull/3872
Refers to: https://github.com/rancher/fleet/issues/3855
### Checklist

### DIFF
--- a/internal/bundlereader/charturl.go
+++ b/internal/bundlereader/charturl.go
@@ -49,6 +49,10 @@ func ChartVersion(location fleet.HelmOptions, a Auth) (string, error) {
 
 		r.Client = authCli
 
+		if a.BasicHTTP {
+			r.PlainHTTP = true
+		}
+
 		tag, err := GetOCITag(r, location.Version)
 
 		if len(tag) == 0 || err != nil {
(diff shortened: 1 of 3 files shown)


internal/bundlereader/charturl.go
```diff
@@ -49,6 +49,10 @@ func ChartVersion(location fleet.HelmOptions, a Auth) (string, error) {
 
 		r.Client = authCli
 
+		if a.BasicHTTP {
+			r.PlainHTTP = true
+		}
```
weyfonk, line 54: How about `InsecureSkipVerify`? Is there a specific reason for not propagating it?

0xavi0: TLS is handled at a Client level (how we deal with the connection).
In line 37... the `getHTTPClient` is the one building the client with the desired TLS config. 

BasicHTTP (or plain HTTP) in handled at a repository level. 
When it builds the URL to access the chart it prefixes it with `http` or `https` depending on that value.

weyfonk: Got it, thanks for clarifying!

PR #4020, opened 2025-08-19
Fix Rancher test setup

thardeck:
small fixes for our dev rancher setup, which should make it more robust.

### DIFF
--- a/dev/setup-rancher-clusters
+++ b/dev/setup-rancher-clusters
@@ -7,10 +7,16 @@ if [ ! -d ./.github/scripts ]; then
   exit 1
 fi
 
-public_hostname="${public_hostname-172.18.0.1.sslip.io}"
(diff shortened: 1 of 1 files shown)


dev/setup-rancher-clusters
```diff
@@ -7,11 +7,17 @@ if [ ! -d ./.github/scripts ]; then
   exit 1
 fi
 
-public_hostname="${public_hostname-172.18.0.1.sslip.io}"
 upstream_ctx="${FLEET_E2E_CLUSTER-k3d-upstream}"
 downstream_ctx="${FLEET_E2E_CLUSTER_DOWNSTREAM-k3d-downstream1}"
 rancherpassword="${RANCHER_PASSWORD-rancherpassword}"
 
+public_hostname="${public_hostname:-}"
```
weyfonk, line 14: Yesss thanks for this :)

PR #4026, opened 2025-08-20
Improve Bundle Value Secret Logging (SURE-10613)

manno:
Refers to https://github.com/rancher/fleet/issues/4028

### DIFF
--- a/internal/cmd/controller/reconciler/bundle_controller.go
+++ b/internal/cmd/controller/reconciler/bundle_controller.go
@@ -221,11 +221,12 @@ func (r *BundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	// The values secret is optional, e.g. for non-helm type bundles.
 	// This sets the values on the bundle, which is safe as we don't update bundle, just its status
 	if bundle.Spec.ValuesHash != "" {
-		if msg, err := loadBundleValues(ctx, r.Client, bundle); err != nil {
+		if err := loadBundleValues(ctx, r.Client, bundle); err != nil {
+			if errors.Is(err, fleetutil.ErrRetryable) {
+				logger.Info(err.Error())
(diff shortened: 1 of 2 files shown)


internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -516,24 +520,28 @@ func (r *BundleReconciler) createBundleDeployment(
 }
 
 // loadBundleValues loads the values from the secret and sets them in the bundle spec
-func loadBundleValues(ctx context.Context, c client.Client, bundle *fleet.Bundle) error {
+func loadBundleValues(ctx context.Context, c client.Client, bundle *fleet.Bundle) (string, error) {
 	secret := &corev1.Secret{}
-	if err := c.Get(ctx, types.NamespacedName{Name: bundle.Name, Namespace: bundle.Namespace}, secret); err != nil {
-		return fmt.Errorf("failed to get values secret for bundle %q, this is likely temporary: %w", bundle.Name, err)
+	if err := c.Get(ctx, types.NamespacedName{Name: bundle.Name, Namespace: bundle.Namespace}, secret); apierrors.IsNotFound(err) {
```
weyfonk, line 526: nit: instead of relying on 2 return values here, since non-empty message implies empty error, we could return a specific error type (e.g. a package-level `var errRetry = errors.New("retrying to get values secret for bundle")`, or even an error wrapper with a flag indicating if it's retriable), and check on L173 if the error returned by `loadBundleValues` is of that type ([errors.Is](https://pkg.go.dev/errors#Is)).

weyfonk (approved):

internal/cmd/controller/reconciler/bundle_controller.go
manno: This seems a bit over engineered to me, just for `loadBundleValues`.  But if we want to use it in every controller and make it a pattern, sure.

PR #4056, opened 2025-08-25
Fix k3d dev scripts

thardeck:
To setup Rancher dev run the following
```
k3d cluster delete downstream1 upstream
./dev/setup-k3d && ./dev/setup-k3ds-downstream && ./dev/setup-rancher-clusters 2.12.0
```

### DIFF
--- a/.github/workflows/e2e-rancher-upgrade-fleet.yml
+++ b/.github/workflows/e2e-rancher-upgrade-fleet.yml
@@ -107,7 +107,6 @@ jobs:
         name: Install Rancher CLI
         if: steps.rancher-cli-cache.outputs.cache-hit != 'true'
         run: |
-          # download an older CLI to avoid https://github.com/rancher/rancher/issues/37574
           mkdir -p /home/runner/.local/bin
           wget -q https://github.com/rancher/cli/releases/download/v2.12.0/rancher-linux-amd64-v2.12.0.tar.gz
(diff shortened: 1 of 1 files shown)


manno:
Please do not modify the CI's .github/scripts

e-sp (dismissed):

p-se:
The changes seem to work me.

manno (changes requested):
If you need changes to the .github CI scripts, clone them to dev and change them there. 

See https://github.com/rancher/fleet/blob/main/.github/scripts/README.md

> Please do not modify them to accommodate your local workflow, as this often ends up in scripts that are hard to read and reason about.

thardeck:
> If you need changes to the .github CI scripts, clone them to dev and change them there.

@manno I agree with the policy but the CI scripts have a race conditions in some of their loops.
```
 Fix race conditions in Rancher dev setup loops

The `$!` variable refers to the PID of the most recent background process,
but in this process substitution pattern, there's a timing race between
when the background process starts and when grep finds the match.

The `kill $!` might execute before $! is properly set, or it might
reference a different process entirely, the script itself for example,
which leads to random exits of the rancher dev setup.
```


The Github action ["E2E Upgrade to Rancher HEAD"](https://github.com/rancher/fleet/actions/runs/17264297296/job/48992955446) still works as expected and I have also fixed the Github action ["E2E Upgrade Fleet in Rancher"](https://github.com/rancher/fleet/actions/runs/17264291315/job/48992687478).

Issue #3491, opened 2025-04-01
Pod Disruption Budgets


manno:
Fleet should support PCs and PDBs. They should taken from the existing `AgentDeploymentCustomization` struct.

see "RFC: Cattle Cluster Agent Priority Class And Pod Disruption Budgets"

p-se:
Reference implementation in Rancher https://github.com/rancher/rancher/issues/48995

p-se:
# QA Template

1. Setup Fleet on a cluster. Rancher is not required.

3. Edit the Fleet cluster resource, add `spec.agentSchedulingCustomization` like so:
   ```yaml
   agentSchedulingCustomization:
   priorityClass:
     value: 777
   podDisruptionBudget:
     minAvailable: "1"
   ```
   This should result in the creation of a `PriorityClass` named `fleet-agent-priority-class` and a `PodDisruptionBudget` with name `fleet-agent-pod-disruption-budget`.

4. Make sure the Agent's Deployment was updated and that the pod is successfully running. If a `spec.priorityClassName` is configured and used in a Deployment before the PriorityClass is actually created, the Pod will hang. This must not happen.3. Check that the values of `fleet-agent-priority-class` and `fleet-agent-pod-disruption-budget` are according to the specification in the cluster resource.
4. Check that the selectors of `fleet-agent-pod-disruption-budget` correctly point to the Fleet agent deployment.
5. Check that the `spec.priorityClassName` field of the Fleet agent deployment correctly points to the PriorityClass. The value for `spec.Priority` in the Fleet agent deployment should reflect the configured value for `priorityClass` in the Fleet cluster resource.
6. Delete the `agentSchedulingCustomization` field and ensure that the PriorityClass and PodDisruptionBudget resources have been removed and that the Deployment of the Fleet agent does not contain a reference to `spec.priorityClassName`.
Optionally extend testing for other configurable values of the `spec.agentSchedulingCustomization` field (`podDisruptionBudget.maxUnavailable` or `priorityClass.preemptionPolicy`). Setting both, `minAvailable` and `maxUnavailable` on a PodDisruptionBudget is supposed to prevent the Fleet agent from being updated, as those values are mutually exclusive and cannot successfully be set on a PodDisruptionBudget.


[...]

PR #4057, opened 2025-08-26
Support for setting PC and PDB of agent

p-se:
Refers to #3491

The changes in this PR enable PriorityClass (PC) and Pod Disruption Budget (PDB) creation for the agent of downstream clusters. It extends the cluster resource with a field `agentSchedulingCustomization` which can be used to configure certain but not all fields for the creation of a PDB and PC. The restrictions of what can and cannot be configured follow the upstream implementation in Rancher.

### Checklist


internal/cmd/controller/agentmanagement/controllers/cluster/import.go
```diff
@@ -266,6 +268,12 @@ func (i *importHandler) deleteOldAgent(cluster *fleet.Cluster, kc kubernetes.Int
 	if err := kc.AppsV1().Deployments(namespace).Delete(i.ctx, config.AgentConfigName, metav1.DeleteOptions{}); err != nil && !apierrors.IsNotFound(err) {
 		return err
 	}
+	if err := kc.SchedulingV1().PriorityClasses().Delete(i.ctx, scheduling.FleetAgentPriorityClassName, metav1.DeleteOptions{}); err != nil && !apierrors.IsNotFound(err) {
```
weyfonk, line 271: Is there a risk that priority classes or pod disruption budgets might be shared by multiple agents, in this case the old and the new one? Or would we expect any PDB/PC to be re-created along with the agent if needed?

internal/cmd/controller/agentmanagement/controllers/cluster/import.go
```diff
@@ -378,9 +386,28 @@ func (i *importHandler) importCluster(cluster *fleet.Cluster, status fleet.Clust
 	clusterLabels := yaml.CleanAnnotationsForExport(cluster.Labels)
 	agentReplicas := cmd.ParseEnvAgentReplicaCount()
 
+	var (
+		objs              []runtime.Object
+		priorityClassName string
+	)
+	if schedulingCustomization := cluster.Spec.AgentSchedulingCustomization; schedulingCustomization != nil {
```
weyfonk: nit: how about a shorter variable name for `schedulingCustomization`, for instance `sc`?

internal/cmd/controller/agentmanagement/scheduling/scheduling.go
```diff
@@ -0,0 +1,60 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+const (
+	FleetAgentPriorityClassName       = "fleet-agent-priority-class"
+	FleetAgentPodDisruptionBudgetName = "fleet-agent-pod-disruption-budget"
+)
+
+func PriorityClass(priorityClass *fleet.PriorityClassSpec) *schedulingv1.PriorityClass {
```
weyfonk, line 21: Interesting, TIL priority classes are non-namespaced...

internal/cmd/controller/agentmanagement/scheduling/scheduling_test.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+func TestPriorityClass(t *testing.T) {
+	tests := []struct {
+		name     string
+		spec     *fleet.PriorityClassSpec
+		expected *schedulingv1.PriorityClass
+	}{
+		{
+			name: "priorotiy 1000 with preemption",
```
weyfonk: nit:
```suggestion
			name: "priority 1000 with preemption",
```

internal/cmd/controller/agentmanagement/scheduling/scheduling_test.go
```diff
@@ -0,0 +1,181 @@
...
+				},
+				Value:            math.MaxInt32,
+				Description:      "Priority class for Fleet Agent",
+				PreemptionPolicy: ptr(corev1.PreemptNever),
+			},
+		},
+		{
+			name: "least possible value allowed",
+			spec: &fleet.PriorityClassSpec{
+				Value:            math.MaxInt32,
```
weyfonk:
```suggestion
				Value:            math.MinInt32,
```

internal/cmd/controller/agentmanagement/scheduling/scheduling_test.go
```diff
@@ -0,0 +1,181 @@
...
+				PreemptionPolicy: ptr(corev1.PreemptNever),
+			},
+			expected: &schedulingv1.PriorityClass{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "fleet-agent-priority-class",
+				},
+				Value:            math.MaxInt32,
```
weyfonk:
```suggestion
				Value:            math.MinInt32,
```

e2e/single-cluster/agent_scheduling_customization_test.go
```diff
@@ -0,0 +1,247 @@
...
+				g.Expect(err).ToNot(HaveOccurred())
+			}).Should(Succeed(), "Should be able to update cluster to remove agentSchedulingCustomization")
+
+			// Wait for PriorityClass to be deleted
+			Eventually(func() bool {
+				pc := &schedulingv1.PriorityClass{}
+				err := clientUpstream.Get(context.TODO(), client.ObjectKey{
+					Name: "fleet-agent-priority-class",
+				}, pc)
+				return client.IgnoreNotFound(err) == nil
```
weyfonk: This will pass even if the priority class is not deleted, because `client.IgnoreNotFound(nil)` is `nil` ;)

e2e/single-cluster/agent_scheduling_customization_test.go
```diff
@@ -0,0 +1,247 @@
...
+			Eventually(func(g Gomega) {
+				// Re-fetch the cluster to get the latest version
+				latestCluster := &fleet.Cluster{}
+				err := clientUpstream.Get(context.TODO(), client.ObjectKey{
+					Namespace: env.Namespace,
+					Name:      "local",
+				}, latestCluster)
+				g.Expect(err).ToNot(HaveOccurred())
+
+				latestCluster.Spec.AgentSchedulingCustomization = nil
```
weyfonk, line 70: Should we perhaps also check the cluster's status, making sure that `AgentSchedulingCustomizationHash` is emptied?

e2e/single-cluster/agent_scheduling_customization_test.go
```diff
@@ -0,0 +1,247 @@
...
```
weyfonk, line 168: Same remark as above on `AgentSchedulingCustomizationHash`.

weyfonk (commented):
Nice one, thanks! Leaving a few nitpicks, mostly testing-related.

internal/cmd/controller/agentmanagement/controllers/cluster/import.go
p-se: The PC and PDB are always recreated if the configuration of the agent changes, which, in this case, is when the cluster values for schedulingCustomization changes. Then the agent is also redeployed with proper configuration (e.g. not pointing to a PC anymore). PCs aren't that easy to handle because the value of the priority is immutable and a pod started with a specific priority will keep it as long as it runs. Also, if a value is removed, we make sure the PC and PDB are deleted, which is especially important for the PDB, which would otherwise still select the agents by labels and configure them with a value that doesn't exist anymore.

Does that answer the question?

internal/cmd/controller/agentmanagement/scheduling/scheduling.go
p-se: Also, the pods started with a reference to a PriorityClass get a priority assigned that does not change for as long as the pod is not recreated. The PC needs to be removed and recreated if the name is be reused, because the value of the PC is immutable (interestingly, other values like the description aren't, but that one the user is not supposed to change in our case).

internal/cmd/controller/agentmanagement/controllers/cluster/import.go
weyfonk: It does, thanks!

e2e/single-cluster/agent_scheduling_customization_test.go
```diff
@@ -0,0 +1,267 @@
...
+				g.Expect(cluster.Status.AgentSchedulingCustomizationHash).To(Equal(""))
+			}).Should(Succeed(), "Should have cleared the AgentSchedulingCustomizationHash")
+
+			// Wait for PriorityClass to be deleted
+			Eventually(func() bool {
+				pc := &schedulingv1.PriorityClass{}
+				err := clientUpstream.Get(context.TODO(), client.ObjectKey{
+					Name: "fleet-agent-priority-class",
+				}, pc)
+				return err == nil || client.IgnoreNotFound(err) == nil
```
weyfonk: This will still pass if the PriorityClass still exists. How about:
```suggestion
                // with "k8s.io/apimachinery/pkg/api/errors" added to imports
				return errors.IsNotFound(err)
```

Issue #4059, opened 2025-08-26
Webhooks for Azure DevOps do not trigger gitjob when URLs don't match old format


matteotumiati:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Here are two valid URLs pointing to the same resource:
1. https://dev.azure.com/<organization>/<project>/_git/<repo>
2. https://<organization>.visualstudio.com/<project>/_git/<repo>

Both are valid and, according to Azure DevOps, interchangeable. The problem is that if I have set the first one in my GitRepo but Azure DevOps sends a webhook using the second format, then Fleet won't be able to recognize the URL and update the GitRepo.

### Expected Behavior

There should be URL matching between the first and the second format because both are valid and supported by Azure DevOps. 

### Steps To Reproduce

1. Create a GitRepo with disablePolling: true and without pollingInterval
2. Enable webhooks with the ingress in Rancher
3. Set the URL of the GitRepo to https://dev.azure.com/org/project/_git/repo
4. Disable the "Use the new URL" toggle in organization settings in the Azure DevOps org
5. Enable service hooks in Azure DevOps to point to the URL configured in Rancher on push against main branch for that URL
6. Enable debug logging for Fleet (I set debug: true, level: 5)
7. Commit a change to the repository in Azure DevOps
8. Observe how in the logs the payload is printed but the commit ID doesn't change and there are no jobs being created to sync the state

### Environment

```markdown
- Architecture: amd64
- Fleet Version: Tested with several versions of fleet above 0.12
- Cluster:
  - Provider: AKS
  - Options: 3 
  - Kubernetes Version: 1.31
```

### Logs

```markdown

```

weyfonk:
Pull request content for Rancher QA verification.

## Additional QA

### Problem
Fleet would not be able to process webhook calls for repos using the legacy URL for Azure DevOps (`<name>.visualstudio.com`), because event payloads would contain the newer URL equivalent (`dev.azure.com`) instead.
 
### Solution
Fleet now maps legacy URLs to their newer equivalent when processing webhook calls from Azure DevOps.
 
### Testing

### Engineering Testing
#### Manual Testing
N/A

#### Automated Testing
Added unit tests validating that, when receiving an event payload for a repo created using a legacy Azure DevOps URL, Fleet is able to process the call and update the GitRepo accordingly.

### QA Testing Considerations
[...]

Issue #4058, opened 2025-08-26
[SURE-10646] Webhooks for Azure DevOps do not trigger gitjob when there's a space in URL


matteotumiati:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Azure DevOps supports spaces in URL for both project names and repositories (even if for repositories is not recommended). However, the Rancher UI has been updated to support spaces (using `%20`) in the GitRepo UI edit page, but the webhook has not been updated for the same fix.

If we try sending a webhook from an Azure DevOps project that contains a space (like https://dev.azure.com/org/my%20project%20name) the payload is received by Fleet but it is not processed to update the commit ID of the repo and trigger a refresh of the GitRepo.

### Expected Behavior

Sending the webhook from an Azure DevOps repository that contains spaces in the URL, the commit ID of a GitRepo should update and the gitjob should create a new pod to sync the state for that GitRepo.

### Steps To Reproduce

1. Create a GitRepo with disablePolling: true and without pollingInterval
2. Enable webhooks with the ingress in Rancher
3. Set the URL of the GitRepo to https://dev.azure.com/org/my%20project%20name/_git/repo
4. Enable service hooks in Azure DevOps to point to the URL configured in Rancher on push against main branch for that URL
5. Enable debug logging for Fleet (I set debug: true, level: 5)
6. Commit a change to the repository in Azure DevOps
7. Observe how in the logs the payload is printed but the commit ID doesn't change and there are no jobs being created to sync the state

### Environment

```markdown
- Architecture: amd64
- Fleet Version: Tested with several versions of fleet above 0.12
- Cluster:
  - Provider: AKS
  - Options: 3
  - Kubernetes Version: 1.31
```

### Logs

```markdown

```

weyfonk:
## Additional QA
 
### Problem
Webhook calls received by Fleet for Azure DevOps would not trigger git jobs for updates when the repo URL contains a space.
 
### Solution
Added support for spaces to AzureDevOps repo URLs to the webhook logic.

### Testing

### Engineering Testing
#### Manual Testing
N/A

#### Automated Testing
Added unit tests validating that such URLs would enable commits to be updated on GitRepos.

### QA Testing Considerations
See reproduction steps above.

Replace step 7. with expectation of success, e.g.:
> Observe how in the logs the payload is printed ~but the commit ID doesn't change and there are no jobs being created to sync the state~ , the commit SHA is updated and a git job is created.
 
#### Regressions Considerations
[...]

PR #4060, opened 2025-08-26
fix: webhooks in azure devops should support spaces and old URL formats

matteotumiati:
Refers to #4058 and #4059 

### Checklist

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -108,11 +108,6 @@ func (w *Webhook) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
 
 		path := strings.Replace(u.EscapedPath()[1:], "/_git/", "(/_git)?/", 1)
 
-		// URLs for Azure DevOps are allowed to contain spaces
-		if u.Hostname() == "dev.azure.com" {
-			path = strings.ReplaceAll(path, "%20", "(?:%20| )")
-		}
-
 		regexpStr := `(?i)(http://|https://|\w+@|ssh://(\w+@)?|git@(ssh\.)?)` + u.Hostname() +
 			"(:[0-9]+|)[:/](v\\d/)?" + path + "(\\.git)?"
 		repoRegexp, err := regexp.Compile(regexpStr)
--- a/pkg/webhook/webhook_test.go
+++ b/pkg/webhook/webhook_test.go
@@ -110,42 +110,60 @@ func TestAzureDevopsWebhook(t *testing.T) {
 }
 
 func TestAzureDevopsWebhookWithURLSpacing(t *testing.T) {
-	const commit = "f00c3a181697bb3829a6462e931c7456bbed557b"
-	const repoURL = "https://dev.azure.com/fleet/git%20test/_git/git%20test"
-	gitRepo := &v1alpha1.GitRepo{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: "test",
+	cases := []struct {
+		name    string
+		repoURL string
+	}{
+		{
+			name:    "legacy URL",
+			repoURL: "https://visualstudio.com/fleet/git%20test/_git/git%20test",
 		},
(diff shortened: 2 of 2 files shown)


mrolmedo:
SURE-10646

pkg/webhook/webhook.go
```diff
@@ -105,7 +105,14 @@ func (w *Webhook) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
 			w.logAndReturn(rw, err)
 			return
 		}
-		path := strings.Replace(u.Path[1:], "/_git/", "(/_git)?/", 1)
+
+		path := strings.Replace(u.EscapedPath()[1:], "/_git/", "(/_git)?/", 1)
+
+		// URLs for Azure DevOps are allowed to contain spaces
+		if u.Hostname() == "dev.azure.com" {
```
weyfonk: Would a URL using the old domain, e.g. `https://fleet.visualstudio.com/`, also support spaces?
We may need a more elaborate condition here, perhaps with a method for determining if a hostname is Azure DevOps, based on 2 constants (one for `dev.azure.com` and another one for `.visualstudio.com`).
Or am I missing something?

pkg/webhook/webhook.go
```diff
@@ -320,6 +327,25 @@ func parsePayload(payload interface{}) (revision, branch, tag string, repoURLs [
...
+		// There are multiple URLs formats and these may overlap in different areas of Azure DevOps
+		for i, u := range repoURLs {
+			parsed, err := url.Parse(u)
+			if err != nil {
+				continue
+			}
+			if strings.HasSuffix(parsed.Hostname(), ".visualstudio.com") {
+				org := strings.SplitN(parsed.Hostname(), ".", 2)[0]
+				parsed.Host = "dev.azure.com"
+				parsed.Path = "/" + org + parsed.Path
```
weyfonk: Wouldn't a slash be missing here?
e.g. `fleet.visualstudio.com/foo` would become `dev.azure.com/fleetfoo`, wouldn't it?
```suggestion
				parsed.Path = fmt.Sprintf("/%s/%s", org, parsed.Path)
```

pkg/webhook/webhook.go
matteotumiati: The way I thought this is to rewrite the URL, e.g. `https://fleet.visualstudio.com/` into `https://dev.azure.com/fleet` when it is loaded from the payload, so it shouldn't be necessary to double-check here if the URL is `.visualstudio.com` and delete spaces from that. But I'm not an expert in Go, so I tried my best, and I might have missed something as well 😄

pkg/webhook/webhook.go
matteotumiati: Yes, good catch! Thanks 👍

matteotumiati:
> Thank you for your contribution! Leaving a couple of questions, happy to discuss :)

It's my first contribution in Go and in Fleet, so I hope this will also work fine for you and I'm happy to integrate any changes after the discussion. The primary goal is to provide better support for Azure DevOps. 😎 

I did integrate some tests and verified that these changes don't break any of the existing ones. Not sure if you want to add more but those two should cover the basic scenarios.

Perhaps a better implementation of this could be done simply by reworking the regex, but I'm not sure how complicated/unreadable that would become later, especially considering that not all providers support spaces like Azure DevOps does.

pkg/webhook/webhook.go
weyfonk: Thanks for clarifying. I've added a test case for spaces in `visualstudio.com` URLs, which validates that both were indeed already supported through the first commits ;)
This made me realise that updating the regex for `dev.azure.com` URLs was not necessary.

pkg/webhook/webhook.go
weyfonk: ... and it seems I'd misled you there, `parsed.Path` is already prefixed by a slash, `%s%s` works as a format string. Fixed now, sorry about that!

PR #4069, opened 2025-08-28
Improve rancher dev script

thardeck:
small improvement after testing the script on a fresh machine.


dev/setup-rancher-clusters
```diff
@@ -2,6 +2,12 @@
 
 set -eux
 
+# Check availability of Rancher CLI
+if ! command -v rancher &>/dev/null && ! command -v rancher-cli &>/dev/null; then
```
weyfonk, line 6: This check is already run by `.github/scripts/register-downstream-clusters.sh`, called from L64; is the idea here to fail faster if the command is not available?

thardeck: When I was testing this on a fresh system with Manuel, the error came after already setting up a lot.

I think it would be nicer to run these verifications in the beginning of "user" facing scripts, so the script fails early and not after running a lot of commands and spending quite some time.

If this is ok then I would also add a check for `kubectl`.

weyfonk: I'd be surprised if someone tried to run this without having `kubectl` installed...

thardeck: That is true, that's why initially did not add it. If this is fine, I can merge it like it is.

Issue #4028, opened 2025-08-20
[SURE-10613] Fleet Bundle not deploying on upgraded Rancher 2.11.4


kkaempf:
### Issue description (shorted):

We are experiencing issues that are affecting our HOSTED Rancher production environment and is preventing business from happening. The issue began when we upgraded to hosted rancher 2.11.4. This updated fleet to v0.12.5 

We are seeing issues with fleet where the bundles are not deploying and not creating bundle deployments. The bundles are just hanging and not updating when we push updates.

For example, when they try to deploy redis and the same thing happens when they deploy anything. 

In addition, we can not use the UI to 'delete' bundles they just hang. In order to delete them we had to remove the finalizers. 

### Workaround:
Is a workaround available and implemented? yes
What is the workaround: Fall back to Fleet `v0.12.0`

### Actual behavior:
It does not deploy and is 'stuck' , When they try to delete the job it fails. 

### Expected behavior:
We should be able to deploy new bundles when the code changes. 

### Additional notes:

See JIRA for more information.

weyfonk:
Waiting for customer feedback.

kkaempf:
Closing for lack of response.

PR #4075, opened 2025-08-29
Fleet apply refuses to update bundle scheduled for deletion

manno:
Refers to https://github.com/rancher/fleet/issues/4028


internal/cmd/cli/apply/apply.go
```diff
@@ -510,9 +510,13 @@ func pushOCIManifest(ctx context.Context, bundle *fleet.Bundle, opts ocistorage.
 func save(ctx context.Context, c client.Client, bundle *fleet.Bundle) (*fleet.Bundle, error) {
 	updated := bundle.DeepCopy()
 	result, err := controllerutil.CreateOrUpdate(ctx, c, bundle, func() error {
-		if bundle != nil && bundle.Spec.HelmOpOptions != nil {
+		if bundle.Spec.HelmOpOptions != nil {
```
manno, line 521: I would be surprised if CreateOrUpdate could pass a nil bundle.

internal/cmd/cli/apply/apply.go
```diff
@@ -567,7 +571,11 @@ func saveOCIBundle(ctx context.Context, c client.Client, r record.EventRecorder,
 
 	updated := bundle.DeepCopy()
 	_, err = controllerutil.CreateOrUpdate(ctx, c, bundle, func() error {
-		if bundle != nil && bundle.Spec.HelmOpOptions != nil {
+		if bundle.DeletionTimestamp != nil {
+			return fmt.Errorf("the bundle %q is being deleted", bundle.Name)
+		}
```
aruiz14, line 584: This only covers the case where the Bundle is marked for deletion during the execution of the function (`DeletionTimestamp` not set before calling `CreateOrUpdate`). Assuming that `bundle` passed to `saveOCIBundle` is "produced" from scratch with the desired state, this still has the risk of doing extra work unnecessarily (`pushOCIManifest`) for a bundle that is already being deleted and will never be used. Should we do an early `get` of the `Bundle` to prevent that?

manno: That's a frightening scenario. As we do delete values secrets before creating the bundle. And in case of pushOCIManifest, that probably takes long, giving lots of time for the deletion to succeed.

I'll add an additional Get as early as possible.

Issue #3746, opened 2025-06-02
Fix deprecation warning: Replace Result.Requeue with explicit requeueing


thardeck:
The Kubernetes controller-runtime library has deprecated `Result.Requeue` in favor of more explicit requeueing patterns (see [PR #3107](https://github.com/kubernetes-sigs/controller-runtime/pull/3107)).

Fleet is currently using the deprecated `Result.Requeue` field, which should be replaced with explicit `RequeueAfter` or by returning an error to trigger requeueing.

## Impact

- Deprecation warnings in logs
- Future compatibility issues when upgrading controller-runtime
- Code clarity improvements with explicit requeueing patterns

## Acceptance Criteria

- [x] Identify all usages of `reconcile.Result{Requeue: true}` in the codebase
- [x] Replace with appropriate alternatives:
  - Use `reconcile.Result{RequeueAfter: time.Duration}` for time-based requeueing
  - Return an error for immediate requeueing with backoff
  - Use `reconcile.Result{}` (no requeue) where appropriate
- [x] Update any related documentation or comments
- [x] Ensure existing reconciliation behavior is preserved
- [x] Add/update tests to cover the new requeueing patterns

kkaempf:
Shouldn't this be in the v2.12.2 milestone ?

weyfonk:
> Shouldn't this be in the v2.12.2 milestone ?

Correct. This is not user-facing though, I don't see why this would need QA review 🤔

PR #4077, opened 2025-08-29
Replace deprecated Requeue with RequeueAfter in controllers

thardeck:
Replace all usages of `reconcile.Result{Requeue: true}` with appropriate `RequeueAfter` durations to align with controller-runtime best practices.

Refers #3746

### DIFF
--- a/internal/cmd/controller/gitops/reconciler/gitjob_controller.go
+++ b/internal/cmd/controller/gitops/reconciler/gitjob_controller.go
@@ -226,7 +226,7 @@ func (r *GitJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	// If so, we need to return a Result with EnqueueAfter set.
 
 	res, err := r.manageGitJob(ctx, logger, gitrepo, oldCommit, repoPolled)
-	if err != nil {
+	if err != nil || res.RequeueAfter > 0 {
 		return res, err
 	}
 


internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -193,7 +193,7 @@ func (r *GitJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		}
 
 		// requeue as adding the finalizer changes the spec
-		return ctrl.Result{Requeue: true}, nil
+		return ctrl.Result{RequeueAfter: time.Second}, nil
```
manno: I would prefer the same duration for most results: https://github.com/rancher/fleet/blob/main/pkg/durations/durations.go#L20 
Though we should fix the type and `DefaultRequeueueAfter -> DefaultRequeueAfter`

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -225,8 +225,7 @@ func (r *GitJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	// If so, we need to return a Result with EnqueueAfter set.
 
 	res, err := r.manageGitJob(ctx, logger, gitrepo, oldCommit, repoPolled)
-	// nolint: staticcheck // Requeue is deprecated; see fleet#3746.
-	if err != nil || res.Requeue {
```
manno, line 229: Can we ignore `res.Requeue`? I thought this method can ask for a requeue even if it does not return an error.

e2e/require-secrets/gitrepo_test.go
```diff
@@ -131,14 +131,13 @@ var _ = Describe("Git Repo", func() {
 			Eventually(func() string {
 				out, _ := k.Get("gitrepo", "gitrepo-test", "-o", "yaml")
 				return out
-			}).Should(ContainSubstring("commit: " + commit))
+			}, "120s", "5s").Should(ContainSubstring("commit: " + commit))
```
manno: I think a few of these durations can be extracted into a constant in suite_test.go or even e2e/testenv/env.go
Not sure about the naming, maybe `ShortTimeout` and `MediumTimeout`

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
thardeck: Maybe I understand the question wrong, but `res.Requeue` is deprecated and `manageGitJob` uses the recommended `RequeueAfter:  durations.DefaultRequeueAfter` which does not depend on `res.Requeue` anymore.

manno: Then it should be 
`if err != nil || res.RequeueAfter ... {` 
so that `manageGitJob` can return and requeue without returning an error.


I am trying to say, this returns early if `manageGitJob` wants to?

e2e/single-cluster/gitrepo_polling_disabled_test.go
```diff
@@ -116,27 +117,58 @@ var _ = Describe("GitRepoPollingDisabled", Label("infra-setup"), func() {
...
+
+			By("Getting the current GitRepo generation before force sync")
+			var currentGeneration int64
+			Eventually(func() error {
+				out, err := k.Get("gitrepo", gitrepoName, "-o", "jsonpath={.metadata.generation}")
+				if err != nil {
+					return err
+				}
+				if out != "" {
+					currentGeneration = int64(len(out)) // Simple way to get a number from string
```
weyfonk: :raised_eyebrow: Sure, this gets _an_ integer, not the right one though: if the generation is bumped to 2, `currentGeneration` will still be 1, right?

integrationtests/helmops/controller/controller_test.go
```diff
@@ -430,6 +437,9 @@ var _ = Describe("HelmOps controller", func() {
 				// update the HelmOp spec
 				helmop.Spec.Helm.Version = "0.1.0"
 
+				// Add a small delay to reduce race conditions with the controller
+				time.Sleep(100 * time.Millisecond)
```
weyfonk: What kind of race condition are these? Perhaps we could fix them instead?

thardeck: Maybe race condition is the wrong term. By changing the requeue system some things take a bit longer then before, partly because we use a default of five seconds, which lead to failing tests.

These two might even not be needed anymore. Initially I was trying different things and did not verify if those sleep commands were still necessary after the overall changes to make the tests more flexible in terms of timing.

I will remove them.

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
thardeck: Good point, thanks. Fixed in https://github.com/rancher/fleet/pull/4077/commits/96f16dd007b1fa8adeb7535f179e49895ebf2d06

PR #4090, opened 2025-09-02
Remove unused fixtures, fix GVK recognition

p-se:
Ad hoc fix/refactoring of tests. Actually test fine tuning readiness detection using environment variables.

### Checklist

### DIFF
--- a/_typos.toml
+++ b/_typos.toml
@@ -12,4 +12,4 @@ extend-exclude = [
 "passin" = "passin"
 "ANDed" = "ANDed"
 "HoTWPQxTJ5dIY31" = "HoTWPQxTJ5dIY31"
-"wre" = "wre" # warning regular expression, not 'were'
+"wre" = "wre"                         # warning regular expression, not 'were'


benchmarks/cmd/csv.go
```diff
@@ -22,8 +22,8 @@ var csvCmd = &cobra.Command{
 		description := sample.Description
 		re := regexp.MustCompile(`\{\{[^}]*\}\}`)
 		description = re.ReplaceAllString(description, "")
-		wre := regexp.MustCompile(`WARNING:.*`)
-		description = wre.ReplaceAllString(description, "")
+		warningRe := regexp.MustCompile(`WARNING:.*`)
```
manno: Short variable names are fine. The typo workflow should not flag this.

internal/cmd/agent/deployer/summary/summarizers_test.go
```diff
@@ -173,28 +173,13 @@ func TestCheckErrors(t *testing.T) {
 					Error: false,
 				},
 			},
-			loadConditions: func() {
```
weyfonk: In my understanding, this one still belongs here, as the test case covers what happens if the GVK is not found in the env var (`helm.cattle.io/v1/HelmChart` vs `sample.cattle.io/v1/Sample`).

internal/cmd/agent/deployer/summary/summarizers_test.go
```diff
@@ -210,28 +195,13 @@ func TestCheckErrors(t *testing.T) {
...
```
weyfonk: Same remark as above: this should be kept here, the test case deals with informational conditions.

PR #4096, opened 2025-09-03
Allow to add any label to local cluster

k0da:
Most real use-case for that is to set a "real" cluster name via ClusterName label. So multiple Rancher/Fleet controller instances have unique name/label.

Refers to #XXX

### Checklist

### DIFF
--- a/charts/fleet/templates/configmap.yaml
+++ b/charts/fleet/templates/configmap.yaml
@@ -24,6 +24,7 @@ data:
         "paths": "{{.Values.bootstrap.paths}}",
         "repo": "{{.Values.bootstrap.repo}}",
         "secret": "{{.Values.bootstrap.secret}}",
+        "clusterLabels": "{{.Values.boostrap.clusterLabels}}",
         "branch":  "{{.Values.bootstrap.branch}}",
         "namespace": "{{.Values.bootstrap.namespace}}",
         "agentNamespace": "{{.Values.bootstrap.agentNamespace}}"
--- a/charts/fleet/values.yaml
+++ b/charts/fleet/values.yaml
@@ -53,6 +53,8 @@ bootstrap:
   # The namespace where the fleet agent for the local cluster will be ran, if empty
   # this will default to cattle-fleet-system
   agentNamespace: ""
+  # Apply extra clusterLables to the local cluster bootstrapped by fleet-controller
+  clusterLabels: {}
   # A repo to add at install time that will deploy to the local cluster. This allows
   # one to fully bootstrap fleet, its configuration and all its downstream clusters
   # in one shot.
--- a/internal/cmd/controller/agentmanagement/controllers/bootstrap/bootstrap.go
+++ b/internal/cmd/controller/agentmanagement/controllers/bootstrap/bootstrap.go
@@ -2,6 +2,7 @@ package bootstrap
 
 import (
 	"context"
+	"maps"
 	"os"
 	"regexp"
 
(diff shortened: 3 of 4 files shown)


Buzzglo (approved):

charts/fleet/values.yaml
```diff
@@ -53,6 +53,8 @@ bootstrap:
   # The namespace where the fleet agent for the local cluster will be ran, if empty
   # this will default to cattle-fleet-system
   agentNamespace: ""
+  # Apply extra clusterLables to the local cluster bootstrapped by fleet-controller
```
weyfonk: nit:
```suggestion
  # Apply extra clusterLabels to the local cluster bootstrapped by fleet-controller
```

weyfonk (commented):
LGTM apart from a small typo, thanks for your contribution!

(optional) Would you mind cherry-picking [this commit](https://github.com/rancher/fleet/commit/4f5fd37a6a2bc81df04b0e944e0a8ac60b3190d7) which adds a couple of test cases?

manno:
Note that Rancher does not use Fleet's bootstrapping to create the local cluster.

weyfonk:
> Note that Rancher does not use Fleet's bootstrapping to create the local cluster.

Further to this comment: @k0da , @Buzzglo : this patch would only work for Fleet standalone, without Rancher. Is that enough from your perspective?

If not, a patch to [rancher/rancher](https://github.com/rancher/rancher) could be needed, for instance, in order to allow Rancher's Fleet cluster [controller](https://github.com/rancher/rancher/blob/main/pkg/controllers/provisioningv2/fleetcluster/fleetcluster.go#L202) to inject custom labels into the local Fleet cluster when creating it.

k0da:
Yes. standalone fleet is what we need.

weyfonk:
> Yes. standalone fleet is what we need.

Thanks for clarifying. Would you be willing to address my [comment](https://github.com/rancher/fleet/pull/4096#pullrequestreview-3403118754) above, allowing us to merge this?

charts/fleet/templates/configmap.yaml
```diff
@@ -24,6 +24,7 @@ data:
         "paths": "{{.Values.bootstrap.paths}}",
         "repo": "{{.Values.bootstrap.repo}}",
         "secret": "{{.Values.bootstrap.secret}}",
+        "clusterLabels": "{{.Values.boostrap.clusterLabels}}",
```
weyfonk: A failing workflow, checking for typos, just flagged this:
```suggestion
        "clusterLabels": "{{.Values.bootstrap.clusterLabels}}",
```

k0da: thanks applied

charts/fleet/templates/configmap.yaml
```diff
@@ -24,6 +24,7 @@ data:
         "paths": "{{.Values.bootstrap.paths}}",
         "repo": "{{.Values.bootstrap.repo}}",
         "secret": "{{.Values.bootstrap.secret}}",
+        "clusterLabels": "{{.Values.bootstrap.clusterLabels}}",
```
weyfonk: Two things need to change here for deployments to succeed:
* the value of `clusterLabels` should not be quoted, otherwise they'd be interpreted as strings, while we want to pass a map of strings to strings to our deployment
* that map must be converted into JSON
```suggestion
        "clusterLabels": {{toJson .Values.bootstrap.clusterLabels}}
```

weyfonk (changes requested):

k0da:
@weyfonk applied, thank you

charts/fleet/templates/configmap.yaml
weyfonk: So sorry, I missed an end-of-line comma in my suggestion. Could you please re-add it?

Issue #3810, opened 2025-06-18
dd workflow to run benchmark suite


manno:
### Is your feature request related to a problem?

To make sure it isn't broken.

PR #4106, opened 2025-09-05
Benchmark workflow

p-se:
Refers to #3810

### Checklist

~~- [ ]  I have updated the documentation via a pull request in the
[fleet-docs](https://github.com/rancher/fleet-docs) repository.~~

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,106 @@
+name: Benchmarks
+
+on:
+  workflow_dispatch:
+    inputs:
+      timeout:
+        description: Per-experiment timeout (e.g., 2m, 5m)
+        required: false
+        default: 5m
+      namespace:
+        description: Fleet workspace namespace
+        required: false
+        default: fleet-local
+  schedule:
+    - cron: "0 15 * * *"
+
+permissions:
+  contents: read
+
+concurrency:
+  group: benchmarks-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  GOARCH: amd64
+  CGO_ENABLED: 0
+  SETUP_K3D_VERSION: 'v5.8.3'
+  # Defaults for both manual and scheduled runs
+  BENCH_TIMEOUT: ${{ github.event.inputs.timeout || '5m' }}
+  BENCH_NAMESPACE: ${{ github.event.inputs.namespace || 'fleet-local' }}
+
+jobs:
+  benchmark:
+    name: Run Fleet Benchmarks
+    runs-on: ubuntu-latest
+    timeout-minutes: 90
+    steps:
+      - name: Checkout
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
+        with:
+          fetch-depth: 0
+
+      - name: Setup Go
+        uses: actions/setup-go@44694675825211faa026b3c33043df3e48a5fa00 # v6
+        with:
+          go-version-file: 'go.mod'
+          check-latest: true
+
+      # No extra host dependencies required
+
+      # Rely on kubectl and helm preinstalled in GitHub runner image
+
+      - name: Install k3d
+        run: curl --silent --fail https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh | TAG=${{ env.SETUP_K3D_VERSION }} bash
+
+      - name: Create k3d cluster
+        run: |
+          # Use default settings; no custom Docker network required
+          k3d_args="" ./dev/setup-k3d upstream
+
+      - name: Build Fleet images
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,106 @@
+name: Benchmarks
+
+on:
+  workflow_dispatch:
+    inputs:
+      timeout:
+        description: Per-experiment timeout (e.g., 2m, 5m)
+        required: false
+        default: 5m
```
weyfonk: Is this long enough?
Edit: oh, _per-experiment_, does that mean that the actual benchmark suite timeout would amount roughly to a multiple of this value?

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,106 @@
...
+
+      # Rely on kubectl and helm preinstalled in GitHub runner image
+
+      - name: Install k3d
+        run: curl --silent --fail https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh | TAG=${{ env.SETUP_K3D_VERSION }} bash
+
+      - name: Create k3d cluster
+        run: |
+          # Use default settings; no custom Docker network required
+          k3d_args="" ./dev/setup-k3d upstream
```
weyfonk: Images from `./dev/` are rather for local use. In a CI context, we should use scripts from `./.github/scripts`, or directly run scripts as we do in other workflows, such as:
```suggestion
          k3d cluster create upstream --wait \
              --agents 1 \
              --network "nw01" # possibly adding a k3s version image argument
```

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,106 @@
...
+
+      - name: Build Fleet images
+        run: |
+          ./dev/build-fleet
```
weyfonk:
```suggestion
          ./.github/scripts/build-fleet-images.sh
```

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,106 @@
...
+
+      - name: Import images to k3d
+        env:
+          FLEET_E2E_CLUSTER: k3d-upstream
+          # Set downstream same as upstream to skip jq usage in script
+          FLEET_E2E_CLUSTER_DOWNSTREAM: k3d-upstream
+        run: |
+          ./dev/import-images-k3d
```
weyfonk: Using `k3d-import-retry` from `.github/scripts`, we wouldn't need to set env vars above ;)
```suggestion
          ./.github/scripts/k3d-import-retry.sh rancher/fleet:dev rancher/fleet-agent:dev -c upstream
```

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,106 @@
...
+
+      - name: Install Fleet to cluster
+        run: |
+          ./dev/setup-fleet upstream
```
weyfonk:
```suggestion
          ./.github/scripts/deploy-fleet.sh
```

.github/workflows/benchmark.yml
p-se: > Is this long enough?

Yes it is. In fact, if no timeout is specified, 2m are used by default. Which I took as opportunity to adjust (lower) the default in this action as well.

> Edit: oh, _per-experiment_, does that mean that the actual benchmark suite timeout would amount roughly to a multiple of this value?

Yes. Every `Eventually` used in the suite that doesn't have a hard-coded timeout is getting this (currently we have 9). I've changed the description to be more intuitive by calling it Eventually Default Timeout.

[Running it takes about 12 minutes.
](https://github.com/p-se/fleet/actions/runs/17667054640)

weyfonk: Nice, thanks for clarifying. This looks good to me, and in case of issues it should be easy to adjust... or may well be an indicator of performance regression at some point :þ

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,114 @@
...
+  SETUP_K3D_VERSION: "v5.8.3"
+  SETUP_K3S_VERSION: "v1.33.1-k3s1"
+  # Defaults for both manual and scheduled runs
+  BENCH_TIMEOUT: ${{ github.event.inputs.timeout || '2m' }}
+  BENCH_NAMESPACE: ${{ github.event.inputs.namespace || 'fleet-local' }}
+
+jobs:
+  benchmark:
+    name: Run Fleet Benchmarks
+    runs-on: ubuntu-latest
```
weyfonk, line 36: Note: this runs for now, great! When we get to storing results, we'll want to make sure we stick to a particular runner type, making a conscious choice between Github's and self-hosted runners.
Not a review comment per se, but a thought for the future...

PR #4116, opened 2025-09-09
Add erroneous resource to perClusterState

p-se:
Refers to #3866

### Checklist


weyfonk (approved):
LGTM. Just to be sure, do we need specific handling of `missing`, as per #3866?

p-se:
> LGTM. Just to be sure, do we need specific handling of `missing`, as per #3866?

That's a good question. Just by looking at the code, I can't figure it out. Missing is handled in the `appendToPerClusterState` function, but it may not work as expected. I will need to test that explicitly.

Edit: Which apparently I also already have tried and couldn't reproduce [here](https://github.com/rancher/fleet/issues/3866#issuecomment-3052326693). The difference is that in the reported issue the actual state is `updating`, which could have been fixed by the [first PR](https://github.com/rancher/fleet/pull/3890).

PR #4127, opened 2025-09-15
Revert "Remove rate limiting removal (#3848)"

p-se:
This reverts commit 2ffb043948706fb9b2e00c7e99b6be0f0e42799b.

Reverting because the default rate limiter might be too slow and since we have not added a configuration option for users to configure the rate limiter, we feel it is safer to wait with rate limiting until it is configurable.

### Checklist


kkaempf:
Hmm, no tests ? 🤔

p-se:
> Hmm, no tests ? 🤔

It's a reintroduction of code that's been there for a long time. We are restoring because it turned that the defaults of the client-go rate limiter are less than we thought. Due to that, and the fact that the user cannot change this value, it appears to be risky to rate limit. We have discussed to reintroduce rate limiting with the option of configuring it.

Issue #3924, opened 2025-07-29
Remove HTML content from Helm Op's error messages


torchiaf:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

The UI gets an extra-long error message:

<img width="1590" height="753" alt="Image" src="https://github.com/user-attachments/assets/5f6210d9-bc8d-4bd7-a2f6-76ce9b1caf80" />

### Steps To Reproduce

1. Create a new Helm OP
  - Source type: Repository
  - URL: https://github.com/rancher
  - Chart: fleet-examples/tree/master/single-cluster/helm

### Environment

```markdown
- Architecture:
- Fleet Version: v0.13.0
- Cluster:
  - Provider:
  - Options:
  - Kubernetes Version:
```

### Logs

```markdown

```

weyfonk:
Idea: only log response body as _trace_, not as error log?

weyfonk:
## Additional QA

### Problem
When failing to resolve a Helm chart version, Fleet's HelmOps would include the server's error response body in the HelmOp's status, leading to unreadable status messages in the Rancher UI

### Solution
The response body is now skipped when populating a HelmOp's status, only including the response status code which, along with the wrapping error message added by the HelmOp controller, should be enough to diagnose issues.

### Testing

### Engineering Testing
#### Manual Testing
N/A

#### Automated Testing
The absence of a server's response body from a HelmOp's error message is validated in unit tests.

### QA Testing Considerations
This should be tested in the Rancher UI, for instance when pointing a HelmOp to a Helm repo located inside a Github repository, where no `index.yaml` file exists (see Francesco's example from the above screenshot).
The following example could be enough to reproduce this:
```
helm:
  repo: https://github.com/rancher
  chart: foo
```
 
#### Regressions Considerations
Successful cases, where a chart can be successfully downloaded, should not be affected.

mmartin24:
Verified fix on 2.12 with fleet `v2.12-67b355c13bc2af254e8d80439dcc036b6938fa73-head` with `fleet:107.0.3+up0.13.3`

The HTML long log is no longer visible and a short 404 message is displayed instead:

`Could not get a chart version: failed to read helm repo from https://github.com/rancher/index.yaml, error code: 404`

Posting before / after fix:

<img width="1905" height="975" alt="Image" src="https://github.com/user-attachments/assets/83f50907-cda0-4632-96c2-cbfba3c86e2c" />

PR #4129, opened 2025-09-15
Remove HTML response body from HelmOps errors

weyfonk:
When a chart version cannot be retrieved from a Helm repository, including the response body in the error message brings little context and can lead to excessively long error status messages. Therefore, Fleet now skips the response body in such cases, relying solely on the status code.

Refers to #3924

### Checklist

- [ ] Backport to `release/v0.13`


internal/cmd/controller/helmops/reconciler/helmop_controller_test.go
```diff
@@ -358,75 +366,167 @@ func TestReconcile_Validate(t *testing.T) {
...
+
+		ctx := context.TODO()
+
+		res, err := r.Reconcile(ctx, ctrl.Request{NamespacedName: namespacedName})
+		if err == nil {
+			t.Errorf("expecting error, got nil")
+		}
+		if !strings.Contains(err.Error(), "404") {
+			t.Errorf("expecting error containing: [404], got %v", err.Error())
+		}
```
0xavi0: nit: Maybe it would be better to check the full error message here.
I mention this because if the error body is still being returned, this check would pass.
If we check for the full string (as done in the unit tests above), it wouldn't even be necessary to verify that the HTML is not present.

0xavi0 (approved):

internal/cmd/controller/helmops/reconciler/helmop_controller_test.go
weyfonk: Nice catch, thanks. I've refined the checks to use a regular expression instead, which is not simpler but should be a bit stricter.

Issue #4143, opened 2025-09-17
Orphaned Bundledeployment Lifecycle Secrets Block Installation


manno:
I am installing 50 helm-style bundles, deleting the gitrepo and re-creating the gitrepo. The second time the installation will get stuck, unless I wait for all lifecycle secrets to disappear. I am using Fleet `v0.13.2-rc.2`. I am testing with 500 clusters (25k bundledeployments&secrets).
 
```
kubectl apply -n fleet-default -f - <<EOF
---
kind: GitRepo
apiVersion: fleet.cattle.io/v1alpha1
metadata:
  name: scale-50-single
spec:
  repo: https://github.com/manno/fleet-experiments
  branch: main
  paths:
    - scale-50-bundles/ten
    - scale-50-bundles/twenty
    - scale-50-bundles/thirty
    - scale-50-bundles/fourty
    - scale-50-bundles/fifty
  targetNamespace: scale-50-single
  targets:
    - clusterSelector: {}
EOF
```

After removing a GitRepo:
* the gitrepo is removed
* fleet-controller removes finalizers  from content resources, then it removes content resources
  * there are lot's of conflicts in the logs, but it works.
  * example only creates three (!) content resources total (`s-9d774f37b4a...` is used by 48 bundledeployments)
  * `fleet-controller-76cb85cf48-qshmp fleet-controller {"level":"error","msg":"Reconciler error","controller":"bundledeployment","controllerGroup":"fleet.cattle.io","controllerKind":"BundleDeployment","BundleDeployment":{"name":"scale-50-single-scale-50-bundles-twenty-seventeen","namespace":"cluster-fleet-default-d0-k3k-downstream002-downstream0161-5314b"},"namespace":"cluster-fleet-default-d0-k3k-downstream002-downstream0161-5314b","name":"scale-50-single-scale-50-bundles-twenty-seventeen","error":"Content.fleet.cattle.io \"s-9d774f37b4a7562f0f91c3acd6409fdee41209cb35d55f76aa1f381ebbb1f\" not found",...}`
* fleet-controller removes finalizers from bundledeployments, then it removes bundledeployments
* fleet-controller removes finalizer from bundles, then bundles are removed
* lifecycle secrets are left behind so that k8s garbage collection can remove them (orphaned owner?)
  * `kubectl get secrets -A --field-selector type=fleet.cattle.io/bundle-deployment/v1alpha1`
  * this takes easily 30 minutes for 25000 secrets

Ideas:
[...]

PR #4146, opened 2025-09-17
Explicitly delete bundle secrets on bundle deletion

manno:
Refers to https://github.com/rancher/fleet/issues/4143

This is a quick fix, to remove the secrets faster and not rely entirely on garbage collection.

### DIFF
--- a/internal/cmd/controller/reconciler/bundle_controller.go
+++ b/internal/cmd/controller/reconciler/bundle_controller.go
@@ -460,6 +460,11 @@ func (r *BundleReconciler) handleDelete(ctx context.Context, logger logr.Logger,
 		return ctrl.Result{}, err
 	}
 
+	// pro-actively delete the bundle's secret. k8s owner garbage collection will handle remaining orphans.
+	if err := r.Delete(ctx, &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Name: bundle.Name, Namespace: bundle.Namespace}}); err != nil && !apierrors.IsNotFound(err) {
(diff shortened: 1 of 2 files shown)


manno:
Works well in my tests. The secrets disappear quickly.

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -565,12 +571,19 @@ func (r *BundleReconciler) createOptionsSecret(ctx context.Context, bd *fleet.Bu
 			Namespace: bd.Namespace,
 		},
 	}
-
-	if err := controllerutil.SetControllerReference(bd, secret, r.Scheme); err != nil {
-		return err
+	owners := []metav1.OwnerReference{
```
weyfonk, line 574: Is the owner reference computed explicitly here to prevent additional computations (GVK and additional struct update) run by `controllerutil.SetControllerReference`? :thinking:

manno: SetControllerReference will add to existing owners, it also returns an error if a controller owner is already set. Seems to complicated for our purpose of overwriting existing owners.


https://github.com/kubernetes-sigs/controller-runtime/blob/v0.22.1/pkg/controller/controllerutil/controllerutil.go#L65-L69

weyfonk: For reference: setting a controller reference would only error when a controller reference pointing to a [_different object_](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.22.4/pkg/controller/controllerutil/controllerutil.go#L98) is already set. In the case of a secret owned by a specific bundle deployment, that should not happen anyway.

PR #4154, opened 2025-09-22
chore: Update kube-version and rancher-version annotations in Fleet c…

thardeck:
…harts

Refers to #4155

### DIFF
--- a/charts/fleet-agent/Chart.yaml
+++ b/charts/fleet-agent/Chart.yaml
@@ -1,11 +1,11 @@
 annotations:
   catalog.cattle.io/certified: rancher
   catalog.cattle.io/hidden: "true"
-  catalog.cattle.io/kube-version: '>= 1.28.0-0 < 1.34.0-0'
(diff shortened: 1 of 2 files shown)


charts/fleet-agent/Chart.yaml
```diff
@@ -1,11 +1,11 @@
 annotations:
   catalog.cattle.io/certified: rancher
   catalog.cattle.io/hidden: "true"
-  catalog.cattle.io/kube-version: '>= 1.28.0-0 < 1.34.0-0'
+  catalog.cattle.io/kube-version: '>= 1.28.0-0 < 1.35.0-0'
```
weyfonk: :thinking: Shouldn't we update the min version as well? Or is that planned for later on?

weyfonk (approved):

charts/fleet-agent/Chart.yaml
thardeck: Technically we would still support it I think. I am not sure what the rule is. I can look what the others do.

Issue #3770, opened 2025-06-05
Resources are missing after changing path pointing to same resources


sbulage:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

After changing `GitRepo`'s path (which contains similar resources) resources are getting **deleted** and `GitRepo` shows `Modified` state.

### Expected Behavior

Even after changing path, it should not delete the resources.

### GitRepo Creation 
- Create a `GitRepo` using this path: [`without-diff`](https://github.com/sbulage/fleet-test-data/tree/master/qa-test-apps/ignore-missing-resources/without-diff)
- Wait for the resources to be created.
- `GitRepo` status will be `Modified` due to one of the `job` is `missing`.
**_Note: `without-diff` path we're not using `diff` in `fleet.yaml`_** 

### GitRepo Update
- Edit above `GitRepo` and point to the [`with-diff`](https://github.com/sbulage/fleet-test-data/tree/master/qa-test-apps/ignore-missing-resources/with-diff)
- Wait for sometime
- `GitRepo` stays in the same state i.e. `Modified` but due to the `nginx` resource missing not due to `job` missing.
_**Note: `with-diff` path we are using `diff` in `fleet.yaml` which will change `Modified` state due to job missing to `Active`.**_

### Environment

```markdown
- Architecture:
- Fleet Version:107.0.0+up0.13.0-alpha.3
- Cluster:
  - Provider:K3s
  - Options:
  - Kubernetes Version:
```

### Logs

```markdown

```

### See this video for more detailed steps to reproduce:

https://github.com/user-attachments/assets/b2da927e-7467-4cce-aeef-6a141c06a984

weyfonk:
I do not have access to the video (Github says `No video with supported format and MIME type found`), but a few tests suggest that this is intermittent. Using the provided example, when editing a `GitRepo` in-place, changing its path from `without-diff` to `with-diff`, config map `test-simple-chart-config` is _sometimes_ deleted, leading to the bundle deployment, bundle and `GitRepo` seeing the resource as missing.

This seems likely to be caused by a race condition, although it needs a closer look.

0xavi0:
[...]

PR #4157, opened 2025-09-23
Detect overlaps between deleted and created bundles

weyfonk:
When a GitRepo is updated such that a path to a bundle changes, but resources overlap between the old and new paths, there was a risk of obsolete bundles being cleaned up after newer bundles had been created. This would happen intermittently due to a race condition, and cause deletion of resources from the newer bundle, leading to the GitRepo status appearing as _Modified_.

To work around this, `fleet apply` now detects overlaps in resources between in-cluster bundles which will be deleted because they no longer match git repository contents, and new bundles to be created. When such overlap is detected, `fleet apply` populates a new `Overwrites` field in the bundle.
This enables the Fleet agent to trigger redeployment automatically when it detects that a missing resource matches that overlap field, as that may be a symptom of a bundle having been deleted at an unfortunate time. Re-deploying the corresponding Helm release, done by the agent, enables the missing resource to then be re-created.

Refers to #3770.

### Checklist

### DIFF
--- a/internal/cmd/cli/apply/apply.go
+++ b/internal/cmd/cli/apply/apply.go
@@ -369,7 +369,7 @@ func pruneBundlesNotFoundInRepo(
 			for _, inClusterRsc := range bundle.Spec.Resources {
 				for _, grb := range gitRepoBundlesMap {
 					logrus.Debugf("gitRepo bundle: %v", grb)
-					for _, grRsc := range grb.Spec.Resources { // FIXME nil pointer here: are resources not populated?
+					for _, grRsc := range grb.Spec.Resources {
 						if inClusterRsc.Name != grRsc.Name {
 							continue
 						}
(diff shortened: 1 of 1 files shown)


internal/cmd/cli/apply/apply.go
```diff
@@ -221,6 +230,18 @@ func CreateBundles(pctx context.Context, client client.Client, r record.EventRec
...
+		egWrite.Go(func() error {
+			return writeBundle(ctx, client, r, b.bundle, b.scans, *b.opts)
+		})
+	}
+
+	if err := egWrite.Wait(); err != nil {
+		return err
+	}
+
 	return nil
```
0xavi0: nit: this can maybe be reduced to:
```go
return eqWrite.Wait()
```

internal/cmd/cli/apply/apply.go
```diff
@@ -302,6 +325,18 @@ func CreateBundlesDriven(pctx context.Context, client client.Client, r record.Ev
...
```
0xavi0: Nit: same as above

0xavi0 (approved):
lgtm

internal/cmd/cli/apply/apply.go
weyfonk: Oh, good point! Done, thanks.

PR #4160, opened 2025-09-24
Limit PerClusterState to less than 1mb

manno:
When deploying to many clusters, the per-cluster state in the GitRepo resource is too big.
I could not find errors in the fleet controller logs, the update of the resource either fails silently, or I missed the log statement.

However there are errors in the rke2 logs:
```
kube-apiserver-ip-172-16-105-67 kube-apiserver E0923 13:39:10.814336       1 status.go:71] "Unhandled Error" err="apiserver received an error that is not an metav1.Status: &status.Error{s:(*status.Status)(0xc286d448f0)}: rpc error: code = ResourceExhausted desc = trying to send message larger than max (3043250 vs. 2097152)" logger="UnhandledError"
```

The gitrepo json is nearly 1.49 million characters:
```
kubectl get gitrepo -n fleet-local scale-50-single -ojson | jq -c --join-output
```

The default etcd object size limit is 1.5 MiB: https://etcd.io/docs/v3.3/dev-guide/limit/

This PR keeps the sum of all cluster ids (`namespace.Name/custer.Name`) below 1mb. Some more bytes are used for the repeating state names, e.g. "ready". 

This PR does not set the incomplete flag, as it is not used in UI and documented as:
> IncompleteState is true if a bundle summary has 10 or more non-ready resources or a non-ready resource has more 10 or more non-ready or modified states.


internal/resourcestatus/resourcekey.go
```diff
@@ -172,7 +173,10 @@ func aggregateResourceStatesClustersMap(resourceKeyStates resourceStatesByResour
 				resource.IncompleteState = true
 			}
 
-			appendToPerClusterState(&resource.PerClusterState, entry.state, entry.clusterID)
+			if size < 1024*1024 { // 1mb
```
aruiz14, line 189: Please consider truncating the result instead. The `resourceKeyStates` map will be iterated in a random order, producing a different result on each run. This could cause an infinite loop of status updates, trying to reconcile it to a value that changes with every execution

aruiz14 (changes requested):

internal/resourcestatus/resourcekey.go
manno: So much for my "naive" fix...

internal/resourcestatus/resourcekey.go
```diff
@@ -172,7 +181,10 @@ func aggregateResourceStatesClustersMap(resourceKeyStates resourceStatesByResour
 				resource.IncompleteState = true
 			}
 
-			appendToPerClusterState(&resource.PerClusterState, entry.state, entry.clusterID)
+			size += len(entry.clusterID)
+			if size < 1024*1024 { // 1mb
```
aruiz14, line 189: nit: please move 1024*1024 to a constant

internal/resourcestatus/resourcekey.go
```diff
@@ -50,10 +53,6 @@ func clusterID(bd fleet.BundleDeployment) string {
 // fromResources inspects a list of BundleDeployments and returns a list of per-cluster states per resource keys.
 // It also returns a list of errors messages produced that may have occurred during processing
 func fromResources(items []fleet.BundleDeployment) resourceStatesByResourceKey {
-	sort.Slice(items, func(i, j int) bool {
-		return clusterID(items[i]) < clusterID(items[j])
-	})
```
aruiz14, line 55: If this is removed, the `for` loop in line 179:
https://github.com/rancher/fleet/pull/4160/files#diff-3a8efc300bcb7f5c51c3277e117e6ccc24baa0a7f17d07e46a27804d41b726a0R179
will observer a different order every time, isn't it?

PR #4161, opened 2025-09-24
Update helm.sh/helm dependency to v3.18.6

thardeck:
to fix [regression](https://github.com/helm/helm/issues/31136) in Helm v3.18.5.


p-se (approved):
lgtm

Issue #4118, opened 2025-09-10
PDB & PC for Fleet


p-se:
Rancher needs to be extended:
- add new configuration variables for Fleet PDB & PC in Rancher
- import of a new cluster should result in the Fleet cluster resource getting the `spec.agentSchedulingCustomization` field properly set.

Within this context, PDB and PCs must first be supported in Fleet.

kkaempf:
@p-se what is this blocked on ?

p-se:
> [@p-se](https://github.com/p-se) what is this blocked on ?

It isn't anymore. #4162 unblocked it.

weyfonk:
* [ ] Document the feature

Jono-SUSE-Rancher:
@kkaempf -- Are we still "In Progress" on this since we're past code freeze or are you just keeping it like that because of the documentation?

kkaempf:
@Jono-SUSE-Rancher it's done in Fleet but lacks changes in Rancher (resp. UI). These will be ready in 2.14 😞

kkaempf:
done

PR #4162, opened 2025-09-24
Downgrade go and k8s

thardeck:
because Rancher 2.13 did not upgrade to Go 1.25 yet and might not before the release.

Kubernetes is only downgraded in `pkg/apis` to prevent issues in bumping our api in Rancher while Rancher did not bump its Kubernetes yet.

### DIFF
--- a/charts/fleet-crd/templates/crds.yaml
+++ b/charts/fleet-crd/templates/crds.yaml
@@ -4809,9 +4809,9 @@ spec:
                             etc.),
 
                             compute a sum by iterating through the elements of this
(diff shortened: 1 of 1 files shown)


p-se (approved):
lgtm

thardeck:
> Why is this mixing doc updates with Go downgrades ?

It does not, I just had to run `go generate` because of dependency changes and in this succession the generated templates were changed.

kkaempf (approved):
🤞🏻

p-se:
Unblocks #4118

Issue #3617, opened 2025-05-02
[SURE-9997][RFE] Pass secrets and config maps to workload cluster


alexander-demicev:
### Is your feature request related to a problem?

We're using HelmApp to install out-of-tree cloud providers across multiple clusters, as shown in this [example](https://github.com/rancher/turtles/blob/main/examples/applications/ccm/vsphere/helm-chart.yaml). For our use case, we need to pass a ConfigMap and a Secret to all workload clusters where HelmApp installs the chart.

Ideally, we'd like to have something similar to how one of Cluster API CRDs handle this, like in this [template](https://github.com/rancher/turtles/blob/release-0.19/test/e2e/data/cluster-templates/vsphere-kubeadm-topology.yaml#L154):

```yaml
resources:
  - kind: Secret
    name: vsphere-config-secret
  - kind: Secret
    name: cloud-provider-vsphere-credentials
  - kind: ConfigMap
    name: cpi-configmap
```

manno:
This is related to https://github.com/rancher/fleet/issues/3399

manno:
We will need to keep in mind that we want to re-use the helmops implementation for gitops later.

weyfonk:
## Additional QA
 
### Problem
Users need a way to propagate resources from the upstream cluster to all downstream clusters targeted by a HelmOp.
 
### Solution
HelmOps now expose a new `downstreamResources` field, which can be used to reference resources in the same namespace, by kind and name.
When those resources exist:
* the Fleet controller will copy them from their source namespace (the HelmOp's namespace) to the namespaces of each downstream cluster, next to corresponding bundle deployment resources
* on each targeted downstream cluster, the Fleet agent will copy those resources from the upstream namespace, for the cluster, into their target namespace, before deploying the bundle deployment. It will also delete those resources from the target namespace when the bundle deployment is deleted, unless the HelmOp was configured with `keepResources` set to `true`.

Resources referenced by `downstreamResources` are not directly monitored. Although the agent will propagate updates to those resources, that will only happen if the bundle deployment itself is reconciled, for instance when the parent HelmOp is updated.
 
### Testing

### Engineering Testing
#### Manual Testing
N/A

#### Automated Testing
##### Unit tests
* referenced secret not found
* referenced config map not found
* unsupported resource referenced (neither a secret nor a config map)

##### End-to-end tests
* HelmOp referencing a secret and a config map through `downstreamResources`, and using their contents to populate Helm values through `valuesFrom`:
[...]

PR #4167, opened 2025-09-26
Improve bundle and GitRepo status error reporting

weyfonk:
This propagates a number of errors, which can happen in the process of reconciling a resource, to that resource's status:
* GitRepo: failure to create a git job
* Bundle: failures happening when:
    * resolving a Helm chart version
    * handling OCI storage manifests
    * targeting the bundle
    * writing Content resources
    * managing values secrets, whether writing or reading them
    * adding a finalizer
    * computing rollout partitions
    * cloning OCI storage or Helm access secrets

Refers to #3617.
This patch is a preliminary step to making the changes referenced in that issue.

Most of its lines come from new unit test cases used to validate error status updates in the cases mentioned above.

A new file in the `reconciler` package, `error_handling.go`, contains extracted logic to be reused across reconcilers for easier and more consistent error reporting.

### Checklist

### DIFF
--- a/internal/cmd/controller/gitops/reconciler/gitjob_controller.go
+++ b/internal/cmd/controller/gitops/reconciler/gitjob_controller.go
@@ -234,7 +234,7 @@ func (r *GitJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		return res, updateErrorStatus(ctx, r.Client, req.NamespacedName, gitrepo.Status, err)
 	}
 
-	setAcceptedCondition(&gitrepo.Status, nil)
+	reconciler.SetCondition(v1alpha1.GitRepoAcceptedCondition, &gitrepo.Status, nil)
 
 	err = updateStatus(ctx, r.Client, req.NamespacedName, gitrepo.Status)
 	if err != nil {
@@ -654,14 +654,10 @@ func setStatusFromGitjob(ctx context.Context, c client.Client, gitRepo *v1alpha1
 	return nil
 }
 
-// setAcceptedCondition sets the condition and updates the timestamp, if the condition changed
-func setAcceptedCondition(status *v1alpha1.GitRepoStatus, err error) {
-	reconciler.SetErrorInCondition(v1alpha1.GitRepoAcceptedCondition, status, err)
-}
-
 // updateErrorStatus sets the condition in the status and tries to update the resource
 func updateErrorStatus(ctx context.Context, c client.Client, req types.NamespacedName, status v1alpha1.GitRepoStatus, orgErr error) error {
-	setAcceptedCondition(&status, orgErr)
+	reconciler.SetCondition(v1alpha1.GitRepoAcceptedCondition, &status, orgErr)
+
 	if statusErr := updateStatus(ctx, c, req, status); statusErr != nil {
 		merr := []error{orgErr, fmt.Errorf("failed to update the status: %w", statusErr)}
 		return errutil.NewAggregate(merr)
--- a/internal/cmd/controller/reconciler/bundle_controller.go
+++ b/internal/cmd/controller/reconciler/bundle_controller.go
@@ -631,7 +631,7 @@ func (r *BundleReconciler) updateErrorStatus(
 	orig, bundle *fleet.Bundle,
 	orgErr error,
 ) error {
-	SetErrorInCondition(string(fleet.Ready), &bundle.Status, orgErr)
+	SetCondition(string(fleet.Ready), &bundle.Status, orgErr)
 
 	if statusErr := r.updateStatus(ctx, orig, bundle); statusErr != nil {
 		merr := []error{orgErr, fmt.Errorf("failed to update the status: %w", statusErr)}
@@ -726,7 +726,7 @@ func (r *BundleReconciler) computeResult(
 
 	err = fmt.Errorf("%s: %w", prefix, err)
 
-	SetErrorInCondition(string(fleet.Ready), &bundle.Status, err)
+	SetCondition(string(fleet.Ready), &bundle.Status, err)
 
 	return ctrl.Result{}, r.updateErrorStatus(ctx, bundleOrig, bundle, err)
 }
--- a/internal/cmd/controller/reconciler/error_handling.go
+++ b/internal/cmd/controller/reconciler/error_handling.go
@@ -30,11 +30,13 @@ func CheckRetryable(err error, logger logr.Logger) (bool, ctrl.Result, error) {
 	return false, ctrl.Result{}, nil
 }
 
-// SetErrorInCondition sets the condition and updates the timestamp, if the condition changed
-func SetErrorInCondition[T any](cond string, s copyable[T], err error) {
+// SetCondition sets the condition and updates the timestamp, if the condition changed
+func SetCondition[T any](cond string, s copyable[T], err error) {
 	c := condition.Cond(cond)
 	origStatus := s.DeepCopy()
+
 	c.SetError(s, "", fleetutil.IgnoreConflict(err))
+
 	if !equality.Semantic.DeepEqual(origStatus, s) {
 		c.LastUpdated(s, time.Now().UTC().Format(time.RFC3339))
 	}


internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -206,11 +165,13 @@ func (r *BundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+	bundleOrig := bundle.DeepCopy()
+
 	if err := r.ensureFinalizer(ctx, bundle); err != nil {
-		return ctrl.Result{}, err
-	}
+		err = fmt.Errorf("failed to add finalizer to bundle: %w", err)
 
-	bundleOrig := bundle.DeepCopy()
+		return ctrl.Result{}, r.updateErrorStatus(ctx, req.NamespacedName, bundleOrig, bundle, err)
+	}
```
0xavi0: Question: what is the value of propagating this error to the user?
This is about adding a finalizer to a resource, something internal that the user can't do anything about, and... it should be retried as it comes from an API call and could be temporary.


I mean, it is set in the Condition and the Reconciler would return `ctrl.Result{}, nil` if calling `r.updateErrorStatus` is successful.
That means the reconciler won't requeue itself and it won't be retried. 
`ensureFinalizer` is doing an `Update()` which calls the API server, and that could incur, for example in conflicts.... so requeuing because the Reconcile call returned an error is still desirable (IMHO).

If adding that kind of internal error is useful to the user (I have my doubts because of giving internal details) I would still return the error so the request is requeued.

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -228,7 +189,10 @@ func (r *BundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 				logger.Info(err.Error())
 				return ctrl.Result{RequeueAfter: durations.DefaultRequeueAfter}, nil
 			}
-			return ctrl.Result{}, err
+
+			err = fmt.Errorf("failed to load values secret for bundle: %w", err)
+
+			return ctrl.Result{}, r.updateErrorStatus(ctx, req.NamespacedName, bundleOrig, bundle, err)
```
0xavi0: Same remark here. We might still be interested on returning the error so the request is requeued.
We could also be showing possible k8s internal error details to the user (he/she can't do anything about them)

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -246,13 +210,9 @@ func (r *BundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
 		if _, err := semver.StrictNewVersion(versionToCheck); err != nil {
-			setReadyCondition(
-				&bundle.Status,
-				fmt.Errorf("chart version cannot be deployed; check HelmOp status for more details: %v", err),
-			)
+			err = fmt.Errorf("chart version cannot be deployed; check HelmOp status for more details: %w", err)
 
-			err := r.updateStatus(ctx, bundleOrig, bundle)
-			return ctrl.Result{}, err
+			return ctrl.Result{}, r.updateErrorStatus(ctx, req.NamespacedName, bundleOrig, bundle, err)
```
0xavi0: This is a good example of errors that might be useful to the user, because semver is something he/she can do something about. And retrying (retuning the error in the Reconcile call) makes no sense because it's an error that is not retry-able (it is always going to fail)

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -300,24 +276,32 @@ func (r *BundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		// agents have access to all resources and use their specific
 		// set of `BundleDeploymentOptions`.
 		if err := r.Store.Store(ctx, resourcesManifest); err != nil {
-			return ctrl.Result{}, err
+			err = fmt.Errorf("could not copy manifest into Content resource: %w", err)
+
+			return ctrl.Result{}, r.updateErrorStatus(ctx, req.NamespacedName, bundleOrig, bundle, err)
```
0xavi0: We also might need to requeue as `Store` calls internally `client.Create` and we could have a temporary issue making calls to the API server

0xavi0 (commented):
I think there are a few errors that are maybe too internal to be propagated to the user (like the finalizer one) and we might be missing the retry mechanism from controller-runtime.

I added comments to a few (not all), but basically I think any error coming from an API server call could be a potential candidate to be requeued.

internal/cmd/controller/reconciler/bundle_controller.go
weyfonk: Thanks, I indeed had (and still have) doubts about exposing this kind of errors to users since, as you've noted, they're internal and not something users can act upon.
I did expose this one thinking of users who may not have access to logs and still want to know what might be going on, but in this particular instance it might be slightly over the top and not very informative.

> the Reconciler would return `ctrl.Result{}, nil` if calling `r.updateErrorStatus` is successful.
> That means the reconciler won't requeue itself and it won't be retried.

That is not my understanding: when successful, `updateErrorStatus` returns the original error passed to it, meaning that it _would_ requeue in such cases.

0xavi0: Apologies, I think I looked at the wrong function and thought it was returning the value of calling `Update`.
Looks good to me, then...

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -318,24 +287,42 @@ func (r *BundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
 		if err := r.Store.Store(ctx, resourcesManifest); err != nil {
-			return ctrl.Result{}, err
+			err = fmt.Errorf("could not copy manifest into Content resource: %w", err)
+
+			if errors.Is(err, fleetutil.ErrRetryable) {
+				logger.Info(err.Error())
+				return ctrl.Result{RequeueAfter: durations.DefaultRequeueAfter}, nil
+			}
+
+			return ctrl.Result{}, r.updateErrorStatus(ctx, req.NamespacedName, bundleOrig, bundle, err)
```
0xavi0: I think this file is using this block more than once. 

Would it make sense to make it a function?
And.. I think it would make sense to make it a function accessible from all controllers so we can use it as the standard for errors that might (or not) be shown to the user via conditions or retried. 

That way, when we detect an error in any controller we do the same.

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -714,6 +657,24 @@ func (r *BundleReconciler) handleContentAccessSecrets(ctx context.Context, bundl
 	return nil
 }
 
+// updateErrorStatus sets the Ready condition in the bundle status and tries to update the resource.
+// Setting that condition makes the error message visible in the Rancher UI.
+func (r *BundleReconciler) updateErrorStatus(
+	ctx context.Context,
+	req types.NamespacedName,
```
0xavi0: I wonder why the linter does not complain about `req`.
It's not used in the function... is it?

0xavi0: Also .... same remark about the other block checking if it's retrayable, etc... 
We might want to add this function to a shared package accesible for all controllers.

I see we have the same in the gitops controller using the `Accepted` condition instead. 
Maybe we can have the condition as a parameter? 

Would be cool to have like a standard way for dealing with errors.

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -228,7 +228,7 @@ func (r *GitJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 
 	res, err := r.manageGitJob(ctx, logger, gitrepo, oldCommit, repoPolled)
 	if err != nil || res.RequeueAfter > 0 {
-		return res, err
+		return res, updateErrorStatus(ctx, r.Client, req.NamespacedName, gitrepo.Status, err)
```
0xavi0, line 234: `manageGitJob` could also return errors coming from calls to the k8s API. 

We could maybe use the same approach as in the Bundle Controller and check if the error is Retryable

internal/cmd/controller/reconciler/bundle_controller.go
weyfonk: Yes, great idea, see the new `reconciler/error_handling.go` :)

internal/cmd/controller/reconciler/bundle_controller.go
weyfonk: Nice catch, parameter removed, thanks!
I've added a new function to set an error on a parametrised condition, see `reconciler/error_handling.go`.

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
weyfonk: Strong agree. As discussed offline, I think this would be a larger effort, not least because we would need to check which of the returned errors should lead to retries in the context of a GitRepo, job, etc.
Let's create a follow-up issue for that.

Issue #3646, opened 2025-05-06
[SURE-9881] CA in a secret referenced in spec.helmSecretName or spec.helmSecretNameForPaths not used when pulling Fleet bundle helm chart from a git repository


kkaempf:
### Issue description:

The fleet documentation (https://fleet.rancher.io/ref-fleet-yaml) notes the ability to pull the chart in a bundle from a git repository:

~~~
 # Git repositories can be downloaded via unauthenticated http, by using for
  # example:
  #
  # `git::http://github.com/rancher/fleet-examples/single-cluster/helm`.
~~~
However, this git method does not use any CA configured in a secret referenced on the GitRepo spec.helmSecretName or spec.helmSecretNameForPaths. By comparison, such a CA is used when pulling the chart with simple GET of chart .tgz URL

### Business impact:
Customer is unable to deploy charts via Fleet that are hosted in their internal gitea instance, which is deployed with a private CA

darkhonor:
Is there a way to configure the GitRepo to ignore the custom CA certificates until this is patched? I have tried adding insecureSkipTLSVerify: true but with no effect. 

This has the effect of rendering Fleet useless in Airgap or STIG environments.

rancherbot:
Not creating port issue, milestone 2.13.1 does not exist or is not an open milestone

p-se:
# QA Template

1. Create a (self-hosted) git server that supports HTTPS. 
2. Provide it with a self-signed certificate.
3. Create a git repository on that self-hosted git server which contains a simple helm chart. It can just contain a ConfigMap.
4. Create a repository that contains a `fleet.yaml` file which points to the git server in its `helm.chart` field (e.g. `git::http://gitserver/foo/bar`). It can but does not have to be on the self-hosted git server. This URL instructs go-getter to (git) clone the repository behind `http://gitserver/foo/bar`.
5. Create a GitRepo resource which points to the repository that contains the `fleet.yaml` file. This GitRepo can have different settings, depending on what should be tested:
    1. `CABundle` field in the GitRepo resource. If this field is set, it will use the certificate to connect to the first git repository that contains the `fleet.yaml` file. It will also try to use this certificate to connect to the git server that the `fleet.yaml`'s `helm.chart` field points to, unless
    1. `helmSecretName` in the GitRepo resource points to an existing secret and has a configured `CABundle` or has `InsecureSkipVerify` set to true, or
[...]

PR #4168, opened 2025-09-26
Support multiple repositories in e2e/nginx-git-setup

p-se:
Which adds support for using several git repositories, for instance when
testing fleet.yaml's helm.chart field with a HTTPS URL.

Refers to #3646 

### Checklist


p-se:
Won't be needed in the tests.

PR #4170, opened 2025-09-26
Add resources section to helm chart values

manno:


charts/fleet/values.yaml
```diff
@@ -76,6 +76,14 @@ tolerations: []
 ## Pod affinity for the controllers.
 affinity: {}
 
+resources: {}
```
weyfonk, line 80: Nice, should we perhaps clarify that these resource constraints would apply to each deployment separately (fleet-controller, gitOps, helmOps controllers) and not be propagated to agents? That last bit may be obvious to most, but just to be on the safe side as the controller also manages agents.

weyfonk (approved):

charts/fleet/values.yaml
manno, line 80:
```suggestion
# Pod resource limits and requests for the controllers
resources: {}
```

Issue #3726, opened 2025-05-27
Simple scheduling of GitOps deployments


weyfonk:
History on this includes PRs #450 and #727, both outdated, and issue #383.

The current idea is to support simple, crontab-like syntax for scheduling deployment of workloads.

### Scope

This applies to GitOps deployments. HelmOps is not covered at this point.

### Assumptions
* A git repository referenced by a GitRepo/bundle with scheduling enabled should not be polled before the scheduled start time
    * In turn, this means that the gitOps controller would be responsible for enforcing scheduling:
        * Delaying polling of the git repository and creation of a job cloning the git repository and creating bundles until the next scheduled start time
        * Updating the GitRepo's status to something like `Scheduled`, perhaps stating the next scheduled start time

* Pausing should still have precedence over scheduling; as per the previous point, enforcing that precedence at GitRepo level should be sufficient

### Open questions
* Where should this configuration be applied? Should it be at cluster or even cluster group level, with options to override it at `GitRepo` or even bundle deployment (through target customisations) levels?
    * Configuring this should be possible at least at GitRepo level; future iterations could add cluster or cluster group-level config and target customisation support, with easy migration paths for users:
        * When target customisations are added: updating `fleet.yaml` files to use them
        * When cluster or cluster group support is added (with overrides at lower levels): configuring cluster/cluster group scheduling, which would still be overridden at lower levels (GitRepo, target customisations) until removed from such levels if needed

### Follow-ups (for future iterations):
* Apply commits one by one even if all have been pushed before the scheduling start time
    * Example: upgrade of a chart across multiple minors, none of which should be skipped
    * This should probably be made configurable, to limit the number of deployments when such a precaution is not necessary
    * Could the git job iterate over commits since the latest deployed one, and create bundles, deploy them, and wait for them to be ready for each such commit? This could prevent the monitored git repository from having to be cloned once per commit.
* Extend this feature to HelmOps

weyfonk:
More information on how Rancher supports upgrade windows: https://github.com/rancher/system-upgrade-controller/pull/335.

kkaempf:
> More information on how Rancher supports upgrade windows: [rancher/system-upgrade-controller#335](https://github.com/rancher/system-upgrade-controller/pull/335).

While great for general guidance, I don't think transferring this approach 1:1 to Fleet is applicable.
As we only can control when Fleet starts a transition, we can in no way guarantee an `endTime`.

mmartin24:
[...]

PR #4184, opened 2025-10-01
Scheduled BundleDeployments experimental version

0xavi0:
This PR adds the ability to control whether a `BundleDeployment` can or cannot be installed, based on defined Schedules.

A new CRD, `Schedule`, is introduced. It defines time intervals during which Clusters—selected via label selectors—are allowed to accept new deployments.

An example `Schedule` would look like this:

```yaml
kind: Schedule
metadata:
  name: schedule1
  namespace: fleet-default
spec:
  schedule: "0 */5 * * * *"
  duration: 1m
  targets:
    clusters:
      - name: local
        clusterSelector:
          matchLabels:
            env: dev
```

This would allow new deployments to be installed every 5 minutes (at second 0) for a duration of 1 minute, and it would apply to all clusters located in the *fleet-default* namespace that have the label *env=dev*.

When a `Schedule` is applied, the controller evaluates which `Cluster`s match the defined selectors. For all matching `Cluster`s, the `Status.Scheduled` field is set to _true_.

This field indicates that the `Cluster` is governed by a `Schedule`, meaning no new `BundleDeployment`s will be applied to it unless the `Status.ActiveSchedule` field — also newly introduced — is explicitly set to _true_.

The `Status.ActiveSchedule` field is set to true at the moment a `Schedule` begins execution and remains _true_ for the duration defined in the `Schedule`’s specification.

The mechanism used to prevent an Agent from applying a `BundleDeployment` while a `Schedule` is active mirrors the existing implementation of the "Pause" feature.

The controller actively monitors changes to any `Cluster` that is part of an existing `Schedule`. If a `Cluster`'s labels or name are modified, the controller re-evaluates and updates the affected `Schedule`’s target set accordingly.

Additionally, when a `Schedule` begins execution, the controller performs a fresh evaluation to detect newly matching `Cluster`s or exclude any that no longer meet the criteria.

The set of `Cluster`s associated with a `Schedule` is computed either at the time of `Schedule` creation, upon modification of the `Schedule`, or at the start of its execution.

---
### 📅 **Schedule Execution Implementation**

The execution logic for the Schedule has been implemented as follows:

* The next execution time is calculated based on the current date and time.
* The duration until that time is computed, and a `quartz.NewRunOnceTrigger` is created.
* When execution occurs, another `quartz.NewRunOnceTrigger` is created for the defined duration, and the `Started` property is set to `true`.
* Upon the next execution, the start time of the following execution is recalculated, and the `Started` property is set back to `false`.

> ❗ `quartz.NewCronTriggerWithLoc` was intentionally not used

The reason for avoiding `quartz.NewCronTriggerWithLoc` is to prevent jobs from remaining in the scheduler indefinitely and to allow more precise control over job execution — both start time and duration.

The goal is to *minimize potential race conditions*, especially in edge cases where a job is scheduled to begin very close to the end of a previous execution.

A constraint has also been introduced to ensure that the **minimum viable execution duration is at least 1 second**, after computing feasibility relative to the execution start time.

That is, there must be **at least 1 second** between consecutive execution start times.
This further helps reduce race conditions, since starting an execution involves multiple API calls to the Kubernetes API server.

> 🧪 *This feature is currently experimental.*

To enable it, set the following environment variable:

```
EXPERIMENTAL_SCHEDULES=true
```

---

### 🔧 **Pending Work**

* [ ] **Overlap detection for Schedules** has not yet been implemented.
* [ ] **Targets** have been implemented using a `clusters` field, anticipating future support for other resources like `GitRepos`, where polling or Bundle updates could be paused.
* [ ] Currently, only **Cron-based Schedules** are supported as execution triggers (with **second-level granularity**).
* [ ] Emit events when a new `BundleDeployment` is ready but did not deploy because the Cluster is scheduled and not active?

Refers to: https://github.com/rancher/fleet/issues/3726

### Checklist

### DIFF
--- a/internal/cmd/controller/reconciler/cronduration_job.go
+++ b/internal/cmd/controller/reconciler/cronduration_job.go
@@ -214,13 +214,6 @@ func (c *CronDurationJob) executeStart(ctx context.Context) error {
 		return err
 	}
 
-	// Sets ActiveSchedule to true for all matching clusters.
-	for _, cluster := range clusters {
-		if err := setClusterActiveSchedule(context.Background(), c.client, cluster, c.Schedule.Namespace, true); err != nil {
-			return err
-		}
-	}
-
 	// Sets Scheduled to false for all clusters that previously matched but no longer do.
 	for _, cluster := range c.MatchingClusters {
 		if !slices.Contains(clusters, cluster) {
@@ -231,9 +224,11 @@ func (c *CronDurationJob) executeStart(ctx context.Context) error {
 		}
 	}
 
-	// Sets Scheduled to true to all the matching clusters
-	if err := setClustersScheduled(ctx, c.client, clusters, c.Schedule.Namespace, true); err != nil {
-		return err
+	// Sets ActiveSchedule to true for all matching clusters.
+	for _, cluster := range clusters {
+		if err := setClusterActiveSchedule(ctx, c.client, cluster, c.Schedule.Namespace, true); err != nil {
+			return err
+		}
 	}
 	c.MatchingClusters = clusters
 
--- a/internal/cmd/controller/reconciler/cronduration_job_test.go
+++ b/internal/cmd/controller/reconciler/cronduration_job_test.go
@@ -2,6 +2,7 @@ package reconciler
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -298,14 +299,13 @@ var _ = Describe("CronDurationJob", func() {
 
 	Context("scheduleJob", func() {
 		It("should schedule a job and update status", func() {
-			// wait a bit if we are too close to second 59...
-			// this is to ensure calculations based on minutes and seconds
-			waitIfTimeIsTooCloseTo59()
-
+			// set the schedule time to now plus 10 seconds so we can check
+			// later knowing the expected value.
+			scheduleTime := time.Now().Add(10 * time.Second)
+			schedule.Spec.Schedule = fmt.Sprintf("%d %d %d * * *", scheduleTime.Second(), scheduleTime.Minute(), scheduleTime.Hour())
 			job, err := newCronDurationJob(ctx, schedule, scheduler, k8sclient)
 			Expect(err).NotTo(HaveOccurred())
 
-			timeBeforeSchedule := time.Now()
 			err = job.scheduleJob(ctx)
 			Expect(err).NotTo(HaveOccurred())
 
@@ -319,21 +319,20 @@ var _ = Describe("CronDurationJob", func() {
 			err = k8sclient.Get(ctx, types.NamespacedName{Name: "test-schedule", Namespace: "default"}, updatedSchedule)
 			Expect(err).NotTo(HaveOccurred())
 			Expect(updatedSchedule.Status.Active).To(BeFalse())
-			// time should be the next minute and 0 seconds
-			Expect(updatedSchedule.Status.NextStartTime.Time.Second()).To(BeZero())
-			Expect(updatedSchedule.Status.NextStartTime.Time.Minute()).To(Equal(timeBeforeSchedule.Add(time.Minute).Minute()))
-			Expect(updatedSchedule.Status.NextStartTime.Time.Hour()).To(Equal(timeBeforeSchedule.Hour()))
-			// Expect(updatedSchedule.Status.NextStartTime.Time).To(BeTemporally("~", timeBeforeSchedule.Add(time.Minute), 2*time.Second))
+
+			Expect(updatedSchedule.Status.NextStartTime.Time.Second()).To(Equal(scheduleTime.Second()))
+			Expect(updatedSchedule.Status.NextStartTime.Time.Minute()).To(Equal(scheduleTime.Minute()))
+			Expect(updatedSchedule.Status.NextStartTime.Time.Hour()).To(Equal(scheduleTime.Hour()))
 			Expect(updatedSchedule.Status.MatchingClusters).To(Equal(job.MatchingClusters))
 		})
 	})
 
 	Context("updateJob", func() {
 		It("should update an existing job", func() {
-			// wait a bit if we are too close to second 59...
-			// this is to ensure calculations based on minutes and seconds
-			waitIfTimeIsTooCloseTo59()
-
+			// set the schedule time to now plus 10 seconds so we can check
+			// later knowing the expected value.
+			scheduleTime := time.Now().Add(10 * time.Second)
+			schedule.Spec.Schedule = fmt.Sprintf("%d %d %d * * *", scheduleTime.Second(), scheduleTime.Minute(), scheduleTime.Hour())
 			job, err := newCronDurationJob(ctx, schedule, scheduler, k8sclient)
 			Expect(err).NotTo(HaveOccurred())
 
@@ -341,11 +340,11 @@ var _ = Describe("CronDurationJob", func() {
 			Expect(err).NotTo(HaveOccurred())
 
 			// Modify schedule to force an update
-			schedule.Spec.Schedule = "0 */2 * * * *"
+			newScheduleTime := time.Now().Add(30 * time.Second)
+			schedule.Spec.Schedule = fmt.Sprintf("%d %d %d * * *", newScheduleTime.Second(), newScheduleTime.Minute(), newScheduleTime.Hour())
 			updatedJob, err := newCronDurationJob(ctx, schedule, scheduler, k8sclient)
 			Expect(err).NotTo(HaveOccurred())
 
-			timeBeforeUpdate := time.Now()
 			err = updatedJob.updateJob(ctx)
 			Expect(err).NotTo(HaveOccurred())
 
@@ -361,18 +360,9 @@ var _ = Describe("CronDurationJob", func() {
 			Expect(err).NotTo(HaveOccurred())
 			Expect(updatedSchedule.Status.Active).To(BeFalse())
 
-			// next fire time should be the next even minute and 0 seconds
-			Expect(updatedSchedule.Status.NextStartTime.Time.Second()).To(BeZero())
-			if timeBeforeUpdate.Minute()%2 == 0 {
-				Expect(updatedSchedule.Status.NextStartTime.Time.Minute()).To(Equal(timeBeforeUpdate.Add(2 * time.Minute).Minute()))
-			} else {
-				Expect(updatedSchedule.Status.NextStartTime.Time.Minute()).To(Equal(timeBeforeUpdate.Add(1 * time.Minute).Minute()))
-			}
-			if timeBeforeUpdate.Minute() < 59 {
-				Expect(updatedSchedule.Status.NextStartTime.Time.Hour()).To(Equal(timeBeforeUpdate.Hour()))
-			} else {
-				Expect(updatedSchedule.Status.NextStartTime.Time.Hour()).To(Equal(timeBeforeUpdate.Add(1 * time.Hour).Hour()))
-			}
+			Expect(updatedSchedule.Status.NextStartTime.Time.Second()).To(Equal(newScheduleTime.Second()))
+			Expect(updatedSchedule.Status.NextStartTime.Time.Minute()).To(Equal(newScheduleTime.Minute()))
+			Expect(updatedSchedule.Status.NextStartTime.Time.Hour()).To(Equal(newScheduleTime.Hour()))
 			Expect(updatedSchedule.Status.MatchingClusters).To(Equal(job.MatchingClusters))
 		})
 	})
@@ -552,14 +542,3 @@ var _ = Describe("CronDurationJob", func() {
 		})
 	})
 })
-
-func waitIfTimeIsTooCloseTo59() {
-	for {
-		now := time.Now()
-		// if we are up to second 55... just break
-		if now.Second() < 55 {
-			break
-		}
-		time.Sleep(200 * time.Millisecond)
-	}
-}
--- a/charts/fleet/ci/debug-values.yaml
+++ b/charts/fleet/ci/debug-values.yaml
@@ -51,6 +51,7 @@ controller:
       gitrepo: "1"
       bundle: "1"
       bundledeployment: "1"
+      schedule: "1"
 
 shards:
   - id: shard0
--- a/charts/fleet/ci/nobootstrap-values.yaml
+++ b/charts/fleet/ci/nobootstrap-values.yaml
@@ -50,6 +50,7 @@ controller:
       gitrepo: "1"
       bundle: "1"
       bundledeployment: "1"
+      schedule: "1"
 
 shards:
   - id: shard0
--- a/charts/fleet/ci/nodebug-values.yaml
+++ b/charts/fleet/ci/nodebug-values.yaml
@@ -50,6 +50,7 @@ controller:
       gitrepo: "1"
       bundle: "1"
       bundledeployment: "1"
+      schedule: "1"
 
 shards:
   - id: shard0
--- a/charts/fleet/ci/nogitops-values.yaml
+++ b/charts/fleet/ci/nogitops-values.yaml
@@ -50,6 +50,7 @@ controller:
       gitrepo: "1"
       bundle: "1"
       bundledeployment: "1"
+      schedule: "1"
 
 shards:
   - id: shard0
--- a/charts/fleet/ci/nohelmops-values.yaml
+++ b/charts/fleet/ci/nohelmops-values.yaml
@@ -50,6 +50,7 @@ controller:
       gitrepo: "1"
       bundle: "1"
       bundledeployment: "1"
+      schedule: "1"
 
 shards:
   - id: shard0
--- a/charts/fleet/templates/deployment.yaml
+++ b/charts/fleet/templates/deployment.yaml
@@ -89,6 +89,10 @@ spec:
         - name: IMAGESCAN_RECONCILER_WORKERS
           value: {{ quote $.Values.controller.reconciler.workers.imagescan }}
         {{- end }}
+        {{- if $.Values.controller.reconciler.workers.schedule }}
+        - name: SCHEDULE_RECONCILER_WORKERS
+          value: {{ quote $.Values.controller.reconciler.workers.schedule }}
+        {{- end }}
 {{- if $.Values.extraEnv }}
 {{ toYaml $.Values.extraEnv | indent 8}}
 {{- end }}
--- a/charts/fleet/values.yaml
+++ b/charts/fleet/values.yaml
@@ -119,6 +119,7 @@ controller:
       cluster: "50"
       clustergroup: "50"
       imagescan: "50"
+      schedule: "50"
 
 gitjob:
   replicas: 1


internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -745,7 +753,9 @@ func (r *BundleReconciler) cleanupOrphanedBundleDeployments(ctx context.Context,
 		return err
 	}
 	toDelete := slices.DeleteFunc(list, func(bd fleet.BundleDeployment) bool {
-		return uidsToKeep.Has(bd.UID)
+		// don't delete BundleDeployments that are not in schedule as
+		// that uninstall the deployment in the agent
```
```suggestion
		// that would uninstall the deployment in the agent
```

pkg/apis/fleet.cattle.io/v1alpha1/bundledeployment_types.go
```diff
@@ -351,6 +351,13 @@ type BundleDeploymentSpec struct {
 	// ValuesHash is the hash of the values used to deploy the bundle.
 	// +nullable
 	ValuesHash string `json:"valuesHash,omitempty"`
+
+	// NotInSchedule specifies if the BundleDeployment can be updated.
+	// If set to true, will stop any BundleDeployments from being
+	// updated.
+	// If true, BundleDeployments will be marked as out of sync
+	// when changes are detected.
+	NotInSchedule bool `json:"notInSchedule,omitempty"`
```
weyfonk: I suppose having a `NotInSchedule` field is easier since it defaults to `false`, but this may get a bit confusing.
Could we make it `InSchedule` instead? Or maybe rename it to something like `OffSchedule`? :thinking:

pkg/apis/fleet.cattle.io/v1alpha1/schedule_types.go
```diff
@@ -0,0 +1,80 @@
...
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Schedule",type=string,JSONPath=`.spec.schedule`
+// +kubebuilder:printcolumn:name="Duration",type=string,JSONPath=`.spec.duration`
+// +kubebuilder:printcolumn:name="Active",type=string,JSONPath=`.status.active`
+// +kubebuilder:printcolumn:name="NextStart",type="string",JSONPath=`.status.nextStartTime`
+
+// Schedule represents a schedule in which deployments are allower or not, depending on its definition.
```
weyfonk:
```suggestion
// Schedule represents a schedule in which deployments are allowed or not, depending on its definition.
```

pkg/apis/fleet.cattle.io/v1alpha1/schedule_types.go
```diff
@@ -0,0 +1,80 @@
...
+	Conditions []genericcondition.GenericCondition `json:"conditions,omitempty"`
+	// MatchingClusters is the list of clusters targeted by the Schedule
+	MatchingClusters []string `json:"matchingClusters,omitempty"`
+}
+
+type ScheduleTargets struct {
+	Clusters []ScheduleTarget `json:"clusters,omitempty"`
+}
+
+// ScheduleTarget represent a resource (or group of resources affected by an Schedule)
```
weyfonk: nit:
```suggestion
// ScheduleTarget represents a resource (or group of resources) affected by a Schedule
```

internal/experimental/experimental.go
weyfonk, line 1: I think it would make sense to move this package to `internal/`, unless we see a use case for it being imported from an external codebase.

e2e/single-cluster/schedule_test.go
```diff
@@ -0,0 +1,267 @@
...
+
+			bdList := &fleet.BundleDeploymentList{}
+			err = env.Unmarshal(out, bdList)
+			g.Expect(err).ToNot(HaveOccurred())
+
+			g.Expect(bdList.Items).To(HaveLen(1))
+
+			bd = bdList.Items[0]
+			g.Expect(bd.Spec.NotInSchedule).To(BeFalse())
+		}).Should(Succeed())
```
weyfonk, line 154: Do we want to set stricter time constraints on this `Eventually` block, just to be on the safe side?
In this case, as we're expecting the schedule to be active for 30s every minute, this block should fail much earlier than after the default 5-minute timeout, i.e. after about a minute, shouldn't it?

go.mod
```diff
@@ -40,9 +40,9 @@ require (
 	github.com/prometheus/client_model v0.6.2
 	github.com/prometheus/common v0.65.0
 	github.com/rancher/fleet/pkg/apis v0.13.0
-	github.com/rancher/lasso v0.2.4
+	github.com/rancher/lasso v0.2.3
 	github.com/rancher/wrangler/v3 v3.2.4
-	github.com/reugn/go-quartz v0.15.2
+	github.com/reugn/go-quartz v0.15.1
```
weyfonk: Interesting that `go-quartz` and `lasso` are _downgraded_ here :thinking:

internal/cmd/agent/controller/drift_controller.go
```diff
@@ -86,6 +86,12 @@ func (r *DriftReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 
 		return ctrl.Result{}, err
 	}
+	if bd.Spec.NotInSchedule {
```
weyfonk: Does this need to be done in _both_ bundle deployment and drift controllers, since the underlying `DriftDetect` is the same in both controllers?
(same question about paused bundle deployments, btw)

internal/cmd/controller/operator.go
```diff
@@ -168,6 +169,21 @@ func start(
...
 	}
 
+	if experimental.SchedulesEnabled() {
+		if err = (&reconciler.ScheduleReconciler{
+			Client:   mgr.GetClient(),
+			Scheme:   mgr.GetScheme(),
+			Recorder: mgr.GetEventRecorderFor(fmt.Sprintf("fleet-schedule-ctrl%s", shardIDSuffix)),
+			ShardID:  shardID,
+
+			Workers:   workersOpts.Bundle,
```
weyfonk: Is this deliberate? If so, why should we use the same worker count as for the bundle controller?

internal/cmd/controller/reconciler/schedule_controller.go
```diff
@@ -0,0 +1,421 @@
...
+
+	return updateScheduleStatus(ctx, c, old, schedule)
+}
+
+// setScheduleReadyCondition sets the Ready condition on the status and updates the timestamp if the condition has changed.
+func setScheduleReadyCondition(status *fleet.ScheduleStatus, err error) {
+	if status == nil {
+		status = &fleet.ScheduleStatus{}
+	}
+	cond := condition.Cond(fleet.Ready)
```
weyfonk, line 305: Side note: we'll probably want to document what `Ready` as a condition means for a Schedule, as it is not directly related to whether workloads have been deployed.

internal/cmd/controller/reconciler/schedule_controller.go
```diff
@@ -0,0 +1,421 @@
...
+	cond.SetError(status, "", err)
+	if !equality.Semantic.DeepEqual(origStatus, status) {
+		cond.LastUpdated(status, time.Now().UTC().Format(time.RFC3339))
+	}
+}
+
+// mapClustersToSchedules is a mapping function used to trigger a reconciliation of Schedules
+// when a targeted Cluster changes. It finds all schedules that target the cluster
+// and enqueues a reconcile request for each of them.
+func (r *ScheduleReconciler) mapClustersToSchedules(ctx context.Context, a client.Object) []ctrl.Request {
```
weyfonk: This is a reconciler method, among static functions; could we move it higher up?
Same remark about `ensureFinalizer` and `handleDelete`.

internal/cmd/controller/reconciler/cronduration_job.go
```diff
@@ -0,0 +1,317 @@
...
+	clusters := &fleet.ClusterList{}
+	if err := c.List(ctx, clusters, client.InNamespace(namespace)); err != nil {
+		return nil, fmt.Errorf("%w, listing clusters: %w", fleetutil.ErrRetryable, err)
+	}
+	var clusterNames []string
+	for _, cluster := range clusters.Items {
+		cluster := cluster
+		cgs, err := target.ClusterGroupsForCluster(ctx, c, &cluster)
+		if err != nil {
+			return nil, fmt.Errorf("%w, getting clusters from cluster groups: %w", fleetutil.ErrRetryable, err)
```
weyfonk: Isn't it the other way around?
```suggestion
			return nil, fmt.Errorf("%w, getting cluster groups from clusters: %w", fleetutil.ErrRetryable, err)
```

internal/cmd/controller/reconciler/suite_test.go
```diff
@@ -0,0 +1,16 @@
+package reconciler
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestGitOpsReconciler(t *testing.T) {
```
weyfonk: Why `TestGitOpsReconciler`?
And more importantly: do we need this file?

internal/cmd/controller/reconciler/schedule_controller.go
```diff
@@ -0,0 +1,421 @@
...
+			r.Client,
+			newJob.MatchingClusters,
+			existingCronJob.MatchingClusters,
+			s.Namespace,
+		)
+	}
+	return nil
+}
+
+func jobNeedsUpdate(newJob, existingJob *CronDurationJob) bool {
```
weyfonk:
```suggestion
// jobNeedsUpdate returns true if there is a discrepancy between newJob and existingJob:
// * either in their descriptions, indicating that the parent schedule for those jobs has been updated
// * or in clusters matched by the jobs, which may result from updates to the clusters themselves,
// or creation/deletion of clusters since the existingJob was created.
func jobNeedsUpdate(newJob, existingJob *CronDurationJob) bool {
```

internal/cmd/controller/reconciler/schedule_controller.go
```diff
@@ -0,0 +1,421 @@
...
+				return err
+			}
+		}
+	}
+
+	// now check for clusters that are flagged as scheduled and should no longer be flagged because
+	// they are no longer targeted by any schedule
+	for _, cluster := range clustersOld {
+		if !slices.Contains(clustersNew, cluster) {
+			noLongerScheduled, err := isClusterScheduled(scheduler, cluster, namespace)
```
weyfonk: This would be more readable to me:
```suggestion
			wasScheduled, err := isClusterScheduled(scheduler, cluster, namespace)
```

internal/cmd/controller/reconciler/schedule_controller.go
```diff
@@ -0,0 +1,421 @@
...
+			if err != nil {
+				return err
+			}
+			if !noLongerScheduled {
```
weyfonk: Isn't this condition the other way around? As in, we'd want to un-schedule a cluster if it used to be scheduled but is no longer matched...
```suggestion
			if noLongerScheduled {
```

internal/cmd/controller/reconciler/schedule_controller.go
```diff
@@ -0,0 +1,421 @@
...
+	if errors.Is(err, quartz.ErrJobNotFound) {
+		return scheduleNewCronDurationJob(ctx, s, r.Scheduler, r.Client)
+	}
+
+	// the job already exists, check if an update is needed
+	newJob, err := NewCronDurationJob(ctx, s, r.Scheduler, r.Client)
+	if err != nil {
+		return err
+	}
+	existingJob := schedJob.JobDetail().Job()
```
weyfonk, line 118: As an edge case optimisation: if the set of matched clusters for the new job is empty, delete the job instead of updating it with no matching clusters?

pkg/apis/fleet.cattle.io/v1alpha1/schedule_types.go
```diff
@@ -0,0 +1,80 @@
...
+// ScheduleList contains a list of Schedule
+type ScheduleList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []Schedule `json:"items"`
+}
+
+type ScheduleSpec struct {
+	Schedule string          `json:"schedule,omitempty"`
+	Duration metav1.Duration `json:"duration,omitempty"`
```
weyfonk, line 41: I am a bit concerned about allowing a duration here, thinking it may cause issues: what happens if the deployment _needs_ longer than the specified duration?
Would the schedule and clusters still be marked as inactive/off-schedule, although deployment operations would still be in progress?
Or would Fleet abort any in-progress deployments (which doesn't seem to be the case)?
I think we may want to drop this field, at least for a first implementation.
Happy to discuss.

internal/cmd/controller/reconciler/cronduration_job.go
```diff
@@ -0,0 +1,317 @@
...
+	if err != nil {
+		return 0, err
+	}
+
+	return time.Duration(nextFireTime - now), nil
+}
+
+// checkScheduleAndDuration verifies that the given schedule start time and the duration are feasible.
+// If the duration is longer than 2 consecutive triggers it is considered as non valid.
+func checkScheduleAndDuration(schedule *fleet.Schedule, location *time.Location) error { //nolint:revive
```
weyfonk, line 126: "feasible" here means mathematically consistent, but doesn't inform us about actual feasibility, as we don't know in advance how long a deployment would take...

internal/cmd/controller/reconciler/schedule_controller_test.go
```diff
@@ -0,0 +1,267 @@
...
+			jobKey := scheduleKey(schedule)
+			_, err = scheduler.GetScheduledJob(jobKey)
+			Expect(err).NotTo(HaveOccurred())
+
+			// Check cluster status
+			updatedCluster := &fleet.Cluster{}
+			err = k8sclient.Get(ctx, client.ObjectKeyFromObject(cluster), updatedCluster)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(updatedCluster.Status.Scheduled).To(BeTrue())
+			Expect(updatedCluster.Status.ActiveSchedule).To(BeFalse())
```
weyfonk, line 122: Doesn't the value of `ActiveSchedule` here depend on when (within a given minute) this test case is started?

internal/cmd/controller/reconciler/cronduration_job.go
```diff
@@ -0,0 +1,317 @@
...
+	fleet "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
+	"github.com/reugn/go-quartz/quartz"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type CronDurationJob struct {
+	Started          bool
+	Schedule         *fleet.Schedule
+	Scheduler        quartz.Scheduler
```
weyfonk, line 24: nit: Each job instance will contain a full scheduler struct; perhaps we could do with a pointer?

internal/cmd/controller/reconciler/cronduration_job.go
```diff
@@ -0,0 +1,317 @@
...
+	hash             string
+	key              *quartz.JobKey
+}
+
+// NewCronDurationJob constructs a new CronDurationJob.
+// It also verifies the validity and correctness of the schedule and duration data.
+// Internally, it assigns Location = Local if no location was specified in the schedule,
+// and from that point onward, any time-related calculations are performed using this location.
+func NewCronDurationJob(ctx context.Context, schedule *fleet.Schedule, scheduler quartz.Scheduler, c client.Client) (*CronDurationJob, error) {
+	locationStr := schedule.Spec.Location
```
weyfonk, line 38: nit: this may not be much of an issue for now, as call locations are relatively few and known, but this line and all subsequent function calls assume that the `schedule` pointer is non-nil ;)
That's something we may want to either be more defensive about, or document as an explicit assumption, or a mix of both: for instance being defensive here, by returning an error if `schedule` is nil, and documenting the assumption in the functions using `schedule` as an argument.

internal/cmd/controller/reconciler/cronduration_job.go
```diff
@@ -0,0 +1,317 @@
...
+}
+
+func (c *CronDurationJob) executeStart(ctx context.Context) error {
+	c.Started = true
+
+	// Recalculate matching clusters at execution time. This ensures that any
+	// changes to cluster labels that occurred since the last reconciliation
+	// are included. The controller's watchers only trigger reconciles for
+	// clusters that are already part of a schedule.
+	clusters, err := matchingClusters(ctx, c.Matcher, c.client, c.Schedule.Namespace)
```
weyfonk, line 212: Do we deliberately not populate `c.MatchingClusters` with `clusters` to prevent concurrent mutation of existing jobs from 2 different places (schedule reconciler and jobs methods themselves)?
There may be a risk of the schedule reconciler updating jobs if matching clusters change, reading outdated pre-update matching clusters from this job, although information about new matching clusters may have been fetched here and used in the meantime.

internal/cmd/controller/reconciler/cronduration_job.go
```diff
@@ -0,0 +1,317 @@
...
+		if !slices.Contains(clusters, cluster) {
+			// this cluster is no longer targeted
+			if err := setClusterScheduled(ctx, c.client, cluster, c.Schedule.Namespace, false); err != nil {
+				return err
+			}
+		}
+	}
+
+	// Update the status of the Schedule resource
+	if err := setScheduleActive(ctx, c.client, c.Schedule, true); err != nil {
```
weyfonk, line 236: As we update clusters _and_ the schedule, doesn't this mean that the schedule will be reconciled _twice_ for each cluster?
Edit: Maybe not, as we're updating the schedule's _status_. That may still well mean one schedule reconcile per cluster though :thinking:

internal/cmd/controller/reconciler/cronduration_job.go
```diff
@@ -0,0 +1,317 @@
...
+	jsonBytes, err := json.Marshal(sched.Spec)
+	if err != nil {
+		return "", err
+	}
+
+	hash := sha256.Sum256(jsonBytes)
+	return hex.EncodeToString(hash[:]), nil
+}
+
+func (c *CronDurationJob) calculateNextTimeBasedOnCron() (time.Duration, error) {
```
weyfonk: I find the name of this method somewhat misleading, as if it computed the start time of the next job, whereas it actually returns the remaining _duration_ until that start time.
How about something like this?
```suggestion
func (c *CronDurationJob) durationToNextStart() (time.Duration, error) {
```

internal/cmd/controller/reconciler/cronduration_job.go
```diff
@@ -0,0 +1,317 @@
...
+	}
+
+	// Schedules the Stop call (now + duration)
+	return c.scheduleStopJob()
+}
+
+func (c *CronDurationJob) executeStop(ctx context.Context) error {
+	c.Started = false
+
+	// Sets ActiveSchedule to true for all matching clusters.
```
weyfonk:
```suggestion
	// Sets ActiveSchedule to false for all matching clusters.
```

internal/cmd/controller/reconciler/cronduration_job_test.go
```diff
@@ -0,0 +1,546 @@
...
+		scheduler.Stop()
+		_ = scheduler.Clear()
+	})
+
+	Context("NewCronDurationJob", func() {
+		It("should create a new job successfully", func() {
+			job, err := NewCronDurationJob(ctx, schedule, scheduler, k8sclient)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(job).NotTo(BeNil())
+			Expect(job.MatchingClusters).To(ContainElement("test-cluster"))
```
weyfonk:
```suggestion
			Expect(job.MatchingClusters).To(ContainElement("test-cluster"))
			Expect(job.MatchingClusters).To(HaveLen(1))
```

internal/cmd/controller/reconciler/cronduration_job_test.go
```diff
@@ -0,0 +1,546 @@
...
+
+			// Check schedule status
+			updatedSchedule := &fleet.Schedule{}
+			err = k8sclient.Get(ctx, types.NamespacedName{Name: "test-schedule", Namespace: "default"}, updatedSchedule)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(updatedSchedule.Status.Active).To(BeFalse())
+			// time should be the next minute and 0 seconds
+			Expect(updatedSchedule.Status.NextStartTime.Time.Second()).To(BeZero())
+			Expect(updatedSchedule.Status.NextStartTime.Time.Minute()).To(Equal(timeBeforeSchedule.Add(time.Minute).Minute()))
+			Expect(updatedSchedule.Status.NextStartTime.Time.Hour()).To(Equal(timeBeforeSchedule.Hour()))
```
weyfonk: This could be flaky, with a job first scheduled at Xh59m ;)
It could be safer to just set the seconds of `timeBeforeSchedule` to 0, and checking that there's a minute between the result and `updatedSchedule.Status.NextStartTime`.

internal/cmd/controller/reconciler/cronduration_job_test.go
```diff
@@ -0,0 +1,546 @@
...
+			Expect(updatedSchedule.Status.NextStartTime.Time.Second()).To(BeZero())
+			if timeBeforeUpdate.Minute()%2 == 0 {
+				Expect(updatedSchedule.Status.NextStartTime.Time.Minute()).To(Equal(timeBeforeUpdate.Add(2 * time.Minute).Minute()))
+			} else {
+				Expect(updatedSchedule.Status.NextStartTime.Time.Minute()).To(Equal(timeBeforeUpdate.Add(1 * time.Minute).Minute()))
+			}
+			if timeBeforeUpdate.Minute() < 59 {
+				Expect(updatedSchedule.Status.NextStartTime.Time.Hour()).To(Equal(timeBeforeUpdate.Hour()))
+			} else {
+				Expect(updatedSchedule.Status.NextStartTime.Time.Hour()).To(Equal(timeBeforeUpdate.Add(1 * time.Hour).Hour()))
```
weyfonk: Again, at the risk of being picky, what if this test case starts running at 23:59? :upside_down_face:

pkg/apis/fleet.cattle.io/v1alpha1/bundledeployment_types.go
0xavi0: That's exactly why I named it NotInSchedule, because of the default value.
I don't mind calling it `OffSchedule`, but would not call it `InSchedule` for the same reason. 

I changed the name to `OffSchedule`

internal/experimental/experimental.go
0xavi0: moved it to `internal/`

e2e/single-cluster/schedule_test.go
0xavi0: I'm not sure I get what you mean. 
That block was just to check that the `BundleDeployment` was created and that its `OffSchedule` (formerly known as `NotInSchedule`) field has the expected value. 

The next block checks that the Pod deployed by the `BundleDeployment` is installed.

I was in doubt of creating this `e2e` test because I don't like tests dependant on time constraints, but I'm afraid times and durations are the base of what the feature implements :/

go.mod
0xavi0: Yeah, interesting... I guess that was a conflict in git. In fact I had problems with that file. 
It was reporting nothing with `git status --porcelain` and there were changes... 

I reverted the versions.

internal/cmd/agent/controller/drift_controller.go
0xavi0: I implemented it the same way as Pause as the functionality is identical, I don't know the answer to that question neither. 

I would prefer to keep this implemented as Pause for now...  and update both in the future.

internal/cmd/controller/operator.go
0xavi0: I think I forgot to implement this for Schedules :P, Thanks for catching this.

Should be implemented via a new env variable now.

internal/cmd/controller/reconciler/schedule_controller.go
0xavi0: `fleet.Schedule` has nothing to do with how deployments are installed.
It just defines a window in which deployment updates are allowed.

internal/cmd/controller/reconciler/schedule_controller.go
0xavi0: Moved

internal/cmd/controller/reconciler/cronduration_job.go
0xavi0: Fixed, thanks.

internal/cmd/controller/reconciler/suite_test.go
0xavi0: Copy/paste... I think we do need it as the test is using Ginkgo and there are more than 1 test file? (not 100% sure, but if I delete the file I get: `Found no test suites, did you forget to run "ginkgo bootstrap"?`

I fixed the copy/paste leftover and deleted the `BeforeSuite` statement as is did nothing

internal/cmd/controller/reconciler/schedule_controller.go
0xavi0: Added

internal/cmd/controller/reconciler/schedule_controller.go
0xavi0: I've changed it to `targeted`, check the commend below

internal/cmd/controller/reconciler/schedule_controller.go
0xavi0: mmmmh, no... I think the name of the variable is indeed confusing.
What that block does is ensure that a cluster that is no longer in the quartz.Scheduler list (but it was before) has the `Status.Scheduled` property set to `false`

I've changed it to `targeted`. 

It translates to... if the cluster is no longer targeted... set `Status.Scheduled=false`

internal/cmd/controller/reconciler/schedule_controller.go
0xavi0: I thought also on deleting the job, but I thought that from a logical point of view... the `fleet.Schedule` is still on the k8s cluster and it's still "alive". 
The Status of the `Schedule` resource has the list of matched clusters that should show as empty... the UI maybe could warn the user that Schedule is not matching any cluster?

pkg/apis/fleet.cattle.io/v1alpha1/schedule_types.go
0xavi0: I'm not sure I understand what you mean. 
If you don't have a duration... how do you define the time in which the Schedule is active?
The schedule field just defines the time a Schedules starts allowing deployments to be updated, but how to you mark the end of it?

In regards to the other question... In my understanding this is working exactly like the Pause feature.
If there's a change to active to inactive in the middle the BundleDeployment reconciled in the Agent simply will stop processing. 
If the deployment is still on progress it will finish (if that doesn't need other reconcile calls)

I checked ArgoCD before implementing this, for example, and they also have a duration.
See https://argo-cd.readthedocs.io/en/stable/user-guide/sync_windows/

internal/cmd/controller/reconciler/cronduration_job.go
0xavi0: We can't control deployments or how much they take (there are way too many parameters involved in how much time a deployment takes), this just controls if the Agent applies new updates from BundleDeployments or not. So similar as concurrent updates to the same GitRepo... we don't know when the previous deployment will be done, the Agents just picks the last BundleDeployment. 
in this case the Agent will be picking updates while the `OffSchedule` property is not set or set to false... otherwise updates won't take effect.

Happy to discuss, but I don't think we can guarantee when a deployment will be fully installed because that doesn't depend on Fleet.

internal/cmd/controller/reconciler/schedule_controller_test.go
0xavi0: Yup, you're right, thanks!

As I said in another comment... I don't like time based tests, but.... 

I've added a function to wait for the second 0 before creating the `fleet.Schedule` so it all makes sense and we don't depend on when is the test executed.

internal/cmd/controller/reconciler/cronduration_job.go
0xavi0: `quartz.Scheduler` is an interface. 
https://github.com/reugn/go-quartz/blob/0c1b09e7c0af7f22a355103f5e3270753809e5f9/quartz/scheduler.go#L24

internal/cmd/controller/reconciler/cronduration_job.go
0xavi0: schedule, like the scheduler and client variables come from calls from the Reconciler, and this was not intended to be used as a standalone package.
If you add a new call that passes a `nil` schedule or scheduler or client we would get a very quick panic. 

Adding so many checks might be interesting when writing a public facing API, so we don't allow users to use it wrongly. 

Out of curiosity I've checked other places and I can see we don't do those protections neither.

For example, the `HelmOps` code also passes pointers and there are not protections:

```go 
// Calculates the bundle representation of the given HelmOp resource
func (r *HelmOpReconciler) calculateBundle(helmop *fleet.HelmOp) *fleet.Bundle {
	spec := helmop.Spec.BundleSpec
```

or 

```go
func (r *HelmOpReconciler) createUpdateBundle(ctx context.Context, helmop *fleet.HelmOp) (*fleet.Bundle, error) {
	b := &fleet.Bundle{}
	nsName := types.NamespacedName{
		Name:      helmop.Name,
		Namespace: helmop.Namespace,
	}
```

It's like that everywhere.... GitRepo GitJob reconciler:
```go 
func (r *GitJobReconciler) createJobRBAC(ctx context.Context, gitRepo *v1alpha1.GitRepo) error {
	saName := names.SafeConcatName("git", gitRepo.Name)
```

even in k8s code:

```go
// ContainsFinalizer checks an Object that the provided finalizer is present.
func ContainsFinalizer(o client.Object, finalizer string) bool {
	f := o.GetFinalizers()
```
In the function above o is an interface, that might be a pointer in the original call

	
To avoid confusions I changed the `NewCronDurationJob` so it is not exported (it was only used by the reconciler)

internal/cmd/controller/reconciler/cronduration_job.go
0xavi0: good catch! I forgot to add that ;) 
Added now, thanks!

internal/cmd/controller/reconciler/cronduration_job.go
0xavi0: The reconciler deliberately "listens" only to generation changes. Updates to the status are not affecting the reconciler.

internal/cmd/controller/reconciler/cronduration_job.go
0xavi0: Fine with me... changed to the suggested name

internal/cmd/controller/reconciler/cronduration_job_test.go
0xavi0: I changed it to 
`Expect(job.MatchingClusters).To(ConsistOf("test-cluster"))` that checks the length and content in 1 line.

internal/cmd/controller/reconciler/cronduration_job_test.go
0xavi0: yup,  I've added a function that waits if the actual second is more or greater than 55, so we have 5 seconds gap guaranteed.

internal/cmd/controller/reconciler/cronduration_job_test.go
0xavi0: same as above

internal/cmd/agent/controller/drift_controller.go
weyfonk: No worries, let's keep this in mind and possibly investigate it as a follow-up.

internal/cmd/controller/operator.go
weyfonk: :D thanks!
That environment variable needs to be set in `fleet-controller` deployments through chart values, as we do for other reconcilers.

internal/cmd/controller/reconciler/schedule_controller.go
weyfonk: Precisely :) And that's a different idea of ready than "workload/user resources ready", which is why I think we'll want to document the distinction, when we document scheduling in `rancher/fleet-docs`.

internal/cmd/controller/reconciler/suite_test.go
weyfonk: Oh, I see. Thanks for clarifying and updating the file :+1:

internal/cmd/controller/reconciler/schedule_controller.go
weyfonk: Thanks; this looks less confusing to me too now, and `targeted` makes sense since Fleet otherwise refers to clusters as targets :+1:

internal/cmd/controller/reconciler/cronduration_job.go
weyfonk: Fair point, thanks!

e2e/single-cluster/schedule_test.go
weyfonk: What I meant is that, if this test case were to fail for some reason (regression, etc), should it take 5 min (the [default timeout](https://github.com/rancher/fleet/blob/main/e2e/testenv/env.go#L15) for `Eventually` blocks) to fail, since test execution will now start at the beginning of a minute, and the schedule is active for 30s from then on?

But I now realise that I was conflating schedule active state and the deployment time span which, as we both agree, is beyond Fleet's control.

TL;DR: feel free to ignore this :)

internal/cmd/controller/reconciler/schedule_controller.go
weyfonk: I see your point. There is no real point in running the job, which will be a no-op, but the schedule still exists.

> the UI maybe could warn the user that Schedule is not matching any cluster?

Yes, and that could eventually be part of checks run by our upcoming troubleshooting CLI :)

internal/cmd/controller/reconciler/cronduration_job.go
weyfonk: Oh right, so `Scheduler` is actually already a pointer. Moving on :)

internal/cmd/controller/reconciler/cronduration_job_test.go
weyfonk: Nice, I saw that syntax later on while reviewing this file, glad it's consistent now :100: Thanks!

pkg/apis/fleet.cattle.io/v1alpha1/schedule_types.go
weyfonk: Oh great, I was afraid that we would falsely pretend to expose durations as "maintenance windows", but as long as we agree on the idea that durations are actually just windows during which deployments are allowed to _start_, without any bearing on when they may _finish_, then I'm fine with the current approach.
That will need to be part of eventual documentation on this.

Thanks for clarifying :)

internal/cmd/controller/reconciler/cronduration_job.go
weyfonk: Glad we're on the same page :)
As discussed on the thread about the schedule API itself, durations are about how long deployments are allowed to _start_ for, without any impact on, or enforcement of, _completion_ times.

internal/cmd/controller/reconciler/schedule_controller_test.go
weyfonk: Understood, thanks. That function lives in end-to-end tests though; I can't see any similar logic in unit tests :thinking:

internal/cmd/controller/reconciler/cronduration_job_test.go
weyfonk: Given the supported [syntax](https://github.com/reugn/go-quartz/blob/v0.11.1/quartz/cron.go#L128), which includes seconds, wouldn't it be safer to build our test schedules using a format string, starting at the current time (or even a few seconds in the future)? We could then compute the first and next start times based on that, and avoid having to wait.

internal/cmd/controller/reconciler/cronduration_job.go
```diff
@@ -0,0 +1,323 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Sets ActiveSchedule to true for all matching clusters.
+	for _, cluster := range clusters {
+		if err := setClusterActiveSchedule(context.Background(), c.client, cluster, c.Schedule.Namespace, true); err != nil {
+			return err
+		}
+	}
```
weyfonk: Isn't this redundant with `setClustersScheduled` called below, which calls `setClusterScheduled` on each of those clusters, setting both `cluster.Status.Scheduled` and `cluster.Status.ActiveSchedule`? :thinking:

internal/cmd/controller/reconciler/cronduration_job.go
weyfonk: Let's discuss, as I think I understand the change you made, but not how it relates to my initial comment :sweat_smile:

internal/cmd/controller/operator.go
0xavi0: Sure! Added.

internal/cmd/controller/reconciler/schedule_controller_test.go
0xavi0: There are only 2 unit tests that check for values of the scheduled time.
Anyway I did as you suggested and changed the scheduled time to some exact time in the future that ensures we will be checking against an expected time values.

internal/cmd/controller/reconciler/cronduration_job_test.go
0xavi0: Good idea, I've implemented that as you suggested and it no longer waits.

internal/cmd/controller/reconciler/cronduration_job_test.go
0xavi0: changed to a fixed time as suggested

internal/cmd/controller/reconciler/cronduration_job.go
0xavi0: Indeed, and moreover, the call in the highlighted lines is not using the right context. :P 

I've deleted the second block that only updated `Status.Scheduled` and fixed the context in the remaining call.

weyfonk (approved):
LGTM :)

PR #4185, opened 2025-10-01
Use CABundle when fetching chart with `helm.chart` field of `fleet.yaml` file

p-se:
Refers to #3646

### Checklist

### DIFF
--- a/e2e/single-cluster/go_getter_custom_ca_test.go
+++ b/e2e/single-cluster/go_getter_custom_ca_test.go
@@ -21,6 +21,14 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
+// Those tests are specifically targeting one feature of go-getter, namely cloning of git
+// repositories using HTTPS. That is, because TLS certificates are only used in HTTPS URLs, not if
+// SSH keys are used to clone those repositories. Since go-getter shells out to the `git` CLI for
+// cloning git repositories, the configuration of TLS certificates or ignoring those needs to be
+// configured with `git` typical environment variables. Those are the tests for that implementation.
+// For go-getter to be used, the `helm.chart` field in `fleet.yaml` needs to point to a URL that
+// tells go-getter to use the git protocol over HTTPS. Such an URL is prefixed with `git::https://`.
+// The contents fetched from those repositories are expected to be helm charts.
 var _ = Describe("Testing go-getter CA bundles and insecureSkipVerify for cloning git repositories", Label("infra-setup"), func() {
 	const (
 		sleeper    = "sleeper"
@@ -150,7 +158,7 @@ var _ = Describe("Testing go-getter CA bundles and insecureSkipVerify for clonin
 		Expect(err).ToNot(HaveOccurred())
 		cloneDir = path.Join(tmpDir, "repo") // Fixed and built into the container image.
 		gitrepoName = testenv.RandomFilename("gitjob-test", r)
-		// Creates the content in the sleeperClusterName directory
+		// Creates the content in the sleeper directory
 		_, err = gh.Create(cloneDir, testenv.AssetPath("gitrepo/sleeper-chart"), sleeper)
 		Expect(err).ToNot(HaveOccurred())
 
@@ -256,21 +264,6 @@ var _ = Describe("Testing go-getter CA bundles and insecureSkipVerify for clonin
 
 			createGitRepo(gitRepoOptions{HelmSecretNameForPaths: secretName})
 			Consistently(expectGitRepoToNotBeReady).Should(Succeed())
-		})
-
-		It("should succeed when using a correct CA bundle provided in helmSecretNameForPath", func() {
-			secretName := "helm-ca-bundle-by-path-" + gitrepoName
-			createHelmSecretForPaths(
-				secretName,
-				AuthConfigByPath{
-					entrypoint: {
-						CABundle: mustReadFileAsBase64(getCertificateFilePath()),
-					},
-				},
-			)
-
-			createGitRepo(gitRepoOptions{HelmSecretNameForPaths: secretName})
-			Eventually(expectGitRepoToBeReady).Should(Succeed())
 		})
 	})
 
--- a/e2e/testenv/githelper/git.go
+++ b/e2e/testenv/githelper/git.go
@@ -190,7 +190,7 @@ func (g *Git) Create(repodir string, from string, subdir string) (*git.Repositor
 		return nil, err
 	}
 
-	err = copy.Copy(from, path.Join(repodir, subdir)) //nolint:gosec // test code should never receive user input
+	err = copy.Copy(from, path.Join(repodir, subdir))
 	if err != nil {
 		return nil, err
 	}


e2e/single-cluster/gitrepo_test.go
```diff
@@ -31,6 +31,17 @@ const (
 	HTTPSPort = 4343
 )
 
+type gitRepoTestValues struct {
```
p-se, line 34: By using this struct instead of inline structs, the additional fields added in `gitrepo.yaml` don't need to be provided. The empty value preserves the previous behavior and it is a bit clearer which values can be provided.

e2e/single-cluster/schedule_test.go
```diff
@@ -249,29 +241,6 @@ var _ = Describe("Schedules", Label("infra-setup"), func() {
...
-	gr, err := k.Get("gitrepo", name, "-o=json")
-
-	g.Expect(err).ToNot(HaveOccurred())
-
-	var gitrepo fleet.GitRepo
-	_ = json.Unmarshal([]byte(gr), &gitrepo)
-
-	return gitrepo.Status
-}
-
```
p-se, line 274: Those are already provided in the `singlecluster_test` package.

e2e/testenv/githelper/git.go
```diff
@@ -190,46 +193,40 @@ func (g *Git) Create(repodir string, from string, subdir string) (*git.Repositor
 		return nil, err
 	}
 
-	w, err := repo.Worktree()
-	if err != nil {
-		return nil, err
-	}
+	_, err = g.Update(repo, UpdateForce)
```
p-se, line 199: I'm reusing the functionality of `Update` in `Create` and the newly added `Add` method. Create can be used a second time to force push an update to an already existing repo.

internal/bundlereader/loaddirectory.go
```diff
@@ -254,6 +254,21 @@ func GetContent(ctx context.Context, base, source, version string, auth Auth, di
...
+		defer os.Remove(tmpFile.Name())
+		if _, err := tmpFile.Write(auth.CABundle); err != nil {
+			return nil, err
+		}
+		if err := os.Setenv("GIT_SSL_CAINFO", tmpFile.Name()); err != nil {
+			return nil, err
+		}
+		defer os.Unsetenv("GIT_SSL_CAINFO")
+	}
+
```
p-se, line 272: I'm setting the data here because the logic to determine which CABundle and value for `SkipInsecureTLSVerify` is supposed to be used is determined in the gitjob controller. The alternative is to map all secrets into the pod of the Job and have it handled here more or less again. The Gitjob controllers needs to do it anyway because we apparently need those values in different places. Those are passed as arguments, environment variables and volumes to the pod. So, I thought it would be simpler to keep the logic there.

e2e/testenv/githelper/git.go
```diff
@@ -190,46 +193,40 @@ func (g *Git) Create(repodir string, from string, subdir string) (*git.Repositor
...
-	if _, err := w.Add("."); err != nil {
-		return nil, err
-	}
+	return repo, err
+}
 
-	_, err = w.Commit("add chart", &git.CommitOptions{
-		Author: author(),
-	})
+// Add adds files to a repo, commits them and pushed the changes.
```
weyfonk: nit:
```suggestion
// Add adds files to a repo, commits them and pushes the changes.
```

e2e/testenv/githelper/git.go
```diff
@@ -190,46 +193,40 @@ func (g *Git) Create(repodir string, from string, subdir string) (*git.Repositor
...
-	po := git.PushOptions{
-		Progress: os.Stdout,
-		// Force push, so our initial state is deterministic
-		RefSpecs: []config.RefSpec{config.RefSpec("+refs/heads/master:refs/heads/" + g.Branch)},
-		// This prevents IP SANs from being needed on the git server cert; the TLS verification we are most
-		// interested in is the one happening in Fleet's controllers and jobs.
-		InsecureSkipTLS: true,
-	}
-	k, err := g.Auth.getKeys()
+	cmd := exec.Command("cp", "-a", from, path.Join(repodir, subdir)) //nolint:gosec // test code should never receive user input
```
weyfonk: Do we need to shell out here?
Would something like [os.CopyFS](https://pkg.go.dev/os@go1.25.3#CopyFS) work instead?

e2e/single-cluster/go_getter_custom_ca_test.go
```diff
@@ -0,0 +1,196 @@
...
+		targetNamespace string
+		// Build git repo URL reachable _within_ the cluster, for the GitRepo
+		host = githelper.BuildGitHostname()
+	)
+
+	getExternalRepoURL := func(repoName string) string {
+		GinkgoHelper()
+		addr, err := githelper.GetExternalRepoAddr(env, HTTPSPort, repoName)
+		Expect(err).ToNot(HaveOccurred())
+		addr = strings.Replace(addr, "http://", fmt.Sprintf("%s://", "https"), 1)
```
weyfonk: nit: this looks more complex than needed; how about something like the following?
```suggestion
		addr = strings.Replace(addr, "http://", "https://", 1)
```

e2e/single-cluster/go_getter_custom_ca_test.go
```diff
@@ -0,0 +1,196 @@
...
+		tmpDir, err := os.MkdirTemp("", "fleet-")
+		Expect(err).ToNot(HaveOccurred())
+		cloneDir = path.Join(tmpDir, "repo") // Fixed and built into the container image.
+		gitrepoName = testenv.RandomFilename("gitjob-test", r)
+		// Creates the content in the sleeperClusterName directory
+		_, err = gh.Create(cloneDir, testenv.AssetPath("gitrepo/sleeper-chart"), sleeper)
+		Expect(err).ToNot(HaveOccurred())
+
+		// Create the second repository
+		Expect(err).ToNot(HaveOccurred())
```
weyfonk:
```suggestion
```

e2e/testenv/githelper/git.go
weyfonk: Also, we may want to clarify (at least for our future selves) that `subdir` is the _destination_ subdirectory, unrelated to the _source_ from which data is added.

e2e/single-cluster/go_getter_custom_ca_test.go
```diff
@@ -0,0 +1,196 @@
...
+		gh = githelper.NewHTTP(addr)
+		tmpDir, err := os.MkdirTemp("", "fleet-")
+		Expect(err).ToNot(HaveOccurred())
+		cloneDir = path.Join(tmpDir, "repo") // Fixed and built into the container image.
+		gitrepoName = testenv.RandomFilename("gitjob-test", r)
+		// Creates the content in the sleeperClusterName directory
+		_, err = gh.Create(cloneDir, testenv.AssetPath("gitrepo/sleeper-chart"), sleeper)
+		Expect(err).ToNot(HaveOccurred())
+
+		// Create the second repository
```
weyfonk: this could be a bit clearer:
```suggestion
		// Create the second repository, referencing the Helm chart from the first repository
		// through a fleet.yaml
```

e2e/single-cluster/go_getter_custom_ca_test.go
```diff
@@ -0,0 +1,196 @@
...
+		// Check that the bundle deployment resource has been deleted
+		Eventually(func(g Gomega) {
+			out, _ := k.Get(
+				"bundledeployments",
+				"-A",
+				"-l",
+				fmt.Sprintf("fleet.cattle.io/repo-name=%s", gitrepoName),
+			)
+			g.Expect(out).To(ContainSubstring("No resources found"))
+		}).Should(Succeed())
```
weyfonk: Do we need this? :thinking:

e2e/single-cluster/go_getter_custom_ca_test.go
```diff
@@ -0,0 +1,196 @@
...
+			// Create and apply GitRepo
+			err = testenv.ApplyTemplate(k, testenv.AssetPath("gitrepo/gitrepo.yaml"), gitRepoTestValues{
+				Name:                  gitrepoName,
+				Repo:                  gh.GetInClusterURL(host, HTTPSPort, "repo"),
+				Branch:                gh.Branch,
+				PollingInterval:       "15s",           // default
+				TargetNamespace:       targetNamespace, // to avoid conflicts with other tests
+				Path:                  entrypoint,
+				InsecureSkipTLSVerify: false,
+				CABundle:              encodedCert,
```
weyfonk: If we want to be strict and/or on the safe side about testing this fix, shouldn't we check that this happy path also works with a `helmSecretName` field referencing a secret containing this CA bundle?
That would of course not invalidate any of these test cases, which are still very relevant.

internal/bundlereader/loaddirectory.go
weyfonk: I wonder if this logic wouldn't find a better home in `apply.go`'s `addAuthToOpts`, where `fleet apply` deals with populating that `Auth.CABundle` field.
Instead of creating an additional file, perhaps this logic could then directly read the `CACertsFile` :thinking: 
Happy to discuss.

p-se: Long story short, the CABundle is replaced per path if `helmSecretNameForPaths` is configured. That happens after `apply.go`'s `addAuthToOpts`.

internal/bundlereader/loaddirectory.go
```diff
@@ -254,6 +255,26 @@ func GetContent(ctx context.Context, base, source, version string, auth Auth, di
...
+			return nil, err
+		}
+		if _, err := file.Write(auth.CABundle); err != nil {
+			return nil, err
+		}
+		file.Close()
+		defer os.Remove(file.Name())
+
+		os.Setenv("GIT_SSL_CAINFO", file.Name())
+		defer os.Unsetenv("GIT_SSL_CAINFO")
```
weyfonk, line 270: Keeping in mind that `fleet apply` may run outside of a git job container, we may want to save backup values of these env vars before processing, and restore those values (instead of simply `unset`ting the env vars) in `defer` statements.

e2e/single-cluster/go_getter_custom_ca_test.go
```diff
@@ -0,0 +1,316 @@
...
+
+	JustBeforeEach(func() {
+		// Create the first repository
+		addr := getExternalRepoURL("repo")
+		gh = githelper.NewHTTP(addr)
+		tmpDir, err := os.MkdirTemp("", "fleet-")
+		Expect(err).ToNot(HaveOccurred())
+		cloneDir = path.Join(tmpDir, "repo") // Fixed and built into the container image.
+		gitrepoName = testenv.RandomFilename("gitjob-test", r)
+		// Creates the content in the sleeperClusterName directory
```
weyfonk: nit:
```suggestion
		// Creates the content in the sleeper directory
```

e2e/single-cluster/go_getter_custom_ca_test.go
```diff
@@ -0,0 +1,316 @@
...
+	. "github.com/onsi/gomega"
+
+	"github.com/rancher/fleet/e2e/testenv"
+	"github.com/rancher/fleet/e2e/testenv/githelper"
+	"github.com/rancher/fleet/e2e/testenv/kubectl"
+
+	"sigs.k8s.io/yaml"
+)
+
+var _ = Describe("Testing go-getter CA bundles and insecureSkipVerify for cloning git repositories", Label("infra-setup"), func() {
```
weyfonk, line 32: It could make sense to mention somewhere, either through a test description or a comment, that these tests are more specific to pulling Helm charts from git repositories.

e2e/single-cluster/go_getter_custom_ca_test.go
```diff
@@ -0,0 +1,316 @@
...
+				AuthConfigByPath{
+					entrypoint: {
+						CABundle: mustReadFileAsBase64(getCertificateFilePath()),
+					},
+				},
+			)
+
+			createGitRepo(gitRepoOptions{HelmSecretNameForPaths: secretName})
+			Eventually(expectGitRepoToBeReady).Should(Succeed())
+		})
```
weyfonk: Isn't this a duplicate of L231-244?

p-se: Apparently so :sweat_smile:

internal/bundlereader/loaddirectory.go
p-se: It seems that changing an environment variable inside a go process does not have any effect on the shell that calls it.

```zsh
user@work ➜  ~ MY_ENV_VAR=foo
user@work ➜  ~ go run main.go
user@work ➜  ~ echo $MY_ENV_VAR
foo
user@work ➜  ~ cat main.go
package main

import "os"

func main() {
	os.Setenv("MY_ENV_VAR", "some_value")
}
user@work ➜  ~
```

I also tested with `os.Unsetenv`.

p-se:
Incidental that all tests pass in the first run when I cherry picked the fix for the race? :thinking: 

Edit: Twice.

internal/bundlereader/loaddirectory.go
weyfonk: Interesting :bulb: Thanks for checking, I stand corrected :)

kkaempf:
As this PR targets `main`, I've adapted the milestone to `v2.14.0`.
Please backport to `v2.13.1` 😉

internal/bundlereader/loaddirectory.go
```diff
@@ -254,6 +255,26 @@ func GetContent(ctx context.Context, base, source, version string, auth Auth, di
...
+		if err != nil {
+			return nil, err
+		}
+		if _, err := file.Write(auth.CABundle); err != nil {
+			return nil, err
+		}
+		file.Close()
+		defer os.Remove(file.Name())
+
+		os.Setenv("GIT_SSL_CAINFO", file.Name())
```
aruiz14, line 269: I think this may be problematic with concurrent bundle processing. However, the git getter does not seem to expose the environment variables to be set when running the git command ([like it does for other settings](https://github.com/hashicorp/go-getter/blob/v2.2.3/get_git.go#L321)).
cc/ @weyfonk since you reviewed this PR

p-se: You are completely right that it doesn't do that. Which means we really don't have another option to set those settings (unless go-getter is extended). So, it's more likely that we will need to ensure that this part isn't run concurrently.

PR #4186, opened 2025-10-02
Add README for benchmarks suite

manno:
Add a README for the benchmarks suite which covers usage and development information.


benchmarks/README.md
```diff
@@ -0,0 +1,317 @@
...
+fleet-benchmark run -n fleet-default -t 1h
+
+# Development mode
+go run ./benchmarks/cmd run -n fleet-default
+```
+
+## Key Features
+
+- **Realistic Testing**: Uses actual Fleet installations, not mocks
+- **Multi-cluster Support**: Tests can run against configurations with thousands of clusters
```
mmartin24, line 94: Suggestion, perhaps "hundreds and potentially thousands of clusters" may be safer, but just a suggestion

benchmarks/README.md
```diff
@@ -0,0 +1,317 @@
...
+# Run benchmarks
+go run ./benchmarks/cmd run -n fleet-default
+
+# View report from JSON output
+go run ./benchmarks/cmd report -i b-1234567890.json
+
+# Generate CSV export
+go run ./benchmarks/cmd csv -i b-1234567890.json
+
+# Print report as JSON
```
mmartin24, line 125: Sorry, just to understand, is "print" or "generate" as expressed for CSV?

Issue #3424, opened 2025-03-05
[SURE-7092] Add username to log messages, created resources


manno:
The Rancher UI provides the users name as a label on resources.
Extract the name, clean it up and use it in log messages. Also propagate the name as a label to other resources, like bundles.

Needs: https://github.com/rancher/dashboard/issues/13189

thardeck:
## Additional QA
 
### Problem
In a multi user environment an admin wants to know which user created which gitrepo or bundle.
 
### Solution
Add the user-id to gitrepo and bundle log messages.

### Testing

### Engineering Testing
#### Manual Testing
N/A

#### Automated Testing
Added integration tests to verify that the user id is added to the log message, if the label `fleet.cattle.io/created-by-user-id` is available.

### QA Testing Considerations
Rancher 2.12 and later add the label `fleet.cattle.io/created-by-user-id` to fleet bundles.
Verify that if the label is there (Rancher UI hides it), it can only be seen in the yaml config, that log messages of gitrepos or bundles provide the user id.
 
#### Regressions Considerations
None

sbulage:
| Rancher Version | Fleet Version |
|--------|--------|
|  v2.13-7faa74a968c24b0cdbd605ad90cbba757855f724-head| 108.0.0+up0.14.0-alpha.5 | 


---
- Admin user:

  <details><summary>Fleet Controller Pod logs showing admin user's User ID</summary>
  <p>
  
  ```logs
[...]

PR #4202, opened 2025-10-08
Add userID to log messages

thardeck:
Refers to #3424

### DIFF
--- a/integrationtests/controller/bundle/userid_logging_test.go
+++ b/integrationtests/controller/bundle/userid_logging_test.go
@@ -99,7 +99,6 @@ var _ = Describe("Bundle UserID logging", func() {
 				ContainSubstring(`"userID": "`+userID+`"`),
 			))
 
-			// Verify the logs contain both bundle context and userID
 			Expect(logs).To(ContainSubstring("bundle"))
 			Expect(logs).To(ContainSubstring(bundle.Name))
 		})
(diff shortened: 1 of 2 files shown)


integrationtests/gitjob/controller/userid_logging_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+		It("includes userID in log output", func() {
+			Eventually(func() string {
+				return logsBuffer.String()
+			}, timeout).Should(Or(
+				ContainSubstring(`"userID":"`+userID+`"`),
+				ContainSubstring(`"userID": "`+userID+`"`),
+			))
+
+			// Verify the logs contain both gitrepo context and userID
```
0xavi0: nit: is this comment correct at this line?

integrationtests/controller/bundle/userid_logging_test.go
```diff
@@ -0,0 +1,125 @@
...
+		It("includes userID in log output", func() {
+			waitForReconciliation()
+
+			logs := logsBuffer.String()
+			Expect(logs).To(Or(
+				ContainSubstring(`"userID":"`+userID+`"`),
+				ContainSubstring(`"userID": "`+userID+`"`),
+			))
+
+			// Verify the logs contain both bundle context and userID
```
0xavi0: nit: same as for the gitjob/controller tests...

integrationtests/gitjob/controller/userid_logging_test.go
thardeck: It is not well phrased, the existence of the userID was verified before and I wanted to make sure that the log line also fits the context.

I can remove the comments if you like.

0xavi0: they are a bit confusing, but it's just a minor thing and in a comment. 

Fine with me if you want to merge.

PR #4207, opened 2025-10-09
Exclude Fleet config files regardless of their naming

weyfonk:
While Fleet config files are typically named `fleet.yaml`, the recent introduction of user-driven bundle scanning enables a config file to be named arbitrarily, in which case it must still be excluded from the corresponding bundle. Fleet now takes care of this without any action needed from the user.

Integration tests also demonstrate that `.fleetignore` files can be leveraged to exclude config files living in the same directory as the config file referenced explicitly through user-driven bundle scanning. Fleet would otherwise not exclude those files from the bundle.

Refers to https://github.com/rancher/fleet-docs/issues/307

### Checklist

[fleet-docs](https://github.com/rancher/fleet-docs) repository: see https://github.com/rancher/fleet-docs/pull/354

### DIFF
--- a/integrationtests/cli/apply/apply_test.go
+++ b/integrationtests/cli/apply/apply_test.go
@@ -355,13 +355,26 @@ var _ = Describe("Fleet apply driven", Ordered, func() {
 				cli.AssetsPath + "driven/kustomize/overlays/dev/secret.yaml",
 				cli.AssetsPath + "driven/kustomize/overlays/prod/kustomization.yaml",
 				cli.AssetsPath + "driven/kustomize/overlays/prod/secret.yaml",
-				cli.AssetsPath + "driven/kustomize/dev.yaml",
-				cli.AssetsPath + "driven/kustomize/prod.yaml",
 			}
-			Expect(kDevBundle.Spec.Resources).To(HaveLen(8))
-			Expect(kProdBundle.Spec.Resources).To(HaveLen(8))
-			for _, r := range kResources {
+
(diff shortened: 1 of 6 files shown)


p-se:
I was wondering `.fleetignore` is being tested separately elsewhere? Because it looks like both `prod.yaml` and `dev.yaml` are supposed to be excluded from the bundle without even having a `.fleetignore`. Which would mean that `.fleetignore` wouldn't be tested at all in that particular test. But I have found separate tests for `.fleetignore` elsewhere in our code.

weyfonk:
> I was wondering `.fleetignore` is being tested separately elsewhere?

Yes, we test its behaviour in `internal/bundlereader/loaddirectory_test.go`.

> Because it looks like both `prod.yaml` and `dev.yaml` are supposed to be excluded from the bundle without even having a `.fleetignore`.

The test case using the `driven/kustomize` directory uses a `.fleetignore` excluding only `dev.yaml`,  to check that it's then excluded from both dev and prod bundles. See [this comment](https://github.com/rancher/fleet/pull/4207/files#diff-0192cb1c9dbf69364e24ddfd9632ad72f2a84a6eddbf0a425bd95be9525fd1a5R360-R365).

Conversely, the test case using `driven2/kustomize` uses a `.fleetignore` excluding only `prod.yaml`, for similar reasons (see [this comment](https://github.com/rancher/fleet/pull/4207/files#diff-0192cb1c9dbf69364e24ddfd9632ad72f2a84a6eddbf0a425bd95be9525fd1a5R421-R426)).

PR #4209, opened 2025-10-09
Fixes schedule update reset active

0xavi0:
There was a bug in the recently merged **Scheduling** PR.

The bug consists of applying a label to the Schedule that initially matches **all clusters**, and then, while the Schedule is active, changing the target so that it matches **fewer clusters**.

The clusters that remained as targets would stay in the **ActiveSchedule** state, which is incorrect, since modifying the Schedule should reset it and set its status to **not active**.

Follow-up to #4184.
Related to #3726.

### Checklist

### DIFF
--- a/internal/cmd/controller/reconciler/schedule_controller.go
+++ b/internal/cmd/controller/reconciler/schedule_controller.go
@@ -273,7 +273,7 @@ func setClusterScheduled(ctx context.Context, c client.Client, name, namespace s
 	old := cluster.DeepCopy()
 	cluster.Status.Scheduled = scheduled
 
-	// when this function is called is either because we're updating a
+	// This function is called either because we're updating a
 	// Schedule or because we're creating it.
 	// In both cases ActiveSchedule should be false as a Schedule
 	// always begins in OffSchedule mode until the first start call is executed.
@@ -414,7 +414,6 @@ func setClustersScheduled(ctx context.Context, c client.Client, clusters []strin
 }
 
 func updateScheduledClusters(ctx context.Context, scheduler quartz.Scheduler, c client.Client, clustersNew []string, clustersOld []string, namespace string) error {
-	// first look for clusters that are not scheduled yet and flag them as scheduled
 	for _, cluster := range clustersNew {
 		if err := setClusterScheduled(ctx, c, cluster, namespace, true); err != nil {
 			return err
--- a/internal/cmd/controller/reconciler/schedule_controller_test.go
+++ b/internal/cmd/controller/reconciler/schedule_controller_test.go
@@ -19,6 +19,12 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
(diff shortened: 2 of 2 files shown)


internal/cmd/controller/reconciler/schedule_controller_test.go
```diff
@@ -339,7 +359,7 @@ func checkClusterIsScheduled(
 	scheduler quartz.Scheduler,
 	k8sclient client.Client,
 	cluster, namespace string,
-	scheduledExpected, flaggedAsScheduledExpected bool) {
+	scheduledExpected, flaggedAsScheduledExpected, activeScheduleExpected bool) {
```
weyfonk: nit: I think creating a struct containing these 3 booleans would make the calls above a bit easier to read.
Example:
```
type expected struct{
    job bool
    statusScheduled bool
    statusActiveSchedule bool
}
```
... and, since these fields would be visible on the calling line, this function could be renamed to e.g. `checkState`
```
// before
checkClusterIsScheduled(scheduler, k8sclient, "test-cluster4", "default", true, true, false)
// after
checkState(scheduler, k8sclient, "test-cluster4", "default", expected{job:true, statusScheduled: true, statusActiveSchedule: true})
```

internal/cmd/controller/reconciler/schedule_controller.go
```diff
@@ -265,16 +265,19 @@ func setClusterScheduled(ctx context.Context, c client.Client, name, namespace s
...
 		return nil
 	}
 
 	old := cluster.DeepCopy()
 	cluster.Status.Scheduled = scheduled
-	if !scheduled {
-		cluster.Status.ActiveSchedule = false
-	}
+
+	// when this function is called is either because we're updating a
```
weyfonk:
```suggestion
	// This function is called either because we're updating a
```

internal/cmd/controller/reconciler/schedule_controller.go
```diff
@@ -413,10 +416,8 @@ func setClustersScheduled(ctx context.Context, c client.Client, clusters []strin
 func updateScheduledClusters(ctx context.Context, scheduler quartz.Scheduler, c client.Client, clustersNew []string, clustersOld []string, namespace string) error {
 	// first look for clusters that are not scheduled yet and flag them as scheduled
```
p-se: Is this comment still up-to-date?

0xavi0: I've deleted the comment because now the code is obvious.

internal/cmd/controller/reconciler/schedule_controller.go
0xavi0: changed as suggested

internal/cmd/controller/reconciler/schedule_controller_test.go
0xavi0: changed as suggested

PR #4221, opened 2025-10-10
Avoids returning an error when deleting OCI artifacts

0xavi0:
Returning an error when deleting an OCI artifact could lead to infinite requeing if there is a permanent
error accessing the registry.

The registry is an external resource out of Fleet and we don't have full control.
We already log an event when there is an error deleting the OCI artifact.

This was leading to errors in e2e tests, because one of the tests was checking if there was an error
when trying to use credentials with no delete permissions for the OCI registry.

If the test was the last one, it was logging the expected event and CI was successful, otherwise the next
test was failing randomly depending on which test was (if the test was reseting the OCI storage env variable then
the controller was restarted and unblocked the situation).

Also orders the tests and now the only test that needs the OCI_STORAGE=false env variable is executed the last one.

The prevents too many pod restarts.

Signed-off-by: Xavi Garcia <xavi.garcia@suse.com>

### DIFF
--- a/internal/cmd/controller/reconciler/bundle_controller.go
+++ b/internal/cmd/controller/reconciler/bundle_controller.go
@@ -477,6 +477,10 @@ func (r *BundleReconciler) handleDelete(ctx context.Context, logger logr.Logger,
 		return ctrl.Result{RequeueAfter: requeueAfterBundleDeploymentCleanup}, batchDeleteBundleDeployments(ctx, r.Client, bds)
 	}
 
+	if err := r.maybeDeleteOCIArtifact(ctx, bundle); err != nil {
+		return ctrl.Result{}, err
+	}
+
 	metrics.BundleCollector.Delete(req.Name, req.Namespace)
 	controllerutil.RemoveFinalizer(bundle, finalize.BundleFinalizer)
 	if err := r.Update(ctx, bundle); err != nil {
@@ -488,7 +492,7 @@ func (r *BundleReconciler) handleDelete(ctx context.Context, logger logr.Logger,
 		logger.V(1).Info("Cannot delete bundle's values secret, owner garbage collection will remove it")
 	}
 
-	return ctrl.Result{}, r.maybeDeleteOCIArtifact(ctx, bundle)
+	return ctrl.Result{}, err
 }
 
 // ensureFinalizer adds a finalizer to a recently created bundle.


internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -488,7 +492,7 @@ func (r *BundleReconciler) handleDelete(ctx context.Context, logger logr.Logger,
 		logger.V(1).Info("Cannot delete bundle's values secret, owner garbage collection will remove it")
 	}
 
-	return ctrl.Result{}, r.maybeDeleteOCIArtifact(ctx, bundle)
+	return ctrl.Result{}, err
```
weyfonk, line 495: If the OCI artifact to be deleted were not found when running `Delete`, we would still be returning a non-nil error here, which would trigger a requeue. How about this instead:
```suggestion
	return ctrl.Result{}, nil
```

0xavi0: I'm not sure I understand what you mean. :) 

The last part of `maybeDeleteOCIArtifact` is:
```go
    err = ocistorage.NewOCIWrapper().DeleteManifest(ctx, opts, bundle.Spec.ContentsID)
	if err != nil {
		r.Recorder.Event(bundle, fleetevent.Warning, "FailedToDeleteOCIArtifact", fmt.Sprintf("deleting OCI artifact %q: %v", bundle.Spec.ContentsID, err.Error()))
	}

	// In case there's an error deleting from the OCI registry,
	// we return nil because otherwise the controller would retry the operation,
	// and since the registry is not a component of Fleet,
	// we don't have full control over it.
	return nil
```

So... when there's any error when deleting the artifact (including it's not found) it will create an event for the error and will return `nil` 

In the `handleDelete` function it will only return a non-nil error if there is any error different from deleting the artifact. 

Also, now the last `err` in the `handleDelete` function comes from this call:
```go
	if err := r.Delete(ctx, &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Name: bundle.Name, Namespace: bundle.Namespace}}); err != nil && !apierrors.IsNotFound(err) {
		logger.V(1).Info("Cannot delete bundle's values secret, owner garbage collection will remove it")
	}
```

weyfonk: Oh, of course :facepalm: . I was wrong, thanks for pointing this out so clearly :D

Issue #3985, opened 2025-08-12
[SURE-10589] Fleet does not delete the "cluster-fleet-local-<clusterName>-<hash>" namespace when a cluster is deleted from Rancher UI.


kkaempf:
### Issue description:
Deleting a cluster from Rancher UI does not clean up the `cluster-fleet-local-<clusterName>-<hash>` namespace from the `SYSTEM` project which is created by fleet. 

### Business impact:
There is no business impact as such but customer has to delete the namespace manually. 

### Troubleshooting steps:
We can clean up this namespace manually or restart the fleet-controller which would clean up the namespace.

### Repro steps:

Create a Rancher provisioned downstream cluster.
This would create a namespace in the SYSTEM project for this cluster in `cluster-fleet-local-<clusterName>-<hash>` format. 
Delete the cluster from Rancher UI. 
This should ideally trigger fleet to remove this namespace once the cluster is deleted. However, the namespace remains as an orphan resource. 

### Workaround:
Is a workaround available and implemented? yes
What is the workaround: Restart the fleet-controller to resolve the issue. 

### Actual behavior:
Upon cluster deletion the `cluster-fleet-local-<clusterName>-<hash>` namespace in SYSTEM project is not cleaned up automatically. 

### Expected behavior:
Once the cluster is deleted, all of its associated resources should be cleaned up from the cluster. 

### Files, logs, traces:
Logs can be uploaded to the Jira on demand since this is easily reproducible.

dharmit:
@kkaempf can I work on this one? I haven't worked on anything related to Fleet yet, so would like to get my hands dirty if that's OK.

kkaempf:
It's yours, good luck 😉

kkaempf:
@dharmit please post your updates here (or file a pull request, referencing this issue 😉 )

dharmit:
I'm still looking into this trying to understand the flow of things. One thing I have been able to deduce so far is that namespace doesn't get deleted for older version too - I tried 2.10.8.

Beyond that, I'm looking for what could be the right place to make this change because a `clusters.fleet.cattle.io` resource seems to be created/owned by a `clusters.provisioning.cattle.io`. My mind doesn't have a full picture of code/architecture of fleet, so I'll keep posting my thoughts/learning here as things become clearer.

dharmit:
[...]

PR #4224, opened 2025-10-14
Deletes the cluster's namespace on Cluster deletion

0xavi0:
Adds a finalizer to the Cluster resource so we can delete the namespace related to it on deletion.

Refers to: https://github.com/rancher/fleet/issues/3985

### DIFF
--- a/internal/cmd/controller/reconciler/cluster_controller.go
+++ b/internal/cmd/controller/reconciler/cluster_controller.go
@@ -10,6 +10,7 @@ import (
 	"sort"
 	"time"
 
+	"github.com/rancher/fleet/internal/cmd/controller/finalize"
 	"github.com/rancher/fleet/internal/cmd/controller/summary"
 	"github.com/rancher/fleet/internal/metrics"
 	"github.com/rancher/fleet/internal/resourcestatus"
@@ -33,6 +34,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
@@ -124,11 +126,18 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 
 	cluster := &fleet.Cluster{}
 	err := r.Get(ctx, req.NamespacedName, cluster)
-	if apierrors.IsNotFound(err) {
-		metrics.ClusterCollector.Delete(req.Name, req.Namespace)
-		return ctrl.Result{}, nil
-	} else if err != nil {
-		return ctrl.Result{}, err
+	if err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	if !cluster.DeletionTimestamp.IsZero() {
+		return r.handleDelete(ctx, cluster)
+	}
+
+	if err := r.ensureFinalizer(ctx, cluster); err != nil {
+		// Retry without updating the status, as this should be a transient error about which users can't do
+		// anything.
+		return ctrl.Result{}, fmt.Errorf("%w, failed to add finalizer to bundle: %w", fleetutil.ErrRetryable, err)
 	}
 
 	if cluster.Status.Namespace == "" {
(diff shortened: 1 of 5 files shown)


internal/cmd/controller/reconciler/cluster_controller.go
```diff
@@ -124,11 +126,18 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
-		metrics.ClusterCollector.Delete(req.Name, req.Namespace)
-		return ctrl.Result{}, nil
-	} else if err != nil {
-		return ctrl.Result{}, err
+	if err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	if !cluster.DeletionTimestamp.IsZero() {
+		return r.handleDelete(ctx, cluster)
```
thardeck, line 134: I think this requires an additional RBAC permission for deleting the namespaces.
```
//+kubebuilder:rbac:groups="",resources=namespaces,verbs=delete
```

internal/cmd/controller/reconciler/cluster_controller.go
```diff
@@ -337,3 +346,42 @@ func (r *ClusterReconciler) mapBundleDeploymentToCluster(ctx context.Context, a
...
+		return nil
+	}
+
+	controllerutil.AddFinalizer(cluster, finalize.ClusterFinalizer)
+
+	return r.Update(ctx, cluster)
+}
+
+// handleDelete runs cleanup the namespace associated to a Cluster,
+// finally removing the finalizer to unblock the deletion of the object from kubernetes.
```
thardeck, line 363: Cleaning up the namespaces should also be done in the [cleanup controller](https://github.com/rancher/fleet/blob/main/internal/cmd/controller/cleanup/controllers/cleanup/controller.go#L93-L100).

I am wondering why this did not work out but we potential could get rid of this code path since the new approach should be more reliable.

internal/cmd/controller/reconciler/cluster_controller.go
```diff
@@ -337,3 +346,42 @@ func (r *ClusterReconciler) mapBundleDeploymentToCluster(ctx context.Context, a
...
+func (r *ClusterReconciler) handleDelete(ctx context.Context, cluster *fleet.Cluster) (ctrl.Result, error) {
+	if !controllerutil.ContainsFinalizer(cluster, finalize.ClusterFinalizer) {
+
+		return ctrl.Result{}, nil
+	}
+
+	if cluster.Status.Namespace != "" {
```
thardeck, line 370: nit pick: It might make sense to check if bundle deployments still exist before deleting the namespace to prevent potentially hanging deletions.

internal/cmd/controller/reconciler/cluster_controller_test.go
```diff
@@ -0,0 +1,169 @@
+package reconciler
```
thardeck, line 1: nit pick: An e2e test or integration test might make sense to have better coverage.

0xavi0: I didn't add any `e2e` test on purpose, because they tend to be more flaky, they're very slow and I think the functionality is covered in the unit test already.
Also...we would need a `multi-cluster e2e` test that it's even slower.

internal/cmd/controller/reconciler/cluster_controller.go
0xavi0: If we check for that in here... we would need to requeue the request until the namespace can be deleted.
But kubernetes is already doing it.
The namespace gets flagged with a DeletionTimestamp != 0 and, when all objects using that namespace are gone, deletes the namepace.

internal/cmd/controller/reconciler/cluster_controller.go
0xavi0: I'm not sure that code works fine, to be honest.

It uses wrangler and, if my understanding is correct, it reacts on Namespace changes calling  cleanupNamespace.
```go
namespaces.OnChange(ctx, "managed-namespace-cleanup", h.cleanupNamespace)
```

Which cleans the namespace only if the Cluster is gone. 
It should be subscribed to changes in Clusters, not namespaces...

I think we might need to review that cleanup controller.

internal/cmd/controller/reconciler/cluster_controller.go
0xavi0: I tested it and it did work fine without any RBAC change.
The `fleet-controller` has already permissions to delete namespaces.
see https://github.com/rancher/fleet/blob/2eba76ad2e4a2fbbbe8baddd2798a6e2a8474af0/charts/fleet/templates/rbac.yaml#L12:L18

internal/cmd/controller/reconciler/cluster_controller.go
thardeck: Yes, but then it is handled by us and we can log on it, but fine with me leaving it to Kubernetes.

PR #4225, opened 2025-10-14
Remove deprecated created-by label

thardeck:
`fleet.cattle.io/created-by-display-name` was added by Rancher UI but was redundant because of the additional label `fleet.cattle.io/created-by-user-id`.

#### Automated Testing
Added integration tests to verify that the deprecated label `fleet.cattle.io/created-by-display-name` is removed if it exists.

### QA Testing Considerations

1. The label was added only in Rancher 2.12.0 and maybe 2.12.1 after adding a Fleet Resource in the Rancher UI.
2. Verify that the label is there (Rancher UI hides it), it can only be seen in the yaml config.
3. Upgrade to a Rancher head with support for the migration (or update the Fleet version manually).
4. Verify that  `fleet.cattle.io/created-by-display-name` is gone but `fleet.cattle.io/created-by-user-id` is still there.

 
`fleet.cattle.io/created-by-user-id`  and other user labels should not be affected by the migration.

### DIFF
--- a/integrationtests/controller/bundle/label_migration_test.go
+++ b/integrationtests/controller/bundle/label_migration_test.go
@@ -55,7 +55,7 @@ var _ = Describe("Bundle label migration", func() {
 	}
 
 	DescribeTable("should remove deprecated label after migration",
-		func(bundleName string, initialLabels map[string]string, shouldHaveDisplayLabel bool) {
+		func(bundleName string, initialLabels map[string]string) {
 			const deprecatedLabel = "fleet.cattle.io/created-by-display-name"
 
 			createBundle(bundleName, initialLabels)
@@ -83,14 +83,12 @@ var _ = Describe("Bundle label migration", func() {
 				"fleet.cattle.io/created-by-display-name": "admin",
 				v1alpha1.CreatedByUserIDLabel:             "user-12345",
 			},
-			true,
 		),
 		Entry("without label present initially",
 			"bundle-without-label",
 			map[string]string{
 				v1alpha1.CreatedByUserIDLabel: "user-12345",
 			},
-			false,
 		),
 	)
 })
--- a/integrationtests/gitjob/controller/label_migration_test.go
+++ b/integrationtests/gitjob/controller/label_migration_test.go
@@ -43,7 +43,7 @@ var _ = Describe("GitRepo label migration", func() {
 	}
 
 	DescribeTable("should remove deprecated label after migration",
-		func(gitRepoName string, initialLabels map[string]string, shouldHaveDisplayLabel bool) {
+		func(gitRepoName string, initialLabels map[string]string) {
 			const deprecatedLabel = "fleet.cattle.io/created-by-display-name"
 
 			createGitRepo(gitRepoName, initialLabels)
@@ -71,14 +71,12 @@ var _ = Describe("GitRepo label migration", func() {
 				"fleet.cattle.io/created-by-display-name": "admin",
 				v1alpha1.CreatedByUserIDLabel:             "user-12345",
 			},
-			true,
 		),
 		Entry("without label present initially",
 			"gitrepo-without-label",
 			map[string]string{
 				v1alpha1.CreatedByUserIDLabel: "user-12345",
 			},
-			false,
 		),
 	)
 })


internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -405,6 +411,28 @@ func (r *GitJobReconciler) addGitRepoFinalizer(ctx context.Context, nsName types
 	return nil
 }
 
+// removeDisplayNameLabel removes the obsolete created-by-display-name label from the gitrepo if it exists.
+func (r *GitJobReconciler) removeDisplayNameLabel(ctx context.Context, nsName types.NamespacedName) error {
```
weyfonk, line 415: Are only GitRepos and bundles affected?
Are bundle deployments' labels updated when the parent bundle's labels are? (I think so, but worth checking)
How about HelmOps?

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -186,6 +186,12 @@ func (r *BundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		return ctrl.Result{}, fmt.Errorf("%w, failed to add finalizer to bundle: %w", fleetutil.ErrRetryable, err)
 	}
 
+	// Migration: Remove the obsolete created-by-display-name label if it exists
+	if err := r.removeDisplayNameLabel(ctx, bundle); err != nil {
+		logger.V(1).Error(err, "Failed to remove display name label")
```
weyfonk: This could be handled similarly to a failure to store a manifest in a `Content` resource, being retryable:
```
return r.computeResult(ctx, logger, bundleOrig, bundle, "failed to remove display name label", err)
```
This also takes care of logging the error message, if `removeDisplayNameLabel` returns a retryable error (see `createOptionsSecret` for an example of that).

integrationtests/controller/bundle/label_migration_test.go
```diff
@@ -0,0 +1,96 @@
...
+						Name:        "cluster",
+						ClusterName: "cluster",
+					},
+				},
+			},
+		}
+		Expect(k8sClient.Create(ctx, bundle)).ToNot(HaveOccurred())
+	}
+
+	DescribeTable("should remove deprecated label after migration",
```
weyfonk, line 57: Nice, TIL :)

integrationtests/controller/bundle/label_migration_test.go
```diff
@@ -0,0 +1,96 @@
...
+		func(bundleName string, initialLabels map[string]string, shouldHaveDisplayLabel bool) {
```
weyfonk: Where is argument `shouldHaveDisplayLabel` used? :thinking: 
(same question in Gitjob tests)

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
thardeck: Bundle deployments might have the label, but they seem to be synchronized with the bundle, because after the migration they did not have the deprecated label in k9s when I checked.

But I will do one specific manual test now to make sure if it is there before.

thardeck: Yes, the label is there before and gone after upgrading Fleet to the version with the migration.

integrationtests/controller/bundle/label_migration_test.go
thardeck: Good point, that was a leftover from a previous implementation.

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
weyfonk: Thanks for checking :)

sbulage:
### System Information

| Rancher Version | Fleet Version |
|--------|--------|
| v2.13-70fc7d2355e78d5571830c4018820b1b085b003c-head | 108.0.0+up0.14.0-beta.1 | 

#### Verified that label: 

| Present | Absent |
|--------|--------|
| `fleet.cattle.io/created-by-user-id` | `fleet.cattle.io/created-by-display-name` | 

:heavy_check_mark: I am able to see the label: 

```yaml
labels:
    fleet.cattle.io/created-by-user-id: <my-user-id>
```

PR #4236, opened 2025-10-16
[Experimental] Support copying user resources to downstream clusters

weyfonk:
A new experimental feature, gated behind environment variable `EXPERIMENTAL_COPY_RESOURCES_DOWNSTREAM`, enables Fleet to copy resources from the upstream cluster to all downstream clusters.

This introduces a new `DownstreamResources` field in bundle deployment options, which can be populated directly from a HelmOp's `spec`. Resources to be copied can be referenced through that field by name and kind, e.g.:
```
kind: HelmOp
[...] # metadata
spec:
  helm:
    [...] # Helm options
  downstreamResources:
    - kind: Secret
      name: my-secret
    - kind: ConfigMap
      name: my-config
```

An important use case for this could be `valuesFrom`, which have requested secrets and/or config maps to be created directly on downstream clusters so far. This feature would eliminate that need.

Restrictions:
* Resources referenced in `downstreamResources` must live in the same namespace as the HelmOp resource.
* Only secrets and config maps are currently supported.

When referencing resources to be copied downstream, and the feature is enabled:
* the Fleet controller will copy those resources from the HelmOp's namespace to the (upstream) namespace of each targeted downstream cluster
* on each downstream cluster, the Fleet agent will:
    * copy those resources from the upstream namespace for the cluster to the deployment namespace, before the bundle deployment is deployed
    * delete these resources when the bundle deployment is deleted, _unless_ `keepResources` is set to `true` on the bundle deployment.

### Limitations:
* copying resources into the _local_ cluster would not make much sense, and has not been tested
* this has only been tested with HelmOps, but it may work with GitOps too, since the implementation is independent of the bundle type
* Fleet does not monitor resources referenced by `downstreamResources` for changes. This could be part of a future iteration of this feature, but in the meantime, changes to secrets and config maps referenced for downstream copy will only be applied when a bundle is updated.

Refers to #3617.

### Checklist

[fleet-docs](https://github.com/rancher/fleet-docs) repository: https://github.com/rancher/fleet-docs/pull/362

### DIFF
--- a/internal/cmd/agent/controller/bundledeployment_controller.go
+++ b/internal/cmd/agent/controller/bundledeployment_controller.go
@@ -157,14 +157,15 @@ func (r *BundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		}
 	}
 
-	if err := r.copyResourcesFromUpstream(ctx, bd, logger); err != nil {
+	forceDeploy, err := r.copyResourcesFromUpstream(ctx, bd, logger)
+	if err != nil {
 		return ctrl.Result{}, err
 	}
 
 	var merr []error
 
 	// helm deploy the bundledeployment
-	if status, err := r.Deployer.DeployBundle(ctx, bd); err != nil {
+	if status, err := r.Deployer.DeployBundle(ctx, bd, forceDeploy); err != nil {
 		logger.V(1).Info("Failed to deploy bundle", "status", status, "error", err)
 
 		// do not use the returned status, instead set the condition and possibly a timestamp
@@ -238,13 +239,13 @@ func (r *BundleDeploymentReconciler) copyResourcesFromUpstream(
 	ctx context.Context,
 	bd *fleetv1.BundleDeployment,
 	logger logr.Logger,
-) error {
+) (bool, error) {
 	if !experimental.CopyResourcesDownstreamEnabled() {
-		return nil
+		return false, nil
 	}
 
 	if len(bd.Spec.Options.DownstreamResources) == 0 {
-		return nil
+		return false, nil
 	}
 
 	destNS := namespaces.GetDeploymentNS(r.DefaultNamespace, bd.Spec.Options)
@@ -253,11 +254,13 @@ func (r *BundleDeploymentReconciler) copyResourcesFromUpstream(
 	if err := r.LocalClient.Get(ctx, types.NamespacedName{Name: ns.Name}, &ns); apierrors.IsNotFound(err) {
 		if err := r.LocalClient.Create(ctx, &ns); err != nil {
 			logger.Info(err.Error())
-			return err
+			return false, err
 		}
 
 		logger.V(1).Info("Created namespace to copy resources from upstream", "namespace", ns.Name)
 	}
+
+	requiresBDUpdate := false
 
 	for _, rsc := range bd.Spec.Options.DownstreamResources {
 		switch strings.ToLower(rsc.Kind) {
@@ -268,7 +271,12 @@ func (r *BundleDeploymentReconciler) copyResourcesFromUpstream(
 				// these objects are copied to the cluster's namespace, hence retries should happen if
 				// they are not found.
 
-				return err
+				return false, fmt.Errorf(
+					"could not get secret %s/%s from upstream namespace for copying: %w",
+					bd.Namespace,
+					rsc.Name,
+					err,
+				)
 			}
 
 			s.Namespace = destNS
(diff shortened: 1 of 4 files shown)


internal/cmd/agent/controller/bundledeployment_controller.go
```diff
@@ -220,6 +231,83 @@ func (r *BundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+
+				return err
+			}
+
+			s.Namespace = destNS
+			s.ResourceVersion = ""
+			if s.Labels == nil {
+				s.Labels = map[string]string{}
+			}
+			s.Labels["fleet.cattle.io/bundledeployment"] = bd.Name
```
0xavi0: Nit: maybe create a constant instead?
I see this same label is used in line 298 and the `deleteResourcesCopiedFromUpstream` function from the `delete.go` file

internal/cmd/agent/controller/bundledeployment_controller.go
```diff
@@ -220,6 +231,83 @@ func (r *BundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+
+			if _, err := controllerutil.CreateOrUpdate(ctx, r.LocalClient, &s, func() error { return nil }); err != nil {
```
0xavi0: Question: 
Why do we use `CreateOrUpdate` with a _MutateFn_ that just returns `nil`? 

If the object already exists it will do nothing even if there are updates, right? 

`CreateOrUpdate` first calls `Get`, retrieves the object from the cache and if the mutation function does nothing the result is that the object is simply not updated.

internal/cmd/agent/controller/bundledeployment_controller.go
```diff
@@ -220,6 +231,83 @@ func (r *BundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+				return err
+			}
+			cm.Namespace = destNS
+			cm.ResourceVersion = ""
+			if cm.Labels == nil {
+				cm.Labels = map[string]string{}
+			}
+			cm.Labels["fleet.cattle.io/bundledeployment"] = bd.Name
+
+			if _, err := controllerutil.CreateOrUpdate(ctx, r.LocalClient, &cm, func() error { return nil }); err != nil {
```
0xavi0: Same question here (see the Secret part)

Also...  isn't the code for _Secrets_ and _ConfigMaps_ exactly the same? (aside from the object type)
I wonder if it could be reused.

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -566,6 +571,52 @@ func (r *BundleReconciler) getOCIReference(ctx context.Context, bundle *fleet.Bu
...
+	}
+
+	updated := targetCM.DeepCopy()
+	if _, err := controllerutil.CreateOrUpdate(ctx, r.Client, targetCM, func() error {
+		targetCM.Data = updated.Data
+		targetCM.BinaryData = updated.BinaryData
+
+		return nil
+	}); err != nil {
+		if !apierrors.IsAlreadyExists(err) {
```
0xavi0: Is this a leftover from the previous implementation?
I think `CreateOrUpdate` can't return an _AlreadyExists_ error

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -582,28 +633,36 @@ func (r *BundleReconciler) cloneSecret(
...
 	}
-	if err := r.Create(ctx, targetSecret); err != nil {
+
+	updated := targetSecret.DeepCopy()
+	if _, err := controllerutil.CreateOrUpdate(ctx, r.Client, targetSecret, func() error {
+		targetSecret.Data = updated.Data
+		targetSecret.StringData = updated.StringData
+
+		return nil
+	}); err != nil {
```
0xavi0, line 660: Same as the config map function. I think `CreateOrUpdate` can't return a _AlreadyExists_

internal/helmdeployer/delete.go
```diff
@@ -152,3 +159,46 @@ func deleteHistory(cfg action.Configuration, logger logr.Logger, bundleID string
...
+	// No information is available about a deleted bundle deployment beside its name and namespace;
+	// in particular, we do not know where its resources copied from the upstream cluster, if any, might live, so we
+	// cannot delete them by name and namespace; instead, we need to resort to labels.
+	opts := client.MatchingLabels{
+		"fleet.cattle.io/bundledeployment": bdName,
+	}
+
+	secrets := corev1.SecretList{}
+
+	// XXX: should we log instead of erroring?
```
0xavi0, line 181: yup, I also wonder if we should log instead...
We can keep it with error for now, the resource was created by the _BundleDeployment_ so I guess it should behave like any other _BundleDeployment_

0xavi0 (commented):
Just minor things and my main concern would be the `CreateOrUpdate` call with a "does nothing" _MutateFn_.

internal/cmd/agent/controller/bundledeployment_controller.go
weyfonk: Yes, thanks for bringing this up. I thought about it but was wondering where that constant should live; finally added it to `pkg/apis`.

internal/cmd/agent/controller/bundledeployment_controller.go
weyfonk: Great catch, thanks! That's yet more evidence that "if it's not tested, it doesn't work" :þ
I've just updated this implementation to use actual mutate functions, and more broadly support updates to `downstreamResources`. Those resources are of course still not directly monitored, but when copying them downstream, the Fleet agent will now detect updates and force-redeploy the bundle deployment when necessary to take those updates into account.

internal/cmd/agent/controller/bundledeployment_controller.go
weyfonk: See my response above.
Also yes, I've had a bit of fun writing generic functions for this, but would eventually just save a couple of lines of code, so left it this way...

However, we could consider it again, especially now that we have more logic involved in each case; the main obstacle for me was finding a way to inject a mutate function into a generic one, so I may have missed something but would rather reconsider this for a future iteration if needed (for instance if we were to support more resource types, beside secrets and config maps).

internal/cmd/controller/reconciler/bundle_controller.go
weyfonk: Good catch, yes, it seems to be. Removed, thanks!

internal/cmd/controller/reconciler/bundle_controller.go
weyfonk: Yup, removed, thanks :)

internal/cmd/agent/controller/bundledeployment_controller.go
0xavi0: sure, we can do it in another iteration.

Issue #4246, opened 2025-10-20
Troubleshoot flaky tests


weyfonk:
We have seen a couple of examples of flaky tests over the past few days:
* single-cluster end-to-end tests: signal handling: [example](https://github.com/rancher/fleet/actions/runs/18612349475/job/53072245205)
    * see https://github.com/rancher/fleet/pull/4247 for an attempt at fixing them
* multi-cluster end-to-end tests: copy of resources downstream: [example](https://github.com/rancher/fleet/actions/runs/18630877784/job/53115496535), which makes both `E2E Upgrade Fleet in Rancher To HEAD` and `Test Fleet in Rancher` workflows fail
* benchmarks: see [this run](https://github.com/rancher/fleet/actions/runs/19434081697/job/55600213613)
* integration tests: HelmOps bundle creation from index, see [this run](https://github.com/rancher/fleet/actions/runs/19419028241/job/55552823632)

Let's troubleshoot and fix them.

PR #4247, opened 2025-10-20
Reduce flakiness of signal handling tests

weyfonk:
End-to-end tests for signal handling depend on availability of deployments, which may be affected by other running test suites, as is the case of the Fleet agent with pod disruption budgets and priority classes.
Therefore, signal handling tests now retry killing a deployment if that deployment or its targeted container is not available.

Refers to #4246.

### Checklist

### DIFF
--- a/e2e/single-cluster/signals_test.go
+++ b/e2e/single-cluster/signals_test.go
@@ -28,8 +28,10 @@ var _ = Describe("Shuts down gracefully", func() {
 		})
 
 		JustBeforeEach(func() {
-			out, err := k.Run("exec", "deploy/fleet-agent", "--", "kill", "1")
-			Expect(err).ToNot(HaveOccurred(), out)
+			Eventually(func(g Gomega) {
+				out, err := k.Run("exec", "deploy/fleet-agent", "--", "kill", "1")
+				g.Expect(err).ToNot(HaveOccurred(), out)
+			}).Should(Succeed())
 		})
 
 		It("exits gracefully", func() {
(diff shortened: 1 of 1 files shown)


p-se:
https://github.com/onsi/ginkgo/issues/1562

e2e/single-cluster/signals_test.go
```diff
@@ -93,11 +97,13 @@ var _ = Describe("Shuts down gracefully", func() {
...
-			Expect(err).ToNot(HaveOccurred(), pod)
+			Eventually(func(g Gomega) {
+				pod, err := k.Get("pod", "-l", labels, "-o", "jsonpath={.items[0].metadata.name}")
+				g.Expect(err).ToNot(HaveOccurred(), pod)
 
-			out, err := k.Run("exec", pod, "--", "kill", "1")
-			Expect(err).ToNot(HaveOccurred(), out)
+				out, err := k.Run("exec", pod, "--", "kill", "1")
+				g.Expect(err).ToNot(HaveOccurred(), out)
+			})
```
p-se:
```suggestion
				out, err := k.Run("exec", pod, "--", "kill", "1")
				g.Expect(err).ToNot(HaveOccurred(), out)
			}).Should(Succeed())
```

e2e/single-cluster/signals_test.go
```diff
@@ -127,11 +133,13 @@ var _ = Describe("Shuts down gracefully", func() {
...
```
p-se:
```suggestion
				out, err := k.Run("exec", pod, "--", "kill", "1")
				g.Expect(err).ToNot(HaveOccurred(), out)
			}).Should(Succeed())
```

p-se (approved):

e2e/single-cluster/signals_test.go
weyfonk: Good catch, thanks! :bow: Fixed.

PR #4249, opened 2025-10-20
 Add UserID logging to HelmOp and BundleDeployment reconciliers

thardeck:
HelmOps resources do not provide the UserID label yet, because it is not implemented in the Rancher UI for now.
But as soon as the label is added, it will be used for HelmOps logging.

Additionally the UserID logging was added to the BundleDeployment controller too, to be consistent. The bundles distribute the label to the BundleDeployments.

Follow up to #4202

Refers to #3424

### QA Consideration

Same as mentioned in https://github.com/rancher/fleet/issues/3424#issuecomment-3385446376. The only difference is that Rancher UI does not add the UserID label for HelmOp bundles yet, so this would need to be done manually to see if it is used for logging.

### DIFF
--- a/integrationtests/controller/bundledeployment/suite_test.go
+++ b/integrationtests/controller/bundledeployment/suite_test.go
@@ -7,7 +7,6 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-	"go.uber.org/zap/zapcore"
 	ctrl "sigs.k8s.io/controller-runtime"
 
 	"github.com/rancher/fleet/integrationtests/utils"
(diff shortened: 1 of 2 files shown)


integrationtests/controller/bundledeployment/suite_test.go
```diff
@@ -33,6 +38,11 @@ func TestFleet(t *testing.T) {
 }
 
 var _ = BeforeSuite(func() {
+	// Configure log capture
+	GinkgoWriter.TeeTo(&logsBuffer)
+	zapLogger := zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true), zap.Level(zapcore.Level(-4)))
```
weyfonk: Why `-4`?
Same question for HelmOps integration tests.

thardeck: Should not be there. This was from debugging the test before.

sbulage:
It is not working as expected also UI implementation is missing as well. But user will not get impacted as it doesn't hurt any `HelmOps` functionality to work.

PR #4257, opened 2025-10-23
Use a scheduler for gitops polling

0xavi0:
This PR replaces the mechanism of using *RequeueAfter* to implement polling with a Quartz Scheduler similar to the one used in *HelmOps*.

An attempt has been made to use the same approach so that the schedulers are as similar as possible; perhaps in a future iteration the scheduler can be further generalized.

A new Quartz trigger (`ControllerTrigger`) has been introduced, which is used in both *GitOps* and *HelmOps*. Besides having the functionality to execute the task when we add the job to the scheduler, it also has the ability to add jitter to avoid collisions in scenarios with many resources.

Some new unit tests have been created, and others have been removed as, given the new implementation, they no longer made sense.

Additionally, a new value has been added to the Status of a GitRepo (`PollingCommit`) so that when the job performing the polling finds a new commit, it can patch only this field, avoiding collisions with other possible Status updates.

Refers to: https://github.com/rancher/fleet/issues/3394

### Checklist

### DIFF
--- a/internal/cmd/controller/gitops/reconciler/gitjob_controller.go
+++ b/internal/cmd/controller/gitops/reconciler/gitjob_controller.go
@@ -846,6 +846,8 @@ func getFleetCLIErrorsFromLine(l string) string {
 	return s
 }
 
+// getNextCommit returns a commit SHA coming either from the status' webhook
+// commit or, with lower precedence, from the polling commit.
 func getNextCommit(status v1alpha1.GitRepoStatus) string {
 	commit := status.Commit
 	if status.PollingCommit != "" && status.PollingCommit != commit {
--- a/internal/cmd/controller/gitops/reconciler/polling_job.go
+++ b/internal/cmd/controller/gitops/reconciler/polling_job.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2024 SUSE LLC
+// Copyright (c) 2025 SUSE LLC
 
 package reconciler
 
--- a/internal/cmd/controller/gitops/reconciler/polling_job_test.go
+++ b/internal/cmd/controller/gitops/reconciler/polling_job_test.go
@@ -54,7 +54,6 @@ func TestPollGitRepo(t *testing.T) {
 				})
 				gf.EXPECT().LatestCommit(gomock.Any(), gomock.Any(), gomock.Any()).Return("new-commit", nil)
 				c.EXPECT().Status().Return(sw)
-				// sw.EXPECT().Patch(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil)
 			},
 			expectedEvents: []string{"Normal GotNewCommit new-commit"},
 			validateGitRepo: func(t *testing.T, gr *v1alpha1.GitRepo) {
@@ -186,7 +185,7 @@ func TestPollGitRepo(t *testing.T) {
 
 			if tc.expectedErr != "" {
 				if err == nil || err.Error() != tc.expectedErr {
-					t.Errorf("expected error '%s', got '%v'", tc.expectedErr, err)
+					t.Errorf("expected error '%q', got '%v'", tc.expectedErr, err)
 				}
 			} else if err != nil {
 				t.Errorf("unexpected error: %v", err)
@@ -199,7 +198,7 @@ func TestPollGitRepo(t *testing.T) {
 				}
 				for i, expectedEvent := range tc.expectedEvents {
 					if event, ok := <-recorder.Events; ok && event != expectedEvent {
-						t.Errorf("expected event %d to be '%s', got '%s'", i, expectedEvent, event)
+						t.Errorf("expected event %d to be '%q', got '%q'", i, expectedEvent, event)
 					}
 				}
 			}
@@ -224,7 +223,7 @@ func TestGitPollingJob_Description(t *testing.T) {
 
 	expected := "gitops-polling-test-ns-test-repo-http://a.b/c.git-develop"
 	if job.Description() != expected {
-		t.Errorf("expected description '%s', got '%s'", expected, job.Description())
+		t.Errorf("expected description '%q', got '%q'", expected, job.Description())
 	}
 }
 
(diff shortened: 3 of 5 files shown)


internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -803,3 +845,17 @@ func getFleetCLIErrorsFromLine(l string) string {
 
 	return s
 }
+
+func getNextCommit(status v1alpha1.GitRepoStatus) string {
```
weyfonk, line 851: nit: a bit of function-level documentation would be welcome here, e.g.:
```suggestion
// getNextCommit returns a commit SHA coming either from the status' webhook commit or, with lower precedence, from the polling commit.
func getNextCommit(status v1alpha1.GitRepoStatus) string {
```

internal/cmd/controller/gitops/reconciler/polling_job.go
```diff
@@ -0,0 +1,163 @@
+// Copyright (c) 2024 SUSE LLC
```
weyfonk: 2024? :upside_down_face:

internal/cmd/controller/quartz/trigger.go
```diff
@@ -0,0 +1,64 @@
...
+
+// jitter returns a random jitter between 0% and +jitterPercent% of the original duration.
+// jitterPercent is an integer percentage (e.g., 10 for 10%).
+func jitter(d time.Duration, jitterPercent int) time.Duration {
+	if jitterPercent <= 0 {
+		return 0
+	}
+
+	// Convert jitter percent to a fraction
+	jitterFraction := float64(jitterPercent) / 100.0
```
weyfonk, line 62: nit: we _could_ consider limiting the jitter to 100%, although that's of course not critical, because this logic is only used internally.

internal/cmd/controller/quartz/trigger.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+// ControllerTrigger is a custom trigger, implementing the quartz.Trigger interface. This trigger is
+// used to schedule jobs to be run both:
+// * periodically, after the first polling interval, as would happen with Quartz's `simpleTrigger`
+// * right away, without waiting for that first polling interval to elapse.
+// It also adds the possibility to add a percentage of jitter to the duration of the trigger to avoid
+// situations in which we have many collisions.
+type ControllerTrigger struct {
+	IsInitRunDone bool
```
weyfonk: Why does this field need to be exported?

internal/cmd/controller/gitops/reconciler/polling_job_test.go
```diff
@@ -0,0 +1,240 @@
...
+					obj.Name = name
+					obj.Namespace = namespace
+					obj.Spec.Repo = repoURL
+					obj.Spec.Branch = branch
+					obj.Status.Commit = "old-commit"
+					return nil
+				})
+				gf.EXPECT().LatestCommit(gomock.Any(), gomock.Any(), gomock.Any()).Return("new-commit", nil)
+				c.EXPECT().Status().Return(sw)
+				// sw.EXPECT().Patch(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil)
```
weyfonk:
```suggestion
```

internal/cmd/controller/gitops/reconciler/polling_job_test.go
```diff
@@ -0,0 +1,240 @@
...
+			}
+
+			close(recorder.Events)
+			if len(tc.expectedEvents) > 0 {
+				if len(recorder.Events) != len(tc.expectedEvents) {
+					t.Errorf("expected %d events, got %d", len(tc.expectedEvents), len(recorder.Events))
+				}
+				for i, expectedEvent := range tc.expectedEvents {
+					if event, ok := <-recorder.Events; ok && event != expectedEvent {
+						t.Errorf("expected event %d to be '%s', got '%s'", i, expectedEvent, event)
```
weyfonk: nit: `'%s'` can be expressed as `%q` (same remark about L189 and L227):
```suggestion
						t.Errorf("expected event %d to be %q, got'%q", i, expectedEvent, event)
```

integrationtests/gitjob/controller/controller_test.go
```diff
@@ -509,6 +509,26 @@ var _ = Describe("GitJob controller", func() {
 					return k8sClient.Delete(ctx, &job)
 				}).Should(Not(HaveOccurred()))
 
+				// Eventually the job should have an "orphan" finalizer
```
weyfonk, line 512: Where does this come from? Does k8s add this finalizer to jobs?

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
0xavi0: Added as suggested

internal/cmd/controller/gitops/reconciler/polling_job.go
0xavi0: I did it long long ago :bowtie:
Fixed

internal/cmd/controller/quartz/trigger.go
0xavi0: Sure, it's only used internally, but in case the number entered is greater it limits it to 100. 
I also added a line in the function's doc

internal/cmd/controller/quartz/trigger.go
0xavi0: It does not need to be exported. I guess it was at some point and forgot to unexport :) 
Changed

internal/cmd/controller/gitops/reconciler/polling_job_test.go
0xavi0: changed as suggested

integrationtests/gitjob/controller/controller_test.go
0xavi0: Yeah, k8s adds the finalizer to ensure the dependant pod is deleted.

Issue #4256, opened 2025-10-23
Improve tests on Release to prevent dev merges into release branches


mmartin24:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

As a need for QA I created a new chart `0.12.9-beta.1` to test into `2.11-head` (note: NOT on the `2.11.7`) on the previous day to the release of Fleet `0.12.8` and it successfully managed to build: https://github.com/rancher/rancher/actions/runs/18684812883

Chart is also in the assets:

https://github.com/rancher/charts/blob/dev-v2.11/assets/fleet/fleet-106.1.7%2Bup0.12.9-beta.1.tgz

However, the chart is not found when attempting to deploy it in Rancher. 
```
k logs -n cattle-system rancher-6656c98b66-qrxdh 

2025/10/22 09:31:16 [INFO] Starting cluster controllers for local
2025/10/22 09:31:18 [ERROR] Failed to install system chart fleet-crd: no chart version found for fleet-crd-106.1.7+up0.12.9-beta.1
2025/10/22 09:31:18 [ERROR] Failed to install system chart fleet: no chart version found for fleet-106.1.7+up0.12.9-beta.1



```

<img width="1070" height="934" alt="Image" src="https://github.com/user-attachments/assets/bb34571f-1b92-4d80-839a-5860f0e0578c" />

Logs proving is not visible on Rancher head:
[rancher_2.11_fleet_0.12.9-beta-1.log](https://github.com/user-attachments/files/23093981/rancher_2.11_fleet_0.12.9-beta-1.log)

The chart did work as standalone, but this is another issue.

---

Now the issue is that this chart should not have reached 2.11.8

Adding comment from @thardeck 

> The reason for the issue is that during the Rancher release, the charts branch are switched from dev to the release branches.
> 
> As you can see here for [Rancher 2.11](https://github.com/rancher/rancher/blob/release/v2.11/package/Dockerfile#L20).
> Rancher 2.10 was not released and is still on the [dev-v2.10 branch](https://github.com/rancher/rancher/blob/release/v2.10/package/Dockerfile#L3).
> 
> After Rancher is released, they usually switch the branch back, but it was not-
> I am just wondering why the release/v2.11 was not locked during this phase, so this merge would have been prevented 

### Expected Behavior

Prevent these types of merges during release

### Environment

```markdown
- Architecture:
- Fleet Version:
- Cluster:
  - Provider:
  - Options:
  - Kubernetes Version:
```

[...]

PR #4261, opened 2025-10-27
 Add code freeze verification for alpha/beta versions in release workflow

thardeck:
Refers to #4256

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,6 +30,29 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Verify Rancher is not in code freeze (alpha/beta only)
+        run: |
+          set -x
+          FLEET_VERSION="${GITHUB_REF_NAME#v}"
+
+          if [[ ! "${FLEET_VERSION}" =~ (alpha|beta) ]]; then
+            echo "Skipping code freeze check for non-alpha/beta release"
+            exit 0
+          fi
+
+          FLEET_MINOR=$(echo "${FLEET_VERSION}" | cut -d'.' -f2)
+          RANCHER_VERSION="2.$((FLEET_MINOR - 1))"
+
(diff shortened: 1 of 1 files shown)


.github/scripts/release-against-rancher.sh
```diff
@@ -24,8 +24,18 @@ if [ ! -e ~/.gitconfig ]; then
     git config --global user.email fleet@suse.de
 fi
 
-# Check if version is available online
 CHART_DEFAULT_BRANCH=$(grep "ARG CHART_DEFAULT_BRANCH=" package/Dockerfile | cut -d'=' -f2)
+
+# Prevent alpha and beta releases during the Rancher release window
+if [[ "${NEW_FLEET_VERSION}" =~ (alpha|beta) ]]; then
+    if [[ ! "${CHART_DEFAULT_BRANCH}" =~ ^dev- ]]; then
```
weyfonk, line 31: Does this deliberately not apply to RCs?

.github/workflows/release.yml
```diff
@@ -30,6 +30,29 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Verify Rancher is not in code freeze (alpha/beta only)
+        run: |
+          set -x
+          FLEET_VERSION="${GITHUB_REF_NAME#v}"
+
+          if [[ ! "${FLEET_VERSION}" =~ (alpha|beta) ]]; then
```
weyfonk, line 38: Same question as above.

.github/scripts/release-against-rancher.sh
thardeck: Yes, because otherwise we could not create new RCs during code freeze or unrc the final version.

.github/workflows/release.yml
thardeck: Yes, because otherwise we could not create new RCs during code freeze or unrc the final version.

Issue #4144, opened 2025-09-17
Fleet rollout gets stuck


manno:
I am using Fleet `v0.13.2-rc.2` to deploy to 500 clusters. The rollout gets stuck only 164/500 bundledeployments are created.
The deployment consists of 25k bundledeployments. The reconcile workers are configured to:

```
  --set controller.reconciler.workers.gitrepo=100 \
  --set controller.reconciler.workers.bundle=200 \
  --set controller.reconciler.workers.bundledeployment=200 \
  --set controller.reconciler.workers.cluster=100 \
```

Status of the stuck bundle:

```
   conditions:                                                                                                                                                              
   - lastUpdateTime: "2025-09-17T12:18:45Z"                                                                                                                                  
     message: 'targeting error: failed to get options secret for bundledeployment cluster-fleet-default-d0-k3k-downstream001-downstream0050-fa83d/scale-50-single-scale-50-bundles-twenty-sixteen,                                                                                                                                                              
       this is likely temporary: secrets "scale-50-single-scale-50-bundles-twenty-sixteen"                                                                      
       not found'                                                                                                                                                  
     reason: Error                                                                                                                                                           
     status: "False"                                                                                                                                                            
     type: Ready                                                                                                                                                                
   display:                                                                                                                                                                  
     readyClusters: 164/500                                                                                                                                                     
     state: WaitApplied
```

When I check, the secret is missing. 
* I waited 30 minutes.
* Restarting the fleet-controller does not create the missing secret.
* Upgrading to [v0.14.0-alpha.2](https://github.com/rancher/fleet/releases/tag/v0.14.0-alpha.2) does not create the secret.
* Increasing the `forceSyncGeneration` re-creates all bundledeployments. Possibly getting stuck again.

innobead:
This takes priority because it affects the upcoming Harvester 1.7.0 scheduled for December due to the impact of https://github.com/harvester/harvester/issues/9349. 

cc @Vicente-Cheng @olblak @ibrokethecloud @rebeccazzzz

ibrokethecloud:
i suspect the issue is arising from the change in behaviour on when `createOrUpdate` bundledeployment fails

https://github.com/rancher/fleet/compare/v0.13.0...v0.14.0-alpha.3#diff-38f599d4b6f84abae0fd9c6582e85b27f3f8a52774a565742e207ef52302274eL362

The controller now logs the error and continues with processing subsequent targets.

[...]

PR #4273, opened 2025-10-29
Track UID of existing bundle deployment to prevent orphaning on failure

thardeck:
Track the UID of existing bundledeployments before attempting the create/update operation.

This should prevent incorrect cleanup of resources while keeping the fix from #4074 (Bundle controller continues when one bundledeployment cannot be created ).

Refers to #4144

### DIFF
--- a/integrationtests/controller/bundle/bundle_error_handling_test.go
+++ b/integrationtests/controller/bundle/bundle_error_handling_test.go
@@ -182,7 +182,7 @@ var _ = Describe("Bundle controller error handling", Ordered, func() {
 			cleanupResources(bundle, bundleName, testID, content, cluster1, cluster2)
 		})
 
-		It("should not delete existing bundledeployments when content resource fails", func() {
+		It("should not delete existing bundledeployments when bundle deployment creation fails", func() {
 			By("capturing the original bundledeployment UIDs")
 			var uid1, uid2 types.UID
 			Eventually(func(g Gomega) {


internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -358,6 +358,14 @@ func (r *BundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 
 		helmvalues.ClearOptions(bd)
 
+		// If there's already a bundledeployment for this target, track its UID
+		// before calling createBundleDeployment, which might fail. This prevents
+		// cleanupOrphanedBundleDeployments from incorrectly removing this bundledeployment
+		// as "orphaned". See https://github.com/rancher/fleet/issues/4144
+		if target.Deployment != nil && target.Deployment.UID != "" {
+			bundleDeploymentUIDs.Insert(target.Deployment.UID)
```
weyfonk, line 366: Right, the bundle deployment's `UID` is also added below if creating the bundle deployment succeeds, but that shouldn't be an issue as `bundleDeploymentUIDs` is a set :+1:

integrationtests/controller/bundle/bundle_error_handling_test.go
```diff
@@ -0,0 +1,317 @@
...
+			bundle = createBundle(bundleName, bundleNS, "default", "test", []v1alpha1.BundleTarget{{
+				ClusterSelector: &metav1.LabelSelector{MatchLabels: map[string]string{"env": "test"}},
+			}})
+		})
+
+		AfterEach(func() {
+			cleanupResources(bundle, bundleName, testID, content, cluster1, cluster2)
+		})
+
+		It("should not delete existing bundledeployments when content resource fails", func() {
```
weyfonk: nit:
```suggestion
		It("should not delete existing bundledeployments when bundle deployment creation fails", func() {
```

weyfonk (approved):

PR #4276, opened 2025-10-29
[v0.14] Track UID of existing bundle deployment to prevent orphaning on failure

thardeck:
Track the UID of existing bundledeployments before attempting the create/update operation.

This should prevent incorrect cleanup of resources while keeping the fix from https://github.com/rancher/fleet/pull/4074 (Bundle controller continues when one bundledeployment cannot be created ).

Refers to https://github.com/rancher/fleet/issues/4144

### DIFF
--- a/integrationtests/controller/bundle/bundle_error_handling_test.go
+++ b/integrationtests/controller/bundle/bundle_error_handling_test.go
@@ -1,0 +1,317 @@
+package bundle
+
+import (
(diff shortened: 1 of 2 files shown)


mmartin24:
It seems verified by Harvester team [here](https://github.com/rancher/fleet/issues/4144#issuecomment-3484611555)
Closing issue

Issue #4275, opened 2025-10-29
ShardID is not taken into account for Watches statements


0xavi0:
This was reported in https://github.com/rancher/fleet/issues/4183 and confirmed. 

The _Bundle_ reconciler is filtering its requests using:
```go
WithEventFilter(sharding.FilterByShardID(r.ShardID))
```
However, a Watches function is also being used, which monitors changes in Clusters and retrieves the Bundles that are deployed in those Clusters.
In that case, the request is being enqueued without filtering by ShardID, which can lead to a controller without an assigned ShardID processing that request.

More controllers might be affected by this.

PR #4282, opened 2025-10-30
Adds a sharding filter when calculating Bundles from Cluster

0xavi0:
The mapping function in the Bundle reconciler calls BundlesForCluster and gets Bundles from Clusters. Once the Bundle is obtained we should also filter that Bundle by shardID.

Refers to: https://github.com/rancher/fleet/issues/4275

### DIFF
--- a/internal/cmd/controller/reconciler/bundle_controller.go
+++ b/internal/cmd/controller/reconciler/bundle_controller.go
@@ -1022,20 +1022,3 @@ func BundleDeploymentMapFunc(r *BundleReconciler) func(ctx context.Context, a cl
 		return nil
 	}
 }
-
-// BundleDeploymentStatusChangedPredicate returns true if the bundledeployment
-// status has changed, or the bundledeployment was created or is being deleted.
-func BundleDeploymentStatusChangedPredicate() predicate.Funcs {
-	return predicate.Funcs{
-		CreateFunc: func(e event.CreateEvent) bool { return true },
-		UpdateFunc: func(e event.UpdateEvent) bool {
-			n, nOK := e.ObjectNew.(*fleet.BundleDeployment)
-			o, oOK := e.ObjectOld.(*fleet.BundleDeployment)
-			if !nOK || !oOK {
-				return false
-			}
-			return !n.DeletionTimestamp.IsZero() || !reflect.DeepEqual(n.Status, o.Status)
-		},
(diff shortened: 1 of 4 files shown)


internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -1008,3 +992,50 @@ func upper(op controllerutil.OperationResult) string {
 		return "Unknown"
 	}
 }
+
+// BundleDeploymentMapFunc returns a function that maps a BundleDeployment to a Bundle.
+func BundleDeploymentMapFunc(r *BundleReconciler) func(ctx context.Context, a client.Object) []reconcile.Request {
```
weyfonk, line 997: nit: we don't need to export this function.

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -1008,3 +992,50 @@ func upper(op controllerutil.OperationResult) string {
...
+			}}
+		}
+
+		return nil
+	}
+}
+
+// BundleDeploymentStatusChangedPredicate returns true if the bundledeployment
+// status has changed, or the bundledeployment was created or is being deleted.
+func BundleDeploymentStatusChangedPredicate() predicate.Funcs {
```
weyfonk: Where is this used?
It looks like an improved duplicate of `bundleDeploymentStatusChangedPredicate()`, which lives in the bundle deployment controller.

internal/cmd/controller/reconciler/bundle_controller.go
0xavi0: I did it this way because the `bundle_controller_test.go` is in the `reconciler_test` package.
I can also add a specific file just to test this part of the code... I thought that was adding too many changes for an export in an internal package.

internal/cmd/controller/reconciler/bundle_controller.go
0xavi0: I forgot to move it to the _BundleDeployment_ file :) 
Added now unexported and I've also moved the unit tests to a _BundleDeployment_ specific file.

weyfonk (approved):

internal/cmd/controller/reconciler/bundle_controller.go
weyfonk: Oh, I'd missed that :þ No worries then, thanks.

PR #4284, opened 2025-11-03
Create options secret before bd

weyfonk:
In order to prevent errors triggered by a missing or outdated options secret, the bundle controller now creates/updates it _before_ creating/updating the bundle deployment owning it.

Updating bundle controller unit tests to fit with this new logic was easier after a bit of simplification, hence the additional diff.

Supersedes #3889.

### Checklist

### DIFF
--- a/internal/cmd/controller/reconciler/bundle_controller.go
+++ b/internal/cmd/controller/reconciler/bundle_controller.go
@@ -344,7 +344,7 @@ func (r *BundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 			bundleDeploymentUIDs.Insert(target.Deployment.UID)
 		}
 
-		bd, err = r.createBundleDeployment(
+		op, bd, err := r.createBundleDeployment(
 			ctx,
 			logger,
 			bd,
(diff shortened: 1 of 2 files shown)


internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -368,19 +364,13 @@ func (r *BundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
-				logger.Info(err.Error())
-
-				return ctrl.Result{RequeueAfter: durations.DefaultRequeueAfter}, nil
+		// At this stage, we know the UID of our bundle deployment, hence we can use it to populate the owner reference in the
+		// options secret.
+		// If the bundle deployment already existed and has simply been updated, the secret will already bear an owner
+		// reference from its creation or latest update.
+		if op == controllerutil.OperationResultCreated {
+			if err := r.ensureOwnerReferences(ctx, bd, optionsSecret); err != nil {
+				return r.computeResult(ctx, logger, bundleOrig, bundle, "failed to ensure owner references are set in options secret", err)
```
0xavi0, line 373: Question: What’s the advantage of setting the owner reference when calling `manageOptionsSecret`? That we skip this call in here when updating?

0xavi0 (approved):
lgtm... just a question in case I missed something

internal/cmd/controller/reconciler/bundle_controller.go
weyfonk: Yes, that's the idea:
```
		// If the bundle deployment already existed and has simply been updated, the secret will already bear an owner
		// reference from its creation or latest update.
```
If the bundle has been updated, its owner reference should already be set by `manageOptionsSecret`, which created or updated the secret. This is an optimisation.

On the other hand, if the bundle was created, it did then not yet exist when `manageOptionsSecret` was called, hence the need to ensure the owner reference is set here.

Issue #3742, opened 2025-05-30
[SURE-10259] Feature Request: scan bundles concurrently in Fleet CLI


aruiz14:
SURE-10259

### Is your feature request related to a problem?

When a new commit is detected in a `GitRepo`, this is cloned and scanned looking for Bundles to be created/updated. This happens sequentially, which can be slow since it's affected by repository size/structure, bundle size, network connectivity, etc.

### Solution you'd like

Run scan and processing of multiple bundles concurrently.

sbulage:
### System Information
| Rancher Version | Fleet Version |
|--------|--------|
| v2.12-21a76a198159bd46a12ec261791bf584194a84a3-head| 107.0.0+up0.13.0-beta.3 | 


For verification, we have seen our CI for 2.12-head version:
- Tests are passing for 4 - 5 runs.
- There is no issue found with tests related to bundles.

sbulage:
### System Information
| Rancher Version | Fleet Version |
|--------|--------|
| v2.12-d1bb158637d9a02a85ee57e98b182b7c3bb6b423-head| 107.0.0+up0.13.0-beta.4 | 


Following Steps performed:
- GitRepo created using @manno's fleet-experiments repo which installs 50 bundles.
- Time took from GitRepo Create to GitRepo Active is around ~20-25 seconds compared to 2.11 which is ~30-35 seconds.
- Also, saw that bundles were created in batches and sequentially.
- No Errors were encountered during the verification testing.
- QA's CI for testing single or more than 2 bundles testing also working fine without any error.

Based on above observation, Verifying this issue.

Thanks @aruiz14 for your guidance and information 😄

kkaempf:
Can this issue be closed now ?

PR #4285, opened 2025-11-03
Aggregate concurrently fetching the same index.yaml

aruiz14:
Refers to #4343

This PR aggregates concurrent Helm repository index fetching into a single call, achieving the same "serialized access" pattern used before #3742 without giving up on parallel processing. This is meant to alleviate the load put on non-static Helm repositories that have to work with many entries (potentially making the index.yaml size grow over 20-30MB or more).

## Additional Information

I found the implementation around getting the chart URL a bit convoluted, since it has grown organically since OCI chart support was added, so I've performed a little refactor of the affected area to make it easier to reason about.

It also contains a fix for what I think is a bug in the auth management. I extracted it to a separate PR (#4336), so this commit should be gone after that one is merged and rebased.

### Experiments

I created a fake registry simulating slow queries for an index.yaml, as certain implementations that calculate it on the fly could become quite heavy for a high number of entries (even reaching 30-40MB or more).
I configured it to limit concurrency to 1, and take 30s to serve every request.
Then, I created a GitRepo monitoring 25 Bundles (a to z) and observed the GitJob pod lods. As expected, bundle creation logs were appearing sequentially and took 12 minutes to complete:
```
time="2025-11-17T16:39:19Z" level=info msg="created (bundle): fleet-local/test-many-bundles-same-repo-test-a"
time="2025-11-17T16:39:49Z" level=info msg="created (bundle): fleet-local/test-many-bundles-same-repo-test-d"
time="2025-11-17T16:40:19Z" level=info msg="created (bundle): fleet-local/test-many-bundles-same-repo-test-c"
time="2025-11-17T16:40:49Z" level=info msg="created (bundle): fleet-local/test-many-bundles-same-repo-test-b"
......
time="2025-11-17T16:51:19Z" level=info msg="created (bundle): fleet-local/test-many-bundles-same-repo-test-x"
time="2025-11-17T16:51:49Z" level=info msg="created (bundle): fleet-local/test-many-bundles-same-repo-test-y"
time="2025-11-17T16:52:19Z" level=info msg="created (bundle): fleet-local/test-many-bundles-same-repo-test-z"
```

However, with this patch, the messages appeared in batches (of 4, as it's configured the max concurrency at the moment), and only took 3m:
```
time="2025-11-17T17:33:40Z" level=info msg="updated (bundle): fleet-local/test-many-bundles-same-repo-test-a"
time="2025-11-17T17:33:40Z" level=info msg="updated (bundle): fleet-local/test-many-bundles-same-repo-test-b"
time="2025-11-17T17:33:40Z" level=info msg="updated (bundle): fleet-local/test-many-bundles-same-repo-test-d"
time="2025-11-17T17:33:40Z" level=info msg="updated (bundle): fleet-local/test-many-bundles-same-repo-test-c"
time="2025-11-17T17:34:10Z" level=info msg="updated (bundle): fleet-local/test-many-bundles-same-repo-test-e"
...
time="2025-11-17T17:36:40Z" level=info msg="updated (bundle): fleet-local/test-many-bundles-same-repo-test-z"
time="2025-11-17T17:36:40Z" level=info msg="updated (bundle): fleet-local/test-many-bundles-same-repo-test-y"
```

### Checklist

### DIFF
--- a/internal/bundlereader/auth.go
+++ b/internal/bundlereader/auth.go
@@ -3,6 +3,7 @@ package bundlereader
 import (
 	"context"
 	"crypto/sha256"
+	"encoding/binary"
 	"encoding/hex"
 	"fmt"
 	"strconv"
@@ -22,8 +23,6 @@ type Auth struct {
 	// remember to update Hash() when adding/modifying fields
 }
 
-var nullBytes = []byte("\x00")
-
 func toByte(v bool) byte {
 	if v {
 		return byte(1)
@@ -33,6 +32,15 @@ func toByte(v bool) byte {
 
 func (a Auth) Hash() string {
 	hash := sha256.New()
+
+	// Use data length as delimiter to avoid possible collisions
+	lenBuf := make([]byte, 8)
+	writeField := func(data []byte) {
+		binary.LittleEndian.PutUint64(lenBuf, uint64(len(data)))
+		hash.Write(lenBuf)
+		hash.Write(data)
+	}
+
 	for _, v := range [][]byte{
 		[]byte(a.Username),
 		[]byte(a.Password),
@@ -41,9 +49,9 @@ func (a Auth) Hash() string {
 		{toByte(a.InsecureSkipVerify)},
 		{toByte(a.BasicHTTP)},
 	} {
-		hash.Write(v)
-		hash.Write(nullBytes)
+		writeField(v)
 	}
+
 	return hex.EncodeToString(hash.Sum(nil))
 
 }


internal/bundlereader/resources.go
```diff
@@ -208,11 +208,11 @@ func mergeGenericMap(first, second *fleet.GenericMap) *fleet.GenericMap {
 func addRemoteCharts(directories []directory, base string, charts []*fleet.HelmOptions, auth Auth, helmRepoURLRegex string) ([]directory, error) {
 	for _, chart := range charts {
 		if _, err := os.Stat(filepath.Join(base, chart.Chart)); os.IsNotExist(err) || chart.Repo != "" {
+			auth := auth
 			shouldAddAuthToRequest, err := shouldAddAuthToRequest(helmRepoURLRegex, chart.Repo, chart.Chart)
 			if err != nil {
 				return nil, fmt.Errorf("failed to add auth to request for %s: %w", downloadChartError(*chart), err)
-			}
-			if !shouldAddAuthToRequest {
+			} else if !shouldAddAuthToRequest {
```
weyfonk: This doesn't make a difference (apart from saving us one line of code :) ), since if `err != nil`, this line will not be reached.

internal/bundlereader/auth.go
```diff
@@ -17,6 +19,33 @@ type Auth struct {
...
+		a.CABundle,
+		a.SSHPrivateKey,
+		{toByte(a.InsecureSkipVerify)},
+		{toByte(a.BasicHTTP)},
+	} {
+		hash.Write(v)
+		hash.Write(nullBytes)
+	}
+	return hex.EncodeToString(hash.Sum(nil))
+
```
0xavi0, line 56: Question: 
`nullBytes` seems to be the delimiter between fields, but technically those fields could also contain `\x00` bytes.

Could that introduce collisions?
This is the critical part of the PR IMHO, as it could potentially introduce a _hard to find_ bug...

Couldn't it be safer (although more expensive, of course) to encode a canonical JSON or CBOR and then hash?

0xavi0 (commented):
review still in process, just a question for now...

internal/bundlereader/auth.go
aruiz14: You are technically correct that could happen, especially for `CABundle` and `SSHPrivateKey` fields (using a username or password containing null would be asking for trouble, isn't it?). Also, this hash is later combined with the actual URL being downloaded, so the probability of conflict are really small.

Serializing into an intermediate format would be valid and may introduce an overhead, but this will mainly run as part of the Fleet CLI, for which this is not so critical (and probably negligible compared with the rest of work it performs). It may also prevent future mistakes, in case someone adds a new field but forgets to modify the hash method (for which I could only add a warning comment. On the other hand, if we use JSON, there would still be risk, since every field must be exported).

An good middle-term from the current proposal could be replacing the nullBytes delimiter with length-padding, which would eliminate the chance for collisions. WDYT?
EDIT: see 235f76d1

0xavi0: yup, I was thinking on the `CABundle` and `SSHPrivateKey` fields, but also for password in case it is generated and the `\x00` byte is allowed (you never know the funny things people can do :) )

I think using length-padding is safer than just `\x00` bytes, and we could even combine both. 
Length, then data, then `\x00`

Other approaches I've seen also add the data type encoded as a number and the name or number of the field. 

My main concern is that if we never have a collision and the index was supposed to be different it could be potentially hard to find in production (Fleet CLI is sometimes harder to track as it's ephemeral compared to the controllers).

The good point I see with the JSON is that the `Hash` function would remain the same. (Although there are always chances to forget exporting the new field)

aruiz14: Addressed in 09c9e861bd757b0261a2d3dc1d49e7b292d9c639

PR #4288, opened 2025-11-04
Specify version for slsactl installation

thardeck:
to prevent unexpected regressions during release.


.github/workflows/release.yml
```diff
@@ -117,6 +117,8 @@ jobs:
       - name: Install Cosign
         uses: sigstore/cosign-installer@faadad0cce49287aee09b3a48701e75088a2c6ad # v4.0.0
       - uses: rancherlabs/slsactl/actions/install-slsactl@2539cebd3fe5c291e5431f042e225c5b65850f50 # v0.1.1
```
weyfonk, line 119: Wouldn't the SHA alone prevent unwanted updates and their possible regressions?

thardeck: By default the action pulls always the latest version of the cli, independent of its own version.

weyfonk (approved):

Issue #4233, opened 2025-10-15
[SURE-10842] Add config option to configure the number of routines when creating bundles in fleet apply


0xavi0:
It is fixed to 4, currently and the code creates an additional routine. 

See https://github.com/rancher/fleet/blob/02ac73869b3ba24bf83ca2da7a8878ea7b9e4273/internal/cmd/cli/apply/apply.go#L108 

We might need to configure this to allow more or less concurrent bundle creation. 

Created after: SURE-10842

aruiz14:
## QA template

### Verify the settings are correctly configured from the `gitjob` deployment config to the Pod running the Fleet CLI

1. Set `FLEET_BUNDLE_CREATION_MAX_CONCURRENCY` in the `gitjob` deployment:
```
kubectl -n cattle-fleet-system set env deploy/gitjob FLEET_BUNDLE_CREATION_MAX_CONCURRENCY=2
```
2. Create a `GitRepo` and wait for Fleet to start monitoring it.
3. Get the logs from the Pod running the Fleet CLI (which prints some env. variables at the top). Then verify the env. variable matches the desired value:
```
❯ kubectl -n fleet-local logs -f test-33d1a-sxxqw
Defaulted container "fleet" out of: fleet, gitcloner-initializer (init)
FLEET_BUNDLE_CREATION_MAX_CONCURRENCY=2
HOSTNAME=test-33d1a-sxxqw
```
(example for a `GitRepo` created in the `fleet-local` workspace).

### Verify that Fleet CLI is effectively respects the configured max concurrency

This would actually part of the original issue that added concurrency to this code, and could probably be skipped.
Nonetheless, it should be enough to create a GitRepo with multiple Bundles and compare the execution times and CPU usage running with concurrency 1 vs. 5, for example.

sbulage:
### System Information

| Rancher Version | Fleet Version |
|--------|--------|
| v2.13-fdbd9c020abdbe37f885fd73433e5abed0e9baf7-head | 108.0.1+up0.14.1-rc.1 |

#### Followed steps mentioned in the [comment](https://github.com/rancher/fleet/issues/4233#issuecomment-3641676400).
- Set env variable `FLEET_BUNDLE_CREATION_MAX_CONCURRENCY` at gitjob or Fleet value level.
- With value set to `1`:
  - I see bundles were creating 1 by 1.
  - <details><summary><b>FLEET_BUNDLE_CREATION_MAX_CONCURRENCY=1</b></summary>
[...]

PR #4292, opened 2025-11-05
Add configurable concurrent bundle creation routines to fleet apply

thardeck:
Issue: #4233

Make the number of concurrent bundle creation routines configurable via:
- CLI flag: --bundle-creation-max-concurrency (default: 4)
- Environment variable: FLEET_BUNDLE_CREATION_MAX_CONCURRENCY

### DIFF
--- a/internal/cmd/controller/gitops/operator.go
+++ b/internal/cmd/controller/gitops/operator.go
@@ -23,7 +23,6 @@ import (
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 
 	command "github.com/rancher/fleet/internal/cmd"
-	fleetapply "github.com/rancher/fleet/internal/cmd/cli/apply"
 	"github.com/rancher/fleet/internal/cmd/controller/gitops/reconciler"
 	fcreconciler "github.com/rancher/fleet/internal/cmd/controller/reconciler"
 	"github.com/rancher/fleet/internal/metrics"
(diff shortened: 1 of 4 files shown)


0xavi0 (approved):

0xavi0:
this part lgtm... We'd also need to use the new parameter when creating the `gitjob`

internal/cmd/cli/apply/apply_concurrency_test.go
```diff
@@ -24,3 +25,71 @@ func TestGetEffectiveMaxConcurrency(t *testing.T) {
...
+			expectedError: false,
+		},
+		{
+			name:          "invalid value returns error",
+			envValue:      "not_a_number",
+			expectedValue: 4,
+			expectedError: true,
+		},
+		{
+			name:          "zero is treated as invalid by atoi",
```
weyfonk: Is it, though, if 0 (not the default value) is returned without an error?

thardeck: True, it is not an error and replaced by the default value.

internal/cmd/controller/gitops/operator.go
```diff
@@ -138,21 +139,27 @@ func (g *GitOperator) Run(cmd *cobra.Command, args []string) error {
...
+		Scheduler:                    sched,
+		Workers:                      workers,
+		ShardID:                      g.ShardID,
+		JobNodeSelector:              g.ShardNodeSelector,
+		GitFetcher:                   &git.Fetch{KnownHosts: kh},
+		Clock:                        reconciler.RealClock{},
+		Recorder:                     mgr.GetEventRecorderFor(fmt.Sprintf("fleet-gitops%s", shardIDSuffix)),
+		SystemNamespace:              namespace,
+		KnownHosts:                   kh,
+		BundleCreationMaxConcurrency: bundleCreationMaxConcurrency,
```
0xavi0: Do we need the variable to be propagated in here?

As it's an env variable that only affects the `gitjob` creation we could just simple read it when calling `argsAndEnvs`, right? 

It is just to keep things simpler.

Issue #1851, opened 2023-10-10
Unable to use `lookup` in Fleet-managed Helm template


skaven81:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

If I create a Fleet Bundle with an embedded Helm chart, and one of the templates in that Helm chart uses Helm's `lookup` feature, it does not work.  The `lookup` function returns nothing, and the template does not render as expected.

### Expected Behavior

Fleet Bundles with Helm charts that use `lookup` should function as expected.

### Steps To Reproduce

Create a simple Helm bundle:
```
$ cat Chart.yaml
apiVersion: v2
name: lookup-test
version: 1.0.0

$ cat fleet.yaml
helm:
  chart: .

$ cat templates/config.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: config-test
  namespace: kube-system
data:
  lookup-output: {{ len (lookup "v1" "Pods" "kube-system" "").items }}
```

When deployed to a cluster, the expected ConfigMap should have a number (the count of Pods in kube-system) as the value in `lookup-output`.  But instead we get an error from Fleet indicating that it was unable to parse the template:

```
    message: 'ErrApplied(10) [Bundle lookup-test-test: template: test/templates/config.yaml:7:20:
      executing "test/templates/config.yaml" at <len (lookup "v1" "Pods"
      "kube-system" "").items>: error calling len: len of nil pointer]'
```

The fact that it complains of a `nil pointer` is important -- that means that the `lookup` function either returned an empty dict, or a nil pointer itself.  If it had executed properly, even if there were no resources matching the query, `items` would have been an empty list, and `len` would have returned `0`.

### Environment

```markdown
- Architecture: amd64
- Fleet Version: 0.7.0
- Cluster:
  - Provider: RKE
  - Options:
  - Kubernetes Version: 1.26
```

skaven81:
I have not seen any solution or workaround for this.

skaven81:
I have managed to get _some_ forms of `lookup` to work.  For example, this works for me in a template that is rendered by Fleet:
```
  {{- range $ns := (lookup "v1" "Namespace" "" "").items -}}
    {{- $prjId := dig "metadata" "annotations" "field.cattle.io/projectId" "" $ns }}
    {{- if has $prjId $excludedProjectIds -}}
      {{- $excludedNamespaces = append $excludedNamespaces $ns.metadata.name -}}
    {{- end -}}
  {{- end -}}
```

[...]

PR #4302, opened 2025-11-06
Adds support for Helm lookup

0xavi0:
Adds support for using templates with the _lookup_ command from _Helm_ as it was previously not working because of the first run that we do for the _install_ command we do it in _dryRun_ mode to check for errors.

In that case the Helm command internally was not doing the call to the cluster and the _lookup_ calls where resolving to `nil`.

This PR adds a new function that checks if a given _Chart_ has templates with the _lookup_ command and, if so, it enables the `DryRunOption="server"` and also sets the option `ClientOnly` to _false_ so Helm connects the cluster and the _lookup_ does not return `nil`

Refers to: https://github.com/rancher/fleet/issues/1851

### Checklist

### DIFF
--- a/internal/helmdeployer/lookup.go
+++ b/internal/helmdeployer/lookup.go
@@ -1,6 +1,8 @@
 package helmdeployer
 
 import (
+	"reflect"
+	"strings"
 	"text/template"
 	"text/template/parse"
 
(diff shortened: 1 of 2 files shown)


internal/helmdeployer/lookup.go
```diff
@@ -0,0 +1,87 @@
...
+		return false
+	}
+
+	switch n := node.(type) {
+	case *parse.ActionNode:
+		return containsLookup(n.Pipe)
+	case *parse.IfNode:
+		return containsLookup(n.Pipe) || containsLookup(n.List) || containsLookup(n.ElseList)
+	case *parse.ListNode:
+		for _, subNode := range n.Nodes {
```
weyfonk: nit: according to my editor, this could be "simplified" using [slices.ContainsFunc](https://pkg.go.dev/slices#ContainsFunc). No strong opinion here though, since saving one line of code does not result in better readability in my opinion.

Same remark about L79 (and I'd add L66, which wasn't flagged).

internal/helmdeployer/lookup.go
```diff
@@ -0,0 +1,87 @@
...
+			}
+		}
+	}
+
+	return false
+}
+
+// containsLookup recursively checks whether a parse.Node (and its children)
+// contains a call to the "lookup" function.
+func containsLookup(node parse.Node) bool {
```
weyfonk: Note: this doesn't work with the `range` example listed [here](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-lookup-function):
```
{{ range $index, $service := (lookup "v1" "Service" "mynamespace" "").items }}
    {{/* do something with each service */}}
{{ end }}
```
Leaving the block as-is leads to a nil reference exception, whereas filling it seems to fail the test.

0xavi0: Good catch, Thank you!

The code related to node types in Helm templates is a bit tedious, in my opinion… I missed a node type that could contain subnodes: _NodeChain_.
I’ve added it and refactored the function.
Now it first checks whether the string representation of the node explicitly contains the string `lookup`. If it doesn’t, it won’t even proceed to inspect the subnodes in depth.

I've also added a unit test covering that specific expression.

internal/helmdeployer/lookup.go
0xavi0: In my humble opinion, it's already clear enough. I get the impression that it would even get more complicated by using ContainsFunc… I don’t know, I don’t really see much benefit in it.

weyfonk (approved):

internal/helmdeployer/lookup.go
weyfonk: Nice, thanks for your detailed comments too :bow:

PR #4316, opened 2025-11-10
Bump Golangci lint to 2.6.1 and add additional linters

thardeck:
fix newly reported issues.
Also add additional potentially useful linters, one per commit with its according fixes.

### DIFF
--- a/internal/cmd/cli/apply.go
+++ b/internal/cmd/cli/apply.go
@@ -2,7 +2,6 @@ package cli
 
 import (
 	"bytes"
-	"context"
 	"fmt"
 	"os"
 	"os/exec"
@@ -242,7 +241,7 @@ func (a *Apply) addAuthToOpts(opts *apply.Options, readFile readFile, helmBasicH
 }
 
 func currentCommit() string {
-	cmd := exec.CommandContext(context.Background(), "git", "rev-parse", "HEAD")
+	cmd := exec.Command("git", "rev-parse", "HEAD") //nolint:noctx // TODO: refactor to use go-git's ResolveRevision
 	buf := &bytes.Buffer{}
 	cmd.Stdout = buf
 	err := cmd.Run()


internal/cmd/cli/apply.go
```diff
@@ -241,7 +242,7 @@ func (a *Apply) addAuthToOpts(opts *apply.Options, readFile readFile, helmBasicH
 }
 
 func currentCommit() string {
-	cmd := exec.Command("git", "rev-parse", "HEAD")
+	cmd := exec.CommandContext(context.Background(), "git", "rev-parse", "HEAD")
```
weyfonk: This context will never be cancelled, so this doesn't really makes any difference.

internal/cmd/controller/agentmanagement/controllers/clusterregistrationtoken/handler.go
```diff
@@ -69,7 +69,7 @@ func Register(ctx context.Context,
 
 func (h *handler) OnChange(token *fleet.ClusterRegistrationToken, status fleet.ClusterRegistrationTokenStatus) ([]runtime.Object, fleet.ClusterRegistrationTokenStatus, error) {
 	if gone, err := h.deleteExpired(token); err != nil || gone {
-		return nil, status, nil
+		return nil, status, err
```
weyfonk: I wonder if this may have unforeseen consequences :thinking:

internal/helmdeployer/install.go
```diff
@@ -331,16 +335,20 @@ func assertRelease(rel interface{}) (*releasev1.Release, error) {
...
 	if err != nil {
-		return false, nil
+		// If the release doesn't exist, there's nothing to uninstall
+		if errors.Is(err, driver.ErrReleaseNotFound) || errors.Is(err, driver.ErrNoDeployedReleases) {
+			return false, nil
+		}
+		return false, err
 	}
-	return r.Info.Status == releasecommon.StatusUninstalling || r.Info.Status == releasecommon.StatusPendingInstall, err
+	return r.Info.Status == releasecommon.StatusUninstalling || r.Info.Status == releasecommon.StatusPendingInstall, nil
```
weyfonk, line 344: Did this line trigger a linter error? :thinking:

internal/metrics/metrics.go
```diff
@@ -84,7 +84,7 @@ func (c *CollectorCollection) Collect(ctx context.Context, obj metav1.ObjectMeta
 			if !ok {
 				msg = "unexpected error"
 			}
-			logger.Error(errors.New("error collecting metrics"), msg, r)
+			logger.Error(errors.New("error collecting metrics"), msg, "recover", r)
```
weyfonk: Not sure we'll want to expose the details here, see also [this approach](https://github.com/rancher/fleet/pull/4337/files)

aruiz14: What about:
```
if r := recover(); r != nil {
	logger.Error(errors.New("error collecting metrics"), "observed panic", "panic", r)
}
```

internal/helmdeployer/install.go
thardeck: Yes, the `nilerr` linter flagged this. The original code returned `err` which was already guaranteed to be nil at that point (all error cases are handled earlier). Changing to explicit nil makes the code clearer and satisfies the linter.

internal/cmd/cli/apply.go
thardeck: True - this is purely for noctx linter compliance. In a CLI context that runs to completion, context.Background() provides no functional benefit, but it follows the pattern of using context-aware functions.
Would you prefer a `nolint` exception?

internal/cmd/controller/agentmanagement/controllers/clusterregistrationtoken/handler.go
thardeck: I was not sure, but better to keep the previous behavior then to end up with retry loops.

internal/cmd/cli/apply.go
weyfonk: I'd be tempted to do the following:
1. Leave a `nolint` comment for now
2. In a separate patch, ideally refactor this by calling `go-git`'s [ResolveRevision](https://github.com/go-git/go-git/blob/2784eeea7f8650557d9811a95945b947df94787e/repository.go#L1577) to avoid shelling out to git if possible

PR #4326, opened 2025-11-13
 Migrate Fleet to Helm v4

thardeck:
Upgrade Helm SDK from v3 to [v4](https://github.com/helm/helm/releases/tag/v4.0.0) and adapt Fleet to [breaking API changes](https://helm.sh/docs/overview/):

**Changes:**

- Replace deprecated fields and types:
  * DryRun (bool) -> DryRunStrategy enum (None=execute, Client=dry-run)
  * Force -> ForceReplace for resource replacement
  * Atomic -> RollbackOnFailure for rollback on upgrade failure
  * Removed ClientOnly (replaced by DryRunStrategy)
  * Removed Log method - migrate to controller-runtime's context-based logging
- Add required WaitStrategy configuration (StatusWatcherStrategy/HookOnlyStrategy)
- Fix ServerSideApply behavior:
  * Set to false when TakeOwnership=true to avoid managedFields conflicts
  * Use "auto" mode for upgrades to preserve initial apply method
- Add ContentCache field to downloader.Manager (required in Helm v4)
- Update drift correction tests to reflect Helm v4's improved client-side apply
- Refactor helmdeployer install() function to reduce cyclomatic complexity

**Followup tasks:**

- Do we want to stay with the `Atomic` option for Fleet or also support the Helm naming of `RollbackOnFailure`? At the moment `Atomic` is automatically mapped.
- Do we want to stay with the `Force` option for Fleet or also support the Helm naming of `ForceReplace`? At the moment `Force` is automatically mapped.
- Do we want to support OCI digest based chart installations, e.g. `oci://registry.example.com/charts/app@sha256:abc123...`?

Refers #4351

### DIFF
--- a/internal/helmdeployer/install.go
+++ b/internal/helmdeployer/install.go
@@ -171,8 +171,24 @@ func (h *Helm) runInstall(
 	return assertRelease(rel)
 }
 
+// configureDryRunStrategy sets the DryRunStrategy based on template mode and dryRunConfig.
+// Template mode requires DryRunClient to render without cluster interaction.
+// If DryRunOption is "server", use DryRunServer to allow lookup functions to query the cluster.
+// Otherwise, use DryRunClient for client-only dry run or DryRunNone for actual execution.
+func (h *Helm) configureDryRunStrategy(dryRunCfg dryRunConfig) action.DryRunStrategy {
+	if h.template {
+		return action.DryRunClient
+	} else if dryRunCfg.DryRun {
+		if dryRunCfg.DryRunOption == "server" {
+			return action.DryRunServer
+		}
+		return action.DryRunClient
+	}
+	return action.DryRunNone
+}
+
 // configureInstallAction configures a Helm Install action with Fleet-specific options,
-// including timeout, wait strategies, and server-side apply settings.
+// including timeout, wait strategies, and dry-run configuration.
 func (h *Helm) configureInstallAction(u *action.Install, cfg *action.Configuration, releaseName, namespace string, timeout time.Duration, options fleet.BundleDeploymentOptions, pr *postRender, dryRunCfg dryRunConfig) {
 	if cfg.Capabilities != nil {
 		if cfg.Capabilities.KubeVersion.Version != "" {
@@ -196,21 +212,7 @@ func (h *Helm) configureInstallAction(u *action.Install, cfg *action.Configurati
 	u.CreateNamespace = true
 	u.Namespace = namespace
 	u.Timeout = timeout
-	// Configure dry-run strategy based on template mode and dryRunConfig.
-	// Template mode (h.template) requires DryRunClient to render without cluster interaction.
-	// If DryRunOption is "server", use DryRunServer to allow lookup functions to query the cluster.
-	// Otherwise, use DryRunClient for client-only dry run or DryRunNone for actual execution.
-	if h.template {
-		u.DryRunStrategy = action.DryRunClient
-	} else if dryRunCfg.DryRun {
-		if dryRunCfg.DryRunOption == "server" {
-			u.DryRunStrategy = action.DryRunServer
-		} else {
-			u.DryRunStrategy = action.DryRunClient
-		}
-	} else {
-		u.DryRunStrategy = action.DryRunNone
-	}
+	u.DryRunStrategy = h.configureDryRunStrategy(dryRunCfg)
 	u.SkipSchemaValidation = options.Helm.SkipSchemaValidation
 	u.PostRenderer = pr
 	u.WaitForJobs = options.Helm.WaitForJobs
@@ -283,21 +285,7 @@ func (h *Helm) configureUpgradeAction(u *action.Upgrade, namespace string, timeo
 	}
 	u.Namespace = namespace
 	u.Timeout = timeout
-	// Configure dry-run strategy based on template mode and dryRunConfig.
-	// Template mode (h.template) requires DryRunClient to render without cluster interaction.
-	// If DryRunOption is "server", use DryRunServer to allow lookup functions to query the cluster.
-	// Otherwise, use DryRunClient for client-only dry run or DryRunNone for actual execution.
-	if h.template {
-		u.DryRunStrategy = action.DryRunClient
-	} else if dryRunCfg.DryRun {
-		if dryRunCfg.DryRunOption == "server" {
-			u.DryRunStrategy = action.DryRunServer
-		} else {
-			u.DryRunStrategy = action.DryRunClient
-		}
-	} else {
-		u.DryRunStrategy = action.DryRunNone
-	}
+	u.DryRunStrategy = h.configureDryRunStrategy(dryRunCfg)
 	u.SkipSchemaValidation = options.Helm.SkipSchemaValidation
 	u.DisableOpenAPIValidation = h.template || dryRunCfg.DryRun
 	u.PostRenderer = pr


integrationtests/agent/bundle_deployment_drift_test.go
```diff
@@ -268,64 +269,30 @@ var _ = Describe("BundleDeployment drift correction", Ordered, func() {
...
+					svc, err := env.getService(svcName)
+					g.Expect(err).NotTo(HaveOccurred())
 
+					g.Expect(svc.Spec.Ports).ToNot(BeEmpty())
+					g.Expect(svc.Spec.Ports[0].Port).Should(Equal(int32(80)))
+					g.Expect(svc.Spec.Ports[0].TargetPort.IntVal).Should(BeEquivalentTo(9376))
+					g.Expect(svc.Spec.Ports[0].Name).Should(Equal("myport"))
 				}).Should(Succeed())
 
-				// TODO: This test is flaky, and the following code is commented out until the flakiness is resolved.
```
p-se, line 297: Is someone working on that issue?

internal/helmdeployer/history.go
```diff
@@ -83,17 +83,21 @@ func getReleaseNameVersionAndNamespace(bundleID, resourcesID string) (string, in
...
 	} else if err != nil {
 		return nil, err
 	}
 
-	for _, release := range releases {
+	for _, releaser := range releasers {
+		release, err := releaserToV1Release(releaser)
+		if err != nil {
+			continue
+		}
```
p-se, line 103: Should we at least log that something unexpected happened over here?

integrationtests/agent/bundle_deployment_drift_test.go
```diff
@@ -245,16 +245,17 @@ var _ = Describe("BundleDeployment drift correction", Ordered, func() {
 			})
 		})
 
-		// Helm rollback uses three-way merge by default (without force), which fails when trying to rollback a
-		// change made on an item in the ports array.
-		Context("Drift correction fails", func() {
+		// Helm v4 with client-side apply can successfully handle drift correction on service ports,
+		// even when there are complex changes. This is an improvement over Helm v3 where three-way
```
weyfonk, line 249: Here's one solid argument for upgrading to v4 if I ever saw one :D

internal/helmdeployer/capabilities.go
```diff
@@ -47,35 +49,39 @@ func getCapabilities(c action.Configuration) (*chartutil.Capabilities, error) {
 	apiVersions, err := getVersionSet(dc)
 	if err != nil {
 		if discovery.IsGroupDiscoveryFailedError(err) {
-			c.Log("WARNING: The Kubernetes server has an orphaned API service. Server reports: %s", err)
-			c.Log("WARNING: To fix this, kubectl delete apiservice <service-name>")
+			logger := log.FromContext(ctx).WithName("helm-capabilities")
+			logger.V(1).Info("WARNING: The Kubernetes server has an orphaned API service", "error", err.Error())
```
weyfonk: Since we're logging an error, I'd be tempted to use `logger.Error()` here.

internal/helmdeployer/release_conversion.go
```diff
@@ -0,0 +1,72 @@
...
+	"fmt"
+
+	"helm.sh/helm/v4/pkg/release"
+	releasev1 "helm.sh/helm/v4/pkg/release/v1"
+	"helm.sh/helm/v4/pkg/storage"
+)
+
+// releaserToV1Release converts a release.Releaser interface to a concrete v1.Release type.
+// This follows the same pattern used in Helm v4's own codebase for handling versioned releases.
+func releaserToV1Release(rel release.Releaser) (*releasev1.Release, error) {
```
weyfonk, line 13: Not sure I understand why we need this :thinking: 
Is this in anticipation for future supported release versions, e.g. enabling migrating Helm releases when updating the Fleet agent?

internal/helmdeployer/release_conversion_test.go
```diff
@@ -0,0 +1,153 @@
...
+			}
+
+			require.NoError(t, err)
+			assert.Len(t, result, tt.wantLen)
+
+			// Verify data preservation for non-nil releases
+			for i, r := range tt.input {
+				if r != nil {
+					switch rel := r.(type) {
+					case *releasev1.Release:
```
weyfonk: nit: do we still need a `switch` block here, given that a single case is supported? :thinking:
On the other hand, we'll probably want to cover input that isn't of type `*releasev1.Release` (one test case would be enough, no need to cover all of them like in `TestReleaserToV1Release`), just so that regressions are more likely to be detected if `releaseListToV1List` stops converting items from the input list.

internal/helmdeployer/rollback.go
```diff
@@ -23,40 +25,44 @@ func (h *Helm) RemoveExternalChanges(ctx context.Context, bd *fleet.BundleDeploy
...
+	r.WaitStrategy = kube.HookOnlyStrategy
 	r.Version = currentRelease.Version
 	r.MaxHistory = bd.Spec.Options.Helm.MaxHistory
 	if r.MaxHistory == 0 {
 		r.MaxHistory = MaxHelmHistory
 	}
-	if bd.Spec.CorrectDrift.Force {
-		r.Force = true
-	}
+	// Force field removed in Helm v4
```
weyfonk: `ForceReplace` is available on rollbacks though :thinking:

internal/helmupdater/helmupdater.go
```diff
@@ -54,6 +60,7 @@ func UpdateHelmDependencies(path string) error {
 				Getters:          getter.All(settings),
 				RepositoryConfig: settings.RegistryConfig,
 				RepositoryCache:  settings.RepositoryCache,
+				ContentCache:     settings.RepositoryCache, // Required in Helm v4
```
weyfonk: Why not use `settings.ContentCache`? :thinking:

p-se (approved):

internal/helmdeployer/install.go
```diff
@@ -112,16 +114,13 @@ func (h *Helm) install(ctx context.Context, bundleID string, manifest *manifest.
...
-		if err != nil {
-			return nil, err
-		}
+		mapper, _ := cfg.RESTClientGetter.ToRESTMapper()
 		pr.mapper = mapper
 	}
 
 	if install {
-		u := action.NewInstall(&cfg)
-		u.ClientOnly = h.template || (dryRunCfg.DryRun && dryRunCfg.DryRunOption == "")
```
p-se, line 124: Wonder if `h.template` should set `u.DryRunStrategy` to `action.DryRunClient`. Or perhaps it's possible to remove it and just use `dryRunCfg`. Currently it appears to be ignored in this file, which seems to change how it previously behaved.

internal/helmdeployer/deployer.go
```diff
@@ -93,6 +91,12 @@ func (h *Helm) Setup(ctx context.Context, client client.Client, getter genericcl
 	return nil
 }
 
+func (h *Helm) Setup(ctx context.Context, client client.Client, getter genericclioptions.RESTClientGetter) error {
```
weyfonk: Just curious: why split `Setup` into two methods, when `Init` is not called elsewhere?

internal/helmdeployer/deployer.go
```diff
@@ -116,60 +120,78 @@ func (h *Helm) getOpts(bundleID string, options fleet.BundleDeploymentOptions) (
...
+func (h *Helm) createCfg(ctx context.Context, namespace string) (*action.Configuration, error) {
+	// Create a logger handler for Helm SDK components.
+	// This uses Fleet's controller-runtime logger (which uses logr/zapr) and adapts it to slog.
+	// The logger level is set to V(1) to match the verbosity level used in Helm v3.
 	logger := log.FromContext(ctx).WithName("helmSDK")
-	info := func(format string, v ...interface{}) {
-		logger.V(1).Info(fmt.Sprintf(format, v...))
-	}
+	handler := slog.NewTextHandler(&logrWriter{logger: logger}, &slog.HandlerOptions{
+		Level: slog.LevelDebug,
```
weyfonk, line 160: Should this be set according to the agent's debug config, or does `LevelDebug` not play any role since `logrWriter` will only ever produce info logs?

internal/helmdeployer/install.go
```diff
@@ -112,96 +127,190 @@ func (h *Helm) install(ctx context.Context, bundleID string, manifest *manifest.
 	}
 
 	if !h.useGlobalCfg {
-		mapper, err := cfg.RESTClientGetter.ToRESTMapper()
-		if err != nil {
-			return nil, err
-		}
+		mapper, _ := cfg.RESTClientGetter.ToRESTMapper()
```
weyfonk: Why ignore possible errors here?

internal/helmdeployer/install.go
```diff
@@ -112,96 +127,190 @@ func (h *Helm) install(ctx context.Context, bundleID string, manifest *manifest.
...
-			}
-			if cfg.Capabilities.APIVersions != nil {
-				u.APIVersions = cfg.Capabilities.APIVersions
-			}
+	return pr
+}
+
+// runInstall executes a Helm install operation with the provided configuration and values.
+// It creates an Install action, configures it, and runs the installation.
+func (h *Helm) runInstall(ctx context.Context, cfg *action.Configuration, chart *chartv2.Chart, values map[string]interface{}, releaseName, namespace string, timeout time.Duration, options fleet.BundleDeploymentOptions, pr *postRender, dryRunCfg dryRunConfig) (*releasev1.Release, error) {
```
weyfonk: nit: this is a long line, it should be reformatted for easier readability, with one parameter per line.

(same remark about `runUpgrade`)

integrationtests/agent/bundle_deployment_drift_test.go
thardeck: I suppose not needed anymore, because we do not need force anymore for the drift correction and the commented out test case is directly working with Helm v4.

weyfonk (commented):
Thanks for this! :rocket:
Leaving a few comments in addition to Patrick's.

internal/helmdeployer/release_conversion.go
thardeck: I think it is required for Helm v4's `release.Releaser` interface. The storage layer returns `release.Releaser` interfaces, but Fleet needs concrete `*releasev1.Release` pointers to access release fields.

internal/helmdeployer/rollback.go
thardeck: Good point, that went by the board after the first adaption. Should be fixed in my follow up commit.

internal/helmdeployer/deployer.go
thardeck: `slog.LevelDebug` doesn't control actual filtering - it just tells slog to accept all levels. Real filtering happens in `logrWriter.Write()` using `logger.V(1).Info()`, which respects Fleet's verbosity configuration (via `-v` flag).

internal/helmdeployer/deployer.go
thardeck: Not sure, definitely makes more sense in one function.

internal/helmdeployer/deployer.go
weyfonk: That makes sense, thanks for confirming.

internal/helmdeployer/install.go
```diff
@@ -119,89 +137,219 @@ func (h *Helm) install(ctx context.Context, bundleID string, manifest *manifest.
...
+	} else if dryRunCfg.DryRun {
+		if dryRunCfg.DryRunOption == "server" {
+			u.DryRunStrategy = action.DryRunServer
+		} else {
+			u.DryRunStrategy = action.DryRunClient
 		}
-		return u.Run(chart, values)
+	} else {
+		u.DryRunStrategy = action.DryRunNone
+	}
```
weyfonk: nit: since this works the same way in both install and upgrade cases, wouldn't it make sense to isolate this logic into a separate function?

weyfonk (approved):

PR #4337, opened 2025-11-18
Fix panic due to mismatching key-value pair expected from logger

aruiz14:
I found multiple panics in the logs of a failed CI run (successful runs omit verbose logs so it was not being printed):
https://github.com/rancher/fleet/actions/runs/19454838598/job/55666698529#step:5:84:
```
"Observed a panic" panic="odd number of arguments passed as key-value pairs for logging"
```


weyfonk:
Should be addressed through #4316.

aruiz14:
> Should be addressed through #4316.

Thanks, I'll comment there and close this one.

PR #4344, opened 2025-11-19
Improve reliability and readability of integration tests

thardeck:
The testcontainers reaper (Ryuk) often causes issues in local development environments with race conditions and Docker container states. This commit disables it for integration tests that use testcontainers.

It has one [change](https://github.com/rancher/fleet/pull/4344/commits/fe13c22e2538c208650d098a7f692e301f9613f1) affecting the regular Fleet code. I do not think that this normalization incident  is an error, since it seems to be a normal use case, but let me know what you think.

Changes:
- Added init() function to disable reaper in cli/apply, gitcloner, and gitjob tests
- Fixed helm push command to use --plain-http flag for HTTP registries
- Fixed Gomega assertions to properly convert byte arrays to strings
- Improved container cleanup with DeferCleanup pattern
- Reduce log noise from JSON patch normalizer
- Specify test chart version in fleet.yaml to eliminate Helm warnings
- Suppress verbose output during test dependency updates and test executions
- Add SELinux support to gitjob and gitcloner integration tests 

Refers to #4246

### DIFF
--- a/integrationtests/gitjob/git/git_test.go
+++ b/integrationtests/gitjob/git/git_test.go
@@ -102,8 +102,21 @@ var _ = BeforeSuite(func() {
 
 	DeferCleanup(func() {
 		if tmpDir != "" {
-			err := os.RemoveAll(tmpDir)
-			Expect(err).NotTo(HaveOccurred())
+			// Attempt to fix permissions recursively. In some CI environments with
+			// user namespace remapping, files created by the container may be owned
+			// by a different UID, but we can still chmod them if we have access.
+			_ = filepath.WalkDir(tmpDir, func(path string, d os.DirEntry, err error) error {
+				if err != nil {
+					return nil // Skip paths we can't access
+				}
+				_ = os.Chmod(path, 0700)
+				return nil
+			})
+
+			// Attempt cleanup - may still fail in some CI environments
+			if err := os.RemoveAll(tmpDir); err != nil {
+				GinkgoWriter.Printf("Warning: failed to remove tmpDir %s: %v\n", tmpDir, err)
+			}
 		}
 	})
 })


internal/cmd/agent/deployer/normalizers/jsonpatch.go
```diff
@@ -36,7 +36,7 @@ func (j JSONPatchNormalizer) Normalize(un *unstructured.Unstructured) error {
 	gvk := un.GroupVersionKind()
 	metaObj, err := meta.Accessor(un)
 	if err != nil {
-		logrus.Errorf("Failed to normalize obj with json patch, error: %v", err)
+		logrus.Debugf("Failed to normalize obj with json patch, error: %v", err)
```
weyfonk: My concern here is that with debug logs disabled by default, these errors may suddenly become invisible to most users, which may result in unexpected (and now harder to troubleshoot) deployment states with bundle diffs.

integrationtests/utils/envtest.go
```diff
@@ -32,11 +35,31 @@ func init() {
 	gomega.SetDefaultEventuallyPollingInterval(PollingInterval)
 }
 
+// ShouldSuppressLogs checks if logs should be suppressed based on the VERBOSE environment variable.
+// Returns true when VERBOSE is not set (suppress logs), false when VERBOSE=1 (show all logs).
+func ShouldSuppressLogs() bool {
+	return os.Getenv("VERBOSE") == ""
```
weyfonk, line 41: We should mention this in `dev/README.md`, as it could save us future headaches ;)

integrationtests/utils/helpers.go
```diff
@@ -77,3 +78,13 @@ func ExtractResourceLogs(allLogs, resourceName string) string {
 	}
 	return strings.Join(resourceLogs, "\n")
 }
+
+// DisableReaper disables the testcontainers reaper (Ryuk) to avoid issues
+// with Docker container state in local development environments.
+// The reaper is mainly useful in CI but often causes race conditions locally.
+// This should be called in init() functions of test packages that use testcontainers.
+func DisableReaper() {
+	if os.Getenv("TESTCONTAINERS_RYUK_DISABLED") == "" {
```
weyfonk, line 87: As per [this](https://blog.worldline.tech/2023/01/04/ryuk.html) (warning from the maintainers) and [that](https://golang.testcontainers.org/features/garbage_collector/#ryuk), I am not entirely comfortable with _entirely disabling_ Ryuk; yet I've tried alternatives, such as increasing the [timeout for cleaning up containers](https://github.com/testcontainers/moby-ryuk?tab=readme-ov-file#ryuk-configuration) from its default value (`10s`) to `5m`, which didn't change anything.

integrationtests/gitjob/git/git_test.go
```diff
@@ -62,309 +65,373 @@ var (
...
+	Expect(err).NotTo(HaveOccurred())
+
+	container, url, err = createGogsContainer(ctx, tmpDir)
+	Expect(err).NotTo(HaveOccurred())
+})
+
+var _ = AfterSuite(func() {
+	if container != nil {
+		ctx := context.Background()
+		err := container.Terminate(ctx)
```
weyfonk: Do we really want to eliminate the option of gathering container logs?
They could be useful for troubleshooting, especially on CI runners.

integrationtests/gitjob/git/git_test.go
```diff
@@ -62,309 +65,373 @@ var (
...
+		})
 
-	tests := map[string]struct {
-		gitrepo        *v1alpha1.GitRepo
-		expectedCommit string
-		expectedErr    error
-	}{
-		"public repo": {
-			gitrepo: &v1alpha1.GitRepo{
+		It("public repo", func() {
```
weyfonk: I prefer block statements reading like sentences, such as:
```suggestion
		It("fetches the latest commit from a public repo", func() {
```

(same remark about other test cases throughout this file)

integrationtests/gitjob/git/git_test.go
```diff
@@ -62,309 +65,373 @@ var (
...
-				require.NoError(t, err)
-			} else {
-				require.Contains(t, err.Error(), test.expectedErr.Error())
-			}
-			if latestCommit != test.expectedCommit {
-				t.Errorf("latestCommit doesn't match. got %s, expected %s", latestCommit, test.expectedCommit)
-			}
+			latestCommit, err := f.LatestCommit(ctx, gitrepo, client)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring(transport.ErrAuthenticationRequired.Error()))
```
weyfonk: This would be more idiomatic:
```suggestion
			Expect(err).To(MatchError(transport.ErrAuthenticationRequired)) 
```
(same remark about other `ContainSubstring` checks on errors throughout this file)

integrationtests/gitjob/git/git_test.go
```diff
@@ -62,309 +65,373 @@ var (
...
-		t.Run(name, func(t *testing.T) {
+		BeforeEach(func() {
+			ctx := context.Background()
+			port, err := container.MappedPort(ctx, "22")
+			Expect(err).NotTo(HaveOccurred())
+			sshPort = port.Port()
+		})
+
+		// Set up SSH key once per describe block
+		BeforeEach(func() {
```
weyfonk: With a `BeforeEach` block, SSH keys will be set up before running each test case.
The behaviour is correct with the check on empty private keys below, but the comment is misleading ;)

integrationtests/gitjob/git/git_test.go
```diff
@@ -62,309 +65,373 @@ var (
...
-			},
-			expectedCommit: latestCommitPublicRepo,
-			expectedErr:    nil,
-		},
-		"private repo": {
-			gitrepo: &v1alpha1.GitRepo{
+			}
+
+			f := fleetgit.Fetch{}
+			client := mocks.NewMockK8sClient(ctlr)
```
weyfonk: nit: if we don't want to repeat this block, [Table Specs](https://onsi.github.io/ginkgo/#table-specs) are a thing ;)

integrationtests/gitjob/git/git_test.go
```diff
@@ -62,309 +65,373 @@ var (
...
 	var err error
-	t := &testing.T{}
-	container, url, err = createGogsContainer(ctx, createTempFolder(t))
-	require.NoError(t, err, "creating gogs container failed")
-
-	return func(code int) {
-		terminateContainer(ctx, t, container, code)
-	}
-}
+	tmpDir, err := os.MkdirTemp("", "gogs")
```
weyfonk: This directory would not be cleaned up after test suite execution.

integrationtests/utils/helpers.go
thardeck: Yes, some OCI tests still fail on my system with Ryuk enabled even with all other fixes.
Cleanup of containers should happen by `DeferCleanup` and we also have some additional log gathering.

integrationtests/gitjob/git/git_test.go
```diff
@@ -60,311 +64,343 @@ var (
...
+
+	container, url, err = createGogsContainer(ctx, tmpDir)
+	Expect(err).NotTo(HaveOccurred())
+
+	// Register cleanup handlers in reverse order (tmpDir last)
+	DeferCleanup(func() {
+		if tmpDir != "" {
+			// Container may create files with different permissions (e.g., Gogs as non-root user).
+			// We attempt cleanup but don't fail the test if we can't delete due to permission issues.
+			// The test container framework will eventually clean up the temp directory.
```
weyfonk: That wouldn't happen for a tmp dir created through `os.MkdirTemp` though, would it?

integrationtests/gitjob/git/git_test.go
```diff
@@ -60,311 +64,343 @@ var (
...
+				f := fleetgit.Fetch{}
+				client := mocks.NewMockK8sClient(ctlr)
+				// May be called multiple times, including calls to get Rancher CA bundle secrets
+				client.EXPECT().Get(ctx, gomock.Any(), gomock.Any()).
+					Return(apierrors.NewNotFound(schema.GroupResource{}, "notfound")).AnyTimes()
+				latestCommit, err := f.LatestCommit(ctx, gitrepo, client)
+
+				if tc.expectError {
+					Expect(err).To(HaveOccurred())
+					Expect(err).To(MatchError(ContainSubstring(transport.ErrAuthenticationRequired.Error())))
```
weyfonk:
```suggestion
					Expect(err).To(MatchError(transport.ErrAuthenticationRequired))
```

integrationtests/gitjob/git/git_test.go
```diff
@@ -60,311 +64,343 @@ var (
...
-			expectedCommit: "",
-			expectedError:  errors.New("commit not found for revision: v10.0.0"),
-		},
-		"RevisionIsACommit": {
-			gitrepo: &v1alpha1.GitRepo{
+			}
+
+			latestCommit, err := f.LatestCommit(ctx, gitrepo, client)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("commit not found for revision: v10.0.0"))
```
weyfonk:
```suggestion
			Expect(err).To(MatchError(errors.New("commit not found for revision: v10.0.0")))
```

integrationtests/gitjob/git/git_test.go
```diff
@@ -60,311 +64,343 @@ var (
...
+		})
 
-func TestLatestCommit_BasicAuth(t *testing.T) {
-	ctlr := gomock.NewController(t)
-	defer ctlr.Finish()
-	ctx := context.Background()
+		type testCase struct {
+			repoPath          string
+			expectError       bool
+			getExpectedCommit func() string
```
weyfonk, line 142: nit: why do we need a function here, as opposed to a simple `string` value (which may be empty)?

integrationtests/gitjob/git/git_test.go
```diff
@@ -60,311 +64,343 @@ var (
...
+					getExpectedCommit: func() string { return latestCommitPrivateRepo },
+				}),
+			Entry("fails to fetch from a private repo with SSH when known host URL is wrong",
+				testCase{
+					repoPath: "/test/private-repo",
+					knownHostsData: func() []byte {
+						return []byte("doesntexist " + gogsFingerPrint)
+					},
+					getExpectedCommit: func() string { return latestCommitPrivateRepo },
+				}),
```
weyfonk: none of these tests cases expects failures though, despite their descriptions :thinking:

integrationtests/gitjob/git/git_test.go
```diff
@@ -60,311 +64,343 @@ var (
...
+		})
 
-	tests := map[string]struct {
-		gitrepo        *v1alpha1.GitRepo
-		expectedCommit string
-		expectedError  error
-	}{
-		"RevisionNotFound": {
-			gitrepo: &v1alpha1.GitRepo{
+		It("revision not found", func() {
```
weyfonk:
```suggestion
		It("fails to fetch the latest commit for a nonexistent revision", func() {
```

integrationtests/gitjob/git/git_test.go
```diff
@@ -60,311 +64,343 @@ var (
...
-			gitrepo: &v1alpha1.GitRepo{
+			}
+
+			latestCommit, err := f.LatestCommit(ctx, gitrepo, client)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("commit not found for revision: v10.0.0"))
+			Expect(latestCommit).To(BeEmpty())
+		})
+
+		It("revision is a commit", func() {
```
weyfonk:
```suggestion
		It("succeeds when the revision is a commit", func() {
```

integrationtests/gitjob/git/git_test.go
```diff
@@ -60,311 +64,343 @@ var (
...
-		"RevisionIsALightweightTag": {
-			gitrepo: &v1alpha1.GitRepo{
+			}
+
+			latestCommit, err := f.LatestCommit(ctx, gitrepo, client)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(latestCommit).To(Equal("319e76a30f012a760aa7f35d125a4eca8a2c8ba2"))
+		})
+
+		It("revision is a lightweight tag", func() {
```
weyfonk:
```suggestion
		It("succeeds when the revision is a lightweight tag", func() {
```

integrationtests/gitjob/git/git_test.go
```diff
@@ -60,311 +64,343 @@ var (
...
-		"RevisionIsAnAnnotatedTag": {
-			gitrepo: &v1alpha1.GitRepo{
+			}
+
+			latestCommit, err := f.LatestCommit(ctx, gitrepo, client)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(latestCommit).To(Equal(tagCommit0))
+		})
+
+		It("revision is an annotated tag", func() {
```
weyfonk:
```suggestion
		It("succeeds when the revision is an annotated tag", func() {
```

integrationtests/gitjob/git/git_test.go
thardeck: The `getExpectedCommit func() string` is required because:
- TableSpec `Entry()` parameters are evaluated at test registration time (when Go initializes the test suite)
- `latestCommitPublicRepo` and `latestCommitPrivateRepo` are set in `BeforeSuite` which runs at test execution time
- If we used a string directly, it would capture an empty value at registration time

I have added a sanity check which makes sure that `latestCommitPublicRepo` and `latestCommitPrivateRepo` is not empty after initialization.

weyfonk: I see, thanks for clarifying :bow:

PR #4349, opened 2025-11-21
Fix remaining linter warnings and add gocritic

thardeck:
There were a few remaining linter warnings and I have added gocritic on top.

### DIFF
--- a/internal/cmd/controller/gitops/reconciler/gitjob.go
+++ b/internal/cmd/controller/gitops/reconciler/gitjob.go
@@ -642,7 +642,7 @@ func readIntEnvVar(logger logr.Logger, getter func() (int, error), envVarName st
 func argsAndEnvs(
 	gitrepo *v1alpha1.GitRepo,
 	logger logr.Logger,
-	caCertsPathOverride string,
+	pathOverrideCACerts string,
 	knownHosts KnownHostsGetter,
 	drivenScanSeparator string,
 	helmInsecureSkipTLS bool,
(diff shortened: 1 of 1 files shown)


benchmarks/cmd/dataset.go
```diff
@@ -88,12 +88,12 @@ func calculate(sample *parser.Sample, dsPop Dataset, scores scoresByXP) {
 			}
 
 			if _, ok := sample.Experiments[experiment]; !ok {
-				//fmt.Printf("missing experiment %s\n", name)
+				// fmt.Printf("missing experiment %s\n", name)
```
weyfonk, line 91: ... do we really care about a diff like this file's?

thardeck: I had the same first impression as you, but then I thought it does not make sense to add a `//nolint` statement for them.
And on top we usually should not comment out code anyway like this, we are using version control and have a history.

internal/cmd/controller/gitops/reconciler/gitjob.go
```diff
@@ -641,7 +642,7 @@ func readIntEnvVar(logger logr.Logger, getter func() (int, error), envVarName st
 func argsAndEnvs(
 	gitrepo *v1alpha1.GitRepo,
 	logger logr.Logger,
-	CACertsPathOverride string,
+	caCertsPathOverride string,
```
weyfonk: Please let me know if I have misunderstood something, but as per the result of our discussion:
```suggestion
	pathOverrideCACerts string,
```

thardeck: No, you understood it correctly. It seems that I missed it in the relatively big diff, when I searched for those items.

For example there were previously introduced local variable names starting with `ca...` which we do not want to change I suppose for now, since they were not affected by this pr directly.

Issue #4251, opened 2025-10-22
Bundle controller hits massive update conflicts removing finalizers from shared content resources at scale


manno:
Noticed this during my last scaling experiment.

When a single content resource is shared by a large number of bundledeployments (e.g., thousands), the bundle reconciler runs into a race condition when trying to remove finalizers from that shared content resource.

This results in a constant stream of update conflicts. The reconciler's exponential backoff kicks in (?), and the queue effectively grinds to a halt. I was seeing only a handful of reconciles completing every few minutes.
This stops all reconciles for that bundle of course. So it probably affects not only deletion, but also updates.

This is the spot in the code I was looking at: https://github.com/manno/fleet/blob/a3c2ba411c3f284f6140cdeb8c82861ca61eb106/internal/cmd/controller/reconciler/bundle_controller.go#L495-L497

Restarting the fleet-controller pod fixes the issue almost immediately.
We probably need to find a way to make this finalizer removal less "chatty" or handle the fan-out scenario more gracefully.

0xavi0:
This will be implemented with https://github.com/rancher/fleet/issues/4192 as they are related.

0xavi0:
## QA Template.

Please refer to https://github.com/rancher/fleet/issues/4192#issuecomment-3596665898

mmartin24:
Tested [here](https://github.com/rancher/fleet/issues/4192#issuecomment-3698761631) with `0.15.0-alpha.4` using:

```
NUM_CLUSTERS=${NUM_CLUSTERS:-10}
NUM_GITREPOS=${NUM_GITREPOS:-1}
BUNDLES_PER_GITREPO=10
```

Repeated same test successfully with:

```
NUM_CLUSTERS=${NUM_CLUSTERS:-50}
NUM_GITREPOS=${NUM_GITREPOS:-3}
BUNDLES_PER_GITREPO=20
```

Clossing issue

Issue #4192, opened 2025-10-03
Improve UX around deletion of GitRepos


weyfonk:
Follow-up to #4143.

Acceptance criteria:
* A GitRepo is not deleted (in particular, not deleted from the Rancher UI's perspective) until its owned resources (bundles, and transitively bundle deployments) have been deleted.
* Removing finalizers should not output as many errors (conflicts) as is the case as per #4143.

Possible starting points:
* set `OwnerReferences` on a bundle deployment, pointing to its bundle of origin, when [creating or updating](https://github.com/rancher/fleet/blob/main/internal/cmd/controller/reconciler/bundle_controller.go#L517) the bundle deployment
    * including `BlockOwnerDeletion` set to `true`
* set that ownership on the bundle [when creating it](https://github.com/rancher/fleet/blob/main/internal/cmd/cli/apply/apply.go#L417-L425), pointing to the GitRepo, in a way which covers all bundles, regardless of whether they use OCI storage.

0xavi0:
## QA Template

When deleting `GitRepos` with many `Bundles` and `BundleDeployments` we could observe a few things:

* When deleting a `GitRepo`, it  disappears in the UI (it is actually deleted) before all its dependant `Bundles` are deleted
* The could observe conflicts when creating or deleting many `BundleDeployment` that are sharing the same `Content` resource

Testing this implies testing Fleet as scale. A few clusters, for being able to test many `BundleDeployments` and `Bundles`

As no real deployment is needed from the perspective of the upstream cluster, you can use the attached script as an example.

[run-fleet-cycle-tests.sh](https://github.com/user-attachments/files/23854711/run-fleet-cycle-tests.sh)

It creates 50 `Cluster` resources (not real downstream `Clusters`, but that's enough for the test), and deploys 60 `Bundles` to all the `Clusters`.
That means creating 3000 `BundleDeployments`, 3000 `BundleDeployment` secrets and also 60 `Bundle` secrets.
The script deploys everything, wait until all the resources are found and then deletes the `GitRepo`. 
It waits until all the resources are deleted (including the `GitRepo`) and prints the time stamp when the `GitRepo` is found as deleted and when all the dependent resources are deleted.

The goal would be to test something similar and check that:

* The `Gitrepo` is not deleted BEFORE the dependent resources are deleted
* There are no error messages like the following one:
```
[...]

PR #4361, opened 2025-11-24
Change GitRepo deletion.

0xavi0:
This PR introduces changes to how GitRepos are deleted, as well as their dependent resources.

Two main changes are introduced:

* The function that deletes the GitRepo has been modified to check for any pending resources, especially *Bundles* and *ImageScans*, since previously the *GitRepo* was removed before these resources were fully deleted.

* A new controller is introduced for *Content* resources to track the number of *BundleDeployments* that reference the same one. This controller aims to eliminate the conflicts that occurred when updating or deleting *BundleDeployments*, as they were being updated or removed concurrently while modifying a shared *Content* resource.

With these changes, the goal is to eliminate both update conflicts (which triggered additional Reconcile calls in the controllers) and potential race conditions when deleting and recreating *GitRepos* that still had dependent resources pending deletion.

Refers to https://github.com/rancher/fleet/issues/4192
Refers to https://github.com/rancher/fleet/issues/4251

### Checklist

### DIFF
--- a/internal/cmd/controller/reconciler/content_controller_test.go
+++ b/internal/cmd/controller/reconciler/content_controller_test.go
@@ -46,6 +46,13 @@ var _ = Describe("ContentReconciler", func() {
 			Expect(reconciler.mapBundleDeploymentToContent(ctx, bd)).To(BeNil())
 		})
 
+		It("returns nil for BundleDeployment with a label using an empty name", func() {
+			reconciler := &ContentReconciler{}
+			bd := &fleet.BundleDeployment{}
+			bd.Labels = map[string]string{fleet.ContentNameLabel: ""}
+			Expect(reconciler.mapBundleDeploymentToContent(ctx, bd)).To(BeNil())
+		})
+
 		It("maps BundleDeployment with label to a single content request", func() {
 			reconciler := &ContentReconciler{}
 			bd := &fleet.BundleDeployment{}


pkg/apis/fleet.cattle.io/v1alpha1/content_types.go
```diff
@@ -27,10 +27,18 @@ type Content struct {
 	// The bundle resources are copied into the bundledeployment's content
 	// resource, so the downstream agent can deploy them.
 	// +nullable
-	Content []byte `json:"content,omitempty"`
+	Content []byte `json:"content,omitempty"` // Content is a byte array, which contains the manifests of a bundle. The bundle resources are copied into the bundledeployment's content resource, so the downstream agent can deploy them.
```
weyfonk: Is this second sentence needed here? It already appears on L27-28.

internal/cmd/controller/root.go
```diff
@@ -134,6 +135,14 @@ func (f *FleetController) Run(cmd *cobra.Command, args []string) error {
 		workersOpts.Schedule = w
 	}
 
+	if d := os.Getenv("CONTENT_RECONCILER_WORKERS"); d != "" {
```
weyfonk, line 138: This env var should probably be set in the `fleet-controller` deployment.

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -600,6 +629,41 @@ func (r *GitJobReconciler) managePollingJob(logger logr.Logger, gitrepo v1alpha1
...
+		},
+	); err != nil {
+		return nil, err
+	}
+	return list, nil
+}
+
+func GetNamespaceAndNameID(namespace, name string) string {
+	return fmt.Sprintf("%s/%s", namespace, name)
+}
```
weyfonk: If we get rid of the commented out code above, then this function is no longer needed.

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -837,3 +901,27 @@ func getNextCommit(status v1alpha1.GitRepoStatus) string {
...
+		}
+
+		if err := c.Delete(ctx, o); err != nil {
+			errs = append(errs, err)
+		}
+
+		return nil // continue iterating no matter what
+	})
+
+	return errors.Join(errs...)
```
weyfonk, line 914: This will format errors with a newline between each error message (skipping `nil` errors). I'm not sure how well that would work in the Rancher UI.
This may not be critical, but perhaps something to test.

internal/cmd/controller/operator.go
```diff
@@ -184,6 +187,22 @@ func start(
...
+		return err
+	}
+
+	if err = (&reconciler.ContentReconciler{
+		Client:  mgr.GetClient(),
+		Scheme:  mgr.GetScheme(),
+		ShardID: shardID,
+		Workers: workersOpts.Content,
+	}).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to content controller", "controller", "Content")
```
weyfonk: This typo is very relatable :D
```suggestion
		setupLog.Error(err, "unable to create controller", "controller", "Content")
```

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -335,6 +335,15 @@ func (r *BundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		// Changes in the values hash trigger a bundle deployment reconcile.
 		bd.Spec.ValuesHash = valuesHash
 
+		// When content resources are stored in etcd, we need to keep track of the content resource so they
+		// are properly gargabe collected by the content controller.
```
```suggestion
		// are properly garbage-collected by the content controller.
```

internal/cmd/controller/reconciler/content_controller.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+//+kubebuilder:rbac:groups=fleet.cattle.io,resources=contents,verbs=get;list;watch;update;patch;delete
+//+kubebuilder:rbac:groups=fleet.cattle.io,resources=contents/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=fleet.cattle.io,resources=bundledeployments,verbs=get;list;watch
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *ContentReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		Named("content-controller").
```
weyfonk: The docs for this state:
> By default, controllers are named using the lowercase version of their kind.

Except for a few (drift, gitrepo status and helmops status), we don't name our reconcilers. We may want to harmonise their names for metrics at some point.

In the meantime, for consistency, I'd be tempted to set:
```suggestion
		Named("content"). // equivalent to the default, as per the docs, so we might as well skip this line
```
or
```suggestion
		Named("content-reconciler"). // same format as in the drift controller
```

That being said, I'm not sure how useful metrics for this controller would be, so that's definitely not a major issue ;)

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -457,45 +463,11 @@ func (r *BundleReconciler) createBundleDeployment(
 	ctx context.Context,
 	l logr.Logger,
 	bd *fleet.BundleDeployment,
-	contentsInOCI bool,
-	contentsInHelmChart bool,
-	manifestID string,
 ) (controllerutil.OperationResult, *fleet.BundleDeployment, error) {
 	logger := l.WithValues("deploymentID", bd.Spec.DeploymentID)
 
-	// When content resources are stored in etcd, we need to add finalizers.
```
weyfonk, line 466: We will want to update unit tests for this controller (e.g. [here](https://github.com/rancher/fleet/blob/main/internal/cmd/controller/reconciler/bundle_controller_test.go#L452)), to reflect the fact that content management is now owned by another controller.

internal/cmd/controller/reconciler/content_controller.go
```diff
@@ -0,0 +1,136 @@
...
+		logger.V(1).Info("Updating Content reference count", "oldCount", content.Status.ReferenceCount, "newCount", newReferenceCount)
+		orig := content.DeepCopy()
+		content.Status.ReferenceCount = newReferenceCount
+		if err := r.Status().Patch(ctx, content, client.MergeFrom(orig)); err != nil {
+			logger.Error(err, "Failed to update Content reference count status")
+			return ctrl.Result{}, err
+		}
+	}
+
+	return ctrl.Result{}, nil
```
weyfonk, line 113: It could be nice to (unit-)test this, just to be on the safe side. WDYT?

pkg/apis/fleet.cattle.io/v1alpha1/content_types.go
0xavi0: yup, deleted

internal/cmd/controller/root.go
0xavi0: Added to the deployment and values. Thanks for caching it.

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
0xavi0: good catch :).

All deleted.

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
0xavi0: It is the same in the _BundleController_ in the  [batchDeleteBundleDeployments](https://github.com/rancher/fleet/blob/b964a38e72aff2d37379f7f080a4d80812bf2681/internal/cmd/controller/reconciler/bundle_controller.go#L886:L904) function.
I did it like this (with a more generic function) for consistency. 
Also, this error should be internal, because we're not propagating it to the resource conditions, so I don't think the UI will show it, right?

internal/cmd/controller/reconciler/content_controller.go
0xavi0: I've changed the name to simply "content" and I've added a comment explaining why the controller must be named. 
In all those cases we're naming because those controllers don't have a `For()` statement. 
If we don't use the `For()` statement nor `Name` the controller, controller-runtime complains.
(In fact I think it gets the name of the controller from the resource type used in the `For()` statement.

internal/cmd/controller/reconciler/bundle_controller.go
0xavi0: I deleted the comments about Contents in that file (I'm not sure if that's what you mean)

internal/cmd/controller/reconciler/content_controller.go
0xavi0: Yeah, I should add them in the first place, thanks for the suggestion.
Added now.

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
weyfonk: Oh, you're right, this will not be propagated to conditions as it's part of the deletion flow, so we should be fine.

internal/cmd/controller/reconciler/bundle_controller.go
weyfonk: I can still see this typo ;)

internal/cmd/controller/reconciler/bundle_controller.go
weyfonk: Yes, all I had seen were comments stating that we were not expecting Content resource creation/updates, there may not have been more. Thanks!

internal/cmd/controller/reconciler/content_controller_test.go
```diff
@@ -0,0 +1,210 @@
...
+		It("returns nil for non-BundleDeployment objects", func() {
+			reconciler := &ContentReconciler{}
+			Expect(reconciler.mapBundleDeploymentToContent(ctx, &fleet.Content{})).To(BeNil())
+		})
+
+		It("returns nil for BundleDeployment without content label", func() {
+			reconciler := &ContentReconciler{}
+			bd := &fleet.BundleDeployment{}
+			Expect(reconciler.mapBundleDeploymentToContent(ctx, bd)).To(BeNil())
+		})
```
weyfonk, line 47: What happens if the label is present but has an empty value? ;)
(we may want to validate that we don't create a request with an empty name in that case)

internal/cmd/controller/reconciler/content_controller.go
weyfonk: Nice, thanks for clarifying and commenting that :bow:

internal/cmd/controller/reconciler/bundle_controller.go
0xavi0: I've searched at the whole project and I can't find it.

internal/cmd/controller/reconciler/content_controller_test.go
0xavi0: case added

internal/cmd/controller/reconciler/bundle_controller.go
weyfonk: It still says "gargabe" instead of "garbage".

0xavi0: will fix in a future PR, sorry for the confusion

PR #4367, opened 2025-11-25
Sort AgentTolerations when creating agent Bundle

0xavi0:
This is a follow-up to: https://github.com/rancher/rancher/pull/52755 , merged in Rancher recently and to avoid the problem when running Fleet standalone.

It simply sorts the AgentTolerations when creating the agent Bundle so they don't differ and create later different _Content_ resources.

It also adds a unit test to test the _Bundle_'s tolerations sorting, that could be used as the base to add more tests about the generated _Bundle_ in the future.

### Checklist

### DIFF
--- a/internal/cmd/controller/agentmanagement/controllers/manageagent/manageagent_test.go
+++ b/internal/cmd/controller/agentmanagement/controllers/manageagent/manageagent_test.go
@@ -1,7 +1,6 @@
 package manageagent
 
 import (
-	"encoding/json"
 	"reflect"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


internal/cmd/controller/agentmanagement/controllers/manageagent/manageagent_test.go
```diff
@@ -107,6 +118,89 @@ func TestOnClusterChangeAffinity(t *testing.T) {
...
+		}
+		if kind, _ := m["kind"].(string); kind == "Deployment" {
+			js, err := yaml.YAMLToJSON([]byte(d))
+			if err != nil {
+				t.Fatalf("failed to convert YAML to JSON: %v", err)
+			}
+			var dep appsv1.Deployment
+			if err := json.Unmarshal(js, &dep); err != nil {
+				t.Fatalf("failed to unmarshal deployment json: %v", err)
+			}
```
weyfonk: nit: why can't we directly unmarshal the deployment YAML? Is JSON necessary here?

weyfonk (approved):

internal/cmd/controller/agentmanagement/controllers/manageagent/manageagent_test.go
0xavi0: yeah, no need to do the json part... I've deleted that extra step

Issue #4274, opened 2025-10-29
valuesFrom: Namespace assignment does not match downstream target namespace


weyfonk:
Follow-up to #3617.

The Rancher UI allows users to select resources (secrets or config maps) to reference through `valuesFrom`.
Selecting a resource is done from a combination of key type (config map or secret reference), name of the corresponding object, and key within that object.
As a result, a `configMapKeyRef` or `secretKeyRef` is added to a HelmOp or GitRepo definition.

Before #3617,  resources referenced through `valuesFrom` were expected to be manually created on each downstream cluster, but that feature eliminates that need.
The `configMapKeyRef` or `secretKeyRef` is created with the namespace of the source, upstream resource, which Fleet will take care of copying if that resource is referenced through `downstreamResources`.
This means that the namespace in the `configMapKeyRef` or `secretKeyRef` may well be different to the target namespace for the deployment on a downstream cluster.
As a result, at deployment time, the agent is unable to deploy a HelmOp referencing a config map (or secret) for its Helm values, as it looks for the resource in a namespace which is different to the target namespace, where the resource had been copied downstream.

The current workaround for this consists in editing the `configMapKeyRef` or `secretKeyRef` to remove the `namespace` field.
Here are two possible options for a more sustainable solution:
* the UI could check, when writing a `configMapKeyRef` or `secretKeyRef` in a HelmOp/GitRepo, if the corresponding resource is also referenced by `downstreamResources`; if so, it could leave the namespace field empty
* the Fleet agent could overwrite that namespace field in `configMapKeyRef`/`secretKeyRef`, in the bundle deployment it is about to deploy, for resources also referenced by `downstreamResources`.
    * This may lead to misleading/mismatching namespaces in `configMapKeyRef`s/`secretKeyRef`s between HelmOps/GitRepos and their bundle deployments, but we could probably live with that, as those namespaces would not be realistic anyway if the target deployment (downstream) namespace does not match the upstream namespace of the GitRepo/HelmOp/bundle.

sbulage:
### System Information
| Rancher Version | Fleet Version |
|--------|--------|
| v2.14-8fbd6fe7a298c1aac9edb35a6b69ddbd2576f6b1-head | 109.0.0+up0.15.0-alpha.4 | 

#### I am able to verify the case insensitive of `ConfigMap` and `Secret` as kind in `downstreamResources`, see below screenshots.

| kind: ConfigMap | kind: Secret | Worked |
|--------|--------|--------|
[...]

PR #4373, opened 2025-11-26
Use the defaultNs for valuesFrom contained in downstream resources

0xavi0:
This PR adds a change to avoid having a "Not Found" error when using valuesFrom coming from COnfigMaps or Secrets that are also contained in the _DownstreamResources_ of the _BundleDeploymentOptions_

Refers to: https://github.com/rancher/fleet/issues/4274
### Checklist

### DIFF
--- a/internal/helmdeployer/install.go
+++ b/internal/helmdeployer/install.go
@@ -520,20 +520,17 @@ func getDryRunConfig(chart *chartv2.Chart, dryRun bool) dryRunConfig {
 	return cfg
 }
 
-// isInDownstreamResources returns true and the matching DownstreamResource
-// when a resource with the provided name exists in the provided
-// BundleDeploymentOptions.DownstreamResources slice. If not found, returns
-// false and the zero value for fleet.DownstreamResource.
+// isInDownstreamResources returns true when a resource with the
+// provided name exists in the provided BundleDeploymentOptions.DownstreamResources slice.
+// If not found, returns false.
 func isInDownstreamResources(resourceName, kind string, options fleet.BundleDeploymentOptions) bool {
 	if !experimental.CopyResourcesDownstreamEnabled() {
 		return false
 	}
 
 	for _, dr := range options.DownstreamResources {
-		if dr.Name == resourceName {
-			if dr.Kind == kind {
-				return true
-			}
+		if dr.Name == resourceName && dr.Kind == kind {
+			return true
 		}
 	}
 	return false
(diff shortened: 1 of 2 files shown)


internal/helmdeployer/install.go
```diff
@@ -514,3 +519,22 @@ func getDryRunConfig(chart *chartv2.Chart, dryRun bool) dryRunConfig {
 
 	return cfg
 }
+
+// isInDownstreamResources returns true and the matching DownstreamResource
```
weyfonk:
```suggestion
// isInDownstreamResources returns true
```

internal/helmdeployer/install.go
```diff
@@ -514,3 +519,22 @@ func getDryRunConfig(chart *chartv2.Chart, dryRun bool) dryRunConfig {
...
+// when a resource with the provided name exists in the provided
+// BundleDeploymentOptions.DownstreamResources slice. If not found, returns
+// false and the zero value for fleet.DownstreamResource.
```
weyfonk:
```suggestion
// false.
```

internal/helmdeployer/install.go
```diff
@@ -514,3 +519,22 @@ func getDryRunConfig(chart *chartv2.Chart, dryRun bool) dryRunConfig {
...
+func isInDownstreamResources(resourceName, kind string, options fleet.BundleDeploymentOptions) bool {
+	if !experimental.CopyResourcesDownstreamEnabled() {
+		return false
+	}
+
+	for _, dr := range options.DownstreamResources {
+		if dr.Name == resourceName {
+			if dr.Kind == kind {
```
weyfonk: nit: this could be a single condition:
```suggestion
		if dr.Name == resourceName && dr.Kind == kind {
```

internal/helmdeployer/install_test.go
```diff
@@ -453,3 +462,145 @@ func TestCorrectDriftForceOption(t *testing.T) {
...
+	a.Equal("cmDefault", vals["cmVal"])
+	a.Equal("secDefault", vals["secVal"])
+}
+
+func TestValuesFromUsesProvidedNamespaceWhenNotCopiedDownstream(t *testing.T) {
+	a := assert.New(t)
+	r := require.New(t)
+
+	scheme := kruntime.NewScheme()
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
```
weyfonk, line 551: It could make sense to enable the experimental feature in this test case as well, just to be on the safe side and ensure that it does not overlap with the next one.

weyfonk (commented):
Nice one, thanks! Leaving a few comments, the rest LGTM :)

internal/helmdeployer/install.go
0xavi0: oups!, leftovers... Now deleted

internal/helmdeployer/install_test.go
0xavi0: done

Issue #4030, opened 2025-08-20
CLI Troubleshooting and Diagnostics


manno:
**As an** On-call Engineer or Developer, **I want to** run a comprehensive diagnostic CLI command that helps troubleshoot common issues and generate information for support tickets.

First and foremost, the CLI should gather relevant diagnostic information, like resource statuses, Fleet's k8s events and logs.

Additional, detailed diagnostics could check agent registration, resource dependencies, and communication problems, so that root causes can be identified.

**Acceptance Criteria:**

* Ability to generate a diagnostic bundle (logs, resource YAMLs) for **offline analysis and sharing with support**.
  * https://fleet.rancher.io/troubleshooting#where-to-look-for-root-causes-of-issues
  * the bundle should snapshot fleet resources over a period of 30s (?)
  * k8s events
  * also dump the metrics
  * idea: instead of dumping _all_ resources, which may result in massive volumes of data for larger setups, the CLI could accept an entrypoint resource (e.g. GitRepo, HelmOp or event a bundle) and navigate through its ownership graph, dumping all relevant resources along the way: BDs and contents, secrets, etc

* A `fleet troubleshoot` or `fleet doctor` command acts as the main entry point to **validate the current cluster**.  
  * Fetch metrics and analyze queue values (are queues healthy)?
  * provide a summary of checks performed (pass/fail), highlights potential issues, and suggests corrective actions or further commands for deeper investigation (e.g., "try `kubectl logs -n cattle-fleet-system fleet-agent-<pod-id>` on cluster X").

PR #4374, opened 2025-11-26
Add dump command to CLI

weyfonk:
This patch provides a new `dump` command to the Fleet CLI, which can be called as:
`fleet dump -p <dest_path.tgz>`

It creates a `.tgz` archive containing:
* upstream cluster resources in YAML format, with file per {resource type,namespace,name} combination:
    * bundles
    * bundle deployments
    * bundle namespace mappings
    * clusters
    * cluster groups
    * GitRepos
    * GitReporestrictions
    * HelmOps
* Events which occurred in the following namespaces, with one file per namespace:
    * `default`
    * `kube-system`
    * `cattle-fleet-system`
    * `cattle-fleet-local-system`
    * any namespace containing a Fleet `Cluster` resource
* Metrics exposed by the Fleet `monitoring-*` services, if those services exist, with one file per service

Refers to #4030

### Open points/Limitations

* Should secrets and `content` resources (at least optionally) be included in the dump too?
* The dumping logic is pretty naive at the moment, keeping all instances of a given resource in memory before dumping them into the archive; this may become an issue for larger setups.
* Resources living on downstream clusters are beyond the scope of this effort, as the upstream cluster may not be able to reach that cluster.

### Checklist

### DIFF
--- a/internal/cmd/cli/dump/dump.go
+++ b/internal/cmd/cli/dump/dump.go
@@ -126,7 +126,7 @@ func addObjectsToArchive(
 			return fmt.Errorf("failed to marshal %s: %w", r, err)
 		}
 
-		fileName := fmt.Sprintf("%s_%s", r, i.GetName())
+		fileName := fmt.Sprintf("%s_%s_%s", r, i.GetNamespace(), i.GetName())
 		if err := addFileToArchive(g, fileName, w); err != nil {
 			return err
 		}
--- a/e2e/single-cluster/cli_dump_test.go
+++ b/e2e/single-cluster/cli_dump_test.go
@@ -51,13 +51,13 @@ var _ = Describe("Fleet dump", Label("sharding"), func() {
 				Expect(err).ToNot(HaveOccurred())
 
 				fileName := strings.Split(header.Name, "_")
-				Expect(fileName).To(HaveLen(2))
 
 				kindLow := fileName[0]
 				if kindLow != "metrics" {
 					continue
 				}
 
+				Expect(fileName).To(HaveLen(2))
 				Expect(content).ToNot(BeEmpty())
 
 				// Run a few basic checks on expected strings, checking full contents would be cumbersome
--- a/integrationtests/cli/dump/dump_test.go
+++ b/integrationtests/cli/dump/dump_test.go
@@ -74,10 +74,17 @@ var _ = Describe("Fleet dump", func() {
 				},
 			}
 
-			testBundleDeployment := fleet.BundleDeployment{
+			testBundleDeployment1 := fleet.BundleDeployment{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      "my-bundledeployment",
-					Namespace: "baz",
+					Namespace: "ns1",
+				},
+			}
+
+			testBundleDeployment2 := fleet.BundleDeployment{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "my-bundledeployment",
+					Namespace: "ns2",
 				},
 			}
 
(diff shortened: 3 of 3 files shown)


internal/cmd/cli/dump/dump.go
```diff
@@ -0,0 +1,491 @@
...
+	types := []string{
+		"bundles",
+		"bundledeployments",
+		"bundlenamespacemappings",
+		"clusters",
+		"clustergroups",
+		"gitrepos",
+		"gitreporestrictions",
+		"helmops",
+		// "contents",
```
0xavi0, line 75: I'm not sure this comment here is intentional to flag as TODO or a left-over.

From my experience when looking for info, it was useful to get the content resource (at least the metadata to inspect deletion timestamps, finalizers, etc...) We can do that in a future iteration.
Also... same for secrets. Maybe at least get the metadata so we can check for possible deletion timestamps, owner references, etc...

internal/cmd/cli/dump/dump.go
```diff
@@ -0,0 +1,491 @@
...
+	ns := config.DefaultNamespace // XXX: support installation in non-default namespace, and check for services across all namespaces, by label?
+
+	var svcs corev1.ServiceList
+	if err := c.List(ctx, &svcs, client.InNamespace(ns)); err != nil {
+		return fmt.Errorf("failed to list services for extracting metrics: %w", err)
+	}
+
+	var monitoringSvcs []corev1.Service
+	for _, svc := range svcs.Items {
+		if !strings.HasPrefix(svc.Name, "monitoring-") {
```
0xavi0, line 259: We will probably need to find ways to identify the metrics services more concisely.
If there is any service that is not a Fleet metrics service but has the same prefix… what would happen?

internal/cmd/cli/dump/dump.go
```diff
@@ -0,0 +1,491 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create dynamic Kubernetes client: %w", err)
+	}
+
+	return CreateWithClients(ctx, cfg, d, c, path)
+}
+
+func CreateWithClients(ctx context.Context, cfg *rest.Config, d dynamic.Interface, c client.Client, path string) error {
+	logger := log.FromContext(ctx).WithName("fleet-dump")
+	// XXX: be smart about memory management, progressively write data to disk instead of keeping everything in memory?
```
0xavi0, line 56: Yeah, maybe better to write data to disk as we iterate for larger clusters.
Also.. maybe good to have different entry points?. For example: a specific GitRepo/HelmOp/Bundle cascade of objects (in very large clusters maybe we're only interested on certain resources)

internal/cmd/cli/dump/dump.go
weyfonk: I wasn't sure either :sweat_smile: 
Anyway, great suggestions, I've added them to #4385. Thanks!

internal/cmd/cli/dump/dump.go
weyfonk: Well, that additional service would need to live in the default namespace (`cattle-fleet-system` for now), in which users should refrain from creating their own resources in the first place.
If such a service exists, and either doesn't expose a `/metrics` endpoint or returns an empty response body on that endpoint, then collecting data across services would fail (that particular service would not be skipped, which could also be an enhancement for a future iteration).

internal/cmd/cli/dump/dump.go
weyfonk: Indeed, I've created #4419 to keep track of these ideas.

PR #4433, opened 2025-12-12
test: Fix flaky cli_dump_test metrics check

thardeck:
The test was sometimes failing because it expected fleet_bundle_desired_ready
metrics to always be present, but these metrics are only exported when:
1. Resources (Bundles/GitRepos) exist in the cluster
2. The controller has reconciled them
   
Since this test did run independently without creating any resources, the metrics
may not be present, especially in fresh test environments or when the test runs
before other tests that create resources.
   
Instead of removing the fleet_*_desired_ready metrics assertion,
properly set up test resources to ensure these metrics are populated.

### DIFF
--- a/e2e/single-cluster/cli_dump_test.go
+++ b/e2e/single-cluster/cli_dump_test.go
@@ -35,11 +35,16 @@ var _ = Describe("Fleet dump", Label("sharding"), func() {
 				Expect(err).ToNot(HaveOccurred(), out)
 			})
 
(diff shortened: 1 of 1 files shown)


0xavi0 (commented):
Question:
Couldn’t we add some resources (if they’re needed) so we can check that metric?

The fix looks good to me, but I’m afraid we might be losing coverage if there’s another way to do it.

0xavi0 (approved):
LGTM

PR #4453, opened 2025-12-16
Use controlled Requeue for non-ready dependencies during deploy

aruiz14:
Refers to #4469

Forward-port of #4459

When the fleet-agent deploys a BundleDeployment that sets `DependsOn`, it will first lookup those dependencies and check if they are in ready state. We are currently treating this as an error, which is subjected to the default rate limiter: very short initial retries which are unlikely to find a different result, and chances to exhaust the retries (and maybe cause delays deploying bundles).

Unlike other type of problems, e.g. networking, contacting the Kubernetes API, etc., this should use a controlled re-schedule (through ctrl.Result's `RequeueAfter`) and not be treated as an error.
In order to minimize the changes, I've added a custom error type that can be detected from the reconciler and apply the approach mentioned above (set 15s interval between checks, trivial but IMO could offer a good balance between reactivity and noise).

## Additional Information

For following versions, we could consider adding smarter triggering when the dependencies transition to Ready state, but that's out of the scope of this PR.

### Checklist

### DIFF
--- a/internal/cmd/agent/controller/bundledeployment_controller.go
+++ b/internal/cmd/agent/controller/bundledeployment_controller.go
@@ -174,7 +174,7 @@ func (r *BundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		bd.Status = setCondition(bd.Status, err, monitor.Cond(fleetv1.BundleDeploymentConditionDeployed))
 
 		// Not-ready dependencies should not be treated as an error.
-		// Instead, a controlled requeue should happen until the conditions
+		// Instead, a controlled requeue should happen until the conditions are met.
 		var notReadyDependenciesError *deployer.NotReadyDependenciesError
 		if errors.As(err, &notReadyDependenciesError) {
 			if err := r.updateStatus(ctx, orig, bd); err != nil {


susesamu (approved):
LGTM! should we also log something? some info/debug logging

internal/cmd/agent/controller/bundledeployment_controller.go
```diff
@@ -174,6 +175,15 @@ func (r *BundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		// do not use the returned status, instead set the condition and possibly a timestamp
 		bd.Status = setCondition(bd.Status, err, monitor.Cond(fleetv1.BundleDeploymentConditionDeployed))
 
+		// Not-ready dependencies should not be treated as an error.
+		// Instead, a controlled requeue should happen until the conditions
```
p-se:
```suggestion
		// Instead, a controlled requeue should happen until the conditions are met
```

internal/cmd/agent/controller/bundledeployment_controller.go
```diff
@@ -174,6 +175,15 @@ func (r *BundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+		if errors.Is(err, &deployer.NotReadyDependenciesError{}) {
```
p-se: As suggested by @susesamu, `errors.Is` doesn't seem to work here.

PR #4457, opened 2025-12-17
Delete finalizers in content resources after upgrade

0xavi0:
This PR adds the functionality to check for finalizers in _Content_ resources that come from previous versions of Fleet.
It deletes those finalizers so _Contents_ can be properly deleted when needed. 

It adds the `For` statement to track for _Contents_ resources when being created (which is triggered when the controller is started after upgrading) 

Follow up to: https://github.com/rancher/fleet/pull/4361
Refers to: https://github.com/rancher/fleet/issues/4472

### DIFF
--- a/internal/cmd/controller/reconciler/content_controller.go
+++ b/internal/cmd/controller/reconciler/content_controller.go
@@ -35,7 +35,7 @@ type ContentReconciler struct {
 // SetupWithManager sets up the controller with the Manager.
 func (r *ContentReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&fleet.Content{}, // using with CreateFunc only to reconcile Contents to check for finalizers
+		For(&fleet.Content{}, // using For with CreateFunc only to reconcile Contents to check for finalizers
 			builder.WithPredicates(
 				predicate.Funcs{
 					CreateFunc:  func(e event.CreateEvent) bool { return true },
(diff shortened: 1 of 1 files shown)


internal/cmd/controller/reconciler/content_controller.go
```diff
@@ -35,7 +35,15 @@ type ContentReconciler struct {
 // SetupWithManager sets up the controller with the Manager.
 func (r *ContentReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
-		Named("content"). // naming because this controller does not use For()
+		For(&fleet.Content{}, // using For with CreateFunc only to reconcile Contents to check for finalizers
```
weyfonk, line 38: I don't understand this: why would we skip reconciles when deleting a Content resource? I'd think that would be precisely the main use case when its finalizers, if any, should be removed to enable the deletion to succeed.

0xavi0: The problem is that the new controller only reacts to changes on BundleDeployments, either when they are created, when their content label is modified, or when they are deleted.
However, the controller was not reacting to any events related to the Content resources themselves.

In addition, for the controller to be able to establish the BundleDeployment → Content relationship, the BundleDeployment must have been created with the new implementation and must have the corresponding label.

The new For reacting only to create events on the Content resources exists so that, when the controller starts after being updated, it can reconcile previous leftovers.

With a normal Bundle there is usually no problem, because the BundleController reconciles it, recreates the BundleDeployments, and those will already have the required label. The issue arises when this comes from an agent Bundle.

In my case, when upgrading from a previous version, I was also passing new values to Helm (for example, to enable debug). That causes the content to change and, consequently, creates another Content resource.
If that happens midway through the upgrade, what I observed is that the Content was created using the previous implementation and finalizers were added to it.

Then, once the upgrade finishes, the Bundle is reconciled again, the content changes again, and another different Content is created—this time without finalizers.

The result is that the Content that was created midway through the upgrade is left orphaned.

In this way, when the controller restarts, the reconciler is triggered (with a Create event), and there we can check whether it had finalizers. If it did, and there is no BundleDeployment referencing it, then it is a leftover that we can safely delete.

I’m not sure if I’ve explained it clearly enough; otherwise, I can do an offline demo.

weyfonk: Oh this makes sense; we're interested in creation events, because those are the events received by the controller when it first starts, as happens after a Fleet update.
Thanks for clarifying :bow:

Issue #4432, opened 2025-12-12
Fleet-Agent with hostNetwork=true might lock up on rollout


webD97:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When enabling host network mode for the agent, agent rollouts can get stuck: If the new agent cannot open the metrics and probe ports due to the ports still being in use on the old instance, it logs an error and stops doing anything (no crash), even if the old instance eventually terminates. This seems to be a race condition as it only happens sometimes.

### Expected Behavior

Agent rollouts should "just work" without locking up and leaving behind a non-functional agent.

### Steps To Reproduce

1. Install Fleet and connect at least one cluster (local should work, too)
2. Configure the cluster to run the agent in hostNetwork mode
3. Repeatedly restart the agent deployment on the cluster and observe the agent logs

### Environment

```markdown
- Architecture: x86_64
- Fleet Version: 0.13.5
- Cluster:
  - Provider: k3s (on agent cluster)
  - Options: no customizations
  - Kubernetes Version: 1.34.1
```

### Logs

```markdown
I1212 07:51:14.423907       1 leaderelection.go:257] attempting to acquire leader lease fleet-system/fleet-agent...
I1212 07:51:14.430758       1 leaderelection.go:271] successfully acquired lease fleet-system/fleet-agent
{"level":"info","ts":"2025-12-12T07:51:14Z","logger":"setup","msg":"renewed leader","identity":"redacted"}
{"level":"info","ts":"2025-12-12T07:51:14Z","logger":"setup","msg":"successfully registered with upstream cluster","namespace":"cluster-redacted"}
{"level":"info","ts":"2025-12-12T07:51:14Z","logger":"setup","msg":"listening for changes on upstream cluster","cluster":"cluster-redacted","namespace":"cluster-redacted"}
{"level":"info","ts":"2025-12-12T07:51:14Z","logger":"setup","msg":"Starting controller","metricsAddr":":8080","probeAddr":":8081","systemNamespace":"fleet-system"}
[...]

PR #4467, opened 2025-12-19
Fix: Set Recreate deployment strategy when hostNetwork is enabled

webD97:
Refers to #4432

The changes ensure that if the fleet-agent is configured to run in the host network, it will also use a deployment strategy of type `Recreate`. This prevents a race condition where during a rollout, the new agent pod tries to open ports that are still in use by the old pod and locking up without any retries - leaving the downstream cluster disconnected from the controller cluster. By using a `Recreate` strategy, the old pod is shut down before the new pod starts.

The PR also extends the unit tests to cover this scenario.

### Checklist


webD97:
I don't think an update in the docs repo is required here. :)

PR #4474, opened 2025-12-19
Add `monitor` and `analyze` commands

weyfonk:
Contrarily to the `dump` CLI, which only saves states of resources from the upstream cluster, these commands offer the following:
* `monitor`: saving of a snapshot of specific fields and issues about resources, with possible periodic saving of snapshots in JSON format, with the possibility to search those snapshots for issues (using tools like `jq`)
* `analyze`: uses snapshots as input (can be chained with `monitor`), and produces summarised, or detailed output, allowing diffs between snapshots for further analysis.

Implementation by @manno and @0xavi0, with a few bits of refactoring on top.

Follow-up to #4374.
Refers to #4030.

Requires #4483.

## Limitations
When checking controller states:
* gitjob and helmops controllers are not taken into account
* neither are sharded controllers

### Checklist

- [x] I have updated the documentation via a [pull request](https://github.com/rancher/fleet-docs/pull/387) in the [fleet-docs](https://github.com/rancher/fleet-docs) repository

### DIFF
--- a/internal/cmd/cli/monitor.go
+++ b/internal/cmd/cli/monitor.go
@@ -240,7 +240,7 @@ Examples:
 type Snapshot struct {
 	// Timestamp is an UTC timestamp of when the snapshot was created.
 	Timestamp         string                 `json:"timestamp"`
-	Controller        *ControllerInfo        `json:"controller,omitempty"`
+	Controller        []ControllerInfo       `json:"controller,omitempty"`
 	GitRepos          []GitRepoInfo          `json:"gitrepos,omitempty"`
 	Bundles           []BundleInfo           `json:"bundles,omitempty"`
 	BundleDeployments []BundleDeploymentInfo `json:"bundledeployments,omitempty"`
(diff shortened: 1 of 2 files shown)


internal/cmd/cli/monitor.go
```diff
@@ -0,0 +1,949 @@
...
+
+func (m *Monitor) convertGitRepos(gitRepos []fleet.GitRepo) []GitRepoInfo {
+	result := make([]GitRepoInfo, 0, len(gitRepos))
+	for _, gr := range gitRepos {
+		info := GitRepoInfo{
+			Namespace:           gr.Namespace,
+			Name:                gr.Name,
+			Generation:          gr.Generation,
+			ObservedGeneration:  gr.Status.ObservedGeneration,
+			Commit:              gr.Status.Commit,
```
0xavi0, line 566: I think it might be interesting to also store the `WebhookCommit`, `PollingCommit` and `LastPollingTime`.
It could be useful to check if the gitrepo is being polled or if the webhook was successfully handled.
Maybe also the labels.

internal/cmd/cli/monitor.go
```diff
@@ -0,0 +1,949 @@
...
+	}
+	return result
+}
+
+func (m *Monitor) convertContents(contents []fleet.Content) []ContentInfo {
+	result := make([]ContentInfo, 0, len(contents))
+	for _, c := range contents {
+		info := ContentInfo{
+			Name:       c.Name,
+			Finalizers: c.Finalizers,
```
0xavi0: Aside from Finalizers (to detect if there are _Content_ resources not being properly migrated to the new implementation) we might need the new `Status.ReferenceCount` field

Issue #4027, opened 2025-08-20
Investigate git cloning optimisations


weyfonk:
When a GitRepo is created, Fleet's gitjob controller creates a job which:
1. [Clones](https://github.com/rancher/fleet/blob/main/internal/cmd/cli/gitcloner/cloner.go) the repository, using the branch configured in the GitRepo
2. Runs `fleet apply` to create bundles from the repository's contents, according to `paths` configured in the GitRepo

Cloning the full git repository is wasteful, especially for larger repositories.

We should investigate how much it can be improved, for instance by:
* only cloning state from the configured branch's latest commit, as tried in #3136 
    * Does this work with submodules?
* using [partial clones](https://github.com/go-git/go-git/issues/1381)
* supporting [sparse checkouts](https://github.com/go-git/go-git/pull/1101)

dydoc:
I reviewed the current support in go-git for cloning optimizations and related problems

It seems that the general case, where a main repo plus n submodules (which may be nested submodules), has some problems [1](https://github.com/go-git/go-git/issues/1785) and [2](https://github.com/go-git/go-git/pull/1789).


The available strategies to reduce git history (faster cloning) are :
- **the partial clone** : This it is not 100% ready in go-git
- **the shallow clone** : There is an[open bug]((https://github.com/go-git/go-git/issues/1785))  in go-git

The available strategy to reduce memory consumption is:
- **sparse-checkout** only the Fleet relevant folder of a repository: the sparse-checkout is supported from go-git and we could use the GitRepoSpec.paths to identify Fleet relevant folders

### Remark 1

An important info is that both partial clone and shallow clone rely on server side features that must be enabled to support client requests: __the server side behavior in case of unsupported features is not standardized and depends on the involved feature__. Some requests are silently ignored, while others return explicit errors.

### Remark 2
In the case of submodule to work around the [go-git bug]((https://github.com/go-git/go-git/issues/1785)), if the server has the *allow-reachable-sha1-in-want** and the *shallow* feature enabled, we can also fetch  a commit not at the tip of the branch (i.e **fetch --depth=1 <sha>:refs/temp**). If one of the two features is not available we could fallback to alternative strategies.


**Remark 1** and **Remark 2** suggest that a phase of feature discovery is needed. Based on it we should select different strategies corresponding to different trade offs for the git history size  that produces the same worktree.


The following diagram visualize a possible approach:

```
┌─────────────────────────────────────────────────────────────────────────────┐
[...]

PR #4488, opened 2026-01-05
improvement of fleet cli subcommand gitcloner

ncuralli:
Refers to #4027 

This is an implementation of the [proposed plan](https://github.com/rancher/fleet/issues/4027#issuecomment-3655872822) in #4027.
The sparseCheckout  is not implemented at the moment.

The MR splits the logic about how the sub command gitcloner  clones a repository:
- it uses a vanilla go-git clone  for the main repository 
- it enables for the submodules a smart layer that selects the best strategy for cloning a submodule and all its nested submodules based on the git server capabilities : it uses primitives from go-git as much as possible.

### Checklist

### DIFF
--- a/internal/cmd/cli/gitcloner/submodule/capability/discovery.go
+++ b/internal/cmd/cli/gitcloner/submodule/capability/discovery.go
@@ -78,9 +78,9 @@ func capabilitiesFromList(caps *capability.List) *Capabilities {
 	}
 }
 
-// The `ChooseStrategy` method in the `CapabilityDetector` struct is
+// The ChooseStrategy method in the CapabilityDetector struct is
 // determining the appropriate strategy type based on the capabilities
-// provided by the Git server. It takes a `Capabilities` struct as input
+// provided by the Git server. It takes a Capabilities struct as input
 // and evaluates the capabilities to decide which strategy should be used
 // for fetching data from the server.
 func (d *CapabilityDetector) ChooseStrategy(caps *Capabilities) StrategyType {
--- a/internal/cmd/cli/gitcloner/submodule/fetcher.go
+++ b/internal/cmd/cli/gitcloner/submodule/fetcher.go
@@ -19,7 +19,7 @@ import (
 type Strategy interface {
 	Type() capability.StrategyType
 	// Execute fetches the specified commit from the remote.
-	Execute(ctx context.Context, r *git.Repository, CommitHash plumbing.Hash) error
+	Execute(ctx context.Context, r *git.Repository, commitHash plumbing.Hash) error
 }
 
 // CapabilityDetectorInterface abstracts capability detection for testing.
--- a/internal/cmd/cli/gitcloner/submodule/capability/discovery_test.go
+++ b/internal/cmd/cli/gitcloner/submodule/capability/discovery_test.go
@@ -111,7 +111,7 @@ func TestChooseStrategy(t *testing.T) {
 		expected StrategyType
 	}{
 		{
-			name: "shallow SHA - can fetch by SHA and swallow",
+			name: "shallow SHA - can fetch by SHA and shallow",
 			caps: &Capabilities{
 				AllowReachableSHA1InWant: true,
 				Shallow:                  true,
--- a/internal/cmd/cli/gitcloner/submodule/strategy/full_clone_test.go
+++ b/internal/cmd/cli/gitcloner/submodule/strategy/full_clone_test.go
@@ -60,8 +60,8 @@ func TestFullCloneStrategy_FetchError(t *testing.T) {
 			return nil
 		},
 	}
-	CommitHash := plumbing.NewHash("abc123")
-	err := s.Execute(context.Background(), nil, CommitHash)
+	commitHash := plumbing.NewHash("abc123")
+	err := s.Execute(context.Background(), nil, commitHash)
 
 	if err == nil {
 		t.Fatal("expected error")
--- a/internal/cmd/cli/gitcloner/submodule/strategy/full_sha_test.go
+++ b/internal/cmd/cli/gitcloner/submodule/strategy/full_sha_test.go
@@ -60,8 +60,8 @@ func TestFullSHAStrategy_FetchError(t *testing.T) {
 			return nil
 		},
 	}
-	CommitHash := plumbing.NewHash("abc123")
-	err := s.Execute(context.Background(), nil, CommitHash)
+	commitHash := plumbing.NewHash("abc123")
+	err := s.Execute(context.Background(), nil, commitHash)
 
 	if err == nil {
 		t.Fatal("expected error")
--- a/internal/cmd/cli/gitcloner/submodule/strategy/incremental_test.go
+++ b/internal/cmd/cli/gitcloner/submodule/strategy/incremental_test.go
@@ -70,8 +70,8 @@ func TestIncrementalDeepenStrategy_CommitFoundAtDepth5(t *testing.T) {
 		},
 	}
 
-	CommitHash := plumbing.NewHash("abc123")
-	err := s.Execute(context.Background(), nil, CommitHash)
+	commitHash := plumbing.NewHash("abc123")
+	err := s.Execute(context.Background(), nil, commitHash)
 
 	if err != nil {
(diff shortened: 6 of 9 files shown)


internal/cmd/cli/gitcloner/submodule/capability/discovery.go
```diff
@@ -0,0 +1,115 @@
...
+	"fmt"
+
+	"github.com/go-git/go-git/v5/plumbing/protocol/packp/capability"
+	"github.com/go-git/go-git/v5/plumbing/transport"
+)
+
+// Capabilities represents fetch-related capabilities advertised by a Git server.
+type Capabilities struct {
+	AllowReachableSHA1InWant bool
+	AllowTipSHA1InWant       bool
```
0xavi0, line 13: `AllowTipSHA1InWant` is not used aside from the unit tests, right?

internal/cmd/cli/gitcloner/submodule/capability/discovery.go
```diff
@@ -0,0 +1,115 @@
...
+		return StrategyShallowSHA
+	}
+
+	if caps.CanFetchBySHA() && !caps.CanFetchShallow() {
+		return StrategyFullSHA
+	}
+
+	if caps.Shallow {
+		return StrategyIncrementalDeepen
+	}
```
0xavi0: nit: for being inline with the previous code... I guess it would make sense to use  the `CanFetchShallow` function.
(or we could also just access the Capabilities fields as they are exported)

internal/cmd/cli/gitcloner/submodule/git-counts.sh
```diff
@@ -0,0 +1,104 @@
+#!/bin/bash
```
0xavi0: I'm not sure we need this file in the repo... 
I see it is mentioned in a couple of tests, but if the git repository used is not updated we would not need to update those values again. 

We have the [fleet-test-data](https://github.com/rancher/fleet-test-data) repository for our tests, so we don't depend of external sources.

internal/cmd/cli/gitcloner/submodule/fetcher.go
```diff
@@ -0,0 +1,134 @@
...
+			capability.StrategyShallowSHA:        strategy.NewShallowSHAStrategy(auth),
+			capability.StrategyFullSHA:           strategy.NewFullSHAStrategy(auth),
+			capability.StrategyIncrementalDeepen: strategy.NewIncrementalStrategy(auth),
+			capability.StrategyFullClone:         strategy.NewFullCloneStrategy(auth),
+		}
+	}
+
+	return f, nil
+}
+
```
0xavi0, line 117: I think it would be good to document all these exported functions a bit.
For example explain that `Fetch` fetches using the detected strategy or the forced one, etc...

internal/cmd/cli/gitcloner/submodule/fetcher.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/plumbing/transport"
+	"github.com/rancher/fleet/internal/cmd/cli/gitcloner/submodule/capability"
+	"github.com/rancher/fleet/internal/cmd/cli/gitcloner/submodule/strategy"
+)
+
+type Strategy interface {
+	Type() capability.StrategyType
+	Execute(ctx context.Context, r *git.Repository, CommitHash plumbing.Hash) error
+}
```
0xavi0, line 23: nit: Wouldn't it make sense to have the `Strategy` interface in the `strategy` package instead?
I was looking for it at the `strategy.go` file and found it in here. 
What do you think?

internal/cmd/cli/gitcloner/cloner.go
```diff
@@ -79,29 +81,45 @@ func (c *Cloner) CloneRepo(opts *GitCloner) error {
...
 		URL:               opts.Repo,
+		Depth:             1,
 		Auth:              auth,
 		InsecureSkipTLS:   opts.InsecureSkipTLS,
 		CABundle:          caBundle,
 		SingleBranch:      true,
 		ReferenceName:     plumbing.ReferenceName(opts.Branch),
-		RecurseSubmodules: git.DefaultSubmoduleRecursionDepth,
+		RecurseSubmodules: git.NoRecurseSubmodules,
+		Tags:              git.NoTags,
```
0xavi0, line 93: Question: why are we doing a `plainClone` and not using all the new code? 
Looks like the new approach is only done for submodules, right? or am I missing something?

0xavi0 (commented):
Code looks very well organised and the feature seems to be well implemented, Thanks!

I've added a few comments and questions for now, mostly minor things.

My main question is why is the new approach only used for submodules and not the entire git repository

ncuralli:
> Code looks very well organised and the feature seems to be well implemented, Thanks!
> 
> I've added a few comments and questions for now, mostly minor things.
> 
> My main question is why is the new approach only used for submodules and not the entire git repository

The main driver for the limitation of the new approach only for the submodules is because the high level implementation of the submodule update operation in go-git have some bugs (https://github.com/go-git/go-git/issues/1785 and https://github.com/go-git/go-git/pull/1789) and it is a bit opinionated as the git cli implementation as well. They require to extend the gitRepo CRD to permit some clone depth setting for the submodules to avoid the clone failure: it is about  giving the customer the tool for tackling with cases where the commit pointed from the submodule is not in the shallow clone of the submodule itself.
On the other hand the plainClone works fine if we avoid the use the go-git submodule management with `RecurseSubmodules: git.NoRecurseSubmodules`.
This approach is to maintain what is working and rework what is not working making minimal change to gitRepo CRD.

internal/cmd/cli/gitcloner/submodule/fetcher.go
ncuralli: The followed approach was that the interface is defined by the consumer, not the implementer. it should help to maintain the packages independent. Do you see any advantage to move it except the naming?

internal/cmd/cli/gitcloner/submodule/capability/discovery.go
ncuralli: yes: I added there for completeness but it isn't used in the code.

internal/cmd/cli/gitcloner/submodule/capability/discovery.go
ncuralli: fixed in commit c607c1b850441b136b18b2cd52c897aa2d8a16a1

internal/cmd/cli/gitcloner/cloner.go
ncuralli: answered in the general comment at the beginning.

internal/cmd/cli/gitcloner/submodule/fetcher.go
ncuralli: documented with commit 120e9e6bbed4051dc356c32504472f7e68d638e8

internal/cmd/cli/gitcloner/submodule/git-counts.sh
ncuralli: moved the script to fleet-test-data repo: https://github.com/rancher/fleet-test-data/pull/35

internal/cmd/cli/gitcloner/submodule/capability/discovery.go
0xavi0: Okay, we can leave it, just in case we need it in the future

internal/cmd/cli/gitcloner/submodule/git-counts.sh
0xavi0: Thanks! 
What I meant also is that the script is referencing an external git repository we have no control. (https://github.com/git-fixtures/basic)
Although it hasn't changed for 11 years, we are not the owners and our CI could break if any external change is done.
I would rely on our own git repositories for tests. 
We can maybe clone that test or create something similar and store it alongside with the script in `fleet-test-data`.

0xavi0: let's leave it as it is for now.

0xavi0 (approved):
LGTM

Issue #4491, opened 2026-01-05
fleet-controller ends up in a crashback loop


susesgartner:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Fleet version: v0.14.2-alpha.1
Rancher version: v2.13-head or v2.14-head

Description: When provisioning a downstream custom cluster via rancher the fleet-controller ends up in a crashback loop due to a segmentation fault. It sometimes recovers from the crashback loop but I have also seen get stuck for at least 10 minutes. 

Details: https://github.com/rancher/rancher/issues/53233

Workaround: restart the fleet-controller

### Expected Behavior

No segmentation fault occurs and the fleet controller remains in a stable state while provisioning downstream clusters. 

### Steps To Reproduce

1. Provision a rancher cluster on 2.13-head or 2.14-head. 
2. Provision a few downstream custom clusters

### Environment

```markdown
- Architecture: 
- Fleet Version: v0.14.2-alpha.1
- Cluster: RKE2/K3s custom
  - Provider: AWS
  - Options:
  - Kubernetes Version: v1.34.2+rke2r1
```

### Logs

<img width="1321" height="406" alt="Image" src="https://github.com/user-attachments/assets/e6af4fd3-792a-473b-b57a-bd65b278ad73" />

p-se:
It looks like there's an issue with the retrieval of the ClusterRegistrationToken. It should try to create one if it doesn't find any, but it seems that the code assumes it did find one and proceeds with that, which eventually leads to the nil pointer dereference. This would also explain why it eventually catches up. The code that seemingly returns the token comes from wrangler and is cached, so it might have to do with a cache, which could mean that a busy API server is part of why this issue occurs. I did not see any recent changes around the code that fails, which supports that theory. However, I'm still confused that the code that fetches the token [doesn't return an error but is seemingly capable of returning a nil value without an error](https://github.com/p-se/fleet/commit/e882c745ebdadf0917ace6875df72c576b6c9af1). It would be valuable to have it reproduced and checked in more detail.

rancherbot:
Not creating port issue, milestone 2.13.2 does not exist or is not an open milestone

weyfonk:
@susesgartner this should be fixed in Fleet v0.14.2/ Rancher 2.13.2.

susesgartner:
@weyfonk I am now observing this issue on v0.15.0-alpha.4 / Rancher 2.14-head.

weyfonk:
[...]

PR #4494, opened 2026-01-07
Prevent nil pointer dereference on cluster import

p-se:
Refers to #4491

### Checklist

### DIFF
--- a/internal/cmd/controller/agentmanagement/controllers/cluster/import.go
+++ b/internal/cmd/controller/agentmanagement/controllers/cluster/import.go
@@ -378,10 +378,19 @@ func (i *importHandler) importCluster(cluster *fleet.Cluster, status fleet.Clust
 			},
 		})
 		// Ignore AlreadyExists errors (race condition with another reconcile)
-		if err != nil && !apierrors.IsAlreadyExists(err) {
-			logrus.Debugf("Failed to create ClusterRegistrationToken for cluster %s/%s: %v (requeuing)", cluster.Namespace, cluster.Name, err)
-			i.clusters.EnqueueAfter(cluster.Namespace, cluster.Name, durations.TokenClusterEnqueueDelay)
-			return status, err
+		if err != nil {
+			if apierrors.IsAlreadyExists(err) {
+				token, err = i.tokens.Get(cluster.Namespace, tokenName)
+				if err != nil {
+					logrus.Debugf("Failed to get existing ClusterRegistrationToken for cluster %s/%s: %v (requeuing)", cluster.Namespace, cluster.Name, err)
+					i.clusters.EnqueueAfter(cluster.Namespace, cluster.Name, durations.TokenClusterEnqueueDelay)
+					return status, err
+				}
+			} else {
+				logrus.Debugf("Failed to create ClusterRegistrationToken for cluster %s/%s: %v (requeuing)", cluster.Namespace, cluster.Name, err)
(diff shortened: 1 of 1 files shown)


internal/cmd/controller/agentmanagement/controllers/cluster/import.go
aruiz14: I truly think the problem is here, as the program will continue if `err` is not nil but different than `AlreadyExists` (and `token` will be empty). This was recently changed as part of a massive PR:
https://github.com/rancher/fleet/pull/4316/changes#diff-28a4af52a53de89516ada67a34c6382270d9ee7c39e6fa4164c64ecefd7aeead
but the rationale behind the change is not clear / lost during squash. @thardeck do you remember that?

weyfonk: If the error is not nil but different to `AlreadyExists`, execution returns from within this block :thinking: 
The issue may happen if the token _does_ already exist, in which case I think we may want to `Get` it to be on the safe side, rather than reusing a token which only exists in memory (created on L363) but not in the cluster.

aruiz14: Oh, yeah, sorry I read this wrong. My point is that this code should not continue whenever `err` is not `nil`, as `token` cannot be used in that case, and that may cause a nil pointer dereference.
It's up to you guys whether the `AlreadyExists` error should be treated differently (doing an extra `Get` or retry everything from scratch, maybe immediately instead of after a delay). Maybe it will help moving this logic to its own separate function `GetOrCreate...`.

On a side note, is there any concurrent controller that may create this object? The comment says ("race condition with another reconcile"), but the workqueue should prevent more than one concurrent processing of the same key, so if that is true this could hiding a bigger problem?

weyfonk: Agent management does not use controller-runtime at this point; I'm not sure what Wrangler's workqueue implementation (if any) may look like.

aruiz14: Wrangler (through lasso), also [uses a workqueue](https://github.com/rancher/lasso/blob/aacb7deb9ecd0b57016a160544989be6fe383866/pkg/controller/controller.go#L212-L237), so we can assume the same controller won't be reconciling the same object in multiple workers at the same time.

thardeck: > @aruiz14 ... but the rationale behind the change is not clear / lost during squash. @thardeck do you remember that?

The non squashed commit which touched this related coded path was [this](https://github.com/rancher/fleet/pull/4316/changes/f85cf1a4cb827d326bb8a2e4d8c14b24c12452cb#diff-28a4af52a53de89516ada67a34c6382270d9ee7c39e6fa4164c64ecefd7aeeadR361).

I think the original intention of the fix was to prevent legitimate failures (API error, permissions, etc.) to be silently ignored.

The exception for `AlreadyExists` was added because this should not be a real error state, since the token does exist and there is no need to retry creating it.

What was overlooked is the issue Patrick fixed.

weyfonk: For reference, what has been decided is to `Get` the token when it already exists, and to carry on if that retrieval is successful.

weyfonk (approved):

PR #4499, opened 2026-01-07
[v0.14] Support cluster display name in clusterName selector (#4495)

thardeck:
Fleet now matches clusters by both the cluster resource name (e.g., c-12345) and the display name stored in the management.cattle.io/cluster-display-name label. This restores the expected behavior in Rancher where GitRepos can target clusters by their display names.

When a clusterName selector is specified in a GitRepo target, Fleet will now:
1. First try to match against the cluster resource name (existing behavior)
2. If that doesn't match, check if the clusterName matches the value in the management.cattle.io/cluster-display-name label

This fixes the regression where imported clusters, AKS/EKS clusters, and other clusters created through Rancher could no longer be targeted by their display names.

Refers to rancher/rancher#51239


weyfonk (approved):

sbulage:
Respective issue is verified, moving to Done.

PR #4501, opened 2026-01-08
Fix crust gatherer caching

thardeck:
The `date +%j` format returns values with leading zeros (e.g., `008` for January 8th). Bash interprets numbers with leading zeros as octal in arithmetic expressions, causing an error since octal only supports digits `0-7`.

Use `date +%-j` instead to get the day of year without leading zeros, avoiding the octal interpretation issue.

### DIFF
--- a/.github/scripts/determine-cache-key.sh
+++ b/.github/scripts/determine-cache-key.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+# Determine cache key that changes every 28 days
+# Outputs to GITHUB_OUTPUT in GitHub Actions context
+#
+# Usage: determine-cache-key.sh
+
+set -e
+
+# Get the day of year (1-366)
+DAY_OF_YEAR=$(date +%-j)
+
+# Use modulo to create a key that changes every 28 days
+if [ $((DAY_OF_YEAR % 28)) -eq 0 ]; then
+  # On the 28th, 56th, 84th... day of the year, use a date-based key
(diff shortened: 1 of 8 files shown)


.github/scripts/determine-cache-key.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+# Determine cache key that changes every 28 days
```
aruiz14, line 2: Just out of curiosity, is there any particular reason to use exactly 28 days?

If I understand correctly, this makes all builds happening on every 28th day to create a different cache key value, hence creating a new cache (`latest` is a special keyword pointing that will then point to this newest value, right?). What happens if by any chance, there are no builds happening on that specific day? Wouldn't the cache "survive" that expiration event and live for longer than 28 days?

Also, why bothering with this if we could use the month number instead? Couldn't we just use year+month as cache key?

thardeck: I think the original plan was four weeks. But the main goal was to prevent the rate limiting and downloading the same version over and over again.
CI is run daily so there should not be an issue to miss that day.
We could switch to a year+month approach if preferred.
Let me know.

thardeck:
Merging to prevent failing CI, future improvements can be done in a separate pr.

.github/scripts/determine-cache-key.sh
thardeck: I have [changed](https://github.com/rancher/fleet/pull/4509) the caching to a simpler month based scheduling, as suggested by you. Thanks.

PR #4522, opened 2026-01-12
Update v0.11

thardeck:
https://artifacthub.io/packages/helm/fleet/fleet/0.11.11?modal=security-report


go.mod
```diff
@@ -46,8 +46,8 @@ require (
 	github.com/testcontainers/testcontainers-go v0.34.0
 	go.uber.org/mock v0.5.0
 	go.uber.org/zap v1.27.0
-	golang.org/x/crypto v0.36.0
-	golang.org/x/sync v0.12.0
+	golang.org/x/crypto v0.40.0
```
weyfonk, line 49: There is a CVE with medium severity, marked as fixed in v0.45.0; shouldn't we bump to that one?
```suggestion
	golang.org/x/crypto v0.45.0
```

Also, more importantly: why not bump Helm to v3.18.5, which should contain fixes for high-severity vulns?

thardeck: It seems that crypto can not be bumped without bumping Wrangler which requires a new Kubernetes version.

We usually do not bump minor Helm versions for a specific Rancher version.

weyfonk: Ack, thanks for clarifying.

weyfonk (approved):

Issue #2085, opened 2024-01-17
Fleet doesn't detect when Helm values from Secrets or ConfigMaps change


thomashoell:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I have a bundle where I import values from a secret like this:
```yaml
helm:
  valuesFrom:
    - secretKeyRef:
        name: secret-values
        key: values.yaml
```

When this secret changes, Fleet doesn't detect this and still shows my bundle as in sync. Even when I hit _force update_, the old values are redeployed, as the BundleDeployment does not update. I need to change something in my bundle to force Fleet to reread the referenced secret.

This is especially unsatisfying as this specific secret is automatically generated by the External Secrets Operator. 

### Expected Behavior

Fleet should detect that the values from a referenced Secret/ConfigMap change and redeploy the Bundle if needed.

### Steps To Reproduce

1. Create a Fleet Bundle and reference Helm values from a Secret or ConfigMap
2. Deploy the Bundle
3. Change the values in the Secret/ConfigMap
4. Nothing happens

### Environment

```markdown
- Architecture: amd64
- Fleet Version: 0.9.0
- Cluster:
  - Provider: RKE2
  - Options:
  - Kubernetes Version: 1.26.12
```

und-miller:
Experiencing similar issues when updating the `fleet.yaml` file.

matmitch:
@thomashoell I am using a similar setup (values from secret generated by external secret operator) did you find a workaround for this issue ?

thomashoell:
> [@thomashoell](https://github.com/thomashoell) I am using a similar setup (values from secret generated by external secret operator) did you find a workaround for this issue ?

Not really. You could try to make the deployed application itself aware of changes in a secret using https://github.com/jimmidyson/configmap-reload or something similar.

shane-davidson:
I am having the same issue. Force sync is the only way I can get fleet to update the bundle deployment

Does the fleet controller / BundleDeployment template the valuesFrom? and would it make more sense for the "values" to be added to the bundle deployment and the "valuesFrom" to be added/templated by the agent running downstream? Then the agent could subscribe to changes to any secret/configmaps that change.

bundle deployment would just pass the reference to any required downsteam configmap/secret


Is there any work around to force the deployments on a time etc? 

Does "force update" just increase the `forceSyncGeneration` option on the gitrepo resource? (I was looking at making a workaround to automatically increase the value on a set time interval) 

[...]

PR #4530, opened 2026-01-14
Watch downstream resources for changes.

0xavi0:
This PR adds code so that Fleet reacts to changes in Secrets and ConfigMaps that are used by resources copied to downstream clusters.

For the implementation, Watch statements were added to listen for changes in Secrets and ConfigMaps, using an index and a predicate that only reconciles Bundles when the changed Secret or ConfigMap belongs to a downstream resource of a Bundle.

A new field is added to the BundleDeployment spec (DownstreamResourcesGeneration) and another one to the status to know when downstream resources have been synchronized.

The value in the spec is incremented every time Secrets or ConfigMaps are updated or created.

Refers to: https://github.com/rancher/fleet/issues/2085

Tested with the following script: 
[cm-sync-check.sh](https://github.com/user-attachments/files/24719673/cm-sync-check.sh)

You can use: 
```bash
./cm-sync-check.sh 100
``` 
For 100 config map updates.
The default is 50

### Checklist

### DIFF
--- a/e2e/multi-cluster/downstream_clone_objects_test.go
+++ b/e2e/multi-cluster/downstream_clone_objects_test.go
@@ -449,6 +449,17 @@ var _ = Describe("Downstream objects cloning", Ordered, func() {
 				// The value "dXBkYXRlZC12YWx1ZQ==" is base64 encoded "updated-value"
 				g.Expect(s).To(Equal("dXBkYXRlZC12YWx1ZQ=="))
 			}).WithTimeout(30 * time.Second).WithPolling(testenv.LongPollingInterval).Should(Succeed())
+
+			By("propagating the update to the deployment once the secret is updated")
+			Eventually(func(g Gomega) {
+				cms, err := kd.Get("configmaps")
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(cms).To(ContainSubstring(cmName))
+
+				name, err := kd.Get("configmaps", cmName, "-o", "jsonpath={.data.name}")
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(name).To(Equal("updated-secret-value"))
+			}).Should(Succeed())
 		})
 
 		AfterEach(func() {


internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -759,11 +789,32 @@ func (r *BundleReconciler) handleDownstreamObjects(ctx context.Context, bundle *
...
 
+	// Only patch if the value changed to true
+	if resourcesUpdated {
+		// Create a patch to update the BundleDeployment
+		patch := client.MergeFrom(bd.DeepCopy())
+		bd.Spec.DownstreamResourcesGeneration++
+
+		// Check if patch is empty before applying
+		if patchData, err := patch.Data(bd); err == nil && string(patchData) != "{}" {
+			if err := r.Patch(ctx, bd, patch); err != nil {
```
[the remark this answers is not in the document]
0xavi0: I think this is wrong.
`path` is obtained from a deep copy of `bd`, then `bd` is updated and the patch is applied using the patch data obtained with the deep copy.

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -759,11 +789,32 @@ func (r *BundleReconciler) handleDownstreamObjects(ctx context.Context, bundle *
...
+				return fmt.Errorf(
+					"%w: failed to patch BundleDeployment with DownstreamResourcesGeneration: %w",
+					fleetutil.ErrRetryable,
+					err,
+				)
+			}
+		}
```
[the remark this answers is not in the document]
0xavi0: We can't do that, because the secret or configmap needs a _BundleDeployment_ with a UID because the BD owns the secret or configmap. 
We don't get the UID until the BD is created, so we can't update this value yet. 

We could check if the secrets are different before even calling the BundleDeployment creation or update, but that would be more inefficient. 

If there's an error when patching (because of a conflict) the function returns an error, which ends up in another reconcile call.

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -734,13 +755,18 @@ func (r *BundleReconciler) updateErrorStatus(
 
 func (r *BundleReconciler) handleDownstreamObjects(ctx context.Context, bundle *fleet.Bundle, bd *fleet.BundleDeployment) error {
 	if !experimental.CopyResourcesDownstreamEnabled() {
+		bd.Spec.DownstreamResourcesGeneration = 0
```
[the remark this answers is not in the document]
0xavi0: 0 is the default value, so deleted that statement. 
If the env variable is not set we don't need it.

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -134,6 +147,8 @@ func (r *BundleReconciler) SetupWithManager(mgr ctrl.Manager) error {
 //+kubebuilder:rbac:groups=fleet.cattle.io,resources=bundles,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=fleet.cattle.io,resources=bundles/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=fleet.cattle.io,resources=bundles/finalizers,verbs=update
+//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch
+//+kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch
```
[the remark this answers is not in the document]
0xavi0: I've deleted both lines entirely, because they're not needed. 
The controller uses the rbac defined in the helm chart.

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -126,6 +127,18 @@ func (r *BundleReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			// referenced in DownstreamResources changes.
+			&corev1.Secret{},
+			handler.EnqueueRequestsFromMapFunc(r.downstreamResourceMapFunc("Secret")),
+		).
+		Watches(
+			// Fan out from configmap to bundle, reconcile bundles when a configmap
+			// referenced in DownstreamResources changes.
+			&corev1.ConfigMap{},
+			handler.EnqueueRequestsFromMapFunc(r.downstreamResourceMapFunc("ConfigMap")),
+		).
```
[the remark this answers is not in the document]
0xavi0: I've added predicates to only use modifications on secrets or configmaps

internal/cmd/agent/controller/bundledeployment_controller.go
```diff
@@ -390,6 +391,9 @@ func (r *BundleDeploymentReconciler) copyResourcesFromUpstream(
 		}
 	}
 
+	// update DownstreamResourcesGeneration to reflect that resources were copied/updated
+	bd.Status.DownstreamResourcesGeneration = bd.Spec.DownstreamResourcesGeneration
```
weyfonk, line 395: Given the conditions in [ShouldUpdateStatus](https://github.com/rancher/fleet/blob/main/internal/cmd/agent/deployer/monitor/updatestatus.go#L72), I think this should work, i.e. this update would be written to the bundle deployment status in the upstream cluster.
We'd need an integration or end-to-end test to validate this though.

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -1024,3 +1107,39 @@ func BundleDeploymentMapFunc(r *BundleReconciler) func(ctx context.Context, a cl
 		return nil
 	}
 }
+
+// downstreamResourceMapFunc returns a function that maps a Secret or ConfigMap to Bundles
+// that reference it in their DownstreamResources.
+func (r *BundleReconciler) downstreamResourceMapFunc(kind string) func(ctx context.Context, obj client.Object) []reconcile.Request {
```
weyfonk: In the interest of grouping methods vs functions for readability, this one being a method, it would make sense to move it further up in this file.

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -1024,3 +1107,39 @@ func BundleDeploymentMapFunc(r *BundleReconciler) func(ctx context.Context, a cl
...
+	kind = strings.ToLower(kind) // the index uses lowercase kind
```
```suggestion
	kind = strings.ToLower(kind) // the index uses lowercase kind
	
```

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -877,6 +927,39 @@ func batchDeleteBundleDeployments(ctx context.Context, c client.Client, list []f
...
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			switch new := e.ObjectNew.(type) {
+			case *corev1.Secret:
+				old, ok := e.ObjectOld.(*corev1.Secret)
+				if !ok {
+					return false
+				}
+				// Secrets only have Data field (map[string][]byte)
+				return !reflect.DeepEqual(new.Data, old.Data)
```
weyfonk, line 982: To be on the safe side, we may want to take the `Type` field into account as well for secrets.

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -877,6 +927,39 @@ func batchDeleteBundleDeployments(ctx context.Context, c client.Client, list []f
 	return errors.Join(errs...)
 }
 
+// dataChangedPredicate filters Secret and ConfigMap events to only trigger reconciliation
+// when Data or BinaryData fields have changed.
+func dataChangedPredicate() predicate.Funcs {
+	return predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return false
```
weyfonk: Couldn't reconciling upon secret/config map creation still be useful, in case users accidentally reference nonexistent resources, and create them afterwards? Reconciling upon creation would prevent them from needing to re-create the bundle in such cases.

internal/cmd/controller/reconciler/bundle_controller_test.go
```diff
@@ -1108,3 +1108,456 @@ func expectGetWithFinalizer(mockCli *mocks.MockK8sClient, bundle fleetv1.Bundle)
...
+	os.Setenv(envVar, "true")
+
+	testCases := []struct {
+		name                     string
+		downstreamResources      []fleetv1.DownstreamResource
+		initialGeneration        int64
+		existingBundleDeployment *fleetv1.BundleDeployment
+		setupResourceMocks       func(*gomock.Controller, *mocks.MockK8sClient)
+		expectGeneration         int64
+		expectPatch              bool
```
weyfonk: Unless I've missed something, these values are not checked anywhere, and actually redundant with mocked `Patch` calls made in `setupResourceMocks` and checks made within those `Patch` calls, or the absence of such calls.
```suggestion
```

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -1024,3 +1107,39 @@ func BundleDeploymentMapFunc(r *BundleReconciler) func(ctx context.Context, a cl
...
+	kind = strings.ToLower(kind) // the index uses lowercase kind
+	return func(ctx context.Context, obj client.Object) []reconcile.Request {
+		// Create the index key for this resource (Kind/Name)
+		indexKey := fmt.Sprintf("%s/%s", kind, obj.GetName())
+
+		// Find all bundles that reference this resource
+		bundleList := &fleet.BundleList{}
+		err := r.List(ctx, bundleList,
+			client.InNamespace(obj.GetNamespace()),
+			client.MatchingFields{config.BundleDownstreamResourceIndex: indexKey},
```
weyfonk: I would like to see this one in action, just to be sure that it does what we expect it to, for instance in and end-to-end test case. Although, since we're dealing with a single reconciler, integration tests may well be enough.

internal/cmd/agent/controller/bundledeployment_controller.go
0xavi0: I tested it before and it did work (I've attached the script I used for hours in case it is useful to QA). Also we are already updating other status fields from downstream (to show errors or progress, for instance).

I agree it makes sense to add `e2e` tests.
I've added tests for checking that an existing secret or configmap update is propagated and that when deploying non-existing DownstreamResources, they are deployed when they are created.

I've also added explicit checks for the `Status.DownstreamResourcesGeneration` value update.

internal/cmd/controller/reconciler/bundle_controller.go
0xavi0: moved

internal/cmd/controller/reconciler/bundle_controller.go
0xavi0: We can't update the `type` field of a Secret in Kubernetes, right?
I think you have to delete the secret and re-create it...

internal/cmd/controller/reconciler/bundle_controller.go
0xavi0: Good point, I think it's a valid use case.

I've also added an e2e test for testing that case

internal/cmd/controller/reconciler/bundle_controller_test.go
0xavi0: I guess those were leftovers :P. Removed

internal/cmd/controller/reconciler/bundle_controller.go
0xavi0: I've added a couple of `e2e` tests that should cover those lines.

internal/cmd/controller/reconciler/bundle_controller.go
weyfonk: Oh, it seems you're right:
```
$ kc patch secret fleet-agent -n cattle-fleet-local-system --type=merge -p '{"type": "generic"}'
The Secret "fleet-agent" is invalid: type: Invalid value: "generic": field is immutable
```

e2e/multi-cluster/downstream_clone_objects_test.go
```diff
@@ -282,4 +292,237 @@ var _ = Describe("Downstream objects cloning", Ordered, func() {
...
+
+			By("verifying the secret data change propagates to downstream automatically")
+			Eventually(func(g Gomega) {
+				s, err := kd.Get("secret", "secret-image-pull", "-o=jsonpath={.data.new-key}")
+				g.Expect(err).ToNot(HaveOccurred())
+				// The value "dXBkYXRlZC12YWx1ZQ==" is base64 encoded "updated-value"
+				g.Expect(s).To(Equal("dXBkYXRlZC12YWx1ZQ=="))
+			}).WithTimeout(30 * time.Second).WithPolling(testenv.LongPollingInterval).Should(Succeed())
+		})
+
```
weyfonk, line 464: Having checked that the secret update is propagated downstream, can we also verify that the deployment reflects that new value?
e.g. through a new `Eventually` block featuring `name, err := kd.Get("configmaps", cmName, "-o", "jsonpath={.data.name}")`.

weyfonk (commented):
Thank you for adding end-to-end tests :bow: 
I have one last suggestion, please let me know if I've missed something.

e2e/multi-cluster/downstream_clone_objects_test.go
0xavi0: Added as requested

Issue #4559, opened 2026-01-22
DownstreamResources don't work for the local cluster


0xavi0:
There is a problem when copying Downstream Resources into the local cluster, because `DeepCopy()` is being used, which also copies the metadata (including the UID).

When calling `CreateOrUpdate`, the function does not return any error, but internally Kubernetes is not performing the copy because a resource with the same ID already exists.

kkaempf:
This probably needs a backport to 2.13 🤔

sbulage:
### System Information
| Rancher Version | Fleet Version |
|--------|--------|
| v2.14-3d66af043d1a52a83aeb5c4aa67ada04f0f23ce4-head | 109.0.0+up0.15.0-alpha.6 | 


### Verification steps:
- Enabled `EXPERIMENTAL_COPY_RESOURCES_DOWNSTREAM`
  - <details><summary>Enable Experimental feature</summary>
    
    ```
    extraEnv:
      - name: EXPERIMENTAL_COPY_RESOURCES_DOWNSTREAM
        value: "true"
    ```
    </details>

---
- Create ConfigMap using below command:
  -  ConfigMap data file `values.yaml`
      ```yaml
      name: testing-downstream-cluster-configmap
      ```
      ```bash
      kubectl create configmap -n fleet-local config-values --from-file=values.yaml=values.yaml
      ```

---
- Below `HelmOps` YAML used
  - <details><summary>HelmOps YAML</summary>
    
    ```yaml
    apiVersion: fleet.cattle.io/v1alpha1
    kind: HelmOp
    metadata:
      name: test-values-from
      namespace: fleet-local
    spec:
      correctDrift: {}
      downstreamResources:
        - kind: configmap
          name: config-values
      helm:
        chart: >-
          https://github.com/rancher/fleet/raw/refs/heads/main/integrationtests/cli/assets/helmrepository/config-chart-0.1.0.tgz
        valuesFrom:
          - configMapKeyRef:
              key: values.yaml
              name: config-values
      pollingInterval: 5s
      targets:
        - clusterSelector:
            matchExpressions:
              - key: provider.cattle.io
                operator: NotIn
                values:
                  - harvester
    ```
    </details> 
---

<details><summary>ConfigMaps in fleet-local</summary>
<p>

<img width="1919" height="768" alt="Image" src="https://github.com/user-attachments/assets/5abad733-62db-44f9-a565-027f683f2a56" />

</p>
</details>

PR #4560, opened 2026-01-22
Avoid using DeepCopy when copying Downstream Resources

0xavi0:
This PR avoids copying ConfigMaps and Secrets using `DeepCopy`, since this is problematic when you want to copy the object into a different namespace but within the same cluster (this usually happens when using a single-cluster configuration).

E2E tests have been added to verify the copying of downstream resources also in a single-cluster environment, by reusing the multi-cluster tests and adapting them.

Refers to: https://github.com/rancher/fleet/issues/4559

### Checklist

### DIFF
--- a/e2e/single-cluster/downstream_clone_objects_test.go
+++ b/e2e/single-cluster/downstream_clone_objects_test.go
@@ -14,11 +14,6 @@ import (
 	fleet "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
 )
 
-type DownstreamResource struct {
-	Kind string
-	Name string
-}
-
 // This test uses a single cluster to demonstrate cloning of configured resources within the same cluster.
 var _ = Describe("Downstream objects cloning", Ordered, func() {
 	var (
@@ -29,7 +24,7 @@ var _ = Describe("Downstream objects cloning", Ordered, func() {
 		deployNamespace     string
 		keepResources       bool
 		valuesFrom          []fleet.ValuesFrom
-		downstreamResources []DownstreamResource
+		downstreamResources []fleet.DownstreamResource
 		cmName              = "test-simple-chart-config"
 	)
 
@@ -41,11 +36,6 @@ var _ = Describe("Downstream objects cloning", Ordered, func() {
 		cmAsset := path.Join(testenv.AssetPath("single-cluster/values-cm.yaml"))
 		out, err := k.Create("configmap", "config-values", fmt.Sprintf("--from-file=values.yaml=%s", cmAsset))
 		Expect(err).ToNot(HaveOccurred(), out)
-
-		DeferCleanup(func() {
-			_, _ = k.Delete("secret", "secret-values")
-			_, _ = k.Delete("configmaps", "config-values")
-		})
 
 		secretAsset := path.Join(testenv.AssetPath("single-cluster/values-secret.yaml"))
 		out, err = k.Create("secret", "generic", "secret-values", fmt.Sprintf("--from-file=values.yaml=%s", secretAsset))
@@ -64,7 +54,7 @@ var _ = Describe("Downstream objects cloning", Ordered, func() {
 
 		// Use default downstream resources if not set
 		if downstreamResources == nil {
-			downstreamResources = []DownstreamResource{
+			downstreamResources = []fleet.DownstreamResource{
 				{
 					Kind: "Secret",
 					Name: "secret-values",
@@ -90,7 +80,7 @@ var _ = Describe("Downstream objects cloning", Ordered, func() {
 			PollingInterval       time.Duration
 			HelmSecretName        string
 			InsecureSkipTLSVerify bool
-			DownstreamResources   []DownstreamResource
+			DownstreamResources   []fleet.DownstreamResource
 			KeepResources         bool
 			ValuesFrom            []fleet.ValuesFrom
 		}{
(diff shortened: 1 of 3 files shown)


internal/cmd/agent/controller/bundledeployment_controller.go
```diff
@@ -397,6 +339,115 @@ func (r *BundleDeploymentReconciler) copyResourcesFromUpstream(
...
+		s.Type = source.Type // important for e.g. image pull secrets
+		s.Labels = source.Labels
+		s.Annotations = source.Annotations
+		s.Data = source.Data
+		s.StringData = source.StringData
+		// Ensure ownership label is preserved
+		if s.Labels == nil {
+			s.Labels = map[string]string{}
+		}
+		s.Labels[fleetv1.BundleDeploymentOwnershipLabel] = bd.Name
```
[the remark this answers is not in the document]
0xavi0: I think the code suggested is more complex and I don't see so much profit in it.

internal/cmd/agent/controller/bundledeployment_controller.go
```diff
@@ -397,6 +339,115 @@ func (r *BundleDeploymentReconciler) copyResourcesFromUpstream(
...
+		cm.Labels = source.Labels
+		cm.Annotations = source.Annotations
+		cm.Data = source.Data
+		cm.BinaryData = source.BinaryData
+		// Ensure ownership label is preserved
+		if cm.Labels == nil {
+			cm.Labels = map[string]string{}
+		}
+		cm.Labels[fleetv1.BundleDeploymentOwnershipLabel] = bd.Name
+
```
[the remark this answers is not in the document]
0xavi0: same as above

e2e/single-cluster/downstream_clone_objects_test.go
```diff
@@ -0,0 +1,534 @@
...
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/rancher/fleet/e2e/testenv"
+	"github.com/rancher/fleet/e2e/testenv/kubectl"
+	fleet "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
+)
+
+type DownstreamResource struct {
```
weyfonk: nit: this type is already defined [here](https://github.com/rancher/fleet/blob/main/pkg/apis/fleet.cattle.io/v1alpha1/bundledeployment_types.go#L392), with that module already being imported :thinking:

e2e/single-cluster/downstream_clone_objects_test.go
```diff
@@ -0,0 +1,534 @@
...
+
+	JustBeforeEach(func() {
+		cmAsset := path.Join(testenv.AssetPath("single-cluster/values-cm.yaml"))
+		out, err := k.Create("configmap", "config-values", fmt.Sprintf("--from-file=values.yaml=%s", cmAsset))
+		Expect(err).ToNot(HaveOccurred(), out)
+
+		DeferCleanup(func() {
+			_, _ = k.Delete("secret", "secret-values")
+			_, _ = k.Delete("configmaps", "config-values")
+		})
```
weyfonk: Do we still need this given the `AfterEach` block below? :thinking:

e2e/single-cluster/downstream_clone_objects_test.go
```diff
@@ -0,0 +1,534 @@
...
+						Key:                  "values.yaml",
+					},
+				},
+			}
+		})
+
+		It("deploys the HelmOp and clones resources", func() {
+			By("copying resources")
+			Eventually(func(g Gomega) {
+				s, err := k.Get("secrets")
```
weyfonk: I think this check for the values config map should happen against the target namespace, otherwise we'd just be checking that the _source_ secret/config maps exist, not that they've actually been copied.

This applies to this test case and other test cases throughout this file.

e2e/single-cluster/downstream_clone_objects_test.go
```diff
@@ -0,0 +1,534 @@
...
+
+			By("updating a copied resource upstream")
+			out, err := k.Patch("configmap", "config-values", "--type=merge", "-p", `{"data":{"values.yaml":"name: new-name"}}`)
+			Expect(err).ToNot(HaveOccurred(), out)
+
+			By("reconciling the helmop")
+			// remove the secret as downstream resource; does not have any effect on the
+			// bundle deployment itself, but should trigger a new reconcile of the helmop, hence the bundle.
+			out, err = k.Patch("helmop", name, "--type=json", "-p", `[{"op": "remove", "path": "/spec/downstreamResources/0"}]`)
+			Expect(err).ToNot(HaveOccurred(), out)
```
weyfonk: This test case, and its multi-cluster counterpart, could be considered obsolete, now that monitoring of downstream resources is in place: users should no longer need to wait for a new bundle reconcile to happen (for other reasons than an update to the referenced `DownstreamResource` itself) :)

internal/cmd/agent/controller/bundledeployment_controller.go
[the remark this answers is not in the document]
weyfonk: I agree that what Copilot suggests may be more complicated than what we need, but do we need to set the `BundleDeploymentOwnershipLabel` _twice_? i.e. wouldn't L370-372 be superfluous here?

Same remark about L425-427.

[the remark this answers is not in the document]
0xavi0: I've deleted the initial assignment for both.

e2e/single-cluster/downstream_clone_objects_test.go
0xavi0: Yeah, same thing happened in the multi-cluster one. I've deleted it from both. Thanks.

e2e/single-cluster/downstream_clone_objects_test.go
0xavi0: No, we don't :). Deleted

e2e/single-cluster/downstream_clone_objects_test.go
0xavi0: You're absolutely right, my bad. Changed now.

e2e/single-cluster/downstream_clone_objects_test.go
0xavi0: I thought it was still a valid test given that it exercises the reconciliation in a different way, but yeah...  I've deleted it.

Issue #4300, opened 2025-11-06
[SURE-10633] Add allowedTargetNamespaceSelector in the GitRepoRestriction resource


kkaempf:
### Request description:

User would like to have a new tag added to the GitRepoResttriction resource called `allowedTargetNamespaceSelector`.

### Actual behavior:

The `allowedTargetNamespaceSelector` tag will be a label selector to match labels defined on the downstream cluster's namespaces

### Expected behavior:

When labels defined on the namespace resources match the selector labels, the namespace on the downstream cluster will be a target namespace that bundles can be deployed to

thardeck:
Waiting for feedback from Tan

kkaempf:
@weyfonk 😉

kkaempf:
also related: SURE-6497 and #633

mmartin24:
Checked this on `0.15.0-beta.4` and I am not sure if it is blocking everything or I missed something. Here is what I tried:

- Deployed Rancher with 3 downstream clusters.
- On `cluster 1` created:
```
kubectl create namespace qa-apps
kubectl label namespace qa-apps team=qa --overwrite
```
- On `cluster 2` created only ns (not label)
```
kubectl create namespace qa-apps
namespace/qa-apps created

kubectl label namespace qa-apps team-
label "team" not found.
namespace/qa-apps not labeled
```
 - Created `GitRepoRestricion`
```yaml
apiVersion: fleet.cattle.io/v1alpha1
kind: GitRepoRestriction
metadata:
  name: fleet-qa-restriction
  namespace: fleet-default
allowedTargetNamespaceSelector:
  matchLabels:  
    team: qa
```
- Passed gitrepo on `fleet-default` targeting `qa-apps` namespace:
```yaml
apiVersion: fleet.cattle.io/v1alpha1
kind: GitRepo
metadata:
  name: multi-cluster-test
  namespace: fleet-default
spec:
  repo: https://github.com/rancher/fleet-examples
  paths: ["simple"]
  targets:
  - namespace: qa-apps
```
Expected:
- cluster 1 deploys bundle and cluster 2 does not

Observed:
Restriction applied all over with message: `Empty targetNamespace denied, because allowedTargetNamespaceSelector restriction is present`

<img width="2510" height="803" alt="Image" src="https://github.com/user-attachments/assets/798eaa1d-a3e9-4ea8-9b93-829bb75706ac" />

@thardeck , please let me know if I missed something or if this needs to be reviewed. Thanks

--- 

[...]

PR #4562, opened 2026-01-23
Implement AllowedTargetNamespaceSelector for GitRepoRestriction

thardeck:
Add label selector-based validation of target namespaces on downstream clusters. When GitRepoRestriction.AllowedTargetNamespaceSelector is set, the Fleet agent validates that the target namespace labels match the selector before deploying bundles.

Refers to #4300

### DIFF
--- a/internal/cmd/controller/labelselectors/merge.go
+++ b/internal/cmd/controller/labelselectors/merge.go
@@ -1,6 +1,7 @@
 package labelselectors
 
 import (
+	"maps"
 	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -22,9 +23,7 @@ func Merge(selectors ...*metav1.LabelSelector) *metav1.LabelSelector {
 			if result.MatchLabels == nil {
 				result.MatchLabels = make(map[string]string)
 			}
-			for k, v := range selector.MatchLabels {
-				result.MatchLabels[k] = v
-			}
+			maps.Copy(result.MatchLabels, selector.MatchLabels)
 			result.MatchExpressions = mergeMatchExpressions(result.MatchExpressions, selector.MatchExpressions)
 		}
 	}


thardeck:
Waiting for feedback from Tan

internal/cmd/controller/agentmanagement/agent/manifest_test.go
```diff
@@ -304,6 +304,7 @@ func TestManifestAgentResources(t *testing.T) {
 			agent := getAgentFromManifests("", testCase.getOpts())
 			if agent == nil {
 				t.Fatal("there were no deployments returned from the manifests")
+				return
```
weyfonk: nit: `t.Fatal` should already trigger a return on its own.
```suggestion
```

internal/cmd/controller/labelselectors/merge.go
```diff
@@ -0,0 +1,29 @@
...
+		if result == nil {
+			result = selector.DeepCopy()
+		} else {
+			if result.MatchLabels == nil {
+				result.MatchLabels = make(map[string]string)
+			}
+			for k, v := range selector.MatchLabels {
+				result.MatchLabels[k] = v
+			}
+			result.MatchExpressions = append(result.MatchExpressions, selector.MatchExpressions...)
```
weyfonk: Note: if more than one selectors contain `MatchExpressions` with common keys, they will all apply, including duplicates, but also possibly contradictory operators (e.g. `foo Exists` and `foo DoesNotExist`).

internal/cmd/agent/controller/bundledeployment_controller.go
```diff
@@ -162,6 +163,52 @@ func (r *BundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		}
 	}
 
+	if bd.Spec.Options.AllowedTargetNamespaceSelector != nil {
+		targetNamespace := bd.Spec.Options.DefaultNamespace
+		if targetNamespace == "" {
+			targetNamespace = r.DefaultNamespace
+		}
```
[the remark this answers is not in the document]
weyfonk: I think Copilot has a point here: why not use `GetDeploymentNS` here?

internal/cmd/agent/controller/bundledeployment_controller.go
```diff
@@ -162,6 +163,52 @@ func (r *BundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+
+		ns := &corev1.Namespace{}
+		if err := r.LocalClient.Get(ctx, types.NamespacedName{Name: targetNamespace}, ns); err != nil {
+			if apierrors.IsNotFound(err) {
+				err = fmt.Errorf("target namespace %s does not exist on downstream cluster", targetNamespace)
```
weyfonk, line 176: Should this be an error though? Wouldn't the namespace be created later on (for instance before possibly copying resources from upstream, see L350)?

internal/cmd/agent/controller/bundledeployment_controller.go
```diff
@@ -162,6 +163,52 @@ func (r *BundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+			err = fmt.Errorf("invalid AllowedTargetNamespaceSelector: %w", err)
+			bd.Status = setCondition(bd.Status, err, monitor.Cond(fleetv1.BundleDeploymentConditionDeployed))
+			if statusErr := r.updateStatus(ctx, orig, bd); statusErr != nil {
+				return ctrl.Result{}, fmt.Errorf("%w; failed to update status: %w", err, statusErr)
+			}
+			return ctrl.Result{}, err
+		}
+
+		if !selector.Matches(labels.Set(ns.Labels)) {
+			err = fmt.Errorf("target namespace %s does not match AllowedTargetNamespaceSelector: required labels %v, namespace has %v",
```
weyfonk: In cases where labels match, should we feature `MatchExpressions` from the `AllowedTargetNamespaceSelector` as rejection could then come from those expressions?

integrationtests/agent/bundle_deployment_namespace_selector_test.go
```diff
@@ -0,0 +1,198 @@
...
+				err := k8sClient.Get(context.TODO(), types.NamespacedName{
+					Name:      bd.Name,
+					Namespace: bd.Namespace,
+				}, bd)
+				if err != nil {
+					return false
+				}
+
+				return hasCondition(bd, "Deployed", corev1.ConditionTrue) || bd.Status.Ready
+			}).Should(BeTrue())
```
weyfonk: nit: For `Eventually` blocks, I'd be more comfortable writing them this way, to make possible errors easier to troubleshoot (as opposed to receiving output stating that `false` was returned instead of `true`):
```suggestion
			Eventually(func(g Gomega) {
				err := k8sClient.Get(context.TODO(), types.NamespacedName{
					Name:      bd.Name,
					Namespace: bd.Namespace,
				}, bd)
				g.Expect(err).NotTo(HaveOccurred())

				g.Expect(hasCondition(bd, "Deployed", corev1.ConditionTrue) || bd.Status.Ready).To(BeTrue())
			}).Should(Succeed())
```

internal/cmd/controller/labelselectors/merge.go
thardeck: That behavior is intentional: restrictions are ANDed, so contradictory expressions should yield “no match” rather than be filtered.

internal/cmd/agent/controller/bundledeployment_controller.go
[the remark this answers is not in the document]
weyfonk: I still have doubts here 🤔 (see comment right above this one)

internal/cmd/agent/controller/bundledeployment_controller.go
```diff
@@ -162,6 +163,57 @@ func (r *BundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+		if err := r.LocalClient.Get(ctx, types.NamespacedName{Name: targetNamespace}, ns); err != nil {
+			if apierrors.IsNotFound(err) {
+				err = fmt.Errorf("target namespace %s does not exist on downstream cluster yet", targetNamespace)
+				bd.Status = setCondition(bd.Status, err, monitor.Cond(fleetv1.BundleDeploymentConditionDeployed))
+				if statusErr := r.updateStatus(ctx, orig, bd); statusErr != nil {
+					return ctrl.Result{}, fmt.Errorf("%w; failed to update status: %w", err, statusErr)
+				}
+				return ctrl.Result{RequeueAfter: durations.DefaultRequeueAfter}, nil
+			} else {
+				err = fmt.Errorf("failed to get target namespace %s: %w", targetNamespace, err)
```
weyfonk, line 178: In my understanding, this will not work: if the namespace does not exist, requeueing will bring the controller back to this point, loop after loop... Because the namespace would be created later in the loop, when deploying the bundle.

On the other hand... I think that _not finding the namespace_ should be part of the flow here: if the namespace does not yet exist, then it cannot possibly contain labels matching the selector (provided that the selector prescribes labels, of course), in which case we should reject the deployment.
How do you see it?

thardeck: True, good point. I have reverted the change.

weyfonk: Would this approach still error even if no `AllowedTargetNamespaceSelector` is configured in BundleDeployment options? If so, we would be introducing a regression.

thardeck: No, it should be prevented by [if bd.Spec.Options.AllowedTargetNamespaceSelector != nil {](https://github.com/rancher/fleet/pull/4562/changes#diff-a7e61410edd028f42a8c18792f6f0af173f1545521fa3cd96f4327441d775b09R166)

I have added a test to verify that this does not happen.

internal/cmd/agent/controller/bundledeployment_controller.go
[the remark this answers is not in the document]
thardeck: I thought I fixed that already so it was marked as resolved which prevent me from seeing these two comments.

I have addressed it now.

integrationtests/agent/bundle_deployment_namespace_selector_test.go
```diff
@@ -105,6 +105,47 @@ var _ = Describe("BundleDeployment namespace selector validation", Ordered, func
 			err = k8sClient.Delete(context.TODO(), bd)
 			Expect(err).ToNot(HaveOccurred())
 		})
+
+		// Regression test: Verify that deployments without AllowedTargetNamespaceSelector
+		// can still deploy to missing namespaces via Helm's CreateNamespace feature.
+		// Without the selector guard, the namespace existence check would run unconditionally
+		// and prevent Helm from creating the namespace, breaking existing deployments.
+		It("deploys successfully to a missing namespace", func() {
```
weyfonk, line 113: Nice! 🥇

internal/cmd/agent/controller/bundledeployment_controller.go
weyfonk: Oh, of course 🤦 
Thanks for being very explicit about this :D

internal/cmd/controller/labelselectors/merge.go
```diff
@@ -22,8 +25,36 @@ func Merge(selectors ...*metav1.LabelSelector) *metav1.LabelSelector {
 			for k, v := range selector.MatchLabels {
 				result.MatchLabels[k] = v
 			}
-			result.MatchExpressions = append(result.MatchExpressions, selector.MatchExpressions...)
```
weyfonk: nit: my editor tells me this could be simplified as:
```
maps.Copy(result.MatchLabels, selector.MatchLabels)
```

Issue #4484, opened 2026-01-02
bundle controller shard filtering based on cluster changes


adibrastegarnia:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Previously we had a discussion about bundle controller was not filtering bundles correctly based on shard ID and this PR tried to address it https://github.com/rancher/fleet/pull/4282 (https://github.com/rancher/fleet/issues/4275). however, I don't think this PR solves the issue completely.  

Let’s assume I labeled my GitRepo and Bundle as shard-0, while the cluster resource (which is independent of these objects) is labeled as shard-1. If a GitRepo is created first and a cluster is added later, the cluster change triggers the bundle controller. However, since the cluster is in shard-1, the controller cannot find any bundles to reconcile because the bundles generated by the GitRepo belong to shard-0.

To resolve this, I believe WithEventFilter(sharding.FilterByShardID(r.ShardID)) should be removed so it no longer applies to all resources. Instead, we should apply shard filtering specifically to the Bundle and BundleDeployment watchers. We should allow all clusters to trigger the watch, then filter the bundles based on the shard ID associated with the cluster that triggered the watch.

cc @0xavi0 

### Environment

```markdown
- Architecture:
- Fleet Version:
- Cluster:
  - Provider:
  - Options:
  - Kubernetes Version:
```

### Logs

```markdown

```

johnjcool:
Could this be the right fix for this regression? After applying it, BundleDeployments are created correctly for new clusters in our setup.

```
1 file changed, 5 insertions(+), 1 deletion(-)

diff --git a/internal/cmd/controller/reconciler/bundle_controller.go b/internal/cmd/controller/reconciler/bundle_controller.go
index 0e842da1..54b109c8 100644
--- a/internal/cmd/controller/reconciler/bundle_controller.go
+++ b/internal/cmd/controller/reconciler/bundle_controller.go
@@ -126,7 +126,7 @@ func (r *BundleReconciler) SetupWithManager(mgr ctrl.Manager) error {
                        }),
                        builder.WithPredicates(clusterChangedPredicate()),
                ).
-               WithEventFilter(sharding.FilterByShardID(r.ShardID)).
+               // WithEventFilter(sharding.FilterByShardID(r.ShardID)).
                WithOptions(controller.Options{MaxConcurrentReconciles: r.Workers}).
                Complete(r)
 }
[...]

PR #4567, opened 2026-01-26
bundle controller shard filtering based on cluster changes

johnjcool:
Refers to #4484

This PR updates shard filtering behavior for cluster->bundle fan-out in internal/cmd/controller/reconciler/bundle_controller.go. When a fleet.Cluster changes (labels/annotations/spec/status), the controller now only enqueues reconcile requests for fleet.Bundle objects that should be processed by the current shard (ShardID), preventing cross-shard reconciles triggered by cluster updates.

### DIFF
--- a/internal/cmd/controller/reconciler/bundle_controller.go
+++ b/internal/cmd/controller/reconciler/bundle_controller.go
@@ -100,10 +100,7 @@ func (r *BundleReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			// Fan out from bundledeployment to bundle, this is useful to update the
 			// bundle's status fields.
 			&fleet.BundleDeployment{}, handler.EnqueueRequestsFromMapFunc(BundleDeploymentMapFunc(r)),
-			builder.WithPredicates(
-				bundleDeploymentStatusChangedPredicate(),
-				sharding.FilterByShardID(r.ShardID),
-			),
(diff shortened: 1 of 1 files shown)


weyfonk:
Thanks @johnjcool for your contribution :)
We like the idea, and need to figure out how far we want to take this.
For instance, recently-added monitoring of user-created secrets and config maps should also be impacted by this, yet we may not want to require users to add `fleet.cattle.io/shard-ref` labels to such resources.

weyfonk:
> Thanks @johnjcool for your contribution :) We like the idea, and need to figure out how far we want to take this. For instance, recently-added monitoring of user-created secrets and config maps should also be impacted by this, yet we may not want to require users to add `fleet.cattle.io/shard-ref` labels to such resources.

We've had a closer look, leading to the following conclusions:
* Sharding predicates (`sharding.FilterByShardID(r.ShardID)`) are not necessary against monitored secrets and config maps, because the logic mapping config maps and secrets to bundles monitoring them [already](https://github.com/rancher/fleet/blob/main/internal/cmd/controller/reconciler/bundle_controller.go#L947) takes care of filtering out bundles living in different shards
* Schedules were also affected by this mismatch; see #4730 for more details (@johnjcool , feel free to take a look, your comments are most welcome there too).

weyfonk:
Superseded by #4730.
Thanks for this initiative :)

PR #4569, opened 2026-01-27
[helm] Allow to add custom known_hosts to the configmap

SebPlv:
Hi,

After upgrading from Fleet 0.12.4 to 0.13.4, all our gitrepos were failing with messages containing the following error 

> ssh: handshake failed: knownhosts: key is unknown

This is related to a the new feature #3523 introduced in v0.13.0 which reject connection attempts to unknown hosts if `insecure-skip-host-key-checks` is not activated. 

As many compagnies, we have our own internal gitlab solution from which all our gitrepos are pulled. In that case, it makes sense to globally add the public SSH key of our gitlab instance to the fleet known_hosts configmap.

This configmap is created by the helm chart with the most widely used git providers. This PR brings the ability to add its own keys to the configmap through the Helm chart.

- [ ]  I have updated the documentation via a pull request in the [fleet-docs](https://github.com/rancher/fleet-docs) repository.
No, but if you think I should, let me know.

### DIFF
--- a/charts/fleet/values.yaml
+++ b/charts/fleet/values.yaml
@@ -110,7 +110,7 @@ priorityClassName: ""
 insecureSkipHostKeyChecks: false
 
 ## Adds hosts to the known_hosts configmap if using a custom git provider for gitrepos
-additionalKnownHosts: {}
+additionalKnownHosts: []
 # Custom entries to be added in SSH known_hosts format (see charts/fleet/templates/configmap_known_hosts.yaml for examples):
 #   - <host>[:port] <key-algorithm> <fingerprint>
 
--- a/.github/scripts/update_known_hosts_configmap.sh
+++ b/.github/scripts/update_known_hosts_configmap.sh
@@ -23,8 +23,8 @@ for prov in "${providers[@]}"; do
     ssh-keyscan "$prov" | grep "^$prov" | sort -b | sed 's/^/    /' >> "$dst"
 done
 
-echo '{{- if .Values.additionalKnownHosts }}
+echo '    {{- if .Values.additionalKnownHosts }}
     {{ range .Values.additionalKnownHosts -}}
     {{ . }}
     {{ end -}}
-{{- end }}' >> "$dst"
+    {{- end }}' >> "$dst"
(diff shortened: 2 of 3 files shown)


kkaempf:
This is missing a test case

kkaempf:
tentatively setting milestone, to be finalized after triage.

SebPlv:
Hello @kkaempf ,

First you need an internal git provider (eg. gitlab or [gogs](https://github.com/gogs/gogs)). Create a new project on which your Fleet gitrepo will point to.
Example of gitrepo.yaml : 

```
kind: GitRepo
metadata:
  name: velero
  namespace: fleet-default
spec:
  caBundle: <ca_bundle>
  revision: 0.0.1
  keepResources: true
  repo: ssh://git@gogs.internal.co/path_to/my_project.git  
  clientSecretName: fleet-conf-git-ssh-auth
  helmSecretName: fleet-conf-helm
  paths:
    - run
```

clientSecretName only contains the SSH private key (https://fleet.rancher.io/gitrepo-add#adding-a-private-git-repository).

On a default Fleet configuration, you'll get an `knownhosts: key is unknown` message because Fleet doesn't have the SSH key for this git provider.

Update fleet with helm by adding the SSH key in your values.yaml like that : 

```
additionalKnownHosts: 
  - 'gogs.internal.co ecdsa-sha2-nistp256 AAA....'
```

The way to get the key is described in [the documentation](https://fleet.rancher.io/gitrepo-add#known-hosts).

The error should go away.


Is that what you wanted ?

kkaempf:
Thanks @SebPlv for the reproducer - this is important for the final QA approval.

I was asking for an automated test, something we can run in our CI to ensure this bug is fixed (and future changes don't break it again). 😉

charts/fleet/values.yaml
```diff
@@ -94,6 +94,11 @@ priorityClassName: ""
 ## host verification fails. Insecure when set to true. Default: false.
 insecureSkipHostKeyChecks: false
 
+## Adds hosts to the known_hosts configmap if using a custom git provider for gitrepos
+# additionalKnownHosts:
```
weyfonk: The value needs to be defined to enable config map `known-hosts` to be built.
```suggestion
additionalKnownHosts: {}
```

charts/fleet/values.yaml
```diff
@@ -94,6 +94,11 @@ priorityClassName: ""
 ## host verification fails. Insecure when set to true. Default: false.
 insecureSkipHostKeyChecks: false
 
+## Adds hosts to the known_hosts configmap if using a custom git provider for gitrepos
+# additionalKnownHosts: 
+#   - '[gitlab.internal.co]:2222 ecdsa-sha2-nistp256 AAANhLXNoYTItbmlzdHAyNTdHAyNTYAAABBBJ6cPcyaL4gbQHb1+3EFuDhT1yqmH/IWid60KddIcPLKG6QRnU0UVCZb/SqvM7YqrAMWTJU3382yUbIcvhkZZnU='
+#   - 'github.com ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIOMqqnkVzrm0SdG6UOoqKLsabgH5C9okWi0dh2l9GKJl'
```
weyfonk: nit: For the sake of clarity, how about something like this instead?
```suggestion
# Custom entries to be added in SSH known_hosts format (see charts/fleet/templates/configmap_known_hosts.yaml for examples):
#   - <host>[:port] <key-algorithm> <fingerprint>
```

SebPlv:
Thank you @weyfonk,

I updated the PR with your suggestions.

SebPlv:
I tested locally the indentation changes proposed by Copilot and it works fine.

Issue #4385, opened 2025-11-28
Dump secrets and Content resources in Fleet CLI


weyfonk:
Follow-up to #4030.

The `fleet dump` CLI should provide flags to optionally enable collection of secrets and `Content` resources, which will not be collected by default (e.g. `--with-secrets`, `--with-contents`).

Each of these could even be declined into 2 variants, e.g. `--with-secrets`, collecting full secrets, and `--with-secrets-metadata`, collecting only their metadata (creation time, deletion timestamp, finalizers, etc), without their actual contents.

thardeck:
### QA Template

```
fleetcli dump --help
Dump state of upstream Fleet-managed resources into an archive

Usage:
  fleet dump [flags]

Flags:
      --with-content            Include Content resources with full data
      --with-content-metadata   Include Content resources with metadata only
      --with-secrets            Include secrets with full data
      --with-secrets-metadata   Include secrets with metadata only
```

sbulage:
### System Information
| Fleet CLI Version |
|--------|
| fleet version v0.15.0-beta.4 (cd904248e365b0a8088d853ae586c3f77cfbe96b) | 

### Verification
- Same environment is used to verify `fleet dump` command with various available options.

| Test Scenario | Test Result | Checks |
|--------|--------|--------|
| `fleet dump --with-content -p  fleet-dump-with-content-metadata.tgz ` | All resources collected from `fleet-local` along with `metadata`  except `secrets`| Collected resources: `bundles, bundledeployments, clustergroups, clusters, gitrepos, helmops, metrics and events`  except `secrets`|
| `fleet dump --with-content -p  fleet-dump-with-content-metadata.tgz -n fleet-default` | All resources are collected from `fleet-default` along with `metadata` except `secrets` | Collected resources: `bundles, bundledeployments, clustergroups, clusters, gitrepos, helmops, metrics and events` except `secrets`|
| `fleet dump --with-content-metadata -p  fleet-dump-with-content-metadata.tgz ` | Only `metadata:` of resources collected from `fleet-local` except `secrets`| Collected `metadata` of `bundles, bundledeployments, clustergroups, clusters, gitrepos, helmops, metrics and events`  except `secrets`|
[...]

PR #4570, opened 2026-01-27
Add optional flags for secrets and contents in fleet cli dump

thardeck:
Adds four new flags to fleet dump command:
- --with-secrets: collects secrets with full data
- --with-secrets-metadata: collects secrets without sensitive data
- --with-content: collects Content resources with full data
- --with-content-metadata: collects Content resources metadata only

By default, secrets and Content resources are excluded. The flags give users control over what sensitive information gets included.

Refers to #4385

### DIFF
--- a/integrationtests/cli/dump/dump_test.go
+++ b/integrationtests/cli/dump/dump_test.go
@@ -487,10 +487,8 @@ var _ = Describe("Fleet dump", func() {
 			Expect(c.Name).To(Equal("my-content"))
 			// Content field should be stripped in metadata-only mode
 			Expect(c.Content).To(BeNil())
-			// SHA256Sum and Status are metadata, should be preserved
+			// SHA256Sum is metadata and should be preserved
 			Expect(c.SHA256Sum).To(Equal("abc123def456"))
-			// Verify Status field structure is present (even if zero-valued)
-			Expect(c.Status.ReferenceCount).To(Equal(0))
 		})
 	})
 
--- a/integrationtests/cli/dump/suite_test.go
+++ b/integrationtests/cli/dump/suite_test.go
@@ -57,7 +57,7 @@ var _ = AfterSuite(func() {
 
 // fleetDump simulates fleet dump CLI online execution
 func fleetDump(path string) error {
-	return dump.Create(context.Background(), cfg, path)
+	return dump.Create(context.Background(), cfg, path, dump.Options{})
 }
 
 func fleetDumpWithOptions(path string, opts dump.Options) error {
--- a/internal/cmd/cli/dump/dump.go
+++ b/internal/cmd/cli/dump/dump.go
@@ -44,7 +44,7 @@ type Options struct {
 	WithContentMetadata bool
 }
 
-func Create(ctx context.Context, cfg *rest.Config, path string, opts ...Options) error {
+func Create(ctx context.Context, cfg *rest.Config, path string, opt Options) error {
 	c, err := createClient(cfg)
 	if err != nil {
 		return fmt.Errorf("failed to create Kubernetes client: %w", err)
@@ -55,10 +55,6 @@ func Create(ctx context.Context, cfg *rest.Config, path string, opts ...Options)
 		return fmt.Errorf("failed to create dynamic Kubernetes client: %w", err)
 	}
 
-	var opt Options
-	if len(opts) > 0 {
-		opt = opts[0]
-	}
 	return CreateWithClients(ctx, cfg, d, c, path, opt)
 }
 
--- a/internal/cmd/cli/dump/dump_test.go
+++ b/internal/cmd/cli/dump/dump_test.go
@@ -365,30 +365,34 @@ func Test_addSecretsToArchive(t *testing.T) {
 			if c.metadataOnly && len(c.secrets) > 0 {
 				tw.Close()
 				tr := tar.NewReader(&buf)
-				_, err := tr.Next()
-				if err != nil {
-					t.Fatalf("failed to read tar header: %v", err)
-				}
-				data, err := io.ReadAll(tr)
-				if err != nil {
-					t.Fatalf("failed to read tar content: %v", err)
-				}
 
-				var secret corev1.Secret
-				if err := yaml.Unmarshal(data, &secret); err != nil {
-					t.Fatalf("failed to unmarshal secret: %v", err)
-				}
+				// Validate all secrets in the archive
+				for i := 0; i < len(c.secrets); i++ {
+					_, err := tr.Next()
+					if err != nil {
+						t.Fatalf("failed to read tar header for secret %d: %v", i, err)
+					}
+					data, err := io.ReadAll(tr)
+					if err != nil {
+						t.Fatalf("failed to read tar content for secret %d: %v", i, err)
+					}
 
-				if secret.Data != nil {
-					t.Errorf("expected Data field to be nil in metadata-only mode, got %v", secret.Data)
-				}
+					var secret corev1.Secret
+					if err := yaml.Unmarshal(data, &secret); err != nil {
+						t.Fatalf("failed to unmarshal secret %d: %v", i, err)
+					}
 
-				// Verify metadata is preserved
-				if secret.Name != c.secrets[0].Name {
-					t.Errorf("expected Name to be preserved in metadata-only mode, got %v", secret.Name)
-				}
-				if secret.Namespace != c.secrets[0].Namespace {
-					t.Errorf("expected Namespace to be preserved in metadata-only mode, got %v", secret.Namespace)
+					if secret.Data != nil {
+						t.Errorf("expected Data field to be nil in metadata-only mode for secret %d, got %v", i, secret.Data)
+					}
+
+					// Verify metadata is preserved
+					if secret.Name != c.secrets[i].Name {
+						t.Errorf("expected Name %q to be preserved in metadata-only mode for secret %d, got %v", c.secrets[i].Name, i, secret.Name)
+					}
+					if secret.Namespace != c.secrets[i].Namespace {
+						t.Errorf("expected Namespace %q to be preserved in metadata-only mode for secret %d, got %v", c.secrets[i].Namespace, i, secret.Namespace)
+					}
 				}
 			}
 		})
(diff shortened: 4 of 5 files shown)


internal/cmd/cli/dump/dump.go
```diff
@@ -150,6 +180,90 @@ func addFileToArchive(data []byte, name string, w *tar.Writer) error {
...
+			// Only strip the actual content (manifests), keep sha256sum and status as metadata
+			i.Object["content"] = nil
+		}
+
+		g, err := yaml.Marshal(&i)
+		if err != nil {
+			return fmt.Errorf("failed to marshal content: %w", err)
+		}
+
+		fileName := fmt.Sprintf("contents_%s_%s", i.GetNamespace(), i.GetName())
```
weyfonk: Content resources are not namespaced, therefore a typical file name would be `contents__s-19fa84383f1ee3e85b99763d473753f54f7e804ba4aed4bba9eb6b038c75d` (notice the double underscore)
```suggestion
		fileName := fmt.Sprintf("contents_%s", i.GetName())
```

internal/cmd/cli/dump/dump.go
```diff
@@ -48,12 +55,20 @@ func Create(ctx context.Context, cfg *rest.Config, path string) error {
 		return fmt.Errorf("failed to create dynamic Kubernetes client: %w", err)
 	}
 
-	return CreateWithClients(ctx, cfg, d, c, path)
+	var opt Options
+	if len(opts) > 0 {
+		opt = opts[0]
```
weyfonk: Why do we need this? :thinking:

internal/cmd/cli/dump/dump.go
```diff
@@ -48,12 +55,20 @@ func Create(ctx context.Context, cfg *rest.Config, path string) error {
...
+	}
+	return CreateWithClients(ctx, cfg, d, c, path, opt)
 }
 
-func CreateWithClients(ctx context.Context, cfg *rest.Config, d dynamic.Interface, c client.Client, path string) error {
+func CreateWithClients(ctx context.Context, cfg *rest.Config, d dynamic.Interface, c client.Client, path string, opts ...Options) error {
+	var opt Options
+	if len(opts) > 0 {
+		opt = opts[0]
+	}
```
weyfonk: Same question as above.

internal/cmd/cli/dump/dump.go
```diff
@@ -81,6 +95,22 @@ func CreateWithClients(ctx context.Context, cfg *rest.Config, d dynamic.Interfac
 		}
 	}
 
+	if opt.WithContent || opt.WithContentMetadata {
+		// If both full content and metadata-only are requested, prefer full content
+		contentMetadataOnly := opt.WithContentMetadata && !opt.WithContent
```
weyfonk, line 91: This check may seem redundant, since the flags parser already checks that both arguments are not provided together. Yet it could be useful if this logic ends up being made available as a library.
(same remark about secrets)

internal/cmd/cli/dump/dump.go
```diff
@@ -150,6 +180,90 @@ func addFileToArchive(data []byte, name string, w *tar.Writer) error {
...
+
+func addSecretsToArchive(
+	ctx context.Context,
+	dynamic dynamic.Interface,
+	c client.Client,
+	logger logr.Logger,
+	w *tar.Writer,
+	metadataOnly bool,
+) error {
+	nss, err := getNamespaces(ctx, dynamic, logger)
```
weyfonk, line 222: Good point, we may indeed not want to collect secrets living in _all_ namespaces :+1:

integrationtests/cli/dump/suite_test.go
```diff
@@ -5,7 +5,7 @@ import (
 	"testing"
 
 	"github.com/rancher/fleet/integrationtests/utils"
-	"github.com/rancher/fleet/internal/cmd/cli/dump"
+	dumppkg "github.com/rancher/fleet/internal/cmd/cli/dump"
```
weyfonk: Why do we need this?

internal/cmd/cli/dump/dump.go
```diff
@@ -150,6 +180,90 @@ func addFileToArchive(data []byte, name string, w *tar.Writer) error {
...
+		var secrets corev1.SecretList
+		if err := c.List(ctx, &secrets, client.InNamespace(ns)); err != nil {
+			merr = append(merr, fmt.Errorf("failed to list secrets for namespace %q: %w", ns, err))
+			continue
+		}
+
+		for _, secret := range secrets.Items {
+			if metadataOnly {
+				secret.Data = nil
+				secret.StringData = nil
```
weyfonk: nit: I _think_ we could do without this field, as it is write-only and its values would have been merged into `Data` when writing the secret to the cluster anyway.

internal/cmd/cli/dump/dump_test.go
```diff
@@ -268,3 +272,194 @@ func Test_addMetrics(t *testing.T) {
 		})
 	}
 }
+
+func Test_addSecretsToArchive(t *testing.T) {
+	cases := []struct {
+		name         string
+		secrets      []corev1.Secret
+		clusters     []v1alpha1.Cluster
```
weyfonk: This field doesn't seem to be used... let's either remove it, or add test cases with/without clusters? Although the latter would not make much functional sense, we'd cover collection of secrets in namespaces other than cluster registration namespaces.

internal/cmd/cli/dump/dump_test.go
```diff
@@ -268,3 +272,194 @@ func Test_addMetrics(t *testing.T) {
...
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "my-content",
+						Namespace: "default",
+					},
+					Content:   []byte("test-content-data"),
+					SHA256Sum: "abc123def456",
+				},
+			},
+			metadataOnly: true,
+		},
```
weyfonk, line 439: Something looks off to me about these two test cases, between their respective names, `contents` and `metadataOnly` fields. Should `contents with full data` provide a full-blown Content resources, including data?

internal/cmd/cli/dump/dump_test.go
```diff
@@ -268,3 +272,194 @@ func Test_addMetrics(t *testing.T) {
...
+
+			err := addSecretsToArchive(ctx, fakeDynClient, mockClient, logger, tw, c.metadataOnly)
+
+			if (err == nil) != (c.expErrStr == "") {
+				t.Fatalf("expected err %s, \n\tgot %s", c.expErrStr, err)
+			}
+
+			if err != nil && !strings.Contains(err.Error(), c.expErrStr) {
+				t.Fatalf("expected error containing %q, got %q", c.expErrStr, err)
+			}
```
weyfonk, line 362: Shouldn't we check here whether data is included in the dump depending on `c.metadataOnly`, as done below for Content resources?

integrationtests/cli/dump/dump_test.go
```diff
@@ -336,6 +335,302 @@ var _ = Describe("Fleet dump", func() {
...
+			found, fileContent, err := findFileInArchive(tgzPath, "contents_")
+			Expect(err).ToNot(HaveOccurred())
+			Expect(found).To(BeTrue())
+
+			var c fleet.Content
+			Expect(yaml.Unmarshal(fileContent, &c)).ToNot(HaveOccurred())
+			Expect(c.Name).To(Equal("my-content"))
+			// Content field should be stripped in metadata-only mode
+			Expect(c.Content).To(BeNil())
+			// SHA256Sum and Status are metadata, should be preserved
```
weyfonk: ... so let's check for a non-empty status here? :thinking:

integrationtests/cli/dump/dump_test.go
```diff
@@ -397,3 +692,40 @@ func mustCreateNS(ns string) {
 	}
 	Expect(client.IgnoreAlreadyExists(k8sClient.Create(ctx, &toCreate))).NotTo(HaveOccurred())
 }
+
+// findFileInArchive searches for a file with the given prefix in the tar archive
+// and returns true along with the file contents if found
```
weyfonk:
```suggestion
// and returns true along with the first file matching the prefix, if found, and its contents
```

internal/cmd/cli/dump/dump.go
thardeck: I have changed the number of parameters of an existing function which in theory could break existing callers. (Most likely not relevant in this case)
This pattern would provide backward compatibility without the need for an additional function.

I can remove it, if you prefer it.

internal/cmd/cli/dump/dump.go
thardeck: Yes, you should not end up in this situation but at the same time it should not hurt. Also the integration tests do not verify the user facing parameter check.

integrationtests/cli/dump/suite_test.go
thardeck: Good point, I think initially I had a conflicting variable or function, not sure anymore. Will remove it.

integrationtests/cli/dump/dump_test.go
thardeck: Status is a subresource in Kubernetes, I think we can't set it during resource creation without additional status update calls.

weyfonk: Good point; in this case, I don't think that checking an unmarshalled struct (in this case a Content resource) for a zeroed status makes much sense, since the status would most likely be created by the unmarshalling operation, even if the source marshalled data didn't contain it.

internal/cmd/cli/dump/dump.go
weyfonk: Oh, I get it now. I don't think that updating `Create` calls elsewhere, even to add empty options, would have been an issue. I think it would be clearer than accepting a variable number of options, where only the first one would be taken into account anyway.

internal/cmd/cli/dump/dump_test.go
```diff
@@ -268,3 +272,221 @@ func Test_addMetrics(t *testing.T) {
...
+
+			// For metadata-only test, verify sensitive fields are stripped
+			if c.metadataOnly && len(c.secrets) > 0 {
```
weyfonk, line 365: nit: this only validates the _first_ secret, which is all we need for the currently implemented test cases, but this will break if we decide to cover dumping more than one secret.

Issue #4575, opened 2026-01-28
[backport v2.13] DownstreamResources don't work for the local cluster


rancherbot:
This is a backport issue for #4559, automatically created via [GitHub Actions workflow](https://github.com/rancher/fleet/actions/runs/21442150441) initiated by @0xavi0

Original issue body:

There is a problem when copying Downstream Resources into the local cluster, because `DeepCopy()` is being used, which also copies the metadata (including the UID).

When calling `CreateOrUpdate`, the function does not return any error, but internally Kubernetes is not performing the copy because a resource with the same ID already exists.

sbulage:
### System Information
| Rancher Version | Fleet Version |
|--------|--------|
| v2.13-4795da16e8c26817e1d10d38a7538094ee92e382-head | 108.0.3+up0.14.3-beta.1 | 


### Verification steps:
- Enabled `EXPERIMENTAL_COPY_RESOURCES_DOWNSTREAM`
  - <details><summary>Enable Experimental feature</summary>
    
    ```
    extraEnv:
      - name: EXPERIMENTAL_COPY_RESOURCES_DOWNSTREAM
        value: "true"
    ```
    </details>

---
- Create ConfigMap using below command:
  -  ConfigMap data file `values.yaml`
      ```yaml
      name: testing-downstream-cluster-configmap
      ```
      ```bash
      kubectl create configmap -n fleet-local config-values --from-file=values.yaml=values.yaml
      ```

---
- Below `HelmOps` YAML used
  - <details><summary>HelmOps YAML</summary>
    
    ```yaml
    apiVersion: fleet.cattle.io/v1alpha1
    kind: HelmOp
    metadata:
      name: test-values-from
      namespace: fleet-local
    spec:
      correctDrift: {}
      downstreamResources:
        - kind: configmap
          name: config-values
      helm:
        chart: >-
          https://github.com/rancher/fleet/raw/refs/heads/main/integrationtests/cli/assets/helmrepository/config-chart-0.1.0.tgz
        valuesFrom:
          - configMapKeyRef:
              key: values.yaml
              name: config-values
      pollingInterval: 5s
      targets:
        - clusterSelector:
            matchExpressions:
              - key: provider.cattle.io
                operator: NotIn
                values:
                  - harvester
    ```
    </details> 
---

### Verification Screenshots 

| Rancher 2.13.2 | Rancher 2.13-head (Fixed) |
|--------|--------|
[...]

PR #4576, opened 2026-01-28
[v0.14] - Avoid using DeepCopy when copying Downstream Resources (#4560)

0xavi0:
* Avoid using DeepCopy when copying Downstream Resources

This PR avoids copying ConfigMaps and Secrets using `DeepCopy`, since this is problematic when you want to copy the object into a different namespace but within the same cluster (this usually happens when using a single-cluster configuration).

E2E tests have been added to verify the copying of downstream resources also in a single-cluster environment, by reusing the multi-cluster tests and adapting them.

Refers to: https://github.com/rancher/fleet/issues/4575
Backport of: https://github.com/rancher/fleet/pull/4560

---------

### Checklist

### DIFF
--- a/e2e/assets/single-cluster/helmop_downstream_resources.yaml
+++ b/e2e/assets/single-cluster/helmop_downstream_resources.yaml
@@ -1,0 +1,36 @@
+apiVersion: fleet.cattle.io/v1alpha1
+kind: HelmOp
+metadata:
+  name: {{.Name}}
+  namespace: {{.Namespace}}
+spec:
+  helm:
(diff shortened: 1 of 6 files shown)


thardeck (changes requested):
Function-wise it looks good but it seems that not all tests were backported. Was this on purpose?

0xavi0:
> Function-wise it looks good but it seems that not all tests were backported. Was this on purpose?

Yeah, some of the tests are about monitoring DownstreamResources secrets that's not backported.

Issue #4534, opened 2026-01-14
CLI: Generate BundleDiff From Status


weyfonk:
The Fleet CLI should include a command to generate bundle diffs from a resource (bundle or bundle deployment) status. The resource status already contains the needed patch.

EDIT
Implement output of patches in a format that can be directly inserted into fleet.yaml files and uses the comparePatches structure.

thardeck:
# QA Template:

## Test Cases

### Test 1: Standard Output (Default)
**Objective**: Verify default bundlediff output shows human-readable diff information

**Steps**:
1. Deploy a bundle with a modified resource (e.g., ConfigMap with extra annotation)
2. Run: `fleet bundlediff --bundle-deployment <name> -n <namespace>`
3. Verify output shows bundle deployment name, bundle name, and modified resources with their patches

**Expected Result**:
```
BundleDeployment: cluster-fleet-local-abc123/my-bundle-deployment
Bundle: my-bundle

Modified Resources:
  ConfigMap.v1 default/my-config
    Status: Modified
    Patch: {"metadata":{"annotations":{"timestamp":"2024-01-15T10:30:00Z"}}}
```

---

### Test 2: JSON Output
**Objective**: Verify `--json` flag produces valid JSON output

**Steps**:
1. Deploy a bundle with a modified resource
2. Run: `fleet bundlediff --bundle-deployment <name> -n <namespace> --json`
3. Verify output is valid JSON: `fleet bundlediff --bundle-deployment <name> -n <namespace> --json | jq .`

**Expected Result**:
- Output is valid JSON that can be parsed by `jq`
- Contains bundleDeploymentName, bundleName, namespace, and modifiedResources array

---

### Test 3: Fleet YAML Output
**Objective**: Verify `--fleet-yaml` flag produces valid YAML with comparePatches format

**Steps**:
1. Deploy a bundle with a modified resource
2. Run: `fleet bundlediff --bundle-deployment <name> -n <namespace> --fleet-yaml`
3. Verify output contains:
   - `diff:` section
   - `comparePatches:` list
   - Resource metadata (apiVersion, kind, name, namespace)
   - Operations with `op: remove` and correct `path`

**Expected Result**:
```yaml
diff:
  comparePatches:
  - apiVersion: v1
    kind: ConfigMap
    name: my-config
    namespace: default
    operations:
    - op: remove
      path: /metadata/annotations/timestamp
```

---

### Test 4: GitOps Drift Acceptance Workflow
[...]

PR #4581, opened 2026-01-30
Add fleet-yaml output format to bundlediff command

thardeck:
Implement --fleet-yaml flag that outputs patches in a format that can be directly inserted into fleet.yaml files. The output uses the comparePatches structure.

Refers to #4534

### DIFF
--- a/integrationtests/cli/bundlediff/testdata/fleet-yaml-multiple-ops.yaml
+++ b/integrationtests/cli/bundlediff/testdata/fleet-yaml-multiple-ops.yaml
@@ -7,5 +7,6 @@ diff:
     operations:
     - op: replace
       path: /spec/replicas
+      value: "3"
     - op: remove
       path: /spec/template/metadata/labels/old-label
--- a/integrationtests/cli/bundlediff/testdata/fleet-yaml-multiple-resources.yaml
+++ b/integrationtests/cli/bundlediff/testdata/fleet-yaml-multiple-resources.yaml
@@ -7,6 +7,7 @@ diff:
     operations:
     - op: add
       path: /data/key1
+      value: '"val1"'
   - apiVersion: v1
     kind: Secret
     name: secret1
--- a/internal/cmd/cli/bundlediff.go
+++ b/internal/cmd/cli/bundlediff.go
@@ -102,6 +102,10 @@ func (d *BundleDiff) PersistentPre(_ *cobra.Command, _ []string) error {
 }
 
 func (d *BundleDiff) Run(cmd *cobra.Command, args []string) error {
+	if d.JSON && d.FleetYAML {
+		return fmt.Errorf("cannot specify both --json and --fleet-yaml")
+	}
+
 	cfg, err := ctrl.GetConfig()
 	if err != nil {
 		return fmt.Errorf("failed to get k8s config: %w", err)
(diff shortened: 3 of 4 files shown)


internal/cmd/cli/bundlediff.go
```diff
@@ -161,6 +176,66 @@ func (d *BundleDiff) encodeJSON(out io.Writer, diffs []DiffOutput) error {
...
+	for _, diff := range diffs {
+		for _, mod := range diff.ModifiedResources {
+			if mod.Patch == "" {
+				continue
+			}
+
+			var patchOps []struct {
+				Op   string `json:"op"`
+				Path string `json:"path"`
+			}
```
weyfonk, line 196: As per [this definition](https://github.com/rancher/fleet/blob/main/pkg/apis/fleet.cattle.io/v1alpha1/bundledeployment_types.go#L173-L183):
```suggestion
			var patchOps []struct {
				Op    string `json:"op"`
				Path  string `json:"path"`
				Value string `json:"value,omitempty"`
			}
```

internal/cmd/cli/bundlediff.go
```diff
@@ -161,6 +176,66 @@ func (d *BundleDiff) encodeJSON(out io.Writer, diffs []DiffOutput) error {
...
+			if err := json.Unmarshal([]byte(mod.Patch), &patchOps); err != nil {
+				continue
+			}
+
+			operations := make([]fleet.Operation, 0, len(patchOps))
+			for _, p := range patchOps {
+				if p.Op != "" {
+					operations = append(operations, fleet.Operation{
+						Op:   p.Op,
+						Path: p.Path,
```
weyfonk:
```suggestion
						Path:  p.Path,
						Value: p.Value,
```

integrationtests/cli/bundlediff/testdata/fleet-yaml-multiple-ops.yaml
```diff
@@ -0,0 +1,11 @@
+diff:
+  comparePatches:
+  - apiVersion: apps/v1
+    kind: Deployment
+    name: my-deployment
+    namespace: test-ns
+    operations:
+    - op: replace
+      path: /spec/replicas
```
weyfonk, line 9:
```suggestion
      path: /spec/replicas
      value: 3
```

integrationtests/cli/bundlediff/testdata/fleet-yaml-multiple-resources.yaml
```diff
@@ -0,0 +1,16 @@
+diff:
+  comparePatches:
+  - apiVersion: v1
+    kind: ConfigMap
+    name: cm1
+    namespace: ns1
+    operations:
+    - op: add
+      path: /data/key1
```
weyfonk, line 9:
```suggestion
      path: /data/key1
      value: val1
```

internal/cmd/cli/bundlediff.go
```diff
@@ -186,21 +190,39 @@ func (d *BundleDiff) printFleetYAML(out io.Writer, diffs []DiffOutput) error {
 			}
 
 			var patchOps []struct {
-				Op   string `json:"op"`
-				Path string `json:"path"`
+				Op    string      `json:"op"`
+				Path  string      `json:"path"`
+				Value interface{} `json:"value,omitempty"`
```
weyfonk, line 195: This is actually better than what Fleet's operations support on compare patches, where `Value` is a string... Perhaps something we should revisit in the future.

Issue #4582, opened 2026-01-30
Enforce validation at `fleet.yaml` parsing time when bundle are created from gitrepo


mmartin24:
Enhancement after: https://github.com/rancher/fleet/issues/2552

---

Ref [comment](https://github.com/rancher/fleet/issues/2552#issuecomment-3824397825)

While testing this item it was noted that the statuses from the fleet documentation are at times not exactly as in the UI. For instance, in the documentation the state `NotReady` (without space) is represented in the UI as Not Ready (with space). Another example is if in the UI we pause the gitrepo; it will display `Pause` in the state, however the correct state to be passed in the `dependsOn` is  `OutOfSync`. If an user passes `Paused`  by error as in the screnshot the gitrepo will not be deployed  and will silently fail.:

<img width="2522" height="722" alt="Image" src="https://github.com/user-attachments/assets/5eaea33a-15d1-4e8a-91ed-c9ac12ff388b" />

However, if we pass `OutOfSync`, then it will be deployed:

<img width="2522" height="722" alt="Image" src="https://github.com/user-attachments/assets/e1675577-8a58-4df5-8f4e-c79877f110db" />

To enforce this Fleet could enforce validation at fleet.yaml parsing time when bundle are created from gitrepo

mmartin24:
Verified in Rancher `v2.14-f8f53a596830f5e51217073cd5527cac9612eb01-head` with Fleet `0.15.0-alpha.9`

---

When passed an invalid / non existing accepted state, the dependant repo is not deployed and a message indicating both the  errored state and the accepted one is displayed.

In this example I passed as incorrect state `BadState` and the dependant repo is not deployed even when it has an `OutOfSync` correct state with the main repo in such a state:

```yaml
# Leaf bundle - depends on root with acceptedStates
# This bundle will deploy even if root is in Modified state
defaultNamespace: depends-on-accepted-states-leaf

labels:
  role: leaf-accepted-states
  test: accepted-states

dependsOn:
  - selector:
      matchLabels:
        role: root-modified
    # Accept both Ready and Modified states
    # This allows the leaf to deploy even when root has drifted
    acceptedStates:
    - OutOfSync
    - BadState
```

Error displayed:


```
Failed to process bundle: validating fleet.yaml: dependsOn[0].acceptedStates[1]: invalid state "BadState", valid values are: [Ready NotReady Pending OutOfSync Modified WaitApplied ErrApplied]
```

[...]

PR #4584, opened 2026-02-01
fleet.yaml validation (dependsOn part)

ncuralli:
Refers to #4582

## Summary

This PR adds validation for `fleet.yaml` files to catch invalid `acceptedStates` values in `dependsOn` entries before Bundle creation. Previously, invalid states would only be detected at runtime in the agent, making debugging difficult.

## Problem

When a user specifies an invalid `acceptedStates` value in their `fleet.yaml`:

```yaml
dependsOn:
- name: my-dependency
  acceptedStates:
  - Ready
  - InvalidState  # typo or invalid value
```

The Bundle would be created without any error, and the problem would only surface later when the agent tried to process the dependency. This made it hard to identify configuration issues.

## Design decisions

1. **Validation derives valid states from `StateRank`** - This ensures the validation stays in sync if new states are added in the future, without requiring manual updates to a separate list.

2. **States are sorted by rank in error messages** - Provides consistent, predictable error output.

3. **Validation happens early** - By validating during `fleet apply` (before Bundle creation), users get immediate feedback when using the CLI or when the GitJob processes their repository.


internal/bundlereader/validate_fleetyaml.go
```diff
@@ -0,0 +1,60 @@
...
+// validateFleetYAML validates the semantic content of a parsed FleetYAML.
+// It returns an error if any field contains invalid values.
+func validateFleetYAML(fy *fleet.FleetYAML) error {
+	// Validate DependsOn entries at the bundle level
+	for i, dep := range fy.DependsOn {
+		if err := validateBundleRef(i, dep); err != nil {
+			return err
+		}
+	}
+
```
weyfonk, line 19: Nice, as a side note this would be a great place to validate regex for bundle diffs in the future, as noted [here](https://github.com/rancher/fleet/pull/4563/changes#diff-57d701918a455bfacc0f493a1de52b3b0921c1240eb2c2911e26b61236daa092R57-R58).

integrationtests/cli/apply/apply_test.go
```diff
@@ -815,3 +815,151 @@ func getBundleNamed(name string, bundles []*v1alpha1.Bundle) *v1alpha1.Bundle {
...
+			err := fleetApply(name, dirs, options)
+			Expect(err).NotTo(HaveOccurred())
+
+			bundle, err := cli.GetBundleFromOutput(buf)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(bundle.Spec.DependsOn).To(HaveLen(1))
+			Expect(bundle.Spec.DependsOn[0].Name).To(BeEmpty())
+			Expect(bundle.Spec.DependsOn[0].Selector).NotTo(BeNil())
+			Expect(bundle.Spec.DependsOn[0].Selector.MatchLabels).To(HaveKeyWithValue("app", "database"))
+			Expect(bundle.Spec.DependsOn[0].Selector.MatchLabels).To(HaveKeyWithValue("tier", "backend"))
```
weyfonk, line 906: nit: Should we check that accepted states are as expected here, just to be on the safe side?
(edit: downgraded this to a nitpick, as a similar check is already run as part of the below test case, which already features a selector-based `dependsOn` entry)

ncuralli: extended the check also to the acceptedstate field

Issue #4419, opened 2025-12-10
Improve dump CLI flexibility and efficiency


weyfonk:
Here are a couple of ideas which could be implemented to make collection of resources easier on larger setups:

* instead of dumping all resources, which may result in massive volumes of data for larger setups, the CLI could accept an entrypoint resource (e.g. GitRepo, HelmOp or event a bundle) and navigate through its ownership graph, dumping all relevant resources along the way: BDs and contents, secrets, etc

* when dumping resources to an archive, the CLI currently holds all resources of a given kind in memory at once, before writing them to the archive. Instead, it could write them in batches, either by resource count or by size.

p-se:
## QA Template - Filtering

### What to Test
The `fleet dump` command now supports filtering by namespace. When `--namespace` is specified (without `--all-namespaces`), the archive should contain only resources related to that namespace.

### Setup Requirements
- Fleet installation with GitRepos in at least two namespaces (e.g., `fleet-local` and `fleet-other`)
- Each namespace should have: GitRepo → Bundle → BundleDeployments (in cluster namespaces)
- Verify with: `kubectl get gitrepos,bundles,bundledeployments -A`

### Core Test Scenarios

#### 1. Namespace Filtering
**Command**: `fleet dump -p test.tgz --namespace fleet-local`

**Check These Resources**:
- ✅ GitRepos from `fleet-local` only
- ✅ Bundles from `fleet-local` only  
- ✅ BundleDeployments with label `fleet.cattle.io/bundle-namespace=fleet-local`
- ✅ Clusters/ClusterGroups from `fleet-local` only
- ✅ Contents referenced by `fleet-local` bundles
- ✅ Events from `fleet-local` + system namespaces (kube-system, default, cattle-fleet-*)
- ✅ Secrets from `fleet-local` + system namespaces

#### The Four System Namespaces

1. **`default`** - Kubernetes default namespace
2. **`kube-system`** - Core Kubernetes system components
3. **`cattle-fleet-system`** - Main Fleet controller namespace (config.DefaultNamespace)
4. **`cattle-fleet-local-system`** - Local cluster agent namespace


#### Verification:
```bash
tar -tzf test.tgz | grep -E "(gitrepos|bundles)_fleet-local"  # Should show only fleet-local
tar -tzf test.tgz | grep -E "(gitrepos|bundles)_fleet-other"  # Should show nothing
```

#### 2. All-Namespaces Mode  
[...]

PR #4588, opened 2026-02-02
fleet dump CLI: Internal pagination when fetching resources

p-se:
Instead of holding all resources in memory, write them in batches. Also eases the load on the cluster by only fetching a certain amount of resources at a time.

Also fixes an issue with collecting duplicated namespaces from cluster resources.

Refers to #4419

### Checklist

### DIFF
--- a/internal/cmd/builder.go
+++ b/internal/cmd/builder.go
@@ -108,6 +108,8 @@ func Command(obj Runnable, cmd cobra.Command) *cobra.Command {
 
 		flags := c.PersistentFlags()
 		switch fieldType.Type.Kind() {
+		case reflect.Int64:
+			flags.Int64VarP((*int64)(unsafe.Pointer(v.Addr().Pointer())), name, alias, int64(defInt), usage)
 		case reflect.Int:
 			flags.IntVarP((*int)(unsafe.Pointer(v.Addr().Pointer())), name, alias, defInt, usage)
 		case reflect.String:
@@ -149,6 +151,11 @@ func Command(obj Runnable, cmd cobra.Command) *cobra.Command {
 				case reflect.Int:
 					fv, err := flags.GetInt(name)
 					if err == nil && fv == defInt {
+						_ = flags.Set(name, v)
+					}
+				case reflect.Int64:
+					fv, err := flags.GetInt64(name)
+					if err == nil && fv == int64(defInt) {
 						_ = flags.Set(name, v)
 					}
 				default:
--- a/internal/cmd/cli/dump.go
+++ b/internal/cmd/cli/dump.go
@@ -5,6 +5,7 @@ import (
 	"os"
 
 	"github.com/spf13/cobra"
+
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
@@ -34,6 +35,7 @@ func NewDump() *cobra.Command {
 type Dump struct {
 	FleetClient
 	DumpPath            string `usage:"Destination path for the dump" short:"p"`
+	FetchLimit          int64  `usage:"Limit number of items per resource that are fetched at once (0 means no limit)" short:"l" default:"500"`
 	WithSecrets         bool   `usage:"Include secrets with full data"`
 	WithSecretsMetadata bool   `usage:"Include secrets with metadata only"`
 	WithContent         bool   `usage:"Include Content resources with full data"`
@@ -68,10 +70,15 @@ func (d *Dump) Run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	if d.FetchLimit < 0 {
+		return fmt.Errorf("fetch limit must be non-negative, got %d", d.FetchLimit)
+	}
+
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&zopts)))
 	ctx := log.IntoContext(cmd.Context(), ctrl.Log)
 
 	opts := dump.Options{
+		FetchLimit:          d.FetchLimit,
 		WithSecrets:         d.WithSecrets,
 		WithSecretsMetadata: d.WithSecretsMetadata,
 		WithContent:         d.WithContent,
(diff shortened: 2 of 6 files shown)


internal/cmd/builder.go
```diff
@@ -108,6 +108,8 @@ func Command(obj Runnable, cmd cobra.Command) *cobra.Command {
 
 		flags := c.PersistentFlags()
 		switch fieldType.Type.Kind() {
+		case reflect.Int64:
+			flags.Int64VarP((*int64)(unsafe.Pointer(v.Addr().Pointer())), name, alias, int64(defInt), usage)
 		case reflect.Int:
```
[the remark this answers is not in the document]
p-se: Added to env override logic, don't think we'll ever need to worry about truncating.

internal/cmd/cli/dump/dump.go
```diff
@@ -274,30 +298,45 @@ func getNamespaces(ctx context.Context, dynamic dynamic.Interface, logger logr.L
...
+	}
 
-		res = append(res, c.Namespace)
+	// Convert map to slice
+	res := make([]string, 0, len(nsMap))
+	for ns := range nsMap {
+		res = append(res, ns)
 	}
 
 	return res, nil
```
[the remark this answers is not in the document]
p-se: Added.

internal/cmd/cli/dump.go
```diff
@@ -68,10 +70,15 @@ func (d *Dump) Run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	if d.FetchLimit < 0 {
+		d.FetchLimit = 0
```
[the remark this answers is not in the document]
p-se: Done

internal/cmd/cli/dump/dump.go
```diff
@@ -228,25 +252,30 @@ func addSecretsToArchive(
...
-			merr = append(merr, fmt.Errorf("failed to list secrets for namespace %q: %w", ns, err))
-			continue
-		}
-
-		for _, secret := range secrets.Items {
-			if metadataOnly {
-				secret.Data = nil
+		for {
+			if err := c.List(ctx, &secrets, client.InNamespace(ns), client.Limit(fetchLimit), client.Continue(secrets.Continue)); err != nil {
+				return fmt.Errorf("failed to list secrets for namespace %q: %w", ns, err)
```
weyfonk: This breaks multi-error support. What is the reason for choosing this approach?

internal/cmd/cli/dump/dump.go
```diff
@@ -311,36 +357,49 @@ func addEventsToArchive(
...
-			if err != nil {
-				merr = append(merr, fmt.Errorf("failed to encode event into JSON: %w", err))
+		// Accumulate all pages of events before writing to archive
+		for {
+			if err := c.List(ctx, &NSevts, client.InNamespace(ns), client.Limit(fetchLimit), client.Continue(NSevts.Continue)); err != nil {
+				return fmt.Errorf("failed to list events for namespace %q: %w", ns, err)
 			}
 
-			data = append(data, je...)
+			allEvents = append(allEvents, NSevts.Items...)
```
weyfonk: If we're going to accumulate all events anyway, is there still a point in using pagination?

internal/cmd/cli/dump/dump.go
```diff
@@ -464,26 +531,44 @@ func createDialer(ctx context.Context, cfg *rest.Config, c client.Client, svc *c
...
 	matchingLabels := client.MatchingLabels{
 		"app":    appLabel,
 		shardKey: shardValue,
 	}
+	var pods corev1.PodList
+	for {
+		opts := []client.ListOption{
+			client.InNamespace(svc.Namespace),
+			matchingLabels,
+			client.Limit(fetchLimit),
```
weyfonk, line 602: I'd be surprised if we actually _needed_ this limit here, considering that we're only using it to fetch monitoring services, but I guess it doesn't harm :)

internal/cmd/cli/dump/dump_test.go
```diff
@@ -78,6 +204,154 @@ func Test_getNamespaces(t *testing.T) {
...
+		for _, got := range entries {
+			if got == expected {
+				found = true
+				break
+			}
+		}
+		if !found {
+			t.Fatalf("expected tar entry %q not found", expected)
+		}
+	}
```
weyfonk: nit: this could be simplified as:
```suggestion
    if !slices.Equal(entries, expectedEntries) {
            t.Fatalf("expected tar entries %v, got %v", expectedEntries, entries)
    }
```

internal/cmd/cli/dump/dump.go
p-se: Update the code to paginate, write the result to a temp file and use the temp file for writing to the archive, as tar requires to know the file size ahead in the header. That way we can paginate but don't keep everything in memory.

internal/cmd/cli/dump/dump.go
p-se: Right, fixed it!

internal/cmd/cli/dump/dump_test.go
p-se: Thanks, simplified!

internal/cmd/cli/dump/dump.go
```diff
@@ -311,60 +359,136 @@ func addEventsToArchive(
...
+			// page by page to avoid keeping them all in memory.
+			tmpFile, err := os.CreateTemp("", fmt.Sprintf("events_%s_*.json", ns))
+			if err != nil {
+				return fmt.Errorf("failed to create temp file for events in namespace %q: %w", ns, err)
+			}
+			defer os.Remove(tmpFile.Name()) // Clean up temp file
+			defer tmpFile.Close()           // Close file handle
+
+			var NSevts corev1.EventList
+			eventCount := 0
```
weyfonk: nit: given checks on this variable, do we need an actual count here, or would a boolean (such as `foundEvents`) be enough?

weyfonk (approved):

weyfonk (approved):

PR #4592, opened 2026-02-02
Migrate to GoReleaser dockers_v2 and use direct image references

thardeck:
Migrate from deprecated GoReleaser dockers + docker_manifests to [dockers_v2](https://goreleaser.com/customization/dockers_v2/), add platform-specific image tagging, and pass image digests directly between Github action steps to prevent tag confusion.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -218,7 +218,6 @@ jobs:
 
           while read -r image_with_digest; do
             img_tag="${image_with_digest%@*}"
-            img_digest="${image_with_digest#*@}"
 
             if [[ "${img_tag}" =~ -linux-([^:]+)$ ]]; then
               platform="linux/${BASH_REMATCH[1]}"


.github/workflows/release.yml
```diff
@@ -163,66 +163,88 @@ jobs:
...
-            # Pull the manifest locally to make it available for inspection
-            docker pull "${URL}"
-          done
+          set -euo pipefail
 
-      - name: Attest provenance
-        shell: bash
-        run: |
-          IMAGES=$(yq e '.dockers[].image_templates[0]' .goreleaser.yaml | grep STAGE_REGISTRY | sed "s/{{ .Env.STAGE_REGISTRY }}/${{ env.STAGE_REGISTRY }}/g" | sed "s/{{ .Tag }}/${{ github.ref_name }}/g")
+          > platform_digests.txt
```
weyfonk, line 173: nit: not sure if this still works with a space.
What is the intent here? Redirecting only stdout to `platform_digests.txt`?
```suggestion
          >platform_digests.txt
```

.github/workflows/release.yml
```diff
@@ -163,66 +163,88 @@ jobs:
...
+
+          if [[ ! -f dist/digests.txt ]] || [[ ! -s dist/digests.txt ]]; then
```
weyfonk, line 175: This file is generated by GoReleaser according to [this line](https://github.com/rancher/fleet/pull/4592/changes#diff-7326b55c062b0f46fe9e39aace0a25f4515cf206040fb91a6fd2cae839f5e826R221), isn't it? 
Just validating my understanding here.

.github/workflows/release.yml
```diff
@@ -163,66 +163,88 @@ jobs:
...
-            # Construct fully qualified image name with both tag and digest
-            URL="${IMG_NAME}@${DIGEST}"
+          echo "${staging_images}" | while read -r image_with_digest; do
+            image_url="${image_with_digest%@*}"
+            manifest_digest="${image_with_digest#*@}"
+            image_name="${image_url##*/}"
 
-            max_retries=3
-            retry_delay=5
+            [[ "${image_name}" =~ -linux- ]] && continue
```
weyfonk, line 191: Skipping non-linux images?

.github/workflows/release.yml
```diff
@@ -163,66 +163,88 @@ jobs:
...
-                PLATFORM="linux/${BASH_REMATCH[1]}"
-              else
-                PLATFORM="linux/amd64"
-              fi
+            echo "Creating platform tags for ${image_url} (${manifest_digest})"
 
-              if slsactl download provenance --format=slsav1 --platform="${PLATFORM}" "${URL}" > provenance-slsav1.json; then
-                break
+            docker buildx imagetools inspect "${image_url}@${manifest_digest}" --raw | \
+              jq -r '.manifests[] | select(.platform.os=="linux" and .platform.architecture!="unknown") | "\(.digest)|\(.platform.architecture)"' | \
```
weyfonk, line 196: Perhaps dumb question: is escaping closing parentheses not needed here?

.github/workflows/release.yml
```diff
@@ -163,66 +163,88 @@ jobs:
...
+              while IFS='|' read -r platform_digest arch; do
```
weyfonk, line 197: Are we creating _multiple_ images per source image, i.e. one per platform?
In other words, are source images multiplatform? 🤔

.github/workflows/release.yml
thardeck: It should make sure the file is clean from any previous run - the runner should be in a clean state but better safe then sorry.
I have tried the command in my console - at least in bash it works as intended. I can add an echo in front if you prefer it.

.github/workflows/release.yml
thardeck: Yes, [it](https://github.com/rancher/fleet/blob/improve_image_signing/.goreleaser.yaml#L221) is generated by GoReleaser so we make sure to use the same digests, and not try to get the information from the registry.

If for whatever reason the file might be not there, originally the code would have just finished this part without an error, because nothing to iterate over.

.github/workflows/release.yml
thardeck: We do not build docker images for other platforms at the moment I think.

I hope that we can skip this part in the future anyway, I just want to be backward compatible because the Docker v2. implementation only builds multi arch images.

I am planning to start the discussion soon if arch specific tags are needed for Rancher or our customer.

.github/workflows/release.yml
thardeck: Same as above, GoReleaser creates multi arch images now in one go, but then you can not reference the architecture directly, e.g.: `docker.io/rancher/fleet:v1.0-linux-amd64`, but you have to pull `docker.io/rancher/fleet:v1.0`.

That's why I tag the architectures manually. To do this in GoReleaser I would need to create a release for every architecture.

And most likely we do not even need that arch tag, but I will clarify that.

.github/workflows/release.yml
thardeck: No, the closing parentheses don't need escaping here. The entire jq expression is wrapped in single quotes ('...'), which means bash treats everything inside as literal text.

The `\(.digest)` and `\(.platform.architecture)` syntax is jq's string interpolation format, not bash syntax. The backslash-parenthesis combination is part of jq's templating language and is passed literally to `jq` by bash because of the single quotes.

.github/workflows/release.yml
weyfonk: Thanks for clarifying; I'm feeling more confident knowing that this has been tested locally :)

.github/workflows/release.yml
weyfonk: Understood, thanks for clarifying!

PR #4606, opened 2026-02-06
Add bundlediff --fleet-yaml for GitOps drift acceptance

thardeck:
Implements `--fleet-yaml` flag for bundlediff CLI to generate fleet.yaml
snippets that accept drift in GitOps workflows.

Key changes:
- `bundlediff --fleet-yaml` outputs diff.comparePatches snippets
- Align bundlediff yaml ordering
- Patches use 'remove' operations to ignore fields in drift detection
- Accepts any value for removed fields (not just specific values)
- Add `onlyDiffOptionsChanged()` to skip unavailable counting for diff-only changes
- Diff-only changes don't trigger Helm upgrades or rollout blocking
- Ensure diff option changes requeue to clear drift ignores

The 'remove' operation semantic means "ignore this field during drift
detection" - allowing any drift on that field without reporting it as
Modified. This differs from 'replace' which would only accept one
specific value.

```
Example output:
```yaml
diff:
  comparePatches:
  - apiVersion: v1
    kind: ConfigMap
    name: my-config
    namespace: my-namespace
    operations:
    - op: remove
      path: /data/key
```

Refers #4534

### DIFF
--- a/e2e/assets/single-cluster/bundlediff-fleet-snippet.yaml
+++ b/e2e/assets/single-cluster/bundlediff-fleet-snippet.yaml
@@ -3,7 +3,9 @@ diff:
   - apiVersion: v1
     kind: ConfigMap
     name: test-bundlediff-cm
-    namespace: TARGET_NS
+    namespace: ${TARGET_NAMESPACE}
+    jsonPointers:
+    - /data/key
     operations:
     - op: remove
       path: /data/key
--- a/e2e/single-cluster/cli_bundlediff_test.go
+++ b/e2e/single-cluster/cli_bundlediff_test.go
@@ -107,32 +107,27 @@ var _ = Describe("Fleet bundlediff CLI", func() {
 			Expect(err).ToNot(HaveOccurred(), string(errBuf.Contents()))
 
 			diffSnippet := string(buf.Contents())
-			actualOutput := strings.TrimSpace(diffSnippet)
-			Expect(actualOutput).To(ContainSubstring("diff:"))
-			Expect(actualOutput).To(ContainSubstring("comparePatches:"))
-			Expect(actualOutput).To(ContainSubstring("apiVersion: v1"))
-			Expect(actualOutput).To(ContainSubstring("kind: ConfigMap"))
-			Expect(actualOutput).To(ContainSubstring("name: test-bundlediff-cm"))
-			Expect(actualOutput).To(ContainSubstring("operations:"))
-			Expect(actualOutput).To(SatisfyAny(
-				ContainSubstring("op: remove"),
-				ContainSubstring("op: replace"),
-			))
-			Expect(actualOutput).To(ContainSubstring("path: /data/key"))
-			Expect(actualOutput).To(SatisfyAny(
-				ContainSubstring("namespace: "+targetNs),
-				Not(ContainSubstring("namespace:")),
-			))
+			expectedBytes, err := os.ReadFile(testenv.AssetPath("single-cluster/bundlediff-fleet-snippet.yaml"))
+			Expect(err).ToNot(HaveOccurred())
+			expectedOutput := strings.ReplaceAll(string(expectedBytes), "${TARGET_NAMESPACE}", targetNs)
+
+			var actualYAML map[string]interface{}
+			err = yaml.Unmarshal([]byte(diffSnippet), &actualYAML)
+			Expect(err).ToNot(HaveOccurred())
+			var expectedYAML map[string]interface{}
+			err = yaml.Unmarshal([]byte(expectedOutput), &expectedYAML)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(actualYAML).To(Equal(expectedYAML))
 
 			By("verifying comparePatches hide the drift when applied to Bundle")
 			jsonBytes, err := yaml.YAMLToJSON([]byte(diffSnippet))
 			Expect(err).ToNot(HaveOccurred())
-			patchJSON := fmt.Sprintf(`{"spec":{"options":%s}}`, string(jsonBytes))
-			_, err = k.Namespace(bdNamespace).Patch("bundledeployment", bundleDeploymentName, "--type=merge", "-p", patchJSON)
+			patchJSON := fmt.Sprintf(`{"spec":%s}`, string(jsonBytes))
+			_, err = k.Patch("bundle", bundleName, "--type=merge", "-p", patchJSON)
 			Expect(err).ToNot(HaveOccurred())
 
-			forceSyncJSON := fmt.Sprintf(`{"spec":{"options":{"forceSyncGeneration":%d}}}`, time.Now().UnixNano())
-			_, err = k.Namespace(bdNamespace).Patch("bundledeployment", bundleDeploymentName, "--type=merge", "-p", forceSyncJSON)
+			forceSyncJSON := fmt.Sprintf(`{"spec":{"forceSyncGeneration":%d}}`, time.Now().UnixNano())
+			_, err = k.Patch("bundle", bundleName, "--type=merge", "-p", forceSyncJSON)
 			Expect(err).ToNot(HaveOccurred())
 
 			Eventually(func(g Gomega) {
@@ -311,10 +306,6 @@ var _ = Describe("Fleet bundlediff CLI GitOps workflow", Label("gitrepo", "gitse
 			g.Expect(err).ToNot(HaveOccurred())
 			g.Expect(out).ToNot(BeEmpty())
 		}, testenv.Timeout).Should(Succeed())
-
-		forceSyncJSON := fmt.Sprintf(`{"spec":{"options":{"forceSyncGeneration":%d}}}`, time.Now().UnixNano())
-		_, err = k.Namespace(bdNamespace).Patch("bundledeployment", bundleDeploymentName, "--type=merge", "-p", forceSyncJSON)
-		Expect(err).ToNot(HaveOccurred())
 
 		By("verifying the Bundle is no longer in Modified state after GitOps sync")
 		Eventually(func(g Gomega) {
--- a/e2e/assets/single-cluster/gitrepo-bundlediff.yaml
+++ b/e2e/assets/single-cluster/gitrepo-bundlediff.yaml
@@ -6,5 +6,6 @@ metadata:
 spec:
   repo: {{.Repo}}
   branch: {{.Branch}}
+  pollingInterval: 5s
   paths:
   - {{.Path}}


e2e/single-cluster/cli_bundlediff_test.go
```diff
@@ -0,0 +1,341 @@
...
+				g.Expect(out).ToNot(BeEmpty())
+				bundleDeploymentName = out
+			}, testenv.Timeout).Should(Succeed())
+
+			Eventually(func(g Gomega) {
+				out, err := k.Context("").Get("bundledeployments", "-A", "-l", "fleet.cattle.io/bundle-name="+bundleName, "-o", "jsonpath={.items[0].metadata.namespace}")
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(out).ToNot(BeEmpty())
+				bdNamespace = out
+			}, testenv.Timeout).Should(Succeed())
```
weyfonk, line 66: This could be simplified to return both name and namespace in a single call, e.g.:
```suggestion
			Eventually(func(g Gomega) {
				out, err := k.Get("bundledeployments", "-A", "-l", "fleet.cattle.io/bundle-name="+bundleName, "-o", "jsonpath={.items[0]['metadata.name','metadata.namespace']}")
				g.Expect(err).ToNot(HaveOccurred())
				g.Expect(out).ToNot(BeEmpty())
				
				data := strings.Split(out, " ")
                g.Expect(data).To(HaveLen(2))
                bundleDeploymentName, bdNamespace = data[0], data[1]
			}, testenv.Timeout).Should(Succeed())
```
... also, why specify an empty context?

e2e/assets/single-cluster/bundlediff-fleet.yaml
```diff
@@ -0,0 +1,10 @@
+defaultNamespace: {{.TargetNamespace}}
+resources:
+- content: |
+    apiVersion: v1
+    kind: ConfigMap
+    metadata:
+      name: bundlediff-gitops-cm
+    data:
+      key: original-value
+  name: cm.yaml
```
weyfonk: This shouldn't be needed in a `fleet.yaml` 🤔 
```suggestion
```

e2e/single-cluster/cli_bundlediff_test.go
```diff
@@ -0,0 +1,341 @@
...
+			Expect(actualOutput).To(SatisfyAny(
+				ContainSubstring("namespace: "+targetNs),
+				Not(ContainSubstring("namespace:")),
+			))
+
+			By("verifying comparePatches hide the drift when applied to Bundle")
+			jsonBytes, err := yaml.YAMLToJSON([]byte(diffSnippet))
+			Expect(err).ToNot(HaveOccurred())
+			patchJSON := fmt.Sprintf(`{"spec":{"options":%s}}`, string(jsonBytes))
+			_, err = k.Namespace(bdNamespace).Run("patch", "bundledeployment", bundleDeploymentName, "--type=merge", "-p", patchJSON)
```
weyfonk: nit: This can be simplified as:
```suggestion
			_, err = k.Namespace(bdNamespace).Patch("bundledeployment", bundleDeploymentName, "--type=merge", "-p", patchJSON)
```

e2e/single-cluster/cli_bundlediff_test.go
```diff
@@ -0,0 +1,341 @@
...
+			Expect(err).ToNot(HaveOccurred())
+
+			forceSyncJSON := fmt.Sprintf(`{"spec":{"options":{"forceSyncGeneration":%d}}}`, time.Now().UnixNano())
+			_, err = k.Namespace(bdNamespace).Run("patch", "bundledeployment", bundleDeploymentName, "--type=merge", "-p", forceSyncJSON)
```
weyfonk: nit:
```suggestion
			_, err = k.Namespace(bdNamespace).Patch("bundledeployment", bundleDeploymentName, "--type=merge", "-p", forceSyncJSON)
```

e2e/single-cluster/cli_bundlediff_test.go
```diff
@@ -0,0 +1,341 @@
...
+
+		By("verifying the comparePatches are applied to the BundleDeployment")
+		Eventually(func(g Gomega) {
+			out, err := k.Namespace(bdNamespace).Get("bundledeployment", bundleDeploymentName, "-o", "jsonpath={.spec.options.diff.comparePatches[0].operations[0].op}")
+			g.Expect(err).ToNot(HaveOccurred())
+			g.Expect(out).ToNot(BeEmpty())
+		}, testenv.Timeout).Should(Succeed())
+
+		forceSyncJSON := fmt.Sprintf(`{"spec":{"options":{"forceSyncGeneration":%d}}}`, time.Now().UnixNano())
+		_, err = k.Namespace(bdNamespace).Run("patch", "bundledeployment", bundleDeploymentName, "--type=merge", "-p", forceSyncJSON)
```
weyfonk: nit:
```suggestion
		_, err = k.Namespace(bdNamespace).Patch("bundledeployment", bundleDeploymentName, "--type=merge", "-p", forceSyncJSON)
```

e2e/single-cluster/cli_bundlediff_test.go
```diff
@@ -0,0 +1,341 @@
...
+		_, _ = k.Delete("gitrepo", gitRepoName)
+		_, _ = k.Delete("bundle", bundleName)
+		_, _ = k.Delete("ns", targetNs, "--wait=false")
+	})
+
+	It("detects drift, generates fleet.yaml snippet, and resolves drift via Git commit", func() {
+		By("modifying the deployed configmap to trigger drift")
+		_, err := k.Namespace(targetNs).Run(
+			"patch", "configmap", "bundlediff-gitops-cm",
+			"--type=merge", "-p", `{"data":{"key":"modified-value"}}`,
```
weyfonk: nit:
```suggestion
		_, err := k.Namespace(targetNs).Patch(
			"configmap", "bundlediff-gitops-cm",
			"--type=merge", "-p", `{"data":{"key":"modified-value"}}`,
```

e2e/single-cluster/cli_bundlediff_test.go
```diff
@@ -0,0 +1,341 @@
...
+		}, testenv.Timeout).Should(Succeed())
+
+		Eventually(func(g Gomega) {
+			out, err := k.Context("").Get("bundledeployments", "-A", "-l",
+				"fleet.cattle.io/repo-name="+gitRepoName,
+				"-o", "jsonpath={.items[0].metadata.namespace}")
+			g.Expect(err).ToNot(HaveOccurred())
+			g.Expect(out).ToNot(BeEmpty())
+			bdNamespace = out
+		}, testenv.Timeout).Should(Succeed())
```
weyfonk, line 266: Same remark as above about combining 2 requests into one.

e2e/single-cluster/cli_bundlediff_test.go
```diff
@@ -0,0 +1,341 @@
...
+		By("verifying the comparePatches are applied to the BundleDeployment")
+		Eventually(func(g Gomega) {
+			out, err := k.Namespace(bdNamespace).Get("bundledeployment", bundleDeploymentName, "-o", "jsonpath={.spec.options.diff.comparePatches[0].operations[0].op}")
+			g.Expect(err).ToNot(HaveOccurred())
+			g.Expect(out).ToNot(BeEmpty())
+		}, testenv.Timeout).Should(Succeed())
+
+		forceSyncJSON := fmt.Sprintf(`{"spec":{"options":{"forceSyncGeneration":%d}}}`, time.Now().UnixNano())
+		_, err = k.Namespace(bdNamespace).Run("patch", "bundledeployment", bundleDeploymentName, "--type=merge", "-p", forceSyncJSON)
+		Expect(err).ToNot(HaveOccurred())
```
weyfonk: Is this still necessary, with a GitRepo in place? If the goal is to save time, how about setting a shorter-than-default polling interval on the GitRepo instead?

e2e/assets/single-cluster/bundlediff-fleet-snippet.yaml
```diff
@@ -0,0 +1,9 @@
+diff:
```
weyfonk, line 1: This file doesn't seem to be used anywhere.

integrationtests/cli/bundlediff/bundlediff_test.go
```diff
@@ -776,12 +771,32 @@ var _ = Describe("Fleet bundlediff", func() {
...
+			bundle := &fleet.Bundle{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      bundleName,
+					Namespace: namespace,
+				},
+				Spec: fleet.BundleSpec{
+					Resources: []fleet.BundleResource{
+						{
+							Name:    "simple-chart",
+							Content: "H4sIFAAAAAAA/ydIzEsBAAAAAP//AQAA//8BAAABAAAAAA==",
```
weyfonk, line 787: nit: a comment explaining what this represents could be welcome.

internal/cmd/cli/bundlediff.go
```diff
@@ -241,20 +359,91 @@ func (d *BundleDiff) printFleetYAML(out io.Writer, diffs []DiffOutput) error {
...
 
-	output := struct {
-		Diff fleet.DiffOptions `json:"diff"`
+	return d.outputFleetYAMLDiff(ctx, k8sClient, out, diffs[0], comparePatches)
+}
+
+// outputFleetYAMLDiff outputs a fleet.yaml diff snippet that can be added to a GitRepo's fleet.yaml.
+// This allows users to integrate detected drift as permanent ignores in their GitOps workflow.
+func (d *BundleDiff) outputFleetYAMLDiff(ctx context.Context, k8sClient client.Client, out io.Writer, diff DiffOutput, comparePatches []fleet.ComparePatch) error {
+	// Get the Bundle to merge with existing comparePatches if any
```
weyfonk: I think this comment would be better located above the actual `Get` call for a bundle.

internal/cmd/cli/bundlediff.go
```diff
@@ -241,20 +359,91 @@ func (d *BundleDiff) printFleetYAML(out io.Writer, diffs []DiffOutput) error {
...
+
+	// Merge new comparePatches with existing ones from the Bundle
+	var existingPatches []fleet.ComparePatch
+	if bundle.Spec.Diff != nil && bundle.Spec.Diff.ComparePatches != nil {
+		existingPatches = bundle.Spec.Diff.ComparePatches
+	}
+	mergedPatches := mergeComparePatches(existingPatches, comparePatches)
+
+	// Keep yaml field order stable for readability and test fixtures.
+	type comparePatchYAML struct {
```
weyfonk: Do we need to redefine this type? Why not reuse `fleet.ComparePatch`?

internal/cmd/cli/bundlediff_test.go
```diff
@@ -0,0 +1,68 @@
...
+	for i := range merged {
+		if merged[i].Kind == "ConfigMap" && merged[i].Name == "app-config" {
+			configPatch = &merged[i]
+			break
+		}
+	}
+	if configPatch == nil {
+		t.Fatalf("expected ConfigMap patch to exist")
+	}
+	if len(configPatch.Operations) != 2 {
```
weyfonk: This doesn't validate that merged operations or JSON pointers are free of duplicates, though.
Defining the expected result as an array of patches, as done for the input, and checking that the output matches that result, could be easier and more complete.

internal/cmd/agent/deployer/deployer.go
```diff
@@ -125,7 +124,14 @@ func (d *Deployer) helmdeploy(ctx context.Context, logger logr.Logger, bd *fleet
 			return bd.Status.Release, nil
 		}
 	}
-	manifestID, _ := kv.Split(bd.Spec.DeploymentID, ":")
+
+	// manifestID is used for manifest/OCI lookups
+	// DeploymentID format is "manifestID:optionsHash"
+	manifestID := bd.Spec.DeploymentID
```
weyfonk, line 133: I think I understand this change, getting rid of a Wrangler import, but out of curiosity, why is it needed within this PR?

internal/cmd/controller/target/partition.go
```diff
@@ -77,6 +79,15 @@ func maxUnavailablePartitions(partitions []partition, targets []*Target) (int, e
 // updateDeploymentFromStaged will update DeploymentID and Options for the target to the
 // staging values, if it's in a deployable state
 func updateDeploymentFromStaged(t *Target, bundleStatus *fleet.BundleStatus, partitionStatus *fleet.PartitionStatus) {
+	if t.Deployment != nil &&
+		!t.IsPaused() &&
+		t.Deployment.Spec.StagedDeploymentID != "" &&
+		t.Deployment.Spec.DeploymentID == t.Deployment.Spec.StagedDeploymentID &&
+		!reflect.DeepEqual(t.Deployment.Spec.Options.Diff, t.Deployment.Spec.StagedOptions.Diff) {
+		// Keep diff options in sync even when the DeploymentID is unchanged.
```
weyfonk, line 87: nit: this may help our future selves understand what is going on here 😅 
```suggestion
		// Keep diff options in sync even when the DeploymentID is unchanged.
		// This enables diff updates to be propagated downstream to resolve modified statuses (for instance 
		// after updating bundle diffs in a fleet.yaml)
```

internal/cmd/cli/bundlediff.go
thardeck: Good point.

The comparePatchYAML type controls the field order in YAML output for consistency with Kubernetes conventions (apiVersion first, kind second, etc.). The fleet.ComparePatch type has fields in a different order (Kind, APIVersion, Namespace, Name, Operations, JsonPointers).

I ran into this issue during the tests run, that the structure changed and broke my expectations I think.

But now thinking about it, I should just change the structure for `fleet.CopmparePatch` to follow the Kubernetes style.

internal/cmd/agent/deployer/deployer.go
thardeck: To be honest I am not sure anymore after so many refactorings.
I think it was part of a previous change in regards to "options-only change detection" which was not needed anymore after clarifying the implementation details.
Maybe it was even a linter suggestion.

Anyway I can remove it, if you prefer.

internal/cmd/cli/bundlediff.go
weyfonk: > But now thinking about it, I should just change the structure for `fleet.ComparePatch` to follow the Kubernetes style.

I also think that would be preferable; we may want to document it as a breaking change in v0.15.0 release notes, just in case someone/something relied on that order of fields outside of Fleet, as unlikely as it may be.

internal/cmd/agent/deployer/deployer.go
weyfonk: Thanks for clarifying :)
No strong opinion here. We could keep it or move it to a separate PR.

Issue #4452, opened 2025-12-16
Resources under GitRepo shows 0 when used with CorrectDrift option


sbulage:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When using `correctDrift: true` in GitRepo, Resources under GitRepo shows 0 when modified.

### Expected Behavior

Modified Resources should be auto correct when `correctDrift` used. GitRepo should show exact resource count not 0.

### Steps To Reproduce

- Create a normal `GitRepo` without enabling `correctDrift`.
- Verify the resources are created without any issue on cluster.
- Now edit `GitRepo` and enable `correctDrift`.
- Modify the resources for e.g. update replica of deployment etc.
- Modified resources are not reverting back to original state.
- In `GitRepo` resources count are showing '0'

### Environment

```markdown
- Architecture:
- Fleet Version: 109.0.0+up0.15.0-alpha.1
- Cluster:
  - Provider:k3s/k3d
  - Options:
  - Kubernetes Version:
```

### Logs

```markdown

```

### Anything else?

<details><summary>Screenshot Shows 0 Resources</summary>
<p>

<img width="1918" height="960" alt="Image" src="https://github.com/user-attachments/assets/9d75c30c-ebef-43b3-8a6b-25636f1ea6c6" />

</p>
</details> 

<details><summary>Video showing detailed steps and actual output.</summary>
<p>

https://github.com/user-attachments/assets/9a8e7749-9516-48a3-81fd-cb87a7cd578f

</p>
</details>

sbulage:
### System Information
| Rancher Version | Fleet Version |
|--------|--------|
| v2.14-715edf00e1c950adf92f96ea78c10fb9b3b45e36-head | 109.0.0+up0.15.0-alpha.10 | 

### Screenshot
- Before `correctDrift` enabled.
  <details><summary>Screenshot shows Correct Resource count and shows Modified state of GitRepo and Bundles.</summary>
  <p>
  
  <img width="2559" height="1321" alt="Image" src="https://github.com/user-attachments/assets/385cfbe3-527c-4f36-ae7f-029eafa89d3b" />
  
  </p>
  </details> 

- After `correctDrift` enabled

  <details><summary>Screenshot showing GitRepo Active state after correctDrift removed changes manually made</summary>
  <p>
  
  <img width="2548" height="1357" alt="Image" src="https://github.com/user-attachments/assets/b46a8b35-f6e9-4db6-863f-99c955d8f95c" />
[...]

PR #4624, opened 2026-02-10
Fix Install action config for correctDrift force in Helm v4

thardeck:
When correctDrift.Force is enabled, Helm's Install action now:
- Sets ForceReplace flag from both Helm.Force and CorrectDrift.Force
- Disables ServerSideApply when ForceReplace is set (Helm v4 constraint)
- Enables ServerSideApply explicitly when neither TakeOwnership nor ForceReplace is set

Refers to #4452

### DIFF
--- a/e2e/single-cluster/helmop_test.go
+++ b/e2e/single-cluster/helmop_test.go
@@ -190,7 +190,7 @@ var _ = Describe("HelmOp resource with polling of repo index", Label("infra-setu
 	})
 })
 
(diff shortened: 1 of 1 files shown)


kkaempf:
So that means a backport to 2.13.x is not possible 🤔

thardeck:
> So that means a backport to 2.13.x is not possible 🤔

No need to backport because Helm v4 was only introduced in Fleet v0.15 for Rancher v2.14, and the issue seems to be a regression.

Issue #4637, opened 2026-02-11
[backport v2.13] [SURE-11213] Fleet initiates unintended Longhorn uninstall during longhorn upgrade


rancherbot:
This is a backport issue for #4572, automatically created via [GitHub Actions workflow](https://github.com/rancher/fleet/actions/runs/21915865874) initiated by @0xavi0

Original issue body:

### Issue description

During a Longhorn upgrade managed by Fleet, Fleet unexpectedly initiated a Helm uninstall of the existing Longhorn release instead of performing an in-place upgrade.

The uninstall was blocked by a protective uninstall hook in the Longhorn Helm chart, preventing data loss. However, this left the Helm release in a permanent uninstalling state.

Fleet had already deleted all previous Helm revisions, making rollback impossible. As a result, the Longhorn deployment was stuck and could not be upgraded further without manual intervention.

This behavior occurred without any explicit request to uninstall Longhorn.

### Business impact:

* Fleet initiated a Helm uninstall of Longhorn during an upgrade attempt.
* Helm release status became permanently stuck in uninstalling.
* helm history showed only a single revision in uninstalling state:
~~~
    REVISION  STATUS         DESCRIPTION
    1         uninstalling   Deletion in progress (or silently failed)
~~~
* All previous Helm revisions were removed by Fleet.
* Helm rollback was not possible due to missing revision history.
* Longhorn uninstall hook prevented data loss and blocked uninstall. Otherwise it would have lead to data loss.
* Upgrade to Longhorn 1.10.1 could not proceed while release remained in uninstalling.

### Repro steps:

Reproduction is not straightforward through normal upgrade procedures; however, during testing in a lab environment, the following scenario resulted in a similar failure state as observed in the customer environment and led to identification of the documented workaround:

* Longhorn is installed and managed by Fleet (e.g., version 1.10.1).
* Triggered upgrade to an incorrect or non-existent Longhorn chart version (e.g., 1.10.2) is configured in fleet.yaml.
* Fleet attempts the upgrade and fails due to chart/version not found.
* The version in fleet.yaml is reverted back to a valid version (e.g., 1.10.1).
* Fleet then initiates a Helm uninstall of the existing release instead of recovering.
* Helm release becomes stuck in uninstalling state with no rollback path.

### Workaround:
Is a workaround available and implemented? Yes.

1. Pause the git repo
1. Scale down fleet agent to 0
1. Change helm secret:
~~~
[...]

PR #4647, opened 2026-02-12
[v0.14] - Avoid uninstalling releases in PendingInstall status

0xavi0:
This PR prevents Fleet from uninstalling a release that is in **PendingInstall** status, opting instead to attempt an **Upgrade**.

The following approach is taken:

* If a valid previous version is found, an **Upgrade** is attempted (which would trigger a rollback followed by an upgrade).
* If no valid previous version is found, the current release is patched to mark it as **failed**, and then an **Upgrade** is performed.

This avoids uninstalling any resources in cases where a previous release may have ended up in a stuck state.

Refers to: https://github.com/rancher/fleet/issues/4637
Backport of: https://github.com/rancher/fleet/pull/4600

### Checklist

### DIFF
--- a/integrationtests/cli/deploy/deploy_test.go
+++ b/integrationtests/cli/deploy/deploy_test.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 
 	"github.com/onsi/gomega/gbytes"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	clihelper "github.com/rancher/fleet/integrationtests/cli"
 	"github.com/rancher/fleet/integrationtests/utils"
@@ -15,6 +16,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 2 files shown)


thardeck:
We can not merge this before the release of v0.14.3 because of Code Freeze.

0xavi0:
# QA Template

## How to test 
This issue is bit complex to recreate, because basically implies that the Helm release is in `PendingInstall` status, which normally happens in certain conditions that are not easy to reproduce.

For that reason we can test it by patching the release Helm secret so it shows as `pending-install` and then force a _BundleDeployment_ reconciliation (deleting it)

The following script: 
[test-pending-install-fix.sh](https://github.com/user-attachments/files/25289971/test-pending-install-fix.sh)

Patches the Helm release secret for the given _Bundle_ and forces the reconciliation again.
The next reconciliation it finds the secret in `pending-install` status and Fleet SHOULD NOT uninstall the previous release (as it was doing before the fix).

### Possible steps to test
1. Edit the script so it only patches the Helm secret 
2. Check that the secret is showing status=pending_install
3. Force the Bundle or BundleReconciler reconciliation (you can, for example, delete de BundleDeployment or ForceUpdate the _GItRepo_ related)
4. Verify that the release is not uninstalled

### We need to test this in the following scenarios:
* When there is only 1 Helm release secret for that BundleDeployment (Helm can't rollback)
* Where there is more than 1 Helm release secret for that BundleDeployment (Helm Can rollback)

The most important thing is to check that the release is never deleted.

EXTRA INFO:
Run the script like this:
```bash
./test-pending-install-fix.sh -n BUNDLE_NAMESPACE -b BUNDLE_NAME
```

Issue #4655, opened 2026-02-16
[BUG] Grafana in modified state after installing via HelmOps in Fleet 0.15


mmartin24:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Issue.
Installing this chart via HelmOps leaves the bundle in `modified` state with display message:
```
Modified(1) [Bundle test-grafana]; deployment.apps default/test-grafana modified {"spec":{"strategy":{"rollingUpdate":null}}}
```

<img width="2446" height="761" alt="Image" src="https://github.com/user-attachments/assets/6d3ad94a-479b-47a6-9928-6f7e530c3ee4" />

<img width="2446" height="336" alt="Image" src="https://github.com/user-attachments/assets/0eaab0e3-f4ac-4ebe-8a54-6613274c1454" />

Note: Apparently this issue also happened in Helm 3 versions, but it was not manifested, leaving the bundle in modified state.

### Expected Behavior

To be discussed

### Steps To Reproduce

Deploy Rancher 2.14 with Fleet 0.15.0-alpha.9
Upload the following yaml in HelmOps

```yaml
apiVersion: fleet.cattle.io/v1alpha1
kind: HelmOp
metadata:
  name: test-grafana
  annotations:
  namespace: fleet-local
spec:
  correctDrift:
    enabled: false
  helm:
    chart: grafana
    repo: https://grafana.github.io/helm-charts
```

### Environment

```markdown
- Architecture: amd64
- Fleet Version: 0.15.0-alpha.9
- Cluster:
  - Provider: k3s
  - Options:
  - Kubernetes Version: 1.33.1+k3s1
```

### Logs

```markdown

```

mmartin24:
@marcelofukumoto, pinging you here since this is affecting our e2e in case something can be done. Thank you.

mmartin24:
This seems fixed in `0.15.0-beta.2`

<img width="2347" height="1273" alt="Image" src="https://github.com/user-attachments/assets/b286f5df-ba66-47d5-9815-994c81f10d76" />

Both failing tests that detected this passed: https://github.com/rancher/fleet-e2e/actions/runs/22148881118/job/64034228007

<img width="703" height="72" alt="Image" src="https://github.com/user-attachments/assets/e5945d84-18bc-406d-a1e2-a200025a15c6" />

Closing issue. Thanks for the quick fix.

PR #4664, opened 2026-02-17
Prevent false drift from Helm v4 null field rendering

thardeck:
Helm v4 can emit explicit null values for omitted fields (e.g., spec.strategy.rollingUpdate: null), while Helm v3 omitted these fields entirely. When combined with Kubernetes server-side defaulting, this creates false drift detection: the merge patch interprets null as "delete this field," conflicting with defaulted live resources.

Solution:
- Add normalizeNullPatch to recursively strip null-valued fields from
  merge patches before drift comparison
- Handle nested structures (maps and arrays) to cover all potential
  Helm v4 null emissions beyond just deployment strategy

Refers to #4655

### DIFF
--- a/e2e/single-cluster/helm_recreate_drift_test.go
+++ b/e2e/single-cluster/helm_recreate_drift_test.go
@@ -77,8 +77,8 @@ var _ = Describe("Helm v4 Null Field Drift Detection", Label("infra-setup"), fun
 			// Create git auth secret
 			_, err := k.Create(
 				"secret", "generic", "git-auth", "--type", "kubernetes.io/basic-auth",
-				"--from-literal=username=fleet-ci",
-				"--from-literal=password=foo",
+				"--from-literal=username="+os.Getenv("GIT_HTTP_USER"),
+				"--from-literal=password="+os.Getenv("GIT_HTTP_PASSWORD"),
 			)
 			Expect(err).ToNot(HaveOccurred())
 


e2e/assets/helm-recreate-drift/chart/templates/deployment.yaml
```diff
@@ -0,0 +1,21 @@
...
+kind: Deployment
+metadata:
+  name: test-deployment
+spec:
+  replicas: {{ .Values.replicas | default 1 }}
+  selector:
+    matchLabels:
+      app: test-app
+  strategy:
+    type: RollingUpdate
```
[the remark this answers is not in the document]
thardeck: This is on purpose, because it is supposed to represent a real world chart. And the test fails without the fix with Helm v4, but not Helm v3.

e2e/single-cluster/helm_recreate_drift_test.go
```diff
@@ -0,0 +1,146 @@
...
+		}).Should(Succeed())
+	})
+
+	When("deploying a Helm chart with fields omitted by Helm v4", func() {
+		JustBeforeEach(func() {
+			// Create git auth secret
+			_, err := k.Create(
+				"secret", "generic", "git-auth", "--type", "kubernetes.io/basic-auth",
+				"--from-literal=username=fleet-ci",
+				"--from-literal=password=foo",
```
0xavi0: nit: I think it's better to use the env variables for username and password as, in case we change them at some point, it won't require any extra change.

internal/cmd/agent/deployer/desiredset/diff.go
```diff
@@ -240,7 +240,12 @@ func normalizeActual(
 	key objectset.ObjectKey,
 	patch *[]byte,
 ) (bool, error) {
-	return normalizeReplicasPatch(live, uActual, key, patch)
+	emptied, err := normalizeReplicasPatch(live, uActual, key, patch)
+	if err != nil || emptied {
+		return emptied, err
+	}
+
+	return normalizeNullPatch(key, patch)
```
0xavi0, line 249: We could, maybe, check if there are any `null` string before calling `normalizeNullPatch` as it the function is potentially expensive (un-marshalls, walks and re-marshalls data).

We could have a false positive (for example if there was `null` substring as part of another string, but in general cases it would be better for performance. 

I'm suggesting to do:
```go
if !bytes.Contains(*patch, []byte("null")) {
      return false, nil
  }
```
before calling `normalizeNullPatch`

thardeck: Very good point, makes a huge difference.

According to a simple benchmark this provides ~55x performance improvement for patches without nulls:
    
    - WithoutNulls: 127.9 ns/op (vs 7063 ns/op with nulls)
    - 97% fewer allocations (1 vs 72 allocs/op)
    - 97% less memory (176 B vs 5404 B per op)

PR #4669, opened 2026-02-18
Revert "Use GitHub app tokens in release scripts"

thardeck:
because it does not work yet as expected and the legacy process still works.

This reverts commit 8cb64656d1402b55db456fd7dd3b2edea63b1164.

While cleaning up I have also removed the Renovate entry for Rancher 2.9 respectively Fleet v0.10.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -8,8 +8,7 @@
     "release/v0.14",
     "release/v0.13",
     "release/v0.12",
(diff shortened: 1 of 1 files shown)


sbulage (approved):
LGTM 👍

sbulage (approved):
LGTM 👍

PR #4683, opened 2026-02-19
Fix release against rancher PR creation

thardeck:
Checkout v6 stores credentials via Git's includeIf mechanism in `$RUNNER_TEMP` instead of directly in `.git/config`. The old `git config --unset` approach no longer removed these credentials.

Add `persist-credentials: false` to rancher/rancher checkout so the default `GITHUB_TOKEN` credential is not persisted, allowing the PUSH_TO_FORKS_SUBMIT_PRS token set by 'gh auth setup-git' to be the sole credential when pushing.

### Additional changes:
* Removing obsolete chart release process

### DIFF
--- a/.github/scripts/create-pr.sh
+++ b/.github/scripts/create-pr.sh
@@ -12,7 +12,6 @@ if [ -z "$GITHUB_TOKEN" ]; then
 fi
 
 # Configure git login
-git config --local --unset http.https://github.com/.extraheader ^AUTHORIZATION:
 gh auth setup-git
 
 # Create and push new branch
--- a/.github/workflows/release-against-rancher.yml
+++ b/.github/workflows/release-against-rancher.yml
@@ -39,6 +39,7 @@ jobs:
         with:
           fetch-depth: 0
           path: fleet
+          persist-credentials: false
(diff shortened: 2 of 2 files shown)


thardeck:
> LGTM, but out of curiosity: where is `includeIf` used here? 🤔

### New behavior (v6):

1. Creates a separate credential file in `$RUNNER_TEMP` (e.g., `/tmp/git-credentials-abc123`)
1. Writes the credentials there
1. Adds an includeIf directive to the global git config (`~/.gitconfig`)

### How includeIf Works
`includeIf.gitdir` is a Git feature that conditionally includes another config file based on the repository's directory path.

#### Example
```
# In ~/.gitconfig (global config)
[includeIf "gitdir:/home/runner/fleet/.git"]
    path = /tmp/git-credentials-abc123
```

This means: _"If the current Git repository is located under `/home/runner/fleet/.git`, also load the config from `/tmp/git-credentials-abc123`"_

thardeck:
The app token based approach still [does not work](https://github.com/rancher/fleet/actions/runs/22221939503/job/64279224481), so going back to the static token approach for now.

Issue #4671, opened 2026-02-18
Soft-disable imagescan


weyfonk:
[Imagescan](https://fleet.rancher.io/0.14/how-tos-for-users/imagescan) was developed as an experimental feature, and has not been maintained much nor received many enhancement requests since then.

Therefore, we think it would make sense to disable it, including its [controller](https://github.com/rancher/fleet/blob/main/internal/cmd/controller/reconciler/imagescan_controller.go), by default.

Acceptance criteria:
* A new Helm value is added, such as `imagescan.enabled`, and set to `false` by default.
* When that value is set to `true`, imagescan works as it currently does.
* When that value is set to `false`:
    * the imagecan controller is not instantiated (`internal/cmd/controller/operator.go`)
    * the GitRepo imagescan indexer (`AddImageScanGitRepoIndexer`) is not created (`internal/cmd/controller/gitops/operator.go`)
    * the gitOps controller:
        * when handling GitRepo deletion, skips looking for imagescan jobs, nor does it list image scans to delete (`internal/cmd/controller/gitops/reconciler/gitjob_controller.go`)
            * [Optional] when creating/updating a role, does not add `imagescans` to supported resources (`internal/cmd/controller/gitops/reconciler/rbac.go`)
    * the bundle reader (`internal/bundlereader/read.go`):
        * skips iterating over `imagescan` items in `fleet.yaml` or similar config files
        * warns about such items being skipped, for instance through a log message?

weyfonk:
- [x] Document this change in Fleet docs
- [ ] Mention it as a breaking change in Fleet release notes
- [ ] Mention it as a breaking change in Rancher release notes

PR #4692, opened 2026-02-20
Disable imagescan by default

weyfonk:
A new Helm value, `imagescan.enabled`, is set to `false` by default, leading to the imagescan controller not being instantiated. When imagescan is disabled, the reader will output an error when finding imagescans in a bundle.

Users will still be able to enable imagescan, leading it to work the way it has until now.

Refers to #4671.

### Checklist

- [x]  I have updated the documentation via a [pull request](https://github.com/rancher/fleet-product-docs/pull/175) in the

### DIFF
--- a/charts/fleet/values.yaml
+++ b/charts/fleet/values.yaml
@@ -160,6 +160,9 @@ helmops:
   enabled: true
   replicas: 1
 
+# When enabled, the imagescan controller will run to monitor image repositories.
+# If disabled, config files (e.g. fleet.yaml) containing non-empty imageScan fields will result in errors.
+# This option may be deprecated in a future release, and imagescan removed.
 imagescan:
   enabled: false
 


internal/cmd/controller/operator.go
```diff
@@ -159,18 +161,28 @@ func start(
...
-	if err = (&reconciler.ImageScanReconciler{
-		Client: mgr.GetClient(),
-		Scheme: mgr.GetScheme(),
+	imagescanEnabled := false
+	if v := os.Getenv("IMAGESCAN_ENABLED"); v != "" {
+		enabled, err := strconv.ParseBool(v)
+		if err != nil {
+			setupLog.Error(err, "failed to parse IMAGESCAN_ENABLED", "value", v)
+		}
+		imagescanEnabled = enabled
```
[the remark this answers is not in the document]
weyfonk: If an error occurs, `ParseBool` returns `false`, which is not an invalid value but precisely what we need.

charts/fleet/values.yaml
```diff
@@ -160,6 +160,9 @@ helmops:
   enabled: true
   replicas: 1
 
+imagescan:
```
thardeck, line 166: nit: I would add a comment that this option might get deprecated in the future and what it does.

weyfonk: Good point! Added.

Issue #4696, opened 2026-02-23
gitjob pods fail to read SSH secrets with CIS hardening - missing fsGroup in PodSecurityContext


terwey:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Fleet gitjob pods cannot read mounted SSH key secrets when running on clusters with CIS hardening (PodSecurity `restricted` profile). The job fails with permission denied when trying to access the SSH key.

### Expected Behavior

gitjob should be able to read mounted secrets when running as non-root user.

### Steps To Reproduce

1. Deploy RKE2 cluster with CIS hardening profile
2. Create GitRepo with SSH authentication (`clientSecretName` pointing to SSH key secret)
3. Observe gitjob pod fails to clone - SSH key unreadable

### Environment

```markdown
- Fleet version: v0.14.2
- Kubernetes: RKE2 v1.32 with CIS hardening
- PodSecurity: `restricted` enforcement on namespace
```

### Logs

```markdown

```

### Root Cause

Here the PodSecurityContext
https://github.com/rancher/fleet/blob/6272ff725e9d246d23e952e7310b44de6e26846c/internal/cmd/controller/gitops/reconciler/gitjob.go#L451-L455

sets `RunAsUser: 1000` but does not set `FSGroup`:

When secrets are mounted into a pod running as non-root without `fsGroup`, the files are owned by `root:root` with mode `0400`, making them unreadable by UID 1000.

### Proposed Fix

Add `FSGroup` to match the `RunAsUser`:

```go
SecurityContext: &corev1.PodSecurityContext{
    RunAsUser: &[]int64{1000}[0],
    FSGroup:   &[]int64{1000}[0],
},
```

This ensures mounted secrets are readable by the group, which the non-root user belongs to.

### Workaround

Label the `fleet-default` (or `fleet-local`) namespace as privileged:
```bash
# For GitRepos in fleet-default (downstream clusters)
kubectl label namespace cattle-fleet-system \
  pod-security.kubernetes.io/enforce=privileged \
  pod-security.kubernetes.io/audit=privileged \
  pod-security.kubernetes.io/warn=privileged \
  --overwrite

# For GitRepos in fleet-local (management cluster)
kubectl label namespace cattle-fleet-local-system \
  pod-security.kubernetes.io/enforce=privileged \
  pod-security.kubernetes.io/audit=privileged \
  pod-security.kubernetes.io/warn=privileged \
  --overwrite
```

Or

```
[...]

PR #4697, opened 2026-02-23
fix(gitjob): add fsGroup to PodSecurityContext for secret readability

terwey:
Secrets mounted as non-root (runAsUser: 1000) are unreadable without fsGroup, causing SSH key auth failures on CIS-hardened clusters.

Refers to #4696 

Adds `FSGroup: 1000` to the gitjob PodSecurityContext to match the existing `RunAsUser: 1000`. This allows mounted secrets (SSH keys, tokens) to be readable by the non-root user via group ownership.

### Checklist


weyfonk:
Thanks for your contribution, and [candor](https://github.com/rancher/fleet/issues/4696#issuecomment-3950730904). Closing for now.

Issue #3859, opened 2025-07-03
Fix discrepancy in use of OCI chart fields between GitOps and HelmOps


weyfonk:
HelmOps resources reference Helm charts as follows ([docs](https://fleet.rancher.io/helm-ops#oci-registry)):
* `Repo`: Helm repository or OCI registry
* `Chart`: URL to tarball, or chart name

On the other hand, in a `fleet.yaml` file monitored by a `GitRepo`, an OCI chart will be referenced using the `Chart` field ([docs](https://fleet.rancher.io/ref-fleet-yaml#chart))

This should be made consistent, ideally aligning GitOps on HelmOps for consistency within Fleet.

p-se:
## QA Template

Apply both Bundles and make sure that both work, although one uses the `helm.repo` field for providing an OCI chart URL and the other uses `helm.chart` (deprecated, but still functional).

```yaml
apiVersion: fleet.cattle.io/v1alpha1
kind: Bundle
metadata:
  name: nginx-chart-field
  namespace: fleet-default
spec:
  resources:
  - name: fleet.yaml
    content: |
      namespace: nginx
      helm:
        releaseName: nginx
        chart: "oci://registry-1.docker.io/bitnamicharts/nginx"
        version: "18.3.4"
        values:
          replicaCount: 1
  targets:
  - clusterSelector:
      matchLabels:
        env: test
```

```yaml
apiVersion: fleet.cattle.io/v1alpha1
kind: Bundle
metadata:
  name: nginx-repo-field
  namespace: fleet-default
spec:
  resources:
  - name: fleet.yaml
    content: |
      namespace: nginx
      helm:
        releaseName: nginx
        repo: "oci://registry-1.docker.io/bitnamicharts/nginx"
        version: "18.3.4"
        values:
          replicaCount: 1
  targets:
  - clusterSelector:
      matchLabels:
        env: test
```

sbulage:
### System Information

| Rancher Version | Fleet Version |
|--------|--------|
| v2.15-591ed90b3bc21d16e12cf29be0f8eaef6288c6df-head | 110.0.0+up0.16.0-alpha.5 | 

### Verification Steps
- For verification used [`helm-oci-chart-field`](https://github.com/sbulage/test-fleet/tree/main/helm-oci-chart-field) :

  | Helm fields | Verification Screenshot |
  |--------|--------|
[...]

PR #4698, opened 2026-02-23
Fix discrepancy of OCI chart fields between GitOps and HelmOps

p-se:
Changes GitOps to use the same fields as HelmOps for OCI chart URLs.

Refers to #3859 

### Checklist


weyfonk (approved):

p-se:
> E2E test seems to fail:

They are supposed to. They point to additional tests on `rancher/fleet-test-data` that have [not been merged yet](https://github.com/rancher/fleet-test-data/pull/37).

Issue #4681, opened 2026-02-19
Document additional known_hosts entries


weyfonk:
Availability of the new `additionalKnownHosts` value and its use must be documented.

PR #4706, opened 2026-02-24
[v0.14] Add additionalKnownHosts configuration to values.yaml

thardeck:
Refers to #4681

### DIFF
--- a/charts/fleet/values.yaml
+++ b/charts/fleet/values.yaml
@@ -79,6 +79,11 @@ affinity: {}
 # Pod resource limits and requests for the controllers
 resources: {}
   # limits:
+## Adds hosts to the known_hosts configmap if using a custom git provider for gitrepos
+additionalKnownHosts: []
+# Custom entries to be added in SSH known_hosts format (see charts/fleet/templates/configmap_known_hosts.yaml for examples):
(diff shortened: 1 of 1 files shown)


charts/fleet/values.yaml
```diff
@@ -79,6 +79,11 @@ affinity: {}
 # Pod resource limits and requests for the controllers
 resources: {}
   # limits:
+## Adds hosts to the known_hosts configmap if using a custom git provider for gitrepos
+additionalKnownHosts: []
```
weyfonk, line 83: This is not going to be taken into account anywhere without a templated block in `charts/fleet/templates/configmap_known_hosts.yaml`, is it?

thardeck: Yes, but that change was also [backported](https://github.com/rancher/fleet/blob/release/v0.14/charts/fleet/templates/configmap_known_hosts.yaml) and is already part of the current Fleet releases of the according branches.

weyfonk: Thanks, that makes sense, I may have looked at outdated files. And `.github/scripts/update_known_hosts_configmap.sh` is run from `main`, covering all active release branches, hence no need to update it here.

Issue #3845, opened 2025-06-30
HelmOps: fall back to Rancher-configured CA bundles


weyfonk:
Fleet already supports this for GitOps, see #2750.

Falling back to default values should also work for HelmOps. For consistency with what GitOps supports, HelmOps should use [this logic](https://github.com/rancher/fleet/blob/main/pkg/cert/cabundle.go).

mmartin24:
Verified in Rancher `v2.15-cf5c51c7f14a96de21b0afaccd1b2ea958eb24e1-head` with Fleet `0.16.0-alpha.5`

---

## Prerequisites

0. Testing environment set up:
- **k3d cluster** running k3s v1.34+ for Kubernetes compatibility
- **ChartMuseum** serving Helm charts over HTTPS with self-signed certificates
- **Git server** serving git repositories over HTTPS with self-signed certificates
- **Fleet** configured with Rancher global CA for TLS validation

These were accomplished using fleet repo scripts here: https://github.com/rancher/fleet/tree/main. 
More detailed steps in [this guide](https://confluence.suse.com/spaces/~mamartin/pages/1758920770/Custom+CA+testing+for+Fleet+with+local+Git+server).

1. Rancher global CA installed:
```bash
kubectl create secret generic tls-ca-additional \
  -n cattle-system \
  --from-file=ca-additional.pem=FleetCI-RootCA/root.crt
```

2. Two test secrets:
```bash
# Secret WITH CA (has cacerts field)
kubectl get secret helm-secret -n fleet-local

# Secret WITHOUT CA (no cacerts field, only username/password)
kubectl create secret generic helm-secret-no-ca \
  -n fleet-local \
  --from-literal=username='' \
  --from-literal=password=''
```
---

## Test 1: HelmOp with CA in Secret

**What we're testing**: When a HelmOp references a secret that HAS a `cacerts` field, the bundle should still get `helmOpCABundle` populated from Rancher global CA.

### Steps:

1. **Apply the HelmOp**:
```bash
kubectl apply -f fleet-test-data/helmop-test.yaml
```

The HelmOp YAML:
```yaml
apiVersion: fleet.cattle.io/v1alpha1
kind: HelmOp
metadata:
  name: test-helmop-sleeper
  namespace: fleet-local
spec:
  helm:
    repo: https://chartmuseum-service.default.svc.cluster.local:8081
    chart: sleeper-chart
    version: 0.1.0
[...]

PR #4724, opened 2026-02-27
Fall back to Rancher CA bundles for HelmOps

thardeck:
Resolve the CA bundle in the HelmOps controller and store it in BundleHelmOptions.CABundle so the agent can use it without needing access to cattle-system secrets. The agent service account only has access to its own namespace.

Also restart helmops pods in dev/update-controller-k3d so that redeployments pick up the new controller binary.

Refers to #3845


pkg/apis/fleet.cattle.io/v1alpha1/bundle_types.go
```diff
@@ -457,4 +457,11 @@ type BundleHelmOptions struct {
 
 	// InsecureSkipTLSverify will use insecure HTTPS to clone the helm app resource.
 	InsecureSkipTLSverify bool `json:"helmOpInsecureSkipTLSVerify,omitempty"`
+
+	// CABundle is a PEM encoded CA bundle used to validate TLS connections to
+	// the Helm registry. It is resolved by the controller (which has access to
+	// Rancher's cattle-system CA secrets) and stored here so the agent can use
```
weyfonk: nit: Fleet may be installed standalone...
```suggestion
	// Rancher's cattle-system CA secrets, if any) and stored here so the agent can use
```

e2e/single-cluster/helmop_test.go
```diff
@@ -520,6 +505,91 @@ var _ = Describe("HelmOp resource tests with tarball source", Label("infra-setup
...
+		Eventually(func(g Gomega) {
+			outDeployments, _ := k.Namespace(namespace).Get("deployments")
+			g.Expect(outDeployments).To(ContainSubstring("sleeper"))
+		}).Should(Succeed())
+
+		By("setting the expected version in the helmop Status")
+		Eventually(func() string {
+			out, _ := k.Get("helmop", name, "-o=jsonpath={.status.version}")
+			return out
+		}).Should(Equal("0.1.0"))
```
weyfonk: nit: do we need this? The value specified in the HelmOp is static, not a version constraint.

e2e/single-cluster/helmop_test.go
```diff
@@ -371,21 +371,6 @@ var _ = Describe("HelmOp resource with polling of OCI registry", Label("infra-se
 			})
 		})
 
-		Context("containing a valid helmop description pointing to an oci registry and not TLS", func() {
```
weyfonk, line 374: Why remove this test case?

integrationtests/helmops/controller/suite_test.go
```diff
@@ -128,6 +130,22 @@ var _ = BeforeSuite(func() {
...
+	})
+	Expect(err).ToNot(HaveOccurred())
+
+	DeferCleanup(func() {
+		_ = k8sClient.Delete(ctx, &corev1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "cattle-system",
+			},
+		})
+	})
```
weyfonk, line 149: note: this will only work when using an actual cluster, see [this](https://kubebuilder.io/reference/envtest.html#testing-considerations) for more details. Not an issue here though.

internal/cmd/controller/helmops/reconciler/helmop_controller.go
```diff
@@ -176,7 +177,17 @@ func (r *HelmOpReconciler) createUpdateBundle(ctx context.Context, helmop *fleet
...
+	// the agent (whose service account cannot read cattle-system secrets) can use it.
+	// The resolved bundle is also forwarded to handleVersion so getChartVersion can reuse
+	// it without a second cattle-system lookup.
+	cab, err := cert.GetRancherCABundle(ctx, r.Client)
+	if err != nil {
+		return nil, fmt.Errorf("could not get Rancher CA bundle: %w", err)
+	}
+	bundle.Spec.HelmOpOptions.CABundle = cab
+
+	if err := r.handleVersion(ctx, b, bundle, helmop, cab); err != nil {
```
weyfonk: Is `cab` needed as an additional parameter, given that it has just been populated into `bundle.Spec.HelmOpOptions.CABundle`, with the `bundle` itself being a parameter of `handleVersion`?

internal/cmd/controller/helmops/reconciler/helmop_controller.go
```diff
@@ -530,6 +546,20 @@ func getChartVersion(ctx context.Context, c client.Client, helmop fleet.HelmOp)
...
 	auth.InsecureSkipVerify = helmop.Spec.InsecureSkipTLSverify
 
+	// Fall back to a Rancher-configured CA bundle if no CA bundle is set.
+	// Use a pre-resolved bundle when available to avoid a redundant cattle-system lookup.
+	if len(auth.CABundle) == 0 {
+		if len(caBundle) > 0 {
+			auth.CABundle = caBundle
+		} else {
+			cab, err := cert.GetRancherCABundle(ctx, c)
+			if err != nil {
```
weyfonk, line 551: In the case of Fleet standalone installations, Rancher secrets will not be found, in which case this will return an error. We may want to consider ignoring "not found" errors.

internal/bundlereader/helm.go
```diff
@@ -32,6 +32,13 @@ func GetManifestFromHelmChart(ctx context.Context, c client.Reader, bd *fleet.Bu
 	}
 	auth.InsecureSkipVerify = bd.Spec.HelmChartOptions.InsecureSkipTLSverify
 
+	// Use the Rancher CA bundle that was pre-resolved by the controller and stored in
+	// HelmChartOptions.CABundle. The agent service account cannot read cattle-system
+	// secrets directly, so the controller must pass the bundle through.
```
weyfonk: nit (terminology overload on "bundle" 🙃 ):
```suggestion
	// secrets directly, so the controller must pass the CA bundle through.
```

internal/cmd/controller/helmops/reconciler/helmop_controller.go
thardeck: `GetRancherCABundle` in `pkg/cert/cabundle.go` already handles standalone Fleet correctly: it calls `client.IgnoreNotFound` on the `tls-ca` lookup and uses an explicit `errors.IsNotFound` check on `tls-ca-additional`. When both secrets are absent, both lookups return nil errors and the function returns `nil, nil`. This should be covered by the "no secrets found" unit test case in `cabundle_test.go`.

e2e/single-cluster/helmop_test.go
thardeck: The test is invalid with the new implementation. It verified that a chart would fail to deploy when no explicit CA bundle was configured and TLS verification was enabled. The new implementation now supplies the Rancher CA bundle automatically, so the chart does deploy — the test's assertion became incorrect.

weyfonk: Oh right, because [create-secrets.sh](https://github.com/rancher/fleet/blob/main/.github/scripts/create-secrets.sh) creates a Rancher-like CA bundle secret, and is called as part of the setup for this. Understood, thanks.

internal/cmd/controller/helmops/reconciler/helmop_controller.go
weyfonk: ... and you're right :)

PR #4731, opened 2026-03-02
Mark agentSchedulingCustomization nullable in CRD

thardeck:
Add `// +nullable` marker to the `AgentSchedulingCustomization` field so the generated CRD schema includes `nullable`: true. Without this, the API server may serialize the field as an empty object `{}` on clusters that never had it set, causing the Rancher UI to show it uncommented.

Refers to https://github.com/rancher/rancher/issues/53781

### DIFF
--- a/integrationtests/controller/cluster/nullability_test.go
+++ b/integrationtests/controller/cluster/nullability_test.go
@@ -1,0 +1,145 @@
+package cluster
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/rancher/fleet/integrationtests/utils"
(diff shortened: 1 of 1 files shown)


pkg/apis/fleet.cattle.io/v1alpha1/cluster_types.go
```diff
@@ -77,17 +77,23 @@ type ClusterList struct {
 }
 
 type AgentSchedulingCustomization struct {
-	PriorityClass       *PriorityClassSpec       `json:"priorityClass,omitempty"`
+	// +optional
```
weyfonk: Are these `+optional` markers necessary, given that they don't seem to have any impact on generated CRDs?

thardeck: They are not done by me, they are done by `go generate`.

PR #4751, opened 2026-03-04
Normalize empty agentSchedulingCustomization to nil

thardeck:
Some provisioning layers (e.g. Rancher) always initialize AgentSchedulingCustomization as a non-nil empty struct, which is serialized as agentSchedulingCustomization: {} in etcd.  OpenAPI-aware clients (Rancher UI) then show the field uncommented even when the user has not configured it.

The cluster controller now detects this state and patches the field to null on the next reconciliation cycle, so the stored value reflects the user intent rather than the initialization artifact.

Refers to https://github.com/rancher/rancher/issues/53781

Related to https://github.com/rancher/rancher/pull/54027


integrationtests/controller/cluster/nullability_test.go
```diff
@@ -142,4 +142,33 @@ var _ = Describe("Cluster agentSchedulingCustomization nullability", func() {
 		Expect(cleared.Spec.AgentSchedulingCustomization).To(BeNil(),
 			"agentSchedulingCustomization should be nil after being cleared via null merge patch")
 	})
+
+	It("is normalized from empty struct to nil by the controller", func() {
+		// This simulates what Rancher's provisioning layer does: it initializes
+		// AgentSchedulingCustomization as a non-nil empty struct, which gets
```
weyfonk, line 224: My main doubt here is: would Fleet's Cluster controller always have time to update this field before a user edits a cluster in the Rancher UI? At least, this approach should cover most cases.

integrationtests/controller/cluster/nullability_test.go
```diff
@@ -142,4 +142,33 @@ var _ = Describe("Cluster agentSchedulingCustomization nullability", func() {
...
+		}
+		Expect(k8sClient.Create(ctx, cluster)).ToNot(HaveOccurred())
+		DeferCleanup(func() {
+			Expect(k8sClient.Delete(ctx, cluster)).ToNot(HaveOccurred())
+		})
+
+		Eventually(func(g Gomega) {
+			normalized := &v1alpha1.Cluster{}
+			g.Expect(k8sClient.Get(ctx, types.NamespacedName{Name: cluster.Name, Namespace: namespace}, normalized)).ToNot(HaveOccurred())
+			g.Expect(normalized.Spec.AgentSchedulingCustomization).To(BeNil(),
```
weyfonk, line 246: nit: we could add a couple of test cases validating that, if the priority class or pod disruption budget is non-nil, no patching happens.

weyfonk (approved):

Issue #4756, opened 2026-03-05
Error when creating BundleDeployments under heavy load


0xavi0:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

There is an error when creating many _BundleDeployments_ under heavy load and with rolloutStrategy set to 100%.

The `CreateOrUpdate` call considers the resource does not exist and tries to create it, but it already exists in the k8s database. The problem is that the cache was still not in full sync. 

### Environment

```markdown
- Architecture:
- Fleet Version:
- Cluster:
  - Provider:
  - Options:
  - Kubernetes Version:
```

### Logs

```markdown

```

mmartin24:
Tested in `0.15.0-alpha.10` [here](https://github.com/rancher/fleet/issues/4599#issuecomment-3891707250) and working well. 
Closing issue

PR #4757, opened 2026-03-05
Fix BundleDeployment creation race condition

0xavi0:
Under heavy load and when building with optimizations we can call `CreateOrUpdate` for creating the _BundleDeployment_ twice for the same resource very quickly.

If it's fast enough, `CreateOrUpdate` gets a _NotFound_ error because the resource it's still not cached and tries to create the _BundleDeployment_ again.
    In that case, it calls `Create` and gets an _AlreadyExists_ error.

Because we return an error and retrigger, the `op` returned is not `controllerutil.OperationResultCreated` and the secret is not properly created.

The fix checks if `CreateOrUpdate` returns _AlreadyExists_ and, if so, it considers it as a create operation and the secret is properly handled.

Because the `Get` call returned `NotFound` we need to get the resource from the API Server, bypassing the cache that might still not be in sync. 
A new reader has been added to bypass the cache

~Note: it also readds `--platform=linux/$ARCH ` to the `Dockerfile` files as it is needed to cross-build Docker images for different architectures.~

Refers to: https://github.com/rancher/fleet/issues/4756

### Checklist

### DIFF
--- a/internal/cmd/controller/operator.go
+++ b/internal/cmd/controller/operator.go
@@ -120,7 +120,7 @@ func start(
 	if err = (&reconciler.BundleReconciler{
 		Client:    mgr.GetClient(),
 		Scheme:    mgr.GetScheme(),
-		Recorder:  mgr.GetEventRecorderFor(fmt.Sprintf("fleet-bundle-ctrl%s", shardIDSuffix)), //nolint:staticcheck -- GetEventRecorderFor is deprecated
+		Recorder:  mgr.GetEventRecorderFor(fmt.Sprintf("fleet-bundle-ctrl%s", shardIDSuffix)), //nolint:staticcheck // GetEventRecorderFor is deprecated
 		APIReader: mgr.GetAPIReader(),
 
 		Builder: builder,


package/Dockerfile
```diff
@@ -2,7 +2,7 @@ ARG BUILD_ENV=docker
 ARG TARGETPLATFORM
 ARG ARCH
 
-FROM registry.suse.com/bci/bci-base:15.7 AS base
+FROM --platform=linux/$ARCH registry.suse.com/bci/bci-base:15.7 AS base
```
thardeck: There might be an issue now with the release, there was a reason why I have removed it, but I will look into it.

thardeck: There is an issue using only `ARCH`, both issues fixed in https://github.com/rancher/fleet/pull/4763 .

The `$ARCH` variable is only set in the `docker build --build-arg ARCH=amd64` (direct dev/CI docker mode). In goreleaser mode, buildx doesn't pass `ARCH` — it passes `TARGETPLATFORM` (e.g. `linux/amd64`). Using bare `--platform=$ARCH` would expand to `--platform= (empty string)` under buildx/goreleaser and fail.

0xavi0: Let's address that in https://github.com/rancher/fleet/pull/4763
I've deleted the changes in the Docker files.

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -389,7 +389,7 @@ func (r *BundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 			// bundledeployments, but retry the whole reconcile
 			// afterwards.
 			merr = append(merr, fmt.Errorf("failed to create bundle deployment: %w", err))
-			logger.Info(fmt.Sprintf("failed to create a bundledeployment, skipping and requeuing: %v", err))
+			logger.Info(fmt.Sprintf("failed to create a bundledeployment %s/%s, skipping and requeuing: %v", bd.Namespace, bd.Name, err))
```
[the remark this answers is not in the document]
0xavi0: That was a leftover from testing. 
I've added the namespace and name inside the `createBundleDeployment` function so we add context when something fails. Otherwise we don't know exactly what was the BD that failed.

internal/cmd/controller/reconciler/bundle_controller.go
```diff
@@ -509,6 +509,13 @@ func (r *BundleReconciler) createBundleDeployment(
 		return nil
 	})
 	if err != nil {
+		if apierrors.IsAlreadyExists(err) {
+			// In this case CreateOrUpdate returned NotFound when running Get but
+			// because the BundleDeployment was being created and was still not in sync in the
+			// cache it got AlreadyExists when trying to Create it.
+			// Consider this as a non error and return OperationCreated.
+			return controllerutil.OperationResultCreated, bd, nil
```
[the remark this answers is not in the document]
0xavi0: I've added an API reader to bypass the cache and I get the resource when getting AlreadyExists error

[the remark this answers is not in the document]
0xavi0: Test covering the scenario added

package/Dockerfile
0xavi0: Let's address that in https://github.com/rancher/fleet/pull/4763
I've deleted the changes in the Docker files.

internal/cmd/controller/operator.go
```diff
@@ -118,9 +118,10 @@ func start(
 		shardIDSuffix = fmt.Sprintf("-%s", shardID)
 	}
 	if err = (&reconciler.BundleReconciler{
-		Client:   mgr.GetClient(),
-		Scheme:   mgr.GetScheme(),
-		Recorder: mgr.GetEventRecorderFor(fmt.Sprintf("fleet-bundle-ctrl%s", shardIDSuffix)),
+		Client:    mgr.GetClient(),
+		Scheme:    mgr.GetScheme(),
+		Recorder:  mgr.GetEventRecorderFor(fmt.Sprintf("fleet-bundle-ctrl%s", shardIDSuffix)), //nolint:staticcheck // GetEventRecorderFor is deprecated
```
weyfonk, line 123: This deprecation is addressed by https://github.com/rancher/fleet/pull/4678.

0xavi0: Good stuff.
I created an issue for not forgetting about it.
I'll assign the issue to that PR.

PR #4769, opened 2026-03-06
[v0.12] Separate integration tests module and bump docker/cli

thardeck:
Backports the integration test module separation from main and fixes
CVE-2025-15558 in `github.com/docker/cli`.

- Move `integrationtests/` into its own Go module (`github.com/rancher/fleet/integrationtests`). Go embeds its entire module build list in every binary, so `github.com/docker/docker` and `github.com/testcontainers/testcontainers-go` were appearing in `usr/bin/fleet` and `usr/bin/fleetcontroller` and triggering false-positive CVE reports in ArtifactHub. The new module uses replace directives pointing to the local working tree so `go mod tidy` never downloads a published version.
- Add `go.work` to link the three modules (`.`, `./integrationtests`, `./pkg/apis`) from the repo root. The workspace carries the `imdario/mergo` rename replace, which workspace mode requires for strict module path validation. Unlike v0.13/v0.14, v0.12 does not use the `gopkg.in/go-playground/webhooks.v6` import path, so that replace is omitted.
- Add `go.work.sum` to `.gitignore` and `*.test`. `go.work.sum` is generated dynamically per environment and produces non-substantive diffs; excluding it keeps commits clean.
- Bump `github.com/docker/cli` from `v27.5.0` to `v29.2.0` to fix CVE-2025-15558. CVE-2026-1229 (`cloudflare/circl`) was already resolved at `v1.6.3` by a prior commit.

### DIFF
--- a/integrationtests/go.mod
+++ b/integrationtests/go.mod
@@ -10,7 +10,7 @@ replace (
 )
 
 require (
-	github.com/docker/docker v28.3.3+incompatible
+	github.com/docker/docker v28.5.2+incompatible
 	github.com/docker/go-connections v0.5.0
 	github.com/go-git/go-git/v5 v5.16.5
 	github.com/go-logr/logr v1.4.2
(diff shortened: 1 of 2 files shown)


charts/fleet-crd/templates/crds.yaml
```diff
@@ -4360,10 +4360,7 @@ spec:
                                       and labelSelector.
 
                                       Also, matchLabelKeys cannot be set when labelSelector
-                                      isn''t set.
```
weyfonk, line 4363: Not quite sure what would trigger this change here 🤔

thardeck: Yes, I was also confused about that. But it seems to happen with the integration test part separation, not before.

weyfonk (approved):

PR #4774, opened 2026-03-09
Make sure Renovate updates integrations/go.mod

thardeck:
to prevent issues that it is not bumped by Renovate while `go mod tidy` would update it.

Also adjust the rule to only participates in updates when the same dependency is already being updated elsewhere in the workspace to prevent unwanted indirect update prs.


.github/renovate.json
```diff
@@ -14,6 +14,20 @@
...
+        "gomod"
+      ],
+      "matchDepTypes": [
+        "indirect"
+      ],
+      "matchFileNames": [
+        "integrationtests/go.mod",
+        "pkg/apis/go.mod"
+      ],
+      "enabled": true
```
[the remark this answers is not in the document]
thardeck: It seems to work, e.g. https://github.com/rancher/fleet/pull/4771/changes - this one before only bumped the main go.sum.

.github/renovate.json
```diff
@@ -14,6 +14,20 @@
     "**/assets/**"
   ],
   "packageRules": [
+    {
+      "description": "Update indirect dependencies in go.work child modules together with the root module",
+      "matchManagers": [
+        "gomod"
+      ],
+      "matchDepTypes": [
+        "indirect"
```
weyfonk, line 24: How much sense does it make to update indirect dependencies explicitly?
Wouldn't they either:
* be updated transitively (or at least we could reasonably expect that to happen) when updating direct dependencies?
* or risk breakage when updated explicitly, as direct dependencies may depend on older versions of those indirect dependencies?

thardeck: Renovate is skipping `integrationtests/go.mod` bump because the dependency there is indirect, so it never gets grouped with the root update. 

Without this change, Renovate did only bump the `./go.mod`. For example this pr https://github.com/rancher/fleet/pull/4771/changes resulted in failed CI because `go mod tidy` with `find` bumps the integration test modules.

I could remove `pkg/api/go.mod` because I do not remember having similar issues there.

thardeck: You were right, it resulted in additional prs trying to update indirect dependencies.

thardeck:
I have let renovate run with the new fix and it did not create any new pull requests. But we can leave this open until a new bump is there which has this issue to retry.

thardeck:
Blocked until we can test it with a Renovate pr which has the described issue.

thardeck:
The fix worked for https://github.com/rancher/fleet/pull/4820 and with this run: https://github.com/rancher/fleet/actions/runs/23132138555

PR #4780, opened 2026-03-10
Fix stale bundle error when cluster is offline after bad commit

thardeck:
When a GitRepo contains a YAML parse error and the cluster agent is offline, the bundle's Ready condition retains the error message even after a fix commit is pushed. Three interdependent changes are needed to clear the stale state.

- `deployer.go`: Add `MalformedYAMLError` to the `deployErrToStatus` regex. Helm v4 changed the error format from "YAML parse error" to "MalformedYAMLError"; without this match the error is routed to the `Deployed` condition instead of `Installed`, bypassing the staleness guard entirely.
- `summary.go`: In `MessageFromDeployment`, skip the `Installed` condition message when `AppliedDeploymentID` differs from `Spec.DeploymentID`, so a stale error from a superseded apply attempt is not surfaced. Unit tests added for all three condition precedence cases.
- `target.go`: Add `effectiveDeployment` so `state` and `message` compare against `t.DeploymentID` (the ID the controller is about to write) rather than the stale `Spec.DeploymentID` still held in the cached `BundleDeployment`. The bundle controller calls `SetReadyConditions` before updating BD specs, so the `summary.go` guard would otherwise never trigger while the agent is offline. Uses a shallow struct copy;

Refers #594


internal/cmd/controller/target/target.go
```diff
@@ -136,13 +136,26 @@ func (t *Target) state() fleet.BundleState {
...
+
+// effectiveDeployment returns t.Deployment with Spec.DeploymentID overridden
+// to t.DeploymentID, or t.Deployment unchanged when the IDs already match.
+// Status is computed (via state and message) before the reconciler writes the
+// new BD spec, so the cached Spec.DeploymentID lags at that point.
+func (t *Target) effectiveDeployment() *fleet.BundleDeployment {
+	if t.Deployment == nil || t.DeploymentID == t.Deployment.Spec.DeploymentID {
+		return t.Deployment
+	}
+	bd := *t.Deployment
```
weyfonk, line 156: This mutates the bundle deployment... is that really what we want?

integrationtests/controller/bundle/status_test.go
```diff
@@ -186,21 +186,35 @@ var _ = Describe("Bundle Status Fields", func() {
...
-			Eventually(func(g Gomega) {
-				err = k8sClient.Get(ctx, types.NamespacedName{Namespace: namespace, Name: "name"}, bd)
-				g.Expect(err).NotTo(HaveOccurred())
-				g.Expect(bd.Spec.DeploymentID).ToNot(Equal(deplIDBefore))
+		By("validating that the bundle deployment spec is updated")
+		Eventually(func() bool {
+			err = k8sClient.Get(ctx, types.NamespacedName{Namespace: namespace, Name: "name"}, bd)
+			Expect(err).NotTo(HaveOccurred())
+			return bd.Spec.DeploymentID != deplIDBefore
+		}).Should(BeTrue())
```
weyfonk: What's the reason for the indenting change and rewrite of this `Eventually` block?

thardeck: It was accidentally done, fixed in the new Merge commit.

internal/cmd/controller/target/target.go
thardeck: It should not. `BundleDeploymentSpec` is an inline struct field (not a pointer), so `bd := *t.Deployment` copies it by value. The `bd.Spec.DeploymentID = t.DeploymentID` assignment only touches the local copy; `t.Deployment.Spec.DeploymentID` is left unchanged. The returned `&bd` is passed to `GetDeploymentState` and `MessageFromDeployment`, both of which are read-only.

weyfonk: Oh, just realised my test case validating this did not do what I thought it did. Apologies, no mutation happens :)

PR #4796, opened 2026-03-11
Automate Fleet release against Rancher

thardeck:
Reduces the "release Fleet against rancher/rancher" workflow to a single dropdown input by computing Fleet/chart versions and the Rancher target branch from the `rancher/charts` dev branch via the GitHub API, removing the need to look up four values manually.

- `compute-rancher-versions.sh`: new script that derives `new_fleet`, `new_chart`, and `rancher_ref` from the latest chart directory in the `dev-v2.X` branch of `rancher/charts`; falls back to `main` when the Rancher release branch does not exist yet
- `release-against-rancher.sh`: removes unused Dockerfile fallback; adds explicit `build.yaml` guard; detects whether to bump the Fleet API by checking for a `pkg/apis/v<version>` tag in the fleet repo and derives the commit hash from that same tag, avoiding a mismatch with the root `v<version>` tag; uses explicit `go.mod`/`go.sum` paths instead of wildcards; errors out when the target or a newer version is already present in `build.yaml`
- `release-against-rancher.yml`: replaces four manual text inputs with a single `fleet_branch` choice input; all other values flow from the compute step

[Successful run](https://github.com/rancher/fleet/actions/runs/22988341636/job/66743649430)

### DIFF
--- a/.github/scripts/compute-rancher-versions.sh
+++ b/.github/scripts/compute-rancher-versions.sh
@@ -14,6 +14,7 @@ FLEET_REPO_DIR="${FLEET_REPO_DIR:-./fleet}"
 
 # Derive the Fleet minor version from the branch name.
 # For main, compute it as (highest release branch minor) + 1.
+# NOTE: Update 'v0.*' pattern if Fleet has a major version bump (e.g., search for 'v1.*').
 if [ "$FLEET_BRANCH" = "main" ]; then
     highest_minor=$(git -C "$FLEET_REPO_DIR" ls-remote --heads origin 'refs/heads/release/v0.*' \
         | grep -oE 'v0\.[0-9]+' | cut -d. -f2 | sort -n | tail -1)
@@ -26,7 +27,10 @@ else
     fleet_minor=$(printf '%s' "$FLEET_BRANCH" | grep -oE '[0-9]+$')
 fi
 
+# Calculate Rancher minor version based on Fleet minor.
+# NOTE: Update this formula if the version relationship changes (currently Fleet minor - 1).
 rancher_minor=$((fleet_minor - 1))
+# NOTE: Update '2' to '3' (or next major version) if Rancher has a major version bump.
 charts_branch="dev-v2.${rancher_minor}"
 
 # Fetch the Fleet chart directory listing from the rancher/charts dev branch.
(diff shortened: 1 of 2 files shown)


.github/scripts/compute-rancher-versions.sh
```diff
@@ -0,0 +1,74 @@
...
+    exit 1
+fi
+
+# Chart directory names follow the pattern <chart-version>+up<fleet-version>,
+# e.g. 110.0.1+up0.15.1.
+new_fleet="${latest_chart##*+up}"
+new_chart="${latest_chart%%+*}"
+
+# Target the Rancher release branch when it exists; fall back to main.
+rancher_ref="release/v2.${rancher_minor}"
```
mmartin24, line 60: I know we are far away but given this is hardcoded maybe add a note to be changed once we are in 3.x versioin?

thardeck: Good point, but where should I add it?
I guess when we switch to Rancher 3.x or Fleet 1.x we will have quite some places to adapt.
The calculation is at the moment `Minor Fleet Version - 1 = Rancher Minor Version` - that als would most likely not work anymore in both cases.

Trying to cover a wider variety of future cases would make the script significant more complicated I think, and there still can be changes which require us to fix things, like branch or repo changes and so on.

But the good thing is that we can not really break something here, worst case we'll get a weird pull request that we can close.

thardeck: I have added a comment. The drop-down list for the workflow is static, that we have to change every ~4 month (for every new Fleet minor release).
But since we manually have to branch it off anyway, I thought it is ok in combination with the convenience and higher reliability that the person releasing does not have to enter an exact or minor Fleet version themselves.

.github/scripts/compute-rancher-versions.sh
```diff
@@ -0,0 +1,79 @@
...
+    | sort -V | tail -1)
+
+if [ -z "$latest_chart" ]; then
+    printf 'ERROR: No Fleet chart directories found in rancher/charts branch %s\n' "$charts_branch" >&2
+    exit 1
+fi
+
+# Chart directory names follow the pattern <chart-version>+up<fleet-version>,
+# e.g. 110.0.1+up0.15.1.
+new_fleet="${latest_chart##*+up}"
```
weyfonk, line 55: This is a set of _existing_ Fleet and chart versions, isn't it? Wouldn't we then overwrite an existing chart?

PR #4841, opened 2026-03-18
Remove go-getter from GetContent and switch Docker image to bci-busybox

thardeck:
bundlereader.GetContent previously used `github.com/hashicorp/go-getter` for git and HTTP fetches, which shelled out to the `git` binary and controlled TLS only through process-global environment variables. This PR removes go-getter from go.mod entirely and replaces all fetching with native go-git and stdlib HTTP, preserving Fleet's `?ref=`, `?sshkey=`, `?depth=`, and `//subdir` semantics.

- SSH cloning now enforces known-hosts host-key validation when `SSHKnownHosts` is provided; previously all SSH clones used `InsecureIgnoreHostKey` unconditionally.
- `CABundle` is applied per-clone via `CloneOptions.CABundle`. go-git creates an isolated transport per session, so concurrent clones with different CAs are fully independent — no mutex needed.
- `setGitAuth` now only applies SSH auth when the URL scheme is `ssh` or `git`. An HTTPS URL combined with `?sshkey=` returns a clear error instead of failing with a confusing transport error.
- `extractQueryParams` preserves the original byte encoding and ordering of non-Fleet query parameters instead of round-tripping through url.Values, so presigned URLs are not corrupted.
- HTTP archive extraction covers `tar.gz/tgz`, `tar.bz2/tbz2`, `tar.zst/tzst/tar.zstd/zstd`, `tar.xz/txz`, `tar, gz`, `bz2`, `zst`, `xz`, and `zip`. Unknown extensions are written as plain files.
- `?checksum=<type>:<hex>` verifies the body via `TeeReader` without buffering. Supported types: `md5`, `sha1`, `sha256`, `sha512`. The `file` variant is not supported — it would require a recursive HTTP fetch with no meaningful security benefit over an inline hash.
- `?archive=<ext>` overrides format detection for URLs with no recognisable extension. Both ?checksum= and `?archive=` are stripped before the upstream request is made.
- Symlink and path traversal in tar and zip archives are rejected.
- Source shorthand detection covers GitHub, GitLab, and Bitbucket. SCP-style addresses (user@host:path) are normalised to `ssh://` URLs.
- `addAuthToOpts` in `fleet apply` now reads `FLEET_KNOWN_HOSTS` directly into opts.Auth.SSHKnownHosts instead of writing it to a temp file and mutating `GIT_SSH_COMMAND`.
- `transportForAuth` (Helm chart download) guards the http.DefaultTransport.(*http.Transport) assertion to avoid a panic when another component has replaced the global default.
- Switch controller image to the much smaller `bci-busybox` image with static tini. (**before:** `377 MB` - **after:** `220 MB`)

### Known difference to go-getter implementation
* No checksum url support for checksum files (e.g. `checksum=file:http://example.com/sha.txt`) because it would add quite some complexity. Could be done in a later pr if there is relevant interest.

### Additional go modules
* [ulikunitz/xz](https://github.com/ulikunitz/xz)
* [klauspost/compress](https://github.com/klauspost/compress), but it was already an indirect dependency before

Refers #4877

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -3,15 +3,18 @@ ARG TARGETPLATFORM
 ARG TARGETARCH
 ARG ARCH=${TARGETARCH}
 
-FROM --platform=linux/${ARCH} registry.suse.com/bci/bci-base:15.7 AS base
+FROM --platform=linux/${ARCH} registry.suse.com/bci/bci-busybox:15.7 AS base
+ARG ARCH
+# Download tini-static v0.19.0 (stable final release) and verify via embedded SHA256 checksum.
+RUN set -eux; \
+    wget -qO /usr/bin/tini "https://github.com/krallin/tini/releases/download/v0.19.0/tini-static-${ARCH}"; \
+    case "${ARCH}" in \
+      amd64) echo "c5b0666b4cb676901f90dfcb37106783c5fe2077b04590973b885950611b30ee  /usr/bin/tini" | sha256sum -c - ;; \
+      arm64) echo "eae1d3aa50c48fb23b8cbdf4e369d0910dfc538566bfd09df89a774aa84a48b9  /usr/bin/tini" | sha256sum -c - ;; \
+      *) echo "Unsupported architecture: ${ARCH}" >&2; exit 1 ;; \
+    esac; \
+    chmod 0755 /usr/bin/tini
 COPY package/log.sh /usr/bin/
-RUN zypper rm -y container-suseconnect && \
-    zypper ar --priority=500 https://download.opensuse.org/repositories/Virtualization:containers/5.5/Virtualization:containers.repo && \
-    zypper --gpg-auto-import-keys ref && \
-    zypper -n update && \
-    zypper -n install --no-recommends openssh-clients tini git-core && \
-    zypper -n clean -a && \
-    rm -fr /var/log/zypp* /usr/share/doc
 
 FROM base AS copy_docker
 ONBUILD ARG ARCH
@@ -29,7 +32,6 @@ COPY ${TARGETPLATFORM}/fleetcontroller-linux-* /usr/bin/fleetcontroller
 COPY ${TARGETPLATFORM}/fleet-linux-* /usr/bin/fleet
 
 FROM copy_${BUILD_ENV}
-RUN useradd -u 1000 user
 USER 1000
 ENTRYPOINT ["tini", "--"]
 CMD ["fleetcontroller"]
--- a/package/log.sh
+++ b/package/log.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 set -o pipefail
 env
 "$@" 2>&1 | tee /dev/termination-log


p-se:
`detectShorthand` in `source.go` intentionally omits `BitBucketDetector` with the note that it "makes a live HTTP request to determine the repository VCS type". That was true of go-getter's implementation — it queried `api.bitbucket.org/2.0/repositories/{user}/{repo}` to distinguish git from Mercurial.

However, Bitbucket dropped Mercurial support in 2020, so all Bitbucket repositories are git today. The network request is no longer needed to determine the VCS type, which means a `bitbucket.org/user/repo` shorthand can be unconditionally mapped to an HTTPS git URL — exactly like `detectGitHub` and `detectGitLab` already do.

Since this PR aims to preserve go-getter URL semantics, omitting Bitbucket shorthand is a silent regression: any `GitRepo` that previously relied on `bitbucket.org/user/repo` shorthand resolution will now silently fall through to the local-path handler. A `detectBitBucket` function matching the shape of the existing GitHub/GitLab detectors would restore parity without any network dependency.

p-se:
One more compatibility note on the HTTP archive path in `httpdownload.go`: go-getter supported a broader set of extraction formats — `tar.bz2`/`tbz2`, `tar.xz`/`txz`, `tar.zst`/`tzst`, and the single-file variants `bz2`, `xz`, and `zst`. The new implementation handles `tar`/`tar.gz`/`tgz`, `gz`, and `zip`, so those nine extensions are no longer extracted.

This is probably fine for the majority of users, but worth considering before merging: an HTTP source pointing at a `.tar.xz` or `.tar.zst` archive would previously have been extracted correctly and will now silently be written as a plain file instead — no error, just wrong content.

If re-adding the formats is not in scope, it would be good to document the supported extensions so users are not caught off guard during an upgrade.

p-se:
`httpDownload` dispatches to an extractor based solely on the URL path extension. go-getter supported a `?archive=` query parameter that let users override this — useful when a URL has no recognisable extension (e.g. a download endpoint like `https://host/artifacts/latest?token=abc`). Without it, a URL serving a `.tar.gz` archive under an unrecognisable path will be written as a plain file with no error, and the bundle will fail to apply for unrelated-looking reasons downstream.

p-se:
go-getter stripped `?checksum=<type>:<hash>` from the URL and verified the downloaded content against it before extraction. The new HTTP downloader does not handle this parameter — it is preserved as a non-Fleet query param and passed to the server, which will either ignore it or error. Either way, **no integrity check is performed and no warning is emitted**. Any user relying on `?checksum=` to pin a specific artifact version loses that guarantee silently after upgrading. This is worth an explicit call-out in the release notes at minimum, and ideally a supported replacement.

thardeck:
> go-getter stripped `?checksum=<type>:<hash>` from the URL and verified the downloaded content against it before extraction. The new HTTP downloader does not handle this parameter — it is preserved as a non-Fleet query param and passed to the server, which will either ignore it or error. Either way, **no integrity check is performed and no warning is emitted**. Any user relying on `?checksum=` to pin a specific artifact version loses that guarantee silently after upgrading. This is worth an explicit call-out in the release notes at minimum, and ideally a supported replacement.

I have added the checksum verification except that we do not support the referencing checksum files because this would add quite some complexity without much gain. We can add this in a later pr if explicitly requested.

Issue #4837, opened 2026-03-17
[BUG] Intermitent race condition on webhook with disabling polling true


mmartin24:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Issue:

Webhook seems not working all the time with disabling polling true. Webcommit field is correctly captured by not the updated commit. This is happening in 2.11, 2.12, 2.13 and 2.14 

Workaround: disable polling is disabled with time span largerly increased. 

---

Adding comment from @0xavi0:

> This happens when disablePolling: true and what’s different is that when polling is disabled instead of using the commit that comes in the webhook payload we call the fetcher to get the latest commit:

```go
if gitrepo.Spec.DisablePolling || clientSecretChanged {
			commit, err := monitorLatestCommit(gitrepo, func() (string, error) {
				return r.GitFetcher.LatestCommit(ctx, gitrepo, r.Client)
			})
```

> I suspect something has changed in github and we’re hitting a race condition between the webhook and the commit we get from the API.
> In the case of Github we use this header: req.Header.Set("Accept", "application/vnd.github.v3.sha") and we don’t really check the whole branch… it returns the top commit.
> I think we can fix this by comparing the commit we get from LatestCommit against the webhook one and, if they don’t match, requeue in 5s.
> 
> Reason why it was looking to the top commit is that we could be getting multiple webhooks and get them in the wrong order, so our source of truth should be the repository itself… but something is going on.

### Expected Behavior

Webhook should inmidiately work

### Steps To Reproduce

1. After installing Rancher prepare webhook ingress and and sync it with the provider (in our case in Github [here]2. (https://github.com/fleetqa/webhook-github-test/settings/hooks)
2. Add gitrepo against the specific webhook. I our case we used:
```yaml
apiVersion: fleet.cattle.io/v1alpha1
kind: GitRepo
metadata:
  name: webhook-test-disable-polling
  namespace: fleet-local
spec:
  branch: main
  correctDrift:
    enabled: false
  insecureSkipTLSVerify: false
  repo: https://github.com/fleetqa/webhook-github-test
  disablePolling: true
  targets:
  - clusterSelector:
      matchExpressions:
      - key: provider.cattle.io
        operator: NotIn
        values:
        - harvester
__clone: true
```
3. Change the replicas
4. Observe how the webcommit is captured but not the commit

### Environment

```markdown
- Architecture: AMD
- Fleet Version: 0.15.0-rc3
- Cluster:
  - Provider: k3s
[...]

PR #4842, opened 2026-03-18
Requeue when git API lags behind webhook commit

thardeck:
When `disablePolling: true`, a webhook delivers a new commit via `Status.WebhookCommit` and the reconciler calls `LatestCommit` to confirm it against the git server API. If the provider's API hasn't yet propagated the push, `LatestCommit` returns the previously-deployed commit. The original code wrote that stale value into `Status.Commit`, overwriting the webhook commit and causing `shouldCreateJob` to see no diff — so no deployment job was ever created.

- When the fetched commit differs from `Status.WebhookCommit`, restore `Status.Commit` to its pre-reconcile value before requeueing. Without this restore, the status write would persist `WebhookCommit` as `Status.Commit`, making the following reconcile blind to the change even after the API catches up.
- If the git server API never returns the webhook-announced commit (broken API, network partition, or slow propagation beyond any reasonable deadline), the reconciler previously requeued silently every 5 s with no user-visible indication. A new `GitRepoStatus.WebhookCommitStaleSince` field tracks when the reconciler first detected the mismatch. After `GitPollingStaleTimeout` (default 5 min), the `GitPolling` condition is set to `False` with reason `StaleAPI`, making the problem visible via `kubectl get gitrepo` and the Fleet UI. The condition clears automatically once the API returns the expected commit. The timer resets on `LatestCommit` errors so it tracks the "consistently returns old commit" case rather than an unreachable API (which is already reported separately).

Refers #4837

### DIFF
--- a/internal/cmd/controller/gitops/reconciler/gitjob_controller.go
+++ b/internal/cmd/controller/gitops/reconciler/gitjob_controller.go
@@ -353,6 +353,13 @@ func (r *GitJobReconciler) manageGitJob(ctx context.Context, logger logr.Logger,
 		// If the client secret has changed, we now retrieve the latest commit.
 		// If the secret is still incorrect, we will not need to create
 		// the gitJob (which is more expensive) and we will return an error earlier.
+
+		// If polling is enabled or there is no pending webhook commit, any previous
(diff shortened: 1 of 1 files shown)


thardeck:
Closing for now since we went for a simpler fix.

PR #4890, opened 2026-03-25
fix: honor insecureSkipTLS in OCI storage secret

khushalchandak17:
Refers to rancher/fleet#4898

This PR fixes OCI storage secret parsing to honor `insecureSkipTLS`.

Fleet currently reads the legacy `insecure` field from the OCI storage secret. If the secret is created with `insecureSkipTLS: true`, that value is not recognized, so downstream `fleet-agent` still performs TLS verification and may fail against registries using a self-signed/private CA.

This change:
- prefers `insecureSkipTLS`
- keeps `insecure` as a backward-compatible fallback
- updates test coverage to use `insecureSkipTLS`

## Additional Information

This is intended to be a minimal fix for OCI storage secret parsing and avoids breaking existing setups that may still use the legacy `insecure` field.

### DIFF
--- a/internal/ocistorage/secret.go
+++ b/internal/ocistorage/secret.go
@@ -115,6 +115,8 @@ func getBoolValueFromSecret(data map[string][]byte, key string, required bool) (
 	return boolValue, nil
 }
 
+// getBoolValueFromSecretWithFallback extracts a boolean value from data, using keys in the provided order of priority, and returns the first found value, if any.
+// If no value is found, the function returns false, with an error if the value was required.
 func getBoolValueFromSecretWithFallback(data map[string][]byte, required bool, keys ...string) (bool, error) {
 	for _, key := range keys {
 		if _, ok := data[key]; ok {


internal/ocistorage/secret.go
```diff
@@ -108,3 +114,15 @@ func getBoolValueFromSecret(data map[string][]byte, key string, required bool) (
 
 	return boolValue, nil
 }
+
+func getBoolValueFromSecretWithFallback(data map[string][]byte, required bool, keys ...string) (bool, error) {
```
weyfonk, line 120: Documenting what this function does could make sense, for instance:
```suggestion
// getBoolValueFromSecretWithFallback extracts a boolean value from data, using keys in the provided order of priority, and returns the first found value, if any.
// If no value is found, the function returns false, with an error if the value was required.
func getBoolValueFromSecretWithFallback(data map[string][]byte, required bool, keys ...string) (bool, error) {
```

Issue #3512, opened 2025-04-04
[SURE-10528] Allow more than one targetCustomization to be applied


jstanley1730:
### Is your feature request related to a problem?

We are running into an issue where we need to apply multiple targetCustomizations.  The current implementation applies the first targetCustomization in your list that matches.  We have targetCustomizations for all "edge" clusters vs "enterprise" clusters.  We also need to allow for additional customizations users may require to override values via configMap or secret if a given label is specified.   In many cases both need applied.

Example
```
name: istio-cni
defaultNamespace: kube-system
helm:
  releaseName: istio-cni
  repo: https://istio-release.storage.googleapis.com/charts
  chart: cni
  valuesFiles:
    - values.yaml

targetCustomizations:
  - name: istio-cni-values-override-configmap
    helm: 
      valuesFrom:
        - configMapKeyRef::
            name: istio-cni-values-override
            namespace: istio-system
            key: istio-cni-values.yaml    
    clusterSelector:
      matchLabels:
          istio-cni-values-override-configmap: 'true'
  - name: edge-values
    clusterSelector:
      matchLabels:
        edge: "true"
    helm:
      valuesFiles:
        - values/edge.yaml
```

### Solution you'd like

Would like the ability for all matching target customizations to be applied.

### Alternatives you've considered

We have combined labels for the matchSelection but when there is a longer list of targetCustomizations this gets unmaintainable and repetitive.

mandronic0:
[SURE-10528]

p-se:
# Issue
## Allow more than one `targetCustomization` to be applied

---

## What are `targetCustomizations`?

`fleet.yaml` lets you override Helm values, namespaces, etc. **per cluster**:

```yaml
# fleet.yaml
helm:
  valuesFiles:
    - values.yaml

targetCustomizations:
  - name: edge
    clusterSelector:
      matchLabels:
        edge: "true"
    helm:
      valuesFiles:
        - values/edge.yaml

  - name: enterprise
    clusterSelector:
      matchLabels:
        enterprise: "true"
    helm:
      valuesFiles:
        - values/enterprise.yaml
```

---

## The Problem

Targets are evaluated **in order** and **the first match wins** — no further entries are evaluated.

```go
// internal/cmd/controller/target/matcher/bundlematch.go

// MatchTargetCustomizations returns the first BundleTarget that matches...
func (a *BundleMatch) MatchTargetCustomizations(...) *fleet.BundleTarget {
[...]

PR #4940, opened 2026-04-07
Allow more than one targetCustomization to be applied

p-se:
Introduces TargetCustomizationMode on BundleSpec and fleet.yaml.
The default "FirstMatch" preserves existing behaviour. "AllMatches"
applies all matching targetCustomizations in order, merging their
options into the base deployment options.

Customization targets are now identified via hasMatchingRestriction()
rather than relying on their position in the Targets slice. For bundles created before this change, the position in the Targets slice is used as fallback.

Also reverts the independent doNotDeploy option introduced in https://github.com/rancher/fleet/pull/4688, since AllMatches can be used as a replacement.

Refers to #3512

### Checklist

- [x]  I have updated the documentation via a pull request in the [fleet-product-docs](https://github.com/rancher/fleet-product-docs) repository.

### DIFF
--- a/internal/cmd/controller/target/builder.go
+++ b/internal/cmd/controller/target/builder.go
@@ -73,6 +73,7 @@ func (m *Manager) Targets(ctx context.Context, bundle *fleet.Bundle, manifestID 
 		if err != nil {
 			return nil, false, err
 		}
+	clusterLoop:
 		for _, cluster := range clusters.Items {
 			logger.V(4).Info("Cluster has namespace?", "cluster", cluster.Name, "namespace", cluster.Status.Namespace)
 			clusterGroups, err := m.clusterGroupsForCluster(ctx, &cluster)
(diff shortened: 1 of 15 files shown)


internal/cmd/controller/target/matcher/bundlematch.go
```diff
@@ -125,13 +160,16 @@ func (m *matcher) isRestricted(clusterName, clusterGroup string, clusterGroupLab
...
 	}
 
-	// This target is a targetCustomization from a fleet.yaml
 	return true
 }
 
-// checks if criteria is matched just if the target is inside the targetRestrictions. This is used for Targets defined
-// in the GitRepo, since these targets are also added as targetRestrictions.
+// criteriaWithRestrictions checks that the cluster passes the restriction allowlist
+// and matches the target's cluster selector. Used for GitRepo targets only;
```
weyfonk, line 154: How about HelmOps targets? 🤔
My understanding is that they would be taken into account as well.

internal/cmd/controller/target/matcher/bundlematch.go
```diff
@@ -44,40 +44,41 @@ func (a *BundleMatch) Match(clusterName string, clusterGroups map[string]map[str
...
-	return false
+	return result
 }
 
 type targetMatch struct {
-	bundleTarget *fleet.BundleTarget
-	criteria     *ClusterMatcher
+	bundleTarget    *fleet.BundleTarget
+	criteria        *ClusterMatcher
+	isCustomization bool // true when this target comes from fleet.yaml targetCustomizations, not a GitRepo target
```
weyfonk:
```suggestion
	isCustomization bool // true when this target comes from fleet.yaml targetCustomizations, not a GitRepo or HelmOp target
```

internal/cmd/controller/target/matcher/bundlematch.go
```diff
@@ -113,8 +120,36 @@ func (a *BundleMatch) initMatcher() error {
...
+	// NEW BUNDLES: Source field is populated by bundlereader
+	// This is the preferred method for long-term maintainability
+	if target.Source != "" {
+		return target.Source == "customization"
+	}
+
+	// OLD BUNDLES: Source field is empty (created before this field existed)
+	// Use position-based detection as fallback:
+	// - bundlereader appends targetCustomizations first (read.go:187)
+	// - Then appends GitRepo targets from targets file (read.go:332)
```
weyfonk: nit: using a function name _could_ survive the test of time better than relying on line numbers, e.g.: 
```suggestion
	// - Then appends GitRepo targets from targets file (read.go:appendTargets)
```

internal/cmd/controller/target/matcher/bundlematch.go
```diff
@@ -113,8 +120,36 @@ func (a *BundleMatch) initMatcher() error {
...
+	// - Therefore: first N targets are customizations
+	//   where N = len(Targets) - len(TargetRestrictions)
+	//
+	// SPECIAL CASE: If there are no targetRestrictions, this bundle wasn't
+	// created by a GitRepo (e.g., CLI-loaded bundles, standalone bundles).
```
weyfonk: How about HelmOps? Isn't this logic run from the bundle reconciler, which also handles HelmOps bundles?
```suggestion
	// created by a GitRepo nor by a HelmOp (e.g., CLI-loaded bundles, standalone bundles).
```

internal/cmd/controller/target/matcher/bundlematch_test.go
```diff
@@ -0,0 +1,322 @@
...
+		{
+			name:              "old bundle - position-based gitrepo target",
+			target:            fleet.BundleTarget{Name: "prod", Source: ""}, // empty
+			index:             1,
+			numCustomizations: 1,
+			numRestrictions:   1,
+			want:              false, // index >= numCustomizations
+		},
+		{
+			name:              "collision case - old bundle position fixes bug",
```
weyfonk: Where would there be a collision? I don't follow this case.

internal/cmd/controller/target/builder.go
```diff
@@ -85,25 +86,52 @@ func (m *Manager) Targets(ctx context.Context, bundle *fleet.Bundle, manifestID
...
+				for _, tc := range matchedCustomizations {
+					if tc.DoNotDeploy {
+						logger.V(1).Info("Skipping BundleDeployment creation because doNotDeploy is set to true.",
+							"bundle", bundle.Name,
+							"bundleNamespace", bundle.Namespace,
+							"cluster", cluster.Name,
+							"clusterNamespace", cluster.Namespace,
+							"reason", "doNotDeploy on targetCustomization (AllMatches mode)",
+						)
+						continue clusterLoop
```
weyfonk, line 116: It would be great if this logic around `doNotDeploy` for both matching modes (first match vs all matches) could be tested.

weyfonk (commented):
Mostly LGTM, with a few comments and nitpicks.
In particular, I have doubts about the necessity for a new `Source` field on bundle targets; happy to discuss :)

internal/cmd/controller/target/matcher/bundlematch_test.go
p-se: No collision without Source field.

internal/cmd/controller/target/matcher/bundlematch.go
p-se: HelmOp targets aren't targetCustomizations. That field only exists on FleetYAML, not on the shared BundleSpec. A HelmOp's spec.targets already carry their  BundleDeploymentOptions inline and are applied directly via Match(). And since calculateBundle sets no TargetRestrictions, determineIsCustomization classifies every HelmOp target as a regular target, so they're never treated as customizations to be merged.

weyfonk: Oh, I had missed that bit. Thanks for clarifying :)

weyfonk (approved):

PR #4956, opened 2026-04-10
Catch unknown-field errors in deployErrToStatus

thardeck:
When Helm v4 uses server-side apply with strict field validation and the API server rejects a resource containing an unknown field, the error format is `<resource> is invalid: <field>: unknown field` — it does not contain `"error validating data"`. The old `deployErrToStatus` regex did not match this format, so the error was propagated to the agent reconciler which set `Deployed=False` and left `AppliedDeploymentID` pointing at a prior successful deployment ID.
With `Deployed=False` and mismatched IDs, `GetDeploymentState` returned `ErrApplied` and the staleness guard in `MessageFromDeployment` never triggered — the stale error persisted through every subsequent spec
update including the fix commit.

Adding `"(unknown field)"` to the pattern list ensures server-side-apply strict-validation rejections take the same path as YAML/schema errors: the error is stored in `Installed`, `Deployed` is set to `True`, and `AppliedDeploymentID` advances to the failing commit's ID. The ID-mismatch guard in `MessageFromDeployment` then correctly suppresses the stale message once the fix commit is pushed.

Refers #594

### DIFF
--- a/integrationtests/controller/bundle/bundle_error_handling_test.go
+++ b/integrationtests/controller/bundle/bundle_error_handling_test.go
@@ -442,4 +442,115 @@ var _ = Describe("Bundle controller error handling", Ordered, func() {
 			Expect(readyCond.Reason).To(BeEmpty(), "Ready condition Reason should be empty when Status is True")
 		})
 	})
+
+	// Issue #594 - stale error is suppressed once the spec advances past the bad commit, even
+	// when the cluster agent is offline and has not applied the fix yet.
+	// This covers the scenario where the agent's deployErrToStatus routes an "unknown field"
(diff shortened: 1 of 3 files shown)


p-se (approved):

thardeck:
# QA Template

- Repro once with an invalid manifest field (`server-side apply` strict validation): confirm the error is shown as an install/deploy status message and does not leave the `BundleDeployment` permanently stuck on an old deployment ID.
- Push a fix commit after the invalid one: confirm the stale unknown-field message is cleared and status follows the new deployment revision.
- If possible, do one delayed/offline `downstream-agent` update between bad and fixed commit: confirm reconnect does not re-surface the stale old message.

PR #5071, opened 2026-05-05
Fix doNotDeploy not removing off-schedule BundleDeployments

dockerized-nl:
## Summary

- `cleanupOrphanedBundleDeployments` had a guard `|| bd.Spec.OffSchedule` that preserved any BD with `OffSchedule=true`, even when that BD was no longer targeted (e.g. because `doNotDeploy: true` was set in `fleet.yaml` after the app was already deployed).
- This left the app installed on the downstream cluster indefinitely until the cluster re-entered its maintenance window — which could be days or weeks away.
- The fix removes the `OffSchedule` guard. The `uidsToKeep` set already correctly identifies all actively-targeted BDs; the extra guard was redundant for those and harmful for untargeted ones.

## Root Cause

In `cleanupOrphanedBundleDeployments` (`bundle_controller.go`):

```go
// Before — OffSchedule guard prevents deletion of untargeted BDs
return uidsToKeep.Has(bd.UID) || bd.Spec.OffSchedule

// After — only uidsToKeep determines what is kept
return uidsToKeep.Has(bd.UID)
```

`slices.DeleteFunc` removes elements where the predicate returns `true`, so any BD with `OffSchedule=true` was never added to the delete list — even when its cluster was no longer targeted at all.

## Behaviour

| Scenario | Before | After |
|---|---|---|
| `doNotDeploy` set, cluster in active window | BD deleted ✓ | BD deleted ✓ |
| `doNotDeploy` set, cluster in maintenance gap (`OffSchedule=true`) | BD **not** deleted ✗ | BD deleted ✓ |
| `doNotDeploy` not set, cluster in maintenance gap | BD kept ✓ | BD kept ✓ |

## Test plan

- [x] Added integration test `When("setting doNotDeploy after deployment when the existing BundleDeployment is off-schedule")` that explicitly sets `OffSchedule=true` on the BD before enabling `doNotDeploy`, then asserts the BD is cleaned up.
- [x] Existing `doNotDeploy` integration tests continue to pass.)


kkaempf:
thanks @dockerized-nl, esp. for the extensive comments within the code 💚

d3flex (approved):

p-se:
The intention of the maintenance window (a cluster being off schedule) is to prevent any modifications on a downstream cluster, which includes the removal of bundle deployments. Removing this guard would not only allow the removal of bundles on a downstream cluster that has received a `doNotDeploy` flag while the cluster is off schedule, but would enable the removal of bundles in general while the cluster is off schedule. If updates to the cluster are supposed to happen, the cluster needs to be on schedule.

Issue #5011, opened 2026-04-17
Created Git Jobs not inheriting node selectors from Fleet values.yaml


charleshu-8:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

I've set the Helm chart values `nodeSelector` and `tolerations` to confine Fleet to a certain set of nodes. The deployments generated by Fleet all have these node selectors and tolerations (e.g., `fleet-controller`, `gitjob`) but whenever a new K8s job is created in order to sync a GitRepo, the job is assigned with the tolerations but not the selectors, causing it to get scheduled on untainted nodes.

Digging a little deeper, it appears that the job will inherit tolerations from the Fleet Controller:
```go
// add tolerations from the fleet-controller deployment
jobSpec.Template.Spec.Tolerations = append(
	jobSpec.Template.Spec.Tolerations,
	fleetControllerDeployment.Spec.Template.Spec.Tolerations...,
)
```
([source](https://github.com/rancher/fleet/blob/main/internal/cmd/controller/gitops/reconciler/gitjob.go#L199))

but only inherits node selectors when Fleet is sharded:
```go
nodeSelector := map[string]string{"kubernetes.io/os": "linux"}
if shardID != "" && len(strings.TrimSpace(r.JobNodeSelector)) > 0 {
	var shardNodeSelector map[string]string
	if err := json.Unmarshal([]byte(r.JobNodeSelector), &shardNodeSelector); err != nil {
		return nil, fmt.Errorf("could not decode shard node selector: %w", err)
	}

	maps.Copy(nodeSelector, shardNodeSelector)
}
```
([source](https://github.com/rancher/fleet/blob/main/internal/cmd/controller/gitops/reconciler/gitjob.go#L439))

I don't want to shard Fleet, so the jobs stay selector-less and aren't scheduled correctly for me.

### Expected Behavior

The ability to assign node selectors to generated jobs without having the enable Fleet sharding.

### Environment

```markdown

```

### Logs

```markdown

```

weyfonk:
## Additional QA
 
### Problem
Installing Fleet with a non-empty `nodeSelector` Helm value would constrain Fleet controller deployments, but that would not have any effect on git jobs, which pods could still be created on any node.
 
### Solution
Git job creation now takes the `fleet-controller` deployment's node selector into account, and includes it into the job spec.
 
### Testing

### Engineering Testing
#### Manual Testing
1. Spinned up a k3d cluster
[...]

PR #5086, opened 2026-05-07
fix: inherit node selectors for GitRepo sync jobs

immanuwell:
Refers to #5011

GitRepo sync Jobs copied tolerations from the Fleet controller Deployment, but not its `nodeSelector`.

Repro:
1. Install Fleet with `nodeSelector`, no sharding.
2. Force-sync or create a `GitRepo`.
3. Check the generated sync Job: it only has `kubernetes.io/os=linux`, so it can schedule outside the intended node pool. Not great.

This copies the controller Deployment `nodeSelector` into generated GitRepo sync Jobs. Shard selectors still work the same way.

Related: #2579 and #2620 added GitJob/shard selector wiring; this covers the non-sharded Helm values path.

Tested:
```sh
go test ./internal/cmd/controller/gitops/...
```

AI assistance: ChatGPT helped draft this small patch and test update; I checked the behavior against the existing GitJob scheduling code.

Docs: not needed, behavior-only bug fix

### DIFF
--- a/internal/cmd/controller/gitops/reconciler/gitjob.go
+++ b/internal/cmd/controller/gitops/reconciler/gitjob.go
@@ -200,6 +200,11 @@ func (r *GitJobReconciler) newGitJob(ctx context.Context, obj *v1alpha1.GitRepo)
 		jobSpec.Template.Spec.Tolerations,
 		fleetControllerDeployment.Spec.Template.Spec.Tolerations...,
 	)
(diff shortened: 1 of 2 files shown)


weyfonk (approved):
LGTM; we may still want to document that any shard-level `nodeSelector` will be overwritten by the global, non-sharded `nodeSelector`, if any.

weyfonk:
Thanks @immanuwell for your contribution :)

Issue #4897, opened 2026-03-26
Fleet Agent cannot pull Bundle from OCI Storage with self signed CA


SR4ven:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

In a fully Rancher provisioned setup with an rke2 downstream cluster, the fleet-agent in the downstream cluster fails to pull the fleet bundle from the oci registry because it doesn't trust its certificate which is signed by a custom CA.
```
{"level":"error","ts":"2026-03-26T14:21:21Z","msg":"Reconciler error","controller":"bundledeployment","controllerGroup":"fleet.cattle.io","controllerKind":"BundleDeployment","BundleDeployment":{"name":"infra-common-cert-manager","namespace":"cluster-fleet-default-playground-130f5a23df29"},"namespace":"cluster-fleet-default-playground-130f5a23df29","name":"infra-common-cert-manager","reconcileID":"663446f9-9198-446e-af5f-f2f9b219e8e6","error":"failed deploying bundle: failed to perform \"FetchReference\" on source: Get \"https://redacted/s-904eb49f284e823a11b8c6cf9e952211bca46f4bd1e2b43b62d075a9d471d/manifests/latest\": tls: failed to verify certificate: x509: certificate signed by unknown authority","errorCauses":[{"error":"failed deploying bundle: failed to perform \"FetchReference\" on source: Get \"https://redacted/s-904eb49f284e823a11b8c6cf9e952211bca46f4bd1e2b43b62d075a9d471d/manifests/latest\": tls: failed to verify certificate: x509: certificate signed by unknown authority"}],"stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/internal/controller/controller.go:353\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/internal/controller/controller.go:300\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.1\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/internal/controller/controller.go:202"}
```
The fleet-controller can upload bundles into the oci registry without issues.

[...]

Issue #3554, opened 2025-04-14
Remove logrus usage


manno:
We use the structured logger instead. 

Notes:
* `logrus.Fatal` -> `log.Fatal`
*  `setupLog = ctrl.Log.WithName("setup")` in controllers
* `logrus.Warn` -> `logger.Error`
* `logrus.Debug` -> `logger.V(1).Info`
* ...

VltraHeaven:
Hello. I'd like to be assigned to take ownership of this issue. Let me know what steps I need to go through when you are able. Thank you.

mmartin24:
Checked in `0.16.0-alpha3` with this content in fleet-local

```
repo: https://github.com/fleetqa/fleet-qa-examples-public, branch main, path helm-oci-auth
Helm auth: HTTP, private GitHub username/password
helmRepoURLRegex: empty
```

A failed helm oci auth prints the log in UI in full:

```
response status code 401: unauthorized: authentication required: helmRepoURLRegex is empty, so Helm credentials were not forwarded; set spec.helmRepoURLRegex to allow credential forwarding
```

but it seems in the terminal prints correctly the `401` but silently fails the `Helm credentials will not be forwarded to any repository`

```
kubectl -n fleet-local logs job/local-oci-467-a215f --all-containers | grep "401"

{"time":"2026-09-21T15:09:21.014085928Z","level":"ERROR","msg":"Fleet cli failed","fleetErrorMessage":"helm-oci-auth: failed to process bundle: failed reading resources for \"helm-oci-auth\": loading directory .chart/2ef620094470f3cc4f086ee76d63fcca52a581db7b9ec35fa052fe32d05d9e91, helm-oci-auth: downloading OCI chart from \"oci://ghcr.io/fleetqa/fleet-qa-examples/fleet-test-configmap-chart\": helm chart download: failed to perform \"FetchReference\" on source: GET \"https://ghcr.io/v2/fleetqa/fleet-qa-examples/fleet-test-configmap-chart/manifests/0.1.0\": GET \"https://ghcr.io/token?scope=repository%3Afleetqa%2Ffleet-qa-examples%2Ffleet-test-configmap-chart%3Apull&service=ghcr.io\": response status code 401: unauthorized: authentication required: helmRepoURLRegex is empty, so Helm credentials were not forwarded; set spec.helmRepoURLRegex to allow credential forwarding"}

[...]

PR #5128, opened 2026-05-15
Use Custom CAs when pulling OCI artifact

0xavi0:
This PR adds custom CAs when downloading OCI artifacts in a similar way we already use them when downloading helm charts.

Refers to: https://github.com/rancher/fleet/issues/4897

### Checklist

[x]  I have updated the documentation via a pull request in the [fleet-product-docs](https://github.com/rancher/fleet-product-docs) repository.

### DIFF
--- a/e2e/single-cluster/oci_registry_test.go
+++ b/e2e/single-cluster/oci_registry_test.go
@@ -54,6 +54,7 @@ func createOCIRegistrySecret(
 			Namespace: namespace,
 		},
 		Data: map[string][]byte{
+			ocistorage.OCISecretCABundle:        []byte(""),
 			ocistorage.OCISecretReference:       []byte(reference),
 			ocistorage.OCISecretUsername:        []byte(username),
 			ocistorage.OCISecretPassword:        []byte(password),
--- a/e2e/single-cluster/oci_registry_test.go
+++ b/e2e/single-cluster/oci_registry_test.go
@@ -54,7 +54,6 @@ func createOCIRegistrySecret(
 			Namespace: namespace,
 		},
 		Data: map[string][]byte{
-			ocistorage.OCISecretCABundle:        []byte(""),
 			ocistorage.OCISecretReference:       []byte(reference),
 			ocistorage.OCISecretUsername:        []byte(username),
 			ocistorage.OCISecretPassword:        []byte(password),
--- a/internal/ocistorage/ociwrapper.go
+++ b/internal/ocistorage/ociwrapper.go
@@ -88,7 +88,9 @@ func getHTTPClient(insecureSkipTLS bool, caBundle []byte) *http.Client {
 		if !tmpPool.AppendCertsFromPEM(proxyBytes) {
 			logrus.Warnf("%s is set but contains no valid PEM certificates; ignoring proxy CA bundle", fleetgit.ProxyCABundleEnvVar)
 		} else {
-			caBundle = append(caBundle, '\n')
+			if len(caBundle) > 0 && caBundle[len(caBundle)-1] != '\n' {
+				caBundle = append(caBundle, '\n')
+			}
 			caBundle = append(caBundle, proxyBytes...)
 		}
 	}
@@ -98,18 +100,23 @@ func getHTTPClient(insecureSkipTLS bool, caBundle []byte) *http.Client {
 		return retry.DefaultClient
 	}
 
-	// Build custom transport with TLS config
-	transport := &http.Transport{
-		TLSClientConfig: &tls.Config{
-			InsecureSkipVerify: insecureSkipTLS, // #nosec G402
-		},
+	// Clone the default transport to preserve proxy, timeout, and
+	// connection-pooling settings.
+	baseTransport, ok := http.DefaultTransport.(*http.Transport)
+	if !ok {
+		baseTransport = &http.Transport{
+			Proxy: http.ProxyFromEnvironment,
+		}
+	}
+	transport := baseTransport.Clone()
+	transport.TLSClientConfig = &tls.Config{
+		InsecureSkipVerify: insecureSkipTLS, // #nosec G402
 	}
 
 	// Merge custom CA bundle with system cert pool
 	if len(caBundle) > 0 {
 		pool, err := x509.SystemCertPool()
 		if err != nil {
-			// Fall back to empty pool if system pool unavailable
 			pool = x509.NewCertPool()
 		}
 		if !pool.AppendCertsFromPEM(caBundle) {
@@ -119,7 +126,7 @@ func getHTTPClient(insecureSkipTLS bool, caBundle []byte) *http.Client {
 		transport.TLSClientConfig.MinVersion = tls.VersionTLS12
 	}
 
-	return &http.Client{Transport: transport}
+	return &http.Client{Transport: retry.NewTransport(transport)}
 }
 
 func getAuthClient(opts OCIOpts) *auth.Client {
--- a/internal/ocistorage/ociwrapper.go
+++ b/internal/ocistorage/ociwrapper.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -16,13 +17,16 @@ import (
 
 	"github.com/opencontainers/go-digest"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
-	"github.com/rancher/fleet/internal/manifest"
+	"github.com/sirupsen/logrus"
 	"oras.land/oras-go/v2"
 	"oras.land/oras-go/v2/content"
 	"oras.land/oras-go/v2/content/memory"
 	"oras.land/oras-go/v2/registry/remote"
 	"oras.land/oras-go/v2/registry/remote/auth"
 	"oras.land/oras-go/v2/registry/remote/retry"
+
+	"github.com/rancher/fleet/internal/manifest"
+	fleetgit "github.com/rancher/fleet/pkg/git"
 )
 
 const (
@@ -40,6 +44,7 @@ type OCIOpts struct {
 	AgentPassword   string
 	BasicHTTP       bool
 	InsecureSkipTLS bool
+	CABundle        []byte
 }
 
 type OrasOps interface {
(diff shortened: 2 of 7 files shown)


internal/ocistorage/secret.go
```diff
@@ -21,6 +21,7 @@ const (
 	OCISecretBasicHTTP       = "basicHTTP"
 	OCISecretInsecureSkipTLS = "insecureSkipTLS"
 	OCISecretInsecure        = "insecure" // legacy alias
+	OCISecretCABundle        = "cacerts"
```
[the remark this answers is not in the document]
0xavi0: Documentation will be added when the PR is approved

internal/ocistorage/ociwrapper.go
```diff
@@ -72,20 +77,54 @@ func NewOCIWrapper() *OCIWrapper {
...
+			pool = x509.NewCertPool()
+		}
+		if !pool.AppendCertsFromPEM(caBundle) {
+			logrus.Warnf("CA bundle contains no valid PEM certificates; proceeding with system cert pool only")
+		}
+		transport.TLSClientConfig.RootCAs = pool
+		transport.TLSClientConfig.MinVersion = tls.VersionTLS12
+	}
+
+	return &http.Client{Transport: transport}
```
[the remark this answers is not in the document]
0xavi0: change as suggested

internal/ocistorage/ociwrapper.go
```diff
@@ -72,20 +77,54 @@ func NewOCIWrapper() *OCIWrapper {
...
+	caBundle = append([]byte(nil), caBundle...)
+
+	// Merge proxy CA bundle if present
+	if proxyCAPEM, ok := os.LookupEnv(fleetgit.ProxyCABundleEnvVar); ok && proxyCAPEM != "" {
+		proxyBytes := []byte(proxyCAPEM)
+		tmpPool := x509.NewCertPool()
+		if !tmpPool.AppendCertsFromPEM(proxyBytes) {
+			logrus.Warnf("%s is set but contains no valid PEM certificates; ignoring proxy CA bundle", fleetgit.ProxyCABundleEnvVar)
+		} else {
+			caBundle = append(caBundle, '\n')
```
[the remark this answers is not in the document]
0xavi0: change as suggested

internal/ocistorage/ociwrapper_test.go
```diff
@@ -136,17 +137,146 @@ var _ = Describe("OCIUtils tests", func() {
...
+ByqGSM49AgEGBSuBBAAiA2IABMFEoc8Rl1Ca3iOCNQfN0MsYndLxf3c1TzvdlHJS
+7cI7+Oz6e2tYIOyZrsn8aLN1udsJ7MgT9U7GCh1mMEy7H0cKPGEQQil8pQgO4CLp
+0zVozptjn4S1mU1YoI71VOeVyaNCMEAwHQYDVR0OBBYEFMFRRVBZqz7nLFr6ICIS
+B4CIfBFqMA4GA1UdDwEB/wQEAwIBhjAPBgNVHRMBAf8EBTADAQH/MAoGCCqGSM49
+BAMDA2gAMGUCMQCJao1H5+z8blUD2WdsJk6Dxv3J+ysTvLd6jLRl0mlpYxNjOyZQ
+LgGheQaRnUi/wr4CMEfDFXuxoJGZSZOoPHzoRgaLLPIxAJSdYsiJvRmEFOml+wG4
+DXZDjC5Ty3zfDBeWUA==
+-----END CERTIFICATE-----`
+		os.Setenv("PROXY_CA_BUNDLE", proxyCA)
+		defer os.Unsetenv("PROXY_CA_BUNDLE")
```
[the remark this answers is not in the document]
0xavi0: defer also works fine with Ginkgo and its inline with other tests that also use it this way

internal/ocistorage/ociwrapper.go
```diff
@@ -72,20 +77,61 @@ func NewOCIWrapper() *OCIWrapper {
...
+func getHTTPClient(insecureSkipTLS bool, caBundle []byte) *http.Client {
+	// Defensive copy to avoid mutations
+	caBundle = append([]byte(nil), caBundle...)
+
+	// Merge proxy CA bundle if present
+	if proxyCAPEM, ok := os.LookupEnv(fleetgit.ProxyCABundleEnvVar); ok && proxyCAPEM != "" {
+		proxyBytes := []byte(proxyCAPEM)
+		tmpPool := x509.NewCertPool()
+		if !tmpPool.AppendCertsFromPEM(proxyBytes) {
+			logrus.Warnf("%s is set but contains no valid PEM certificates; ignoring proxy CA bundle", fleetgit.ProxyCABundleEnvVar)
```
weyfonk, line 89: nit: we may eventually want to create our `OCIWrapper`s with a logger, to pass it through the chain of calls from `pushFile` et al. down to `getHTTPClient`, and get rid of the `logrus` dependency.
There's already an [issue](https://github.com/rancher/fleet/issues/3554) to that end though.

internal/ocistorage/ociwrapper.go
```diff
@@ -72,20 +77,61 @@ func NewOCIWrapper() *OCIWrapper {
 	}
 }
 
-func getHTTPClient(insecureSkipTLS bool) *http.Client {
-	if !insecureSkipTLS {
+func getHTTPClient(insecureSkipTLS bool, caBundle []byte) *http.Client {
+	// Defensive copy to avoid mutations
+	caBundle = append([]byte(nil), caBundle...)
```
weyfonk, line 82: Do we really need this? Aren't slices passed by value? (that's what my tests and [this piece](https://trstringer.com/golang-slice-references/) suggest)

internal/ocistorage/ociwrapper_test.go
```diff
@@ -136,17 +137,157 @@ var _ = Describe("OCIUtils tests", func() {
...
+		client := getHTTPClient(true, nil)
+
+		// Custom path wraps transport in retry.Transport
+		retryTransport, ok := client.Transport.(*retry.Transport)
+		Expect(ok).To(BeTrue())
+		innerTransport, ok := retryTransport.Base.(*http.Transport)
+		Expect(ok).To(BeTrue())
+		Expect(innerTransport.TLSClientConfig).ToNot(BeNil())
+		Expect(innerTransport.TLSClientConfig.InsecureSkipVerify).To(BeTrue())
+		Expect(innerTransport.Proxy).ToNot(BeNil())
```
weyfonk, line 149: Why do we expect a non-nil proxy transport here?

internal/ocistorage/ociwrapper_test.go
```diff
@@ -136,17 +137,157 @@ var _ = Describe("OCIUtils tests", func() {
...
+		retryT, ok := client.Transport.(*retry.Transport)
+		Expect(ok).To(BeTrue())
+		transport, ok := retryT.Base.(*http.Transport)
+		Expect(ok).To(BeTrue())
+		Expect(transport.TLSClientConfig).ToNot(BeNil())
+		Expect(transport.TLSClientConfig.RootCAs).ToNot(BeNil())
+		Expect(transport.TLSClientConfig.MinVersion).To(Equal(uint16(tls.VersionTLS12)))
+		Expect(transport.TLSClientConfig.InsecureSkipVerify).To(BeFalse())
+
+		// Verify the CA bundle was actually loaded by testing it can be parsed
```
weyfonk: I don't understand this sentence 🤔 
What are we trying to do here? Verifying the input CA bundle, which is already done by `getHTTPClient`?

Or is this missing a few lines, comparing `testPool` with `tranport.TLSClientConfig.RootCAs`?

... unless this is related to the expectation that slices be passed by reference, which, in my understanding, is not the case. This also means that the last test case of this patch would not be necessary.

e2e/single-cluster/oci_registry_test.go
```diff
@@ -54,6 +54,7 @@ func createOCIRegistrySecret(
 			Namespace: namespace,
 		},
 		Data: map[string][]byte{
+			ocistorage.OCISecretCABundle:        []byte(""),
```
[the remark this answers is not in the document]
weyfonk: Do we need this though?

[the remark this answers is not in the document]
0xavi0: you're right, it's indeed not needed.
It was there because when creating the secret `fleet apply` was setting `opts.CABundle` even when it was nil. 
Fixed now, thanks!

internal/ocistorage/ociwrapper.go
0xavi0: Given that we already have an issue for this... could we implement the changes when taking care of the issue?

internal/ocistorage/ociwrapper.go
0xavi0: Slices are, indeed, passed by value, but there are Go internals related on how slices are backed up by arrays that could make an slice to be mutated.
It's a bit overkill, but I saw we were already using it [here](https://github.com/rancher/fleet/blob/b2593d348fdc711b736a8da513e53d54b943e6f5/internal/bundlereader/charturl.go#L277) 

Some more info and examples [here](https://utcc.utoronto.ca/~cks/space/blog/programming/GoSliceMutability)

internal/ocistorage/ociwrapper_test.go
0xavi0: The Proxy field is non-nil because we build the transport by cloning http.DefaultTransport (in
getHTTPClient), and Go's default transport ships with Proxy: http.ProxyFromEnvironment. 

The assertion is there as a guard: it ensures we keep going through baseTransport.Clone() rather than constructing a fresh &http.Transport{} in a future refactor, which would silently drop
HTTP_PROXY/HTTPS_PROXY/NO_PROXY support for users behind a corporate proxy.

internal/ocistorage/ociwrapper_test.go
0xavi0: I guess that was leftovers of a test that was never finished.... 

Removed

internal/ocistorage/ociwrapper.go
weyfonk: Sure, that can be addressed when we work on that issue.

internal/ocistorage/ociwrapper.go
weyfonk: Oh, right, so whether or not slices are mutated depends on an implementation detail, namely on whether the underlying array has extra capacity, TIL. Thanks for clarifying.

internal/ocistorage/ociwrapper_test.go
weyfonk: Understood, thanks.

weyfonk (approved):

PR #5177, opened 2026-05-20
Propagate per-target DownstreamResources to clusters

### LINKED ISSUE #4727: Experimental `downstreamResources` does not take effect in helmop.spec.targets

- [x] I have searched the existing issues

### Current Behavior

I am attempting to use the experimental `downstreamResources` feature by assigning a unique secret to individual clusters, I am doing this by targeting a specific cluster by name, and assigning a secret to that cluster, see example below for more details. 

### Expected Behavior

I would expect `helmop.spec.targets.downstreamResources` and `helmop.spec.targets.helm` to work in the same way. I expect the secret I define in `helmop.spec.targets.downstreamResources` to be passed to the specific cluster targeted. 

### Steps To Reproduce

The following HelmOp is a slightly cut down version of what I am using. The HelmOp targets clusters in the "developmet" group, with the `fleet.application.deploy/democratic-nfs: "true"` label, all of those clusters should get the `dev-nfs-driver-config` pushed down to them from fleet-default in the rancher cluster, but they do not, no secret is pushed to the development clusters.
```yaml
---
kind: HelmOp
metadata:
  name: democratic-nfs
  namespace: fleet-default
spec:
  defaultNamespace: democratic-nfs
  helm:
    chart: democratic-csi
    releaseName: democratic-nfs
    repo: https://democratic-csi.github.io/charts/
    takeOwnership: true
    values:
      csiDriver:
        fsGroupPolicy: File
        name: org.democratic-csi.nfs
      driver:
        config:
          driver: freenas-api-nfs
      storageClasses:
      - allowVolumeExpansion: true
        defaultClass: false
        mountOptions:
        - noatime
        - nfsvers=3
        - nolock
        name: truenas-nfs
        parameters:
          detachedVolumesFromSnapshots: "false"
          fsType: nfs
        reclaimPolicy: Delete
        volumeBindingMode: Immediate
    version: 0.15.1
  targets:
  - clusterGroup: development
    clusterSelector:
      matchLabels:
        fleet.application.deploy/democratic-nfs: "true"
    downstreamResources:
    - kind: Secret
      name: dev-nfs-driver-config
    helm:
      values:
        driver:
          existingConfigSecret: dev-nfs-driver-config
    name: development
```

### Environment

```markdown
- Architecture: x86_64
- Fleet Version: v0.14.3
- Cluster:
  - Provider: Harvester
  - Options: Nothing special/relevant 
  - Kubernetes Version: v1.34.3+rke2r1
```

### Logs

```markdown

```

### Anything else?

_No response_

thardeck:
Three separate bugs prevented `downstreamResources` in `HelmOp.spec.targets[*]` (and `fleet.yaml` `targetCustomizations[*].downstreamResources`) from being copied to downstream cluster namespaces. `options.Merge` did not copy the `DownstreamResources` field from target customizations, so the merged result in `bd.Spec.Options` was always empty for per-target entries. `handleDownstreamObjects` read only `bundle.Spec.DownstreamResources` (the root-level field) rather than `bd.Spec.Options.DownstreamResources`, which holds the post-merge value. The bundle change-trigger indexer also only walked root-level resources, so modifying a per-target secret or configmap did not cause reconciliation.

The indexer fix iterates `bundle.Spec.Targets` so that changes to per-target secrets/configmaps re-trigger bundle reconciliation.

Fixes #4727

### DIFF
--- a/integrationtests/controller/bundle/bundle_downstream_resources_test.go
+++ b/integrationtests/controller/bundle/bundle_downstream_resources_test.go
@@ -78,7 +78,7 @@ var _ = Describe("Bundle target DownstreamResources", Ordered, func() {
 			Expect(err).NotTo(HaveOccurred())
 			Expect(bundle).NotTo(BeNil())
 
-			By("verifying three BundleDeployments are created")
+			By("verifying that a BundleDeployment is created for each matched cluster")
 			bdLabels := map[string]string{
 				"fleet.cattle.io/bundle-name":      bundleName,
 				"fleet.cattle.io/bundle-namespace": namespace,
@@ -199,7 +199,7 @@ var _ = Describe("Bundle target DownstreamResources", Ordered, func() {
 			}
 			Expect(k8sClient.Create(ctx, bundle)).To(Succeed())
 
-			By("verifying three BundleDeployments are created")
+			By("verifying that a BundleDeployment is created for each matched cluster")
 			bdLabels := map[string]string{
 				"fleet.cattle.io/bundle-name":      bundleName,
 				"fleet.cattle.io/bundle-namespace": namespace,
--- a/integrationtests/controller/bundle/bundle_downstream_resources_test.go
+++ b/integrationtests/controller/bundle/bundle_downstream_resources_test.go
@@ -1,0 +1,253 @@
+package bundle
+
+import (
+	"os"
+	"strings"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/rancher/fleet/integrationtests/utils"
+	"github.com/rancher/fleet/internal/experimental"
+	"github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ = Describe("Bundle target DownstreamResources", Ordered, func() {
+	BeforeAll(func() {
+		var err error
+		namespace, err = utils.NewNamespaceName()
+		Expect(err).ToNot(HaveOccurred())
+
+		ns := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace}}
+		Expect(k8sClient.Create(ctx, ns)).ToNot(HaveOccurred())
+
+		createClustersAndClusterGroups()
+
+		// Enable the experimental downstream copy feature.
+		Expect(os.Setenv(experimental.CopyResourcesDownstreamFlag, "true")).To(Succeed())
+
+		DeferCleanup(func() {
+			os.Unsetenv(experimental.CopyResourcesDownstreamFlag)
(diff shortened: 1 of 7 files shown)


internal/cmd/controller/reconciler/bundle_controller_test.go
```diff
@@ -1477,6 +1480,9 @@ func TestReconcile_DownstreamResourcesGeneration_Increment(t *testing.T) {
 				Return(&k8serrors.StatusError{ErrStatus: metav1.Status{Code: http.StatusNotFound}})
 
 			// Target builder returns our test bundle deployment
+			// Simulate what options.Merge would produce: root-level DownstreamResources
+			// end up in the BD's Options after the first reconcile.
+			tc.existingBundleDeployment.Spec.Options.DownstreamResources = tc.downstreamResources
```
weyfonk, line 1485: nit: we could add a few words to clarify that this would be done by the target builder (which is here mocked).

internal/cmd/controller/options/calculate.go
```diff
@@ -110,6 +110,9 @@ func Merge(base, custom fleet.BundleDeploymentOptions) fleet.BundleDeploymentOpt
 	if custom.CorrectDrift != nil {
 		result.CorrectDrift = custom.CorrectDrift
 	}
+	if len(custom.DownstreamResources) > 0 {
+		result.DownstreamResources = append(result.DownstreamResources, custom.DownstreamResources...)
```
weyfonk, line 114: We could deduplicate downstream resources in case both the base and the customisation reference the same resource. That would prevent duplicate copy operations, and even more with multiple target customisations about to be supported (#4940).
(and by the way we could eventually do it for `ValuesFrom` too, and `ComparePatches`: e.g. in the latter case, if overlap is detected, the customization could take precedence)
Happy to discuss, this could be the object of a separate issue.

integrationtests/controller/bundle/bundle_downstream_resources_test.go
```diff
@@ -0,0 +1,253 @@
...
+				},
+				{
+					ClusterGroup: "all",
+				},
+			}
+			bundle, err := utils.CreateBundle(ctx, k8sClient, bundleName, namespace, targets, targets)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(bundle).NotTo(BeNil())
+
+			By("verifying three BundleDeployments are created")
```
weyfonk: nit: this may be easier to follow:
```suggestion
			By("verifying that a BundleDeployment is created for each matched cluster")
```

integrationtests/controller/bundle/bundle_downstream_resources_test.go
```diff
@@ -0,0 +1,253 @@
...
+								ClusterGroup: t.ClusterGroup,
+							}
+						}
+						return out
+					}(),
+				},
+			}
+			Expect(k8sClient.Create(ctx, bundle)).To(Succeed())
+
+			By("verifying three BundleDeployments are created")
```
weyfonk: nit: Same as previous comment.

weyfonk (approved):
LGTM, thanks for this :)

internal/cmd/controller/options/calculate.go
thardeck: Agreed, deduplication makes sense especially in light of #4940. Left as a follow-up - I have opened the issue #5195 to track it separately.

weyfonk (approved):

Issue #5020, opened 2026-04-20
Regression in drift detection for changes to fields with default values


p-se:
Fleet does not report drift when a field that is absent (or null) in the bundle is manually changed on the live object to a non-default value.

Affected versions: v0.15, main

Steps to reproduce:

1. Deploy a bundle with imagePullPolicy: null (or left absent). For instance:
```yaml
apiVersion: fleet.cattle.io/v1alpha1
kind: Bundle
metadata:
  name: drift-test
  namespace: fleet-local
spec:
  targets:
    - clusterSelector: {}

  resources:
    - content: |
        apiVersion: apps/v1
        kind: Deployment
        metadata:
          name: drift-test
          namespace: drift-test
        spec:
          selector:
            matchLabels:
              app: drift-test
          template:
            metadata:
              labels:
                app: drift-test
            spec:
              containers:
                - name: nginx
                  image: nginx:stable-alpine
                  imagePullPolicy: Always

  defaultNamespace: drift-test


```

2. Patch the live deployment:
```
kubectl patch deployment ... -p '{"spec":{"template":{"spec":{"containers":[{"name":"nginx","imagePullPolicy":"Always"}]}}}}'
```

3. Bundle remains Ready / not modified — drift is not reported

**Expected**: Fleet detects that the live object differs from the desired state.

**Root cause**: Commit 69284410 ("chore(deps): bump to kubernetes 1.35", Jan 2026) switched scheme.go from k8s.io/kubernetes's legacyscheme to client-go's scheme on main (and release/v0.15), making Scheme.Default() a no-op. PR #4664 introduced normalizeNullPatch to fix the resulting false-positive drift, but as a side effect it strips all null entries from the diff patch — including nulls that represent a legitimate user change to a field the bundle left unset.

Detailed of the investigation are shared in https://github.com/rancher/fleet/issues/4969#issuecomment-4261944202.

**Workaround**: Explicitly set every field you want drift protection on in the bundle or chart.

rancherbot:
Not creating port issue, milestone v0.15 does not exist or is not an open milestone

p-se:
## QA Template

- Follow the reproduction steps in the issue description
- Unpatched should not report drift whereas patched should see a drift.

sbulage:
Changed Milestone as backport of same issue for 2.14.3 is verified already.

sbulage:
### System Information

| Rancher Version | Fleet Version |
|--------|--------|
[...]

PR #5179, opened 2026-05-21
fix: preserve array nulls in drift patches

puneetdixit200:
Refers to #5020

Preserve arrays while normalizing null-only map fields in drift patches. JSON merge patch arrays are replacement values, so recursively deleting nulls inside them can hide real drift such as container `imagePullPolicy` changes. Map-level null cleanup still handles Helm/default noise.

AI assistance: Used OpenAI GPT-5 to inspect and validate this focused change; I reviewed the code and tests.

## Additional Information

Tested:
- `go test ./internal/cmd/agent/deployer/desiredset -run 'Test_Diff_NullPatch|Test_Diff_RemoveNullPatchFields' -count=1`\n- `go test ./internal/cmd/agent/deployer/desiredset -run 'Test_Diff_HPA' -count=1`\n- `go test ./internal/cmd/agent/deployer/desiredset -count=1`\n- `go test -shuffle=on ./internal/cmd/agent/deployer/desiredset -count=1`\n- `git diff --check`\n\n### Checklist\n\n- [ ] I have updated the documentation via a pull request in the [fleet-product-docs](https://github.com/rancher/fleet-product-docs) repository. Not applicable to this bug fix.

### DIFF
--- a/internal/cmd/agent/deployer/desiredset/diff.go
+++ b/internal/cmd/agent/deployer/desiredset/diff.go
@@ -452,17 +452,16 @@ func normalizeNullPatch(
 	return false, nil
 }
 
-// removeNullPatchFields recursively removes null values from patch data structures.
-// It handles both maps (objects) and slices (arrays), preserving non-null values.
+// removeNullPatchFields recursively removes null values from patch objects.
 //
 // Map behavior:
 //   - Skips entries with nil values
 //   - Skips entries that become empty maps after cleaning
 //
 // Slice behavior:
-//   - Skips nil elements
-//   - Skips elements that become empty maps after cleaning
-//   - Preserves non-map elements (strings, numbers, etc.)
+//   - Preserves arrays as-is. JSON merge patches replace arrays atomically, so
+//     changing nulls inside an array changes the replacement value and can hide
+//     real drift in keyed lists such as container specs.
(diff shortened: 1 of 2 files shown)


puneetdixit200:
AI-assisted follow-up on behalf of Puneet: pushed `2f8807d` to clarify the `removeNullPatchFields` doc comment so it matches the array-preserving behavior.

Verified:
- `go test ./internal/cmd/agent/deployer/desiredset -run 'Test_Diff_NullPatch|Test_Diff_RemoveNullPatchFields' -count=1`\n- `gofmt -w internal/cmd/agent/deployer/desiredset/diff.go`\n- `git diff --check`

p-se:
I'm afraid we took a wrong turn here by introducing `normalizeNullPatch` in the first place. By removing `legacyscheme` we have removed Fleet's ability to precisely calculate the default values that Kubernetes adds to resources. `normalizeNullPatch` was trying to resolve this issue, but couldn't replace the missing functionality, which is why we got false positives. Instead of trying to mitigate those false positives, which may not reliably work either, we can either reintroduce those defaults from `legacyscheme`, or do a dry-run of a server-side apply to let Kubernetes provide those defaults for us. The former will be reintroduced by https://github.com/rancher/fleet/pull/5215.

Thanks for your time on this. Closing.

PR #5180, opened 2026-05-21
Normalize YAML flow-style docs in post-renderer [SURE-11642]

thardeck:
Helm v4's kyaml round-trips rendered manifests through its YAML parser, converting JSON output from `toJson` template functions into YAML flow-style with unquoted keys (e.g. `{apiVersion: v1, kind: ConfigMap}`). `k8s.io/apimachinery`'s `IsJSONBuffer` treats the leading `{` as a signal that the document is already valid JSON and returns it unchanged; `json.Unmarshal` then fails with `invalid character 'a' looking for beginning of object key string`, causing the bundle deployment to fail in the post-render phase.

- Add `normalizeFlowStyleDocs` in `postrender.go`: for each rendered document starting with `{` that does not pass `json.Valid`, round-trip it through `sigs.k8s.io/yaml` to produce block-style YAML before passing to `yaml.ToObjects`. Valid JSON documents are not touched.
- Add `hasFlowStyleCandidate` as a zero-allocation fast-path: a raw byte scan checks whether any document boundary starts with `{`; when none does (the common case), the function returns immediately without allocating or rebuilding the buffer.

Refers #5212

### DIFF
--- a/internal/helmdeployer/postrender.go
+++ b/internal/helmdeployer/postrender.go
@@ -28,18 +28,23 @@ import (
 
 const CRDKind = "CustomResourceDefinition"
 
+var (
+	yamlDocMarker    = []byte("---")
+	newlineDocMarker = []byte("\n---")
+)
+
 // isYAMLDocMarker reports whether b starts with a valid YAML document-start
 // marker: "---" followed by whitespace, CR, LF, or end of input. Per the
 // YAML spec, "---" is only a marker when it is the complete token on a line;
 // a sequence like "---foo" is NOT a marker and must not be stripped.
 func isYAMLDocMarker(b []byte) bool {
-	if !bytes.HasPrefix(b, []byte("---")) {
+	if !bytes.HasPrefix(b, yamlDocMarker) {
 		return false
 	}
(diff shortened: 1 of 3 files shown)


weyfonk (approved):
Nice one, thanks!

internal/helmdeployer/postrender.go
```diff
@@ -18,10 +22,117 @@ import (
...
+		if firstNonSpace(rest) == '{' {
+			return true
+		}
+		// Searching for "\n---" also covers CRLF separators ("\r\n---") because
+		// "\n---" is a substring of "\r\n---".
+		i := bytes.Index(rest, []byte("\n---"))
+		if i < 0 {
+			return false
+		}
+		rest = rest[i+4:]
```
d3flex: is it better to use the length of  the marker?

thardeck: Since the values are static, it is not necessary to use `len()` I suppose. It might increases the readability but then we probably should move the DocMarker to their own variables.

weyfonk (approved):

PR #5188, opened 2026-05-24
fix: collect dump metrics services across namespaces

immanuwell:
fixes a small hole in `fleet dump` metrics collection.

right now it only looks in `cattle-fleet-system`. that misses metrics when Fleet is installed in another Helm release namespace, which is a normal setup.

how to repro:
1. install Fleet with metrics enabled into a namespace like `custom-fleet-system`
2. run `fleet dump -p dump.tgz`
3. no `metrics_*` files show up, even though `monitoring-fleet-controller*` and `monitoring-gitjob*` services exist

what changed:
- list monitoring services across namespaces
- keep the `monitoring-` name filter, so scope stays tight
- include namespace in metrics archive filenames, just in case names overlap

tests:
- added a regression test for a monitoring service in a non-default namespace
- ran `go test ./internal/cmd/cli/dump ./internal/cmd/cli ./internal/cmd/cli/apply -count=1`

### DIFF
--- a/internal/cmd/cli/dump/dump.go
+++ b/internal/cmd/cli/dump/dump.go
@@ -483,12 +483,10 @@ func addEventsToArchive(
 }
 
 func addMetricsToArchive(ctx context.Context, c client.Client, logger logr.Logger, cfg *rest.Config, w *tar.Writer, opt Options) error {
-	ns := config.DefaultNamespace // XXX: support installation in non-default namespace, and check for services across all namespaces, by label?
-
 	var monitoringSvcs []corev1.Service
 	var svcs corev1.ServiceList
 	for {
-		opts := []client.ListOption{client.InNamespace(ns), client.Limit(opt.FetchLimit), client.Continue(svcs.Continue)}
+		opts := []client.ListOption{client.Limit(opt.FetchLimit), client.Continue(svcs.Continue)}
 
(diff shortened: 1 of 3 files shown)


internal/cmd/cli/dump/dump.go
```diff
@@ -483,12 +483,10 @@ func addEventsToArchive(
 }
 
 func addMetricsToArchive(ctx context.Context, c client.Client, logger logr.Logger, cfg *rest.Config, w *tar.Writer, opt Options) error {
-	ns := config.DefaultNamespace // XXX: support installation in non-default namespace, and check for services across all namespaces, by label?
-
 	var monitoringSvcs []corev1.Service
 	var svcs corev1.ServiceList
 	for {
-		opts := []client.ListOption{client.InNamespace(ns), client.Limit(opt.FetchLimit), client.Continue(svcs.Continue)}
+		opts := []client.ListOption{client.Limit(opt.FetchLimit), client.Continue(svcs.Continue)}
```
weyfonk, line 489: Right, fetching monitoring services across _all_ namespaces may be overkill in some setups, but we could revisit this to extract the namespace in which the Fleet release is installed, if needed.

weyfonk (approved):

0xavi0 (approved):

Issue #5198, opened 2026-05-27
[BUG]  `hasLookupFunction() ` does not detect lookup usage inside helper templates (define blocks)


susesamu:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Fleet currently fails to detect Helm lookup usage when the lookup call is located inside helper templates defined in `_helpers.tpl`.
This causes charts that rely on helper-based lookup logic (such as `gha-runner-scale-set`) to fail when deployed through Fleet, even though they succeed with native Helm CLI.
The issue appears to be in `hasLookupFunction()` only inspecting `t.Tree.Root` and not traversing all parsed templates returned by `t.Templates()`.

### Expected Behavior

Fleet should detect lookup usage regardless of whether the function appears directly in a manifest template, inside a define block / _helpers.tpl or helper templates referenced through include/template.

### Steps To Reproduce

A minimal reproducer:

_helpers.tpl
```
{{- define "mychart.lookupHelper" -}}
{{ lookup "v1" "Service" "default" "kubernetes" }}
{{- end -}}
```
configmap.yaml
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: test
data:
  value: {{ include "mychart.lookupHelper" . }}
```

Expected:

`hasLookupFunction(chart) == true`

Actual:

`hasLookupFunction(chart) == false`

### Environment

```markdown
- Architecture:
- Fleet Version:
- Cluster:
  - Provider:
  - Options:
  - Kubernetes Version:
```

### Logs

```markdown

```

sbulage:
### System Information

| Rancher Version | Fleet Version |
|--------|--------|
| v2.15-fab1af47577c3a5b41f730c9aa19debd6ce346e6-head | 110.0.0+up0.16.0-alpha.10 | 

### Verification
- Created custom chart which installs `configmap`
- It also has `_helper.tpl` which fetches Service information in the configmap.
- After creating GitRepo, Fleet properly scan all template files, allowing the lookup function to run and pull live cluster data.
- Verified that it updates the data from helper funtion. (See below screenshot for more information).

<details><summary>Screenshot showing <code>configmap</code> with data fetch using helper function</summary>
<p>

<img width="1626" height="942" alt="Image" src="https://github.com/user-attachments/assets/043a405e-3435-4201-91db-d7fb4ec6e057" />

</p>
</details>

PR #5199, opened 2026-05-27
Fix lookup detection for helper templates and define blocks

susesamu:
Refers to https://github.com/rancher/fleet/issues/5198
Fix `hasLookupFunction()` failing to detect Helm lookup usage inside helper templates (define blocks), such as templates located in `_helpers.tpl`.
Previously Fleet only inspected` t.Tree.Root,` which represents the root parse tree for a template file. However, Go template define blocks are internally stored as separate named templates and exposed through` t.Templates()`.
As a result, lookup calls inside helper templates were ignored, causing Fleet to incorrectly conclude that a chart did not use lookup.

### Changes
Iterate over all parsed templates returned by `t.Templates()`. Detect lookup usage inside:
- helper templates
- define blocks
- _helpers.tpl

### DIFF
--- a/internal/helmdeployer/lookup.go
+++ b/internal/helmdeployer/lookup.go
@@ -31,8 +31,12 @@ func hasLookupFunction(ch *chartv2.Chart) bool {
 		}
 
 		// Walk all parse trees in this template and look for lookup invocations.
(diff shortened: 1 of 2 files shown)


0xavi0 (approved):

PR #5202, opened 2026-05-28
Add tests for namespace label sync

thardeck:
Adds pod-security label filtering tests and a table case verifying no-op behaviour when the namespace already reflects the desired state.

- `TestAddLabelsFromOptions_PodSecurityLabelsFiltered`: covers the rule that `pod-security.kubernetes.io/*` labels from `NamespaceLabels` options are never applied to the namespace, while existing pod-security labels on the namespace are preserved.
- `TestSetNamespaceLabelsAndAnnotations_PodSecurityLabelsPreserved`: end-to-end check through `setNamespaceLabelsAndAnnotations` confirming pod-security labels on the namespace survive a sync.


sbulage (approved):
LGTM 👍

PR #5203, opened 2026-05-28
[v0.15] Add tests for namespace label sync

thardeck:
Backports test coverage for the namespace label synchronisation fix in #5172.

- `TestSetNamespaceLabelsAndAnnotations_NoUpdateWhenAlreadyCorrect`: uses an interceptor to assert no `Update` call is made when the namespace already reflects the desired state, guarding against a regressed equality check.
- `TestAddLabelsFromOptions_PodSecurityLabelsFiltered`: verifies that `pod-security.kubernetes.io/*` labels from `NamespaceLabels` options are never applied to the namespace while existing pod-security labels are preserved.
- `TestSetNamespaceLabelsAndAnnotations_PodSecurityLabelsPreserved`: end-to-end check through `setNamespaceLabelsAndAnnotations` confirming pod-security labels on the namespace survive a sync.

Backports #5202

### DIFF
--- a/internal/cmd/agent/deployer/deployer_test.go
+++ b/internal/cmd/agent/deployer/deployer_test.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"testing"
 
+	"github.com/go-logr/logr"
+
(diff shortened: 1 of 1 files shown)


sbulage (approved):
LGTM 👍

Issue #5036, opened 2026-04-22
Make downstream resource copy non-experimental


weyfonk:
Since v0.14.0 (and Rancher v2.13.0), Fleet has the ability to automatically copy resources to downstream clusters ([docs](https://fleet.rancher.io/experimental-features/experimental-downstream-resource)).

See #3617 for more info.

Note that #4727 should ideally be addressed within the same release or before this issue.

With Application Collection integration [work](https://github.com/rancher/dashboard/issues/14627) planned in the Rancher UI for v2.15.0, it would make sense for this feature not to be experimental anymore.

Instead, it should be considered regular, mature functionality, and enabled without a specific environment variable needing to be set on Fleet deployments.

sbulage:
QA Testing for `downstreamResources` is already done --> https://github.com/rancher/fleet/issues/3617#issuecomment-3443124372

PR #5207, opened 2026-05-28
Remove experimental guard from downstream resource copy

thardeck:
The downstream resource copy feature (copying secrets and config maps to
downstream clusters via `downstreamResources`) is promoted from experimental
to stable. The `EXPERIMENTAL_COPY_RESOURCES_DOWNSTREAM` environment variable
is removed and the feature is always enabled.

- The configmaps ClusterRole rule in RBAC bootstrap is now unconditional
- The env var is no longer propagated from the controller to agent pod manifests

Refers #5036

### DIFF
--- a/internal/cmd/controller/agentmanagement/controllers/resources/data.go
+++ b/internal/cmd/controller/agentmanagement/controllers/resources/data.go
@@ -31,12 +31,7 @@ func ApplyBootstrapResources(systemNamespace, systemRegistrationNamespace string
 		{
 			Verbs:     []string{"get"},
 			APIGroups: []string{""},
-			Resources: []string{"secrets"},
-		},
-		{
-			Verbs:     []string{"get"},
-			APIGroups: []string{""},
-			Resources: []string{"configmaps"},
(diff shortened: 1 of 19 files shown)


internal/cmd/controller/agentmanagement/controllers/resources/data.go
```diff
@@ -34,14 +33,11 @@ func ApplyBootstrapResources(systemNamespace, systemRegistrationNamespace string
...
 			Resources: []string{"secrets"},
 		},
-	}
-
-	if experimental.CopyResourcesDownstreamEnabled() {
-		rules = append(rules, rbacv1.PolicyRule{
+		{
 			Verbs:     []string{"get"},
 			APIGroups: []string{""},
 			Resources: []string{"configmaps"},
```
weyfonk: nit: this could probably be collapsed into a single rule for `"get"` with `Resources: []string{"secrets", "configmaps"},`

weyfonk (approved):
LGTM :)

Issue #4877, opened 2026-03-20
Consolidate Fleet's git client implementations to go-git


thardeck:
Fleet uses different git implementations depending on the operation:

- `internal/cmd/cli/apply/` (resolve current commit SHA): shell out to `git rev-parse HEAD`
- `internal/cmd/cli/gitcloner/` (gitjob clone): go-git
- `internal/bundlereader/` (Helm chart from `helm.chart: git::https://…`): go-getter, which shells out to the `git` binary for SSH
- `pkg/git/` (ref listing, liveness): go-git

The shell-out paths have several problems. go-getter uses `GIT_SSH_COMMAND` for SSH clones (global process state requiring a mutex), replaces the system cert pool with any custom CA bundle (breaking connections to public CAs), does not enforce SSH known-host verification, and does not propagate `InsecureSkipVerify` from the GitRepo spec. The `github.com/hashicorp/go-getter` dependency also generates false-positive CVE reports in ArtifactHub. The `git rev-parse` shell-out in `apply/` is a minor leftover with a `TODO` to migrate.

All git operations should use go-git directly so the same auth settings (CA bundle, credentials, known-hosts, InsecureSkipVerify) apply across all code paths.

After consolidating the git implementation in Fleet it should be possible to remove at least the git package from the Fleet Docker image.

thardeck:
# QA Template

Most basic go-git clone/reconcile scenarios are already well covered in integration and E2E tests.

## Suggested QA checks

- Verify one real `helm.chart: git::...` deployment using your typical production auth path (secret/CA/known_hosts). This is the key end-to-end behavior changed by the go-getter removal.
- Verify HTTPS git source with custom CA and with `insecureSkipTLSVerify` disabled. Expect clear cert failure; then add correct CA and expect recovery without recreating objects.
- Verify SSH source with `SSHKnownHosts`: success with correct host key, hard failure with wrong key. Focus on condition/event clarity.
- Verify HTTP artifact source that relies on `?archive=` and `?checksum=` query params (no file extension URL is best) - e.g. specifying a `helm.chart` url. Confirm extraction/integrity still works and those params are not forwarded upstream.
- Verify one negative auth case (bad token/user/pass or wrong SSH key) and confirm no credential leakage in logs/events/status while error text remains actionable.

## Deeper verification hints

- Run two GitRepos concurrently with different CA bundles to spot cross-request TLS state leaks.
[...]

PR #5211, opened 2026-06-01
Improve go-getter replacement

thardeck:
Three potential issues were identified in the unreleased go-getter replacement code from #4841 and #4891.

- `httpDownload` extracted archive content to disk before verifying `?checksum=`, so a tampered server response could write files even when the hash would not match. The response is now written to a temporary file, the hash verified, and only then extracted. Streaming is preserved for the common case (no checksum).
- `GetContent`'s `WalkDir` callback passed every non-directory entry to `os.ReadFile`, which follows symlinks. A git repository or HTTP archive containing a symlink to `/etc/something` would include the target's content in the bundle. A `filepath.EvalSymlinks` check now rejects symlinks whose resolved target escapes the bundle root, while allowing legitimate intra-bundle symlinks (consistent with what `extractTar` does for tar archives).
- `redactURL` returned its input unchanged when `url.Parse` failed, and the wrapped `*url.Error` echoed the full URL verbatim — leaking an embedded password or `?sshkey=` private key into logs. `redactURL` now redacts textually on parse failure, and a new `redactParseError` strips the raw URL from the wrapped error.

Refers #4877

### DIFF
--- a/internal/bundlereader/gitclone_test.go
+++ b/internal/bundlereader/gitclone_test.go
@@ -271,7 +271,7 @@ func TestGitDownloadMalformedURLRedaction(t *testing.T) {
 	)
 	rawURL := "ssh://user:" + password + "@example.com:notaport/repo?sshkey=" + privateKey
 
-	err := gitDownload(context.Background(), t.TempDir(), rawURL, Auth{})
+	err := gitDownload(t.Context(), t.TempDir(), rawURL, Auth{})
 	require.Error(t, err)
 	assert.NotContains(t, err.Error(), password, "password must not leak in error")
 	assert.NotContains(t, err.Error(), privateKey, "sshkey must not leak in error")
--- a/internal/bundlereader/gitclone_test.go
+++ b/internal/bundlereader/gitclone_test.go
@@ -257,3 +257,22 @@ func TestSetGitAuth(t *testing.T) {
 		assert.NotNil(t, opts.Auth)
 	})
 }
+
+// TestGitDownloadMalformedURLRedaction verifies that a malformed git URL
+// containing an embedded password and ?sshkey= private key does not leak those
+// secrets through the returned error (which is logged). url.Parse fails on the
+// invalid port, and its error otherwise echoes the full URL verbatim.
+func TestGitDownloadMalformedURLRedaction(t *testing.T) {
+	t.Parallel()
+
+	const (
+		password   = "s3cr3tpassword"
+		privateKey = "PRIVATE_KEY_CONTENT"
+	)
+	rawURL := "ssh://user:" + password + "@example.com:notaport/repo?sshkey=" + privateKey
+
+	err := gitDownload(context.Background(), t.TempDir(), rawURL, Auth{})
+	require.Error(t, err)
+	assert.NotContains(t, err.Error(), password, "password must not leak in error")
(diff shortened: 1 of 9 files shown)


d3flex (approved):

internal/bundlereader/gitclone_test.go
```diff
@@ -257,3 +257,22 @@ func TestSetGitAuth(t *testing.T) {
...
+func TestGitDownloadMalformedURLRedaction(t *testing.T) {
+	t.Parallel()
+
+	const (
+		password   = "s3cr3tpassword"
+		privateKey = "PRIVATE_KEY_CONTENT"
+	)
+	rawURL := "ssh://user:" + password + "@example.com:notaport/repo?sshkey=" + privateKey
+
+	err := gitDownload(context.Background(), t.TempDir(), rawURL, Auth{})
```
bigkevmcd:
```suggestion
	err := gitDownload(t.Context(), t.TempDir(), rawURL, Auth{})
```

internal/bundlereader/loaddirectory_test.go
```diff
@@ -716,3 +716,44 @@ func createDirStruct(t *testing.T, basePath string, node fsNode) string {
...
+// Intra-bundle symlinks are valid in both git repositories and tar archives
+// (e.g. values.yaml -> defaults.yaml) and must not be rejected.
+func TestGetContent_InternalSymlinkAllowed(t *testing.T) {
+	bundleDir := t.TempDir()
+	realContent := []byte("real content")
+	require.NoError(t, os.WriteFile(filepath.Join(bundleDir, "real.yaml"), realContent, 0644))
+	// Symlink within the bundle pointing to the sibling real.yaml.
+	require.NoError(t, os.Symlink(filepath.Join(bundleDir, "real.yaml"), filepath.Join(bundleDir, "link.yaml")))
+
+	files, err := bundlereader.GetContent(context.Background(), bundleDir, bundleDir, "", bundlereader.Auth{}, false, nil)
```
bigkevmcd:
```suggestion
	files, err := bundlereader.GetContent(t.Context(), bundleDir, bundleDir, "", bundlereader.Auth{}, false, nil)
```

internal/bundlereader/loaddirectory_test.go
```diff
@@ -716,3 +716,44 @@ func createDirStruct(t *testing.T, basePath string, node fsNode) string {
...
+	secretFile := filepath.Join(outsideDir, "secret.txt")
+	require.NoError(t, os.WriteFile(secretFile, []byte("sensitive content"), 0600))
+
+	// Build a bundle directory that contains one legitimate file and one
+	// symlink pointing to the secret file outside the bundle.
+	bundleDir := t.TempDir()
+	require.NoError(t, os.WriteFile(filepath.Join(bundleDir, "legit.txt"), []byte("ok"), 0644))
+	require.NoError(t, os.Symlink(secretFile, filepath.Join(bundleDir, "evil-link.txt")))
+
+	_, err := bundlereader.GetContent(context.Background(), bundleDir, bundleDir, "", bundlereader.Auth{}, false, nil)
```
bigkevmcd:
```suggestion
	_, err := bundlereader.GetContent(t.Context(), bundleDir, bundleDir, "", bundlereader.Auth{}, false, nil)
```

Issue #5216, opened 2026-06-01
Protective Sleep in webhooks handling introduces a bug


0xavi0:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

The change introduced in PR #4854 introduces a bug when there are enough GitRepos to cause the context to be canceled.

Webhook handlers are typically expected to respond within a maximum of 20 seconds. When there are enough GitRepos, this timeout can be exceeded, causing the handler's context to be canceled. As a result, some GitRepos may completely lose the webhook registration.

### Environment

```markdown
- Architecture:
- Fleet Version:
- Cluster:
  - Provider:
  - Options:
  - Kubernetes Version:
```

### Logs

```markdown

```

mmartin24:
Closing. Same reason as https://github.com/rancher/fleet/issues/5217#issuecomment-4671056433

Issue #5077, opened 2026-05-05
Webhook not working properly in Fleet 0.16


mmartin24:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

QA CI has been failing several days on test [fleet-152](https://github.com/rancher/fleet-e2e/actions/runs/25348912048/job/74324125024#step:10:357) related to webhook creation.

---
Manual check confirms the gitrepo gets the Webhook Commit but it is not updating it: Attaching screenshot: 

<img width="2517" height="1189" alt="Image" src="https://github.com/user-attachments/assets/2d5d795e-9c57-43ad-a077-9c2bb48e2a46" />

### Expected Behavior

Commit hash should be updated on Gitrepo

### Environment

```markdown
- Architecture: amd64
- Fleet Version: 0.16.0-alpha.5
- Cluster:
  - Provider: k3s
  - Options: single cluster
  - Kubernetes Version: 1.35.2+k3s1
```

### Logs

```markdown

```

0xavi0:
I could not recreate so far running this script [webhook-test.sh](https://github.com/user-attachments/files/27448198/webhook-test.sh) with `main`.

The script pushes a new commit and waits for the `GitRepo` to the commit in the `.status.webhookCommit` and `.status.commit`.

It simply deploys the fleet-examples simple folder. 

I'll try also running with the UI

mmartin24:
@thardeck , this is still ocurring on `0.16.0-alpha.5`

<img width="1961" height="1247" alt="Image" src="https://github.com/user-attachments/assets/636fa55a-adc2-4ccc-8170-ae9bf8d91787" />

mmartin24:
> [@thardeck](https://github.com/thardeck) , this is still ocurring on `0.16.0-alpha.5`
> 
[...]

PR #5218, opened 2026-06-01
webhooks handling update

0xavi0:
This PR changes the way webhooks are processed by adding an extra layer to handle the following scenarios:

* A webhook arrives before the repository HEAD has been updated.
* Multiple webhooks are received in a different order than the corresponding commits in the repository.

Webhooks now act solely as triggers for the reconciler. Upon receiving a webhook, the reconciler updates the `status.webhookCommit` field as well as a new field that stores the timestamp of the most recently received webhook.

This timestamp is used to implement an active wait mechanism whenever a mismatch is detected between the commit received in the latest webhook and the repository HEAD.

The repository HEAD is always treated as the single source of truth. If, after a reasonable waiting period, the HEAD still does not match the commit reported by the webhook, the user is notified about the potential loss of a commit reference. This situation should never occur unless there is a serious issue with the Git repository. 

Refers to: https://github.com/rancher/fleet/issues/5216 
Refers to: https://github.com/rancher/fleet/issues/5077

### Checklist

### DIFF
--- a/integrationtests/gitjob/controller/controller_test.go
+++ b/integrationtests/gitjob/controller/controller_test.go
@@ -960,6 +960,9 @@ var _ = Describe("GitJob controller", func() {
 			}).Should(BeTrue())
 
 			By("Setting WebhookCommit to simulate a webhook event")
+			// Make the mock fetcher return the webhook commit so LatestCommit
+			// resolves HEAD to the new tip rather than stableCommit.
+			expectedCommit = webhookCommit
 			Expect(setGitRepoWebhookCommit(gitRepo, webhookCommit)).To(Succeed())
 		})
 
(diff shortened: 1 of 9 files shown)


pkg/webhook/webhook.go
```diff
@@ -166,6 +166,7 @@ func (w *Webhook) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
 				}
 				orig := gitRepoFromCluster.DeepCopy()
 				gitRepoFromCluster.Status.WebhookCommit = revision
+				gitRepoFromCluster.Status.LastWebhookTime = metav1.Now()
 				if err := w.client.Status().Patch(ctx, &gitRepoFromCluster, client.MergeFrom(orig)); err != nil {
```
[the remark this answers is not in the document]
0xavi0: We're only interested in commit changes. If the patch is not updating the commit, it is better to avoid it.

internal/cmd/controller/gitops/reconciler/gitjob.go
```diff
@@ -179,7 +179,22 @@ func (r *GitJobReconciler) createJob(ctx context.Context, gitRepo *v1alpha1.GitR
...
+	// error. The owning GitRepo and commit are encoded in the name, so we are not
+	// adopting a foreign job.
+	if err := r.Create(ctx, job); err != nil {
+		if apierrors.IsAlreadyExists(err) {
+			log.FromContext(ctx).V(1).Info(
+				"Git job already exists for this commit, skipping create",
+				"job", job.Name,
+			)
+			return nil
+		}
```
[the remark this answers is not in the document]
0xavi0: I've changed the `createJob` function to return a boolean and the caller checks if the job was indeed created

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -70,6 +70,11 @@ const (
 	// in order to wait for the dependent resources cleanup to finish
 	requeueAfterResourceCleanup = 2 * time.Second
 
+	// max time to wait for a git host to propagate a pushed commit before giving up
+	webhookMaxWait = 30 * time.Second
```
weyfonk, line 74: How have these waiting times been determined?

integrationtests/gitjob/controller/controller_test.go
```diff
@@ -975,6 +978,100 @@ var _ = Describe("GitJob controller", func() {
...
+				Spec: v1alpha1.GitRepoSpec{
+					Repo:            webhookTestRepo,
+					PollingInterval: &metav1.Duration{Duration: 24 * time.Hour},
+				},
+			}
+			Expect(k8sClient.Create(ctx, &gitRepo)).To(Succeed())
+
+			By("Waiting for the initial job")
+			Eventually(func() error {
+				jobName := names.SafeConcatName(gitRepoName, names.Hex(webhookTestRepo+stableCommit, 5))
```
weyfonk, line 1013: nit: the job name could be computed just once, outside of `Eventually` blocks (as done below for the new job name, btw), instead of multiple times.

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
```diff
@@ -324,6 +366,76 @@ func monitorLatestCommit(obj metav1.Object, fetch func() (string, error)) (strin
...
+	// create any job, so we never act on the stale commit and so a webhook
+	// arriving mid-job can advance Status.Commit and supersede the running job.
+	webhookPending := gitrepo.Status.WebhookCommit != "" && gitrepo.Status.WebhookCommit != oldCommit
+	if webhookPending {
+		fetchErr := r.fetchLatestCommit(ctx, gitrepo, oldCommit)
+
+		// Propagation wait: HEAD hasn't moved yet after a recent webhook.
+		// The git host may still be replicating the pushed commit. Requeue
+		// with a short delay instead of deploying a stale HEAD. Give up after
+		// webhookMaxWait so a miss-delivered webhook never blocks indefinitely.
```
weyfonk, line 397: My understanding of this is that if webhook calls arrive out of order (e.g. with commits A→B→C and webhook calls B→A→C), then:
1. Webhook server receives commit B, sets it in the GitRepo status
2. This controller does not create a job until B has reached the remote
3. Commits B and C are then processed in order.

The assumption is that however many webhook calls are sent out-of-order, the latest commit for which a webhook call has been received will _eventually_ match the latest commit in the branch.

Please correct me if wrong :)

weyfonk (commented):
LGTM, thanks for the helpful comments 🙇

weyfonk (approved):
LGTM, thanks for the helpful comments 🙇

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
0xavi0: There's not rule here. We considered before that 1s was enough. 
If a git server sent a webhook and didn't update the HEAD accordingly in 30 seconds I thought it was more than enough to consider something wrong had happened

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
0xavi0: yup, that's correct :)

internal/cmd/controller/gitops/reconciler/gitjob_controller.go
weyfonk: Ack, good to know, and I agree with your assessment.

Issue #5059, opened 2026-04-29
[Bug] fleet-agent and controller fail to sync BundleDeployment cache due to "event bookmark expired" / WatchListClient (WaitApplied(1))


francesco-z:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

After updating Rancher to 2.14.0 and upgrading the underlying clusters (upstream and downstreams) from RKE2 1.34.3 to 1.35.3, the `fleet-agent` and `fleet-controller` frequently fail to sync their caches for `BundleDeployment`. 

The fleet bundles and `BundleDeployment` resources get stuck in a `WaitApplied(1)` state. The logs show that the controller fails to receive the required bookmark event marking the end of the initial list stream (due to the new `WatchListClient` streaming behavior in `client-go` v0.35). The cache sync eventually times out, causing the manager to completely shut down.

### Expected Behavior

Fleet successfully syncs its informer caches using the Kubernetes API, continuously watches the `BundleDeployment` resources, and transitions bundles from `WaitApplied` to `Active` without timing out or crashing.

### Steps To Reproduce

1. Update Rancher server to `v2.14.0`.
2. Upgrade the upstream and downstream clusters to RKE2 `v1.35.3+rke2r3`.
3. Have active Fleet bundles deployed to the downstream or local clusters.
4. Check the state of the deployed bundles in the Rancher UI or via `kubectl get bundles -A`.
5. Observe the `WaitApplied(1)` state and check the `fleet-agent` (downstream) or `fleet-controller` (upstream) logs for bookmark expiration and cache sync timeout errors.

### Environment

```markdown
- Architecture: [e.g., amd64/x86_64]
- Fleet Version: Bundled with Rancher 2.14.0
- Cluster:
  - Provider: RKE2 (Local and Downstream)
  - Options: Helm Chart RKE2 install, CNI: rke2-cilium-1.19.101
  - Kubernetes Version: v1.35.3+rke2r3
```

### Logs

```markdown
{"level":"info","ts":"2026-04-28T16:53:50Z","logger":"controller-runtime.cache","msg":"Warning: event bookmark expired","err":"pkg/mod/k8s.io/client-go@v0.35.2/tools/cache/reflector.go:289: hasn't received required bookmark event marking the en...
[...]

PR #5226, opened 2026-06-03
Disable WatchListClient in Fleet pods

thardeck:
Set `KUBE_FEATURE_WatchListClient=false` on Fleet controller and agent pods rendered by the `fleet` and `fleet-agent` charts so informer cache startup does not depend on WatchList bookmark behavior.

This returns to the old behavior after k8s client-go 0.35 changed the [default to true](https://github.com/kubernetes/client-go/blob/v0.35.1/features/known_features.go#L101-L104).

Refers #5059

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,7 @@ jobs:
         run: |
           helm plugin install https://github.com/helm-unittest/helm-unittest.git --version v1.0.2
           helm unittest ./charts/fleet
+          helm unittest ./charts/fleet-agent
       -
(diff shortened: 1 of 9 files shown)


0xavi0 (commented):
I think we also need to pass the env variable when using manager initiated agent registration as the Deployment is generated at `internal/cmd/controller/agentmanagement/agent/manifest.go`

0xavi0 (commented):
The `gitjob` and `helmops` pods would need also the env variable, right? 
Sorry, I didn't catch that in my previous review

0xavi0 (approved):

thardeck:
Not sure if changing the new default back to the old is feasible in the midterm. Closing for now.

PR #5240, opened 2026-06-05
Sensible agent check in defaults

weyfonk:
This ensures that an agent check-in interval cannot be set to a zero duration.

### Checklist

### DIFF
--- a/internal/cmd/controller/agentmanagement/controllers/cluster/import.go
+++ b/internal/cmd/controller/agentmanagement/controllers/cluster/import.go
@@ -361,6 +361,10 @@ func (i *importHandler) importCluster(cluster *fleet.Cluster, status fleet.Clust
 		apiServerCA  = secret.Data[config.APIServerCAKey]
 	)
 
+	if cfg.AgentCheckinInterval.Seconds() == 0 {
+		return status, errors.New("agent check-in interval cannot be 0")
+	}
+
 	if apiServerURL == "" {
 		if len(cfg.APIServerURL) == 0 {
 			// Current config cannot be deployed, so remove the "config changed" mark
--- a/internal/cmd/controller/agentmanagement/controllers/cluster/import.go
+++ b/internal/cmd/controller/agentmanagement/controllers/cluster/import.go
@@ -495,6 +495,7 @@ func (i *importHandler) importCluster(cluster *fleet.Cluster, status fleet.Clust
 				AgentResources:       cluster.Spec.AgentResources,
 				HostNetwork:          *cmp.Or(cluster.Spec.HostNetwork, new(false)),
 				AgentReplicas:        agentReplicas,
+				CheckinInterval:      cfg.AgentCheckinInterval.Duration.String(),
 				PriorityClassName:    priorityClassName,
 				ImagePullSecrets:     pullSecrets,
 				PropagatePullSecrets: propagate,
--- a/internal/cmd/controller/agentmanagement/controllers/cluster/import.go
+++ b/internal/cmd/controller/agentmanagement/controllers/cluster/import.go
@@ -361,8 +361,8 @@ func (i *importHandler) importCluster(cluster *fleet.Cluster, status fleet.Clust
 		apiServerCA  = secret.Data[config.APIServerCAKey]
 	)
 
-	if cfg.AgentCheckinInterval.Seconds() == 0 {
-		return status, errors.New("agent check-in interval cannot be 0")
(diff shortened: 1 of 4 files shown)


kkaempf:
@weyfonk tentatively setting milestone to 2.16. 🤔

p-se:
Should we preserve defaulting to 15m if the check-in interval is zero? 

https://github.com/rancher/fleet/blob/e53ea86a5de594842c8e5d98286f5268a43f3c52/internal/cmd/agent/clusterstatus/ticker.go#L46-L48

weyfonk:
> Should we preserve defaulting to 15m if the check-in interval is zero?
> 
> https://github.com/rancher/fleet/blob/e53ea86a5de594842c8e5d98286f5268a43f3c52/internal/cmd/agent/clusterstatus/ticker.go#L46-L48

The agent check-in interval used by the cluster status ticker comes from the `CHECKIN_INTERVAL` environment variable, fed from chart configuration.

I would be tempted to say that setting a check-in interval to 0 should not be silently ignored, but checks are actually run against the config _before_ the agent manifests are created, so unless someone manually creates an agent bundle, it seems unlikely to me that such a value would make it into the agent.

Therefore, I would keep the current check as it is, as a last resort option.
How much sense would that make to you?

internal/cmd/controller/agentmanagement/controllers/cluster/import.go
```diff
@@ -491,6 +495,7 @@ func (i *importHandler) importCluster(cluster *fleet.Cluster, status fleet.Clust
 				AgentResources:       cluster.Spec.AgentResources,
 				HostNetwork:          *cmp.Or(cluster.Spec.HostNetwork, new(false)),
 				AgentReplicas:        agentReplicas,
+				CheckinInterval:      cfg.AgentCheckinInterval.Duration.String(),
```
d3flex, line 496: I wonder if this the right level to set the defaults. it looks to me that internal/config/config.go and particular in the `DefaultConfig` serves this better, as cfg will always get initialized with a default value. thinking of it, the assignment comes after the check (l364) inside a function which reconciles. doesnt this create a loop with no exit?

internal/cmd/controller/agentmanagement/controllers/manageagent/manageagent_test.go
d3flex, line 1: This test (e2e+unit) pass even if I remove https://github.com/rancher/fleet/pull/5240/changes#diff-28a4af52a53de89516ada67a34c6382270d9ee7c39e6fa4164c64ecefd7aeeadR498. do the tests really exercise the change in this case?

internal/cmd/controller/agentmanagement/controllers/cluster/import.go
weyfonk: This is a user-configured value, which may differ from the default.

The check on L364 only returns if the configured agent check-in interval is 0 or less, which are unsupported values.
With correct (i.e. strictly positive) values, the check passes and execution proceeds, enabling manifest options to be created with that configuration.

internal/cmd/controller/agentmanagement/controllers/manageagent/manageagent_test.go
weyfonk: Those are 2 different code paths: the one you've changed refers to cluster import, whereas this test case covers `newAgentBundle`, which has its own logic for creating an agent manifest.

0xavi0 (commented):
I think we have an inconsistence with [the default](https://github.com/rancher/fleet/blob/4a608c1525314a2f25381807c4411e56e67c8fd5/charts/fleet/values.schema.json#L6:L17) value accepted in the charts.

Why 0 is not accepted? 
`time.NewTicker` panics on non-positive but not on 0.

If 0 is not accepted for whatever reason we can always set a new default different to 0 for this value.

weyfonk:
> I think we have an inconsistence with [the default](https://github.com/rancher/fleet/blob/4a608c1525314a2f25381807c4411e56e67c8fd5/charts/fleet/values.schema.json#L6:L17) value accepted in the charts.
> 
> Why 0 is not accepted? `time.NewTicker` panics on non-positive but not on 0.
> 
> If 0 is not accepted for whatever reason we can always set a new default different to 0 for this value.

Indeed, upon further consideration this PR does not bring much value, given that the default value is applied [anyway](https://github.com/rancher/fleet/blob/main/internal/cmd/agent/clusterstatus/ticker.go#L50) on the agent's side if the check-in interval is 0. Closing.

Issue #4807, opened 2026-03-12
[SURE-11011] Allow a way to disable Rancher fleet-agent on management cluster


kkaempf:
## SURE-11011

see JIRA for issue details.
Discussed in today's daily, tentatively scheduling for 2.15.0

0xavi0:
# QA Template

## 1. Test that the new option in installation time disables the agent and Fleet does not create related resources (RBAC,etc) in the agent namespace.

There is a new option in the helm values:
```yaml
  # If true, the local cluster stays registered (Cluster + ClusterGroup) but the
  # fleet-agent is NOT deployed into it. An already-running local agent will be removed.
  # The local cluster will not receive deployments.
  localAgentDisabled: false
```
Try passing `--set bootstrap.localAgentDisabled=true` when running the `helm install` command to install Fleet.

It should not run the agent in the local cluster nor create related resources.

## 2. Disable the agent after it has been deployed.
Deploy Fleet normally (check that the upstream local agent is running and everything is as usual)
Disable the agent running:
```bash
helm -n cattle-fleet-system upgrade fleet fleet \
    --reuse-values \
    --set bootstrap.localAgentDisabled=true
```
This should remove the local agent and also related resources. (including the agent's namespace)

## 3. Enable the agent again.
Now run:
```bash
helm -n cattle-fleet-system upgrade fleet fleet \
    --reuse-values \
    --set bootstrap.localAgentDisabled=false
```
And check that the agent is back again and its related resources (if the agent works fine its resources should also be there)

sbulage:
# Verification Result: 
  - Status: ✅ VERIFIED  

## System Information

| Rancher Version | Fleet Version |
|-----------------|---------------|
| v2.15-e564a92c78efaa2051f98cb9f06afa05d655c929-head | 110.0.0+up0.16.0-rc.4 |

---

## Test Results

| Test | Configuration | Result |
|------|--------------|--------|
| ✅ Test 1 | `localAgentDisabled: true` (install time) | Agent not deployed, no resources created |
| ✅ Test 2 | `localAgentDisabled: true` (post-install) | Agent removed, resources cleaned up |
| ✅ Test 3 | `localAgentDisabled: false` (re-enable) | Agent redeployed, resources restored |

---

## Summary

The `localAgentDisabled` option works correctly:

**When `localAgentDisabled: true`:**
- ✅ fleet-agent deployment not created on local cluster
- ✅ Agent namespace and RBAC resources not created
[...]

PR #5250, opened 2026-06-08
Add option to disable management cluster's agent

0xavi0:
This PR adds an option in the helm chart to disable/enable the local agent. If the agent was already deployed it tears down the agent.

Refers to: https://github.com/rancher/fleet/issues/4807

- [x]  I have updated the documentation via a pull request in the [fleet-product-docs](https://github.com/rancher/fleet-product-docs) repository.

### DIFF
--- a/charts/fleet/tests/agent_disabled_test.yaml
+++ b/charts/fleet/tests/agent_disabled_test.yaml
@@ -1,0 +1,71 @@
+suite: bootstrap.agentDisabled is rendered into the fleet-controller configmap
+# These tests guard the chart side of the "disable the local fleet-agent" feature.
+templates:
+  - configmap.yaml
+tests:
+  - it: defaults agentDisabled to false
+    asserts:
+      - hasDocuments:
+          count: 1
+      - isKind:
+          of: ConfigMap
+      - matchRegex:
+          path: data.config
+          pattern: '"agentDisabled":\s*false'
+
+  - it: emits agentDisabled true when set
+    set:
+      bootstrap.agentDisabled: true
+    asserts:
+      - matchRegex:
+          path: data.config
+          pattern: '"agentDisabled":\s*true'
+      # Must be the JSON boolean true, NOT the string "true": the Go
+      # Bootstrap.AgentDisabled field is bool and json.Unmarshal would reject
+      # the quoted form.
+      - notMatchRegex:
+          path: data.config
+          pattern: '"agentDisabled":\s*"true"'
+
+  - it: keeps the other bootstrap keys intact alongside agentDisabled
+    # Regression guard for the trailing-comma change on the agentNamespace line:
+    # if a future edit drops a comma or removes a field, this test fails before
+    # the broken configmap reaches a cluster.
+    set:
+      bootstrap.agentDisabled: true
+      bootstrap.namespace: fleet-local
+      bootstrap.agentNamespace: cattle-fleet-local-system
+      bootstrap.repo: https://example.com/repo
+      bootstrap.branch: main
+      bootstrap.paths: deploy
+      bootstrap.secret: my-secret
+    asserts:
+      - matchRegex:
+          path: data.config
+          pattern: '"namespace":\s*"fleet-local"'
+      - matchRegex:
+          path: data.config
+          pattern: '"agentNamespace":\s*"cattle-fleet-local-system"'
+      - matchRegex:
+          path: data.config
+          pattern: '"repo":\s*"https://example.com/repo"'
+      - matchRegex:
+          path: data.config
+          pattern: '"branch":\s*"main"'
+      - matchRegex:
+          path: data.config
+          pattern: '"paths":\s*"deploy"'
+      - matchRegex:
+          path: data.config
+          pattern: '"secret":\s*"my-secret"'
+      - matchRegex:
+          path: data.config
+          pattern: '"agentDisabled":\s*true'
+
+  - it: still emits agentDisabled when explicitly set to false
+    set:
+      bootstrap.agentDisabled: false
+    asserts:
+      - matchRegex:
+          path: data.config
+          pattern: '"agentDisabled":\s*false'
--- a/charts/fleet/tests/agent_disabled_test.yaml
+++ b/charts/fleet/tests/agent_disabled_test.yaml
@@ -1,9 +1,9 @@
-suite: bootstrap.agentDisabled is rendered into the fleet-controller configmap
+suite: bootstrap.localAgentDisabled is rendered into the fleet-controller configmap
 # These tests guard the chart side of the "disable the local fleet-agent" feature.
 templates:
   - configmap.yaml
 tests:
-  - it: defaults agentDisabled to false
+  - it: defaults localAgentDisabled to false
     asserts:
       - hasDocuments:
           count: 1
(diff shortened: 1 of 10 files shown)


internal/cmd/controller/agentmanagement/controllers/cluster/import.go
```diff
@@ -525,6 +546,144 @@ func (i *importHandler) importCluster(cluster *fleet.Cluster, status fleet.Clust
...
+	if err := ignoreNotFound(kc.AppsV1().StatefulSets(agentNamespace).Delete(i.ctx, config.AgentConfigName, metav1.DeleteOptions{})); err != nil {
+		return err
+	}
+	if err := ignoreNotFound(kc.PolicyV1().PodDisruptionBudgets(agentNamespace).Delete(i.ctx, scheduling.FleetAgentPodDisruptionBudgetName, metav1.DeleteOptions{})); err != nil {
+		return err
+	}
+
+	// Also clean up the default SA on older installs that left it with a modified
+	// AutomountServiceAccountToken. Candidate namespaces: agent's namespace plus
+	// the legacy default if the system namespace was customized.
```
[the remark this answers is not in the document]
0xavi0: change as suggested

internal/cmd/controller/agentmanagement/controllers/cluster/import.go
```diff
@@ -525,6 +546,144 @@ func (i *importHandler) importCluster(cluster *fleet.Cluster, status fleet.Clust
...
+		if err := kc.CoreV1().Namespaces().Delete(
+			i.ctx,
+			fleetns.SystemRegistrationNamespace(config.DefaultNamespace),
+			metav1.DeleteOptions{},
+		); err != nil && !apierrors.IsNotFound(err) {
+			return err
+		}
+	}
+
+	logrus.Infof("Cluster import for '%s/%s'. Removed local agent (agentDisabled=true)", cluster.Namespace, cluster.Name)
```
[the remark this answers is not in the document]
0xavi0: changed as suggested

charts/fleet/values.yaml
```diff
@@ -60,6 +60,10 @@ bootstrap:
   # The namespace where the fleet agent for the local cluster will be ran, if empty
   # this will default to cattle-fleet-system
   agentNamespace: ""
+  # If true, the local cluster stays registered (Cluster + ClusterGroup) but the
+  # fleet-agent is NOT deployed into it. An already-running local agent will be removed.
+  # The local cluster will not receive deployments.
+  agentDisabled: false
```
weyfonk: nit: I wonder if the name of the value may be slightly misleading, as if the agent were to be disabled on all clusters... On the other hand, that would not make much sense, defeating the purpose of Fleet itself.

internal/cmd/controller/agentmanagement/controllers/cluster/import.go
```diff
@@ -300,6 +304,23 @@ func (i *importHandler) importCluster(cluster *fleet.Cluster, status fleet.Clust
 		return status, nil
 	}
 
+	if localAgentDisabled(cluster) {
```
weyfonk, line 312: Is there a risk that someone might label a non-local cluster with `fleet.cattle.io/local-agent-disabled`, which would then also prevent the agent from being deployed, and actually tear it down, in that cluster?

In other words, should we check that the cluster we are dealing with here is actually the local cluster?

internal/cmd/controller/agentmanagement/controllers/cluster/import.go
```diff
@@ -525,6 +546,144 @@ func (i *importHandler) importCluster(cluster *fleet.Cluster, status fleet.Clust
...
+	if agentNamespace == "" {
+		agentNamespace = i.systemNamespace
+	}
+
+	if err := i.deleteAgentResources(kc, agentNamespace); err != nil {
+		return err
+	}
+
+	// Mirror the deploy path: when the system namespace was customized, drop the
+	// now-orphaned legacy system-registration namespace too.
```
weyfonk: I was confused for a bit, thinking that this referred to `fleetns.SystemRegistrationNamespace(config.LegacyDefaultNamespace)`. Wondering if `legacy` makes sense here.

internal/cmd/controller/agentmanagement/controllers/cluster/import.go
```diff
@@ -525,6 +546,144 @@ func (i *importHandler) importCluster(cluster *fleet.Cluster, status fleet.Clust
...
+	crbName := names.SafeConcatName(agentNamespace, agent.DefaultName, "role", "binding")
+	if err := ignoreNotFound(kc.RbacV1().ClusterRoles().Delete(i.ctx, crName, metav1.DeleteOptions{})); err != nil {
+		return err
+	}
+	if err := ignoreNotFound(kc.RbacV1().ClusterRoleBindings().Delete(i.ctx, crbName, metav1.DeleteOptions{})); err != nil {
+		return err
+	}
+
+	// Cluster-scoped scheduling resources.
+	if err := ignoreNotFound(kc.SchedulingV1().PriorityClasses().Delete(i.ctx, scheduling.FleetAgentPriorityClassName, metav1.DeleteOptions{})); err != nil {
```
weyfonk, line 631: It seems to me that `deleteOldAgent` could reuse at least some of this logic :)

internal/cmd/controller/agentmanagement/controllers/cluster/import_test.go
```diff
@@ -508,3 +509,177 @@ func TestAllowedKubeConfigSecretNamespace(t *testing.T) {
...
+		{
+			name:    "nil labels",
+			cluster: &fleet.Cluster{},
+			want:    false,
+		},
+		{
+			name: "label absent",
+			cluster: &fleet.Cluster{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{"name": "local"},
```
weyfonk: Would it make sense to test this on a non-local cluster?

charts/fleet/values.yaml
0xavi0: the value was under `bootstrap` that only applies to the management cluster, but I've changed the value to `localAgentDisabled` to make it more clear.

internal/cmd/controller/agentmanagement/controllers/cluster/import.go
0xavi0: You're right, thanks for the suggestion!
I've added 2 checks. One that only uses the cluster name and namespace to identify if the cluster can be considered as local and a second one that makes sure 100% the cluster is the local one before the agent is removed. It checks if the UID of the `kube-system` namespace matches using the client connection. 
This way only an agent located in the management cluster would be deleted.

internal/cmd/controller/agentmanagement/controllers/cluster/import.go
0xavi0: Yup. I've created a function and that comment was removed

internal/cmd/controller/agentmanagement/controllers/cluster/import.go
0xavi0: yeah, in fact, I've changed it a bit and now references to `deleteOldAgent` are all gone. Thanks! :)

internal/cmd/controller/agentmanagement/controllers/cluster/import.go
weyfonk: Right, so deciding whether to deploy the agent relies on the simpler, faster check, because not deploying an agent is not destructive, whereas tearing down an existing agent is, hence why that latter operation relies on a more thorough check. Makes sense to me, thanks for the update 🙇 !

internal/cmd/controller/agentmanagement/controllers/cluster/import_test.go
```diff
@@ -508,3 +662,336 @@ func TestAllowedKubeConfigSecretNamespace(t *testing.T) {
...
+
+	cases := []struct {
+		name string
+		// localUID is the UID reported by the controller's own kube-system
+		// namespace; localErr, if set, is returned instead.
+		localUID  string
+		localErr  error
+		remoteUID string
+		want      bool
+		wantErr   bool
```
weyfonk, line 812: nit: this field seems never to be set 🙃 
Same remark about `localErr`

weyfonk (approved):
Thanks for adding checks ensuring that this can only be applied to the local cluster 🚀

internal/cmd/controller/agentmanagement/controllers/cluster/import_test.go
0xavi0: changed to use those fields

charts/fleet/tests/agent_disabled_test.yaml
```diff
@@ -0,0 +1,71 @@
...
+    asserts:
+      - hasDocuments:
+          count: 1
+      - isKind:
+          of: ConfigMap
+      - matchRegex:
+          path: data.config
+          pattern: '"agentDisabled":\s*false'
+
+  - it: emits agentDisabled true when set
```
d3flex, line 16: I just wonder whether the other way around should be also covered. wdyt?

d3flex: sorry. alrady merged.

Issue #5254, opened 2026-06-09
Enable agent image pull secrets propagation to be disabled


weyfonk:
#5050 allowed image pull secrets to be referenced for both Fleet _controller_ and _agent_ deployments.
Fleet propagates referenced image pull secrets to all downstream clusters.

However, in some cases involving Fleet in Rancher, distinct sets of image pull secrets may need to be referenced for different downstream clusters, which the current Fleet implementation does not cover.

Suggested solution: Fleet `Cluster` objects could support a new `agentPullSecrets` field, which would override Fleet's `global.cattle.imagePullSecrets` Helm value when both are non-nil.
Rancher controllers already propagate image pull secrets through https://github.com/rancher/rancher/issues/54180 (see https://github.com/rancher/rancher/pull/55071), meaning that in this case, Fleet's propagation of such secrets needs to be disabled.

weyfonk:
## Additional QA
 
### Problem
Fleet only supported image pull secrets, needed by agent deployments, to be specified globally, through Helm value `global.cattle.imagePullSecrets`. This meant that all agents, regardless of the downstream cluster where they live, would receive the same set of image pull secrets. However, different image registries may be available depending on downstream clusters, which justifies configuring image pull secrets by cluster in some cases.
 
### Solution
Fleet `Cluster` resources now support a new `agentPullSecrets` field which, when specified, overrides the `global.cattle.imagePullSecrets` value.
When a cluster's `agentPullSecrets` is specified, Fleet's agent management controller does _not_ propagate those secrets to the corresponding downstream cluster, because Rancher is then expected to handle that propagation.
 
### Testing

### Engineering Testing
#### Manual Testing
N/A (focused on automated testing in the interest of time)

#### Automated Testing
Unit tests validating that image pull secrets are added to the set of Kubernetes resources created by Fleet's agent management controller, along with agent deployments, and referenced by said deployments, depending on whether propagation of image pull secrets is enabled (which is computed separately, resolving to `true` if only global image pull secrets are configured, false otherwise):
| Global image pull secrets configured? | Cluster-level secrets configured? | Image pull secrets created downstream? | Image pull secrets referenced by Fleet agent deployment? | Observations |
|--------------------------------------------------------|-------------------------------------------------|------------------------------------------------------------------------|---------------------------------------------------------------------------------------|--------------------|
|:white_check_mark: | :x: | ✅ | ✅ | |
[...]

PR #5258, opened 2026-06-10
Support cluster-level image pull secrets for agent deployments

weyfonk:
Image pull secrets being configured in the Fleet chart no longer automatically means that Fleet's agent management controller should propagate those secrets downstream.
Instead, when a Fleet `Cluster` resource references its own agent image pull secrets, through a non-empty `agentPullSecrets` field, Rancher is expected to handle that propagation.

Refers to #5254

### Checklist


charts/fleet-crd/templates/crds.yaml
```diff
@@ -6048,6 +6048,36 @@ spec:
                     cattle-fleet-system.
                   nullable: true
                   type: string
+                agentPullSecrets:
+                  description: 'AgentPullSecrets references image pull secrets to
+                    be copied to deploy the Fleet agent on this cluster.
+
+                    If set, it overrides any global image pull secrets set as values
+                    in the Fleet chart.'
```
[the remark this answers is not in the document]
weyfonk: This code is generated; taking the other comment on the source data, i.e. `pkg/apis`, into account instead.

internal/cmd/controller/agentmanagement/agent/agent.go
```diff
@@ -90,10 +104,15 @@ func AgentWithConfig(ctx context.Context, agentNamespace, controllerNamespace, a
 		return objs, err
 	}
 
+	pullSecrets := cfg.ImagePullSecrets
+	if len(opts.ImagePullSecrets) != 0 {
+		pullSecrets = opts.ImagePullSecrets
+	}
```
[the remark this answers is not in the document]
weyfonk: Fixed, with an additional test case covering this.

internal/cmd/controller/agentmanagement/controllers/cluster/import.go
```diff
@@ -451,15 +453,16 @@ func (i *importHandler) importCluster(cluster *fleet.Cluster, status fleet.Clust
...
+				AgentTolerations:     cluster.Spec.AgentTolerations,
+				PrivateRepoURL:       cluster.Spec.PrivateRepoURL,
+				AgentAffinity:        cluster.Spec.AgentAffinity,
+				AgentResources:       cluster.Spec.AgentResources,
+				HostNetwork:          *cmp.Or(cluster.Spec.HostNetwork, new(false)),
+				AgentReplicas:        agentReplicas,
+				PriorityClassName:    priorityClassName,
+				ImagePullSecrets:     pullSecrets,
+				PropagatePullSecrets: propagate,
 			},
```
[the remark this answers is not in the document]
weyfonk: `manageagent` is meant for [agent-initiated registration](https://fleet.rancher.io/how-tos-for-operators/cluster-registration) though, which is used by Fleet standalone, not by Rancher.
When Rancher is not involved, Fleet must handle propagation of image pull secrets for downstream agents.

internal/cmd/controller/agentmanagement/agent/agent_test.go
```diff
@@ -0,0 +1,248 @@
...
+			// Mock watcher (not used in this code path since secretName is already set)
+			mockWatcher := watch.NewFake()
+			crtClient.EXPECT().Watch(gomock.Any(), gomock.Any()).Return(mockWatcher, nil).AnyTimes()
+			crtClient.EXPECT().List(gomock.Any(), gomock.Any()).Return(
+				&fleet.ClusterRegistrationTokenList{
+					Items: []fleet.ClusterRegistrationToken{*mockToken},
+				}, nil).
+				AnyTimes()
+
+			mockFleetControllers := mocks.NewMockFleetControllers(ctrl)
```
weyfonk, line 164: There's probably a better way of doing this, without needing to generate separate mocks... but I don't remember what. Input welcome, happy to reduce the diff :)

HarrisonWAffel (approved):

internal/cmd/controller/agentmanagement/agent/agent.go
```diff
@@ -57,9 +65,15 @@ func AgentWithConfig(ctx context.Context, agentNamespace, controllerNamespace, a
...
-	// Copy image pull secrets, if any, to the agent namespace.
+	// Copy image pull secrets, if any, to the agent namespace, if their propagation is configured in manifest
+	// options. This will typically not be necessary if cluster-level image pull secrets are configured, as they will be
+	// propagated by Rancher.
 	// This ensures that those secrets are available at agent deployment time.
 	for _, ips := range opts.ImagePullSecrets {
+		if !opts.PropagatePullSecrets {
+			break
+		}
+
```
0xavi0, line 76: nit: opts is not part of the item in the loop.
We could maybe do:
```go
if opts.PropagatePullSecrets {
    for _, ips := range opts.ImagePullSecrets {
        ...
    }
}
```

0xavi0 (commented):
I think we should also call the new function `getPullSecrets` when updating the agent builder in [manageagent.go](https://github.com/weyfonk/fleet/blob/4e6fdc9d87f0c3af25ed7d4494ad0930c11908a8/internal/cmd/controller/agentmanagement/controllers/manageagent/manageagent.go#L352).
We still have:
```go
AgentImagePullPolicy:    cfg.AgentImagePullPolicy,
```

internal/cmd/controller/agentmanagement/agent/agent.go
weyfonk: Yes, I wrote it this way to avoid adding one level of indentation to the `for` loop; but happy to resort to your suggestion, if you think it's more readable.

weyfonk:
> I think we should also call the new function `getPullSecrets` when updating the agent builder in [manageagent.go](https://github.com/weyfonk/fleet/blob/4e6fdc9d87f0c3af25ed7d4494ad0930c11908a8/internal/cmd/controller/agentmanagement/controllers/manageagent/manageagent.go#L352). We still have:
> 
> ```go
> AgentImagePullPolicy:    cfg.AgentImagePullPolicy,
> ```

As mentioned in [this comment](https://github.com/rancher/fleet/pull/5258#discussion_r3394367714), my understanding is that `manageagent.go` refers to agent-initiated cluster registration, hence Fleet standalone, without Rancher being involved. Therefore, propagation of image pull secrets would be unchanged, addressed by Fleet.
Happy to discuss :)

internal/cmd/controller/agentmanagement/agent/agent.go
0xavi0: Up to you, I consider it a nit... so feel free to do whatever you think is best :)

0xavi0:
> > I think we should also call the new function `getPullSecrets` when updating the agent builder in [manageagent.go](https://github.com/weyfonk/fleet/blob/4e6fdc9d87f0c3af25ed7d4494ad0930c11908a8/internal/cmd/controller/agentmanagement/controllers/manageagent/manageagent.go#L352). We still have:
> > ```go
> > AgentImagePullPolicy:    cfg.AgentImagePullPolicy,
> > ```
> 
> As mentioned in [this comment](https://github.com/rancher/fleet/pull/5258#discussion_r3394367714), my understanding is that `manageagent.go` refers to agent-initiated cluster registration, hence Fleet standalone, without Rancher being involved. Therefore, propagation of image pull secrets would be unchanged, addressed by Fleet. Happy to discuss :)

You're right, sorry for the noise ...

weyfonk:
> > > I think we should also call the new function `getPullSecrets` when updating the agent builder in [manageagent.go](https://github.com/weyfonk/fleet/blob/4e6fdc9d87f0c3af25ed7d4494ad0930c11908a8/internal/cmd/controller/agentmanagement/controllers/manageagent/manageagent.go#L352). We still have:
> > > ```go
> > > AgentImagePullPolicy:    cfg.AgentImagePullPolicy,
> > > ```
> > 
> > 
> > As mentioned in [this comment](https://github.com/rancher/fleet/pull/5258#discussion_r3394367714), my understanding is that `manageagent.go` refers to agent-initiated cluster registration, hence Fleet standalone, without Rancher being involved. Therefore, propagation of image pull secrets would be unchanged, addressed by Fleet. Happy to discuss :)
> 
> You're right, sorry for the noise ...

It turns out I was not, thanks for pointing this out! Moving this to draft mode in the meantime, working on adding the logic for disabling propagation to that part as well.

internal/cmd/controller/agentmanagement/controllers/manageagent/manageagent.go
```diff
@@ -334,6 +339,30 @@ func (h *handler) newAgentBundle(ns string, cluster *fleet.Cluster) (runtime.Obj
...
 	}
 
+	var bundleObjs []runtime.Object
+	agentPullSecrets, propagate := secretutils.GetAgentPullSecrets(cfg, cluster)
+	if propagate {
+		for _, ips := range agentPullSecrets {
+			source, err := h.secretCache.Get(ns, ips.Name)
+			if err != nil {
+				return bundleObjs, fmt.Errorf("failed to get image pull secret %q from controller namespace: %w", ips.Name, err)
+			}
```
0xavi0: The log references the `controller namespace` but `Get` uses the namespace we get originally from `OnNamespace`. 
Shouldn't it be:
```go
source, err := h.secretCache.Get((h.systemNamespace, ips.Name)
``` 
?

internal/cmd/controller/agentmanagement/controllers/manageagent/manageagent.go
0xavi0, line 420: I realised this line is here commented.
Are the secrets propagated to the downstream clusters?

internal/cmd/controller/agentmanagement/controllers/manageagent/manageagent.go
weyfonk: Good catch, the controller namespace is actually `h.systemNamespace`. Fixed via a new commit.

0xavi0 (approved):

PR #5267, opened 2026-06-11
fix(benchmarks): ensure all required families are present before extraction

d3flex:
The `getMetrics` already uses a guard before move on to extractFromMetricFamilies checking for a single family cluster `controller_runtime_reconcile_total`. However this can't vouch for the existence of the other metric families in `mfs`. To make sure that the `extractFromMetricFamilies` does not panic due to the empty maps, extend the guard with checks to all the metricFamilies. Expect only to fail with timeout in case that the map is not ready eventually.

This likely does not have any other significant overhead for the benchmark results.

Refers to https://github.com/rancher/fleet/actions/workflows/benchmark.yml

### Checklist

- [x] benchmark passes locally

### DIFF
--- a/benchmarks/record/record.go
+++ b/benchmarks/record/record.go
@@ -233,6 +233,24 @@ func Metrics(experiment *gm.Experiment, suffix string) {
 	}
 }
 
+// requiredMetricFamilies lists every metric family that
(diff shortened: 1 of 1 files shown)


p-se (approved):
lgtm

PR #5285, opened 2026-06-15
Validate GitRepo defaults from GitRepoRestriction and Policy against allow-lists

p-se:
When a `GitRepoRestriction` (or `Policy`) defines a `defaultServiceAccount` that is not contained in a non-empty `allowedServiceAccounts` (and likewise `defaultClientSecretName` vs `allowedClientSecretNames`), the GitRepo reconciler applies the default to `gitrepo.spec` and persists it before that value is validated against the allow-list. The disallowed value is validated only on the following reconcile, once it is an explicit spec field, and the GitRepo is rejected then.

This is inconsistent with the HelmOp and Bundle paths, which validate defaults before applying them.

## Steps to reproduce

```yaml
# GitRepoRestriction in fleet-local
kind: GitRepoRestriction
metadata: { name: repro, namespace: fleet-local }
defaultServiceAccount: some-default-sa
allowedServiceAccounts: [some-other-sa]   # non-empty, does NOT contain the default
```

```yaml
# GitRepo in fleet-local with no serviceAccount set
kind: GitRepo
metadata: { name: repro, namespace: fleet-local }
spec:
  repo: https://github.com/rancher/fleet-examples
```

Apply the restriction, then the GitRepo, and inspect it.

## Expected

The GitRepo is rejected without `spec.serviceAccount` being modified; `metadata.generation` stays `1`.

## Actual

`spec.serviceAccount` is set to `some-default-sa`, `metadata.generation` becomes `2`, and only then is the GitRepo rejected (`Accepted=False`, `PolicyViolation` event).

Refers to #5422

### Checklist

### DIFF
--- a/internal/cmd/controller/gitops/reconciler/restrictions.go
+++ b/internal/cmd/controller/gitops/reconciler/restrictions.go
@@ -60,24 +60,26 @@ func AuthorizeAndAssignDefaults(ctx context.Context, c client.Client, gitrepo *f
 		return errors.New("empty targetNamespace denied, because allowedTargetNamespaceSelector restriction is present")
 	}
 
-	targetNamespace, err := isAllowed(gitrepo.Spec.TargetNamespace, "", allowedTargetNS)
-	if err != nil {
+	// Apply defaults before validation, so a default that is not in the allow-list
+	// is rejected rather than silently accepted. This keeps the GitRepo path
+	// consistent with the HelmOp and Bundle paths.
+	serviceAccount := firstNonEmpty(gitrepo.Spec.ServiceAccount, defaultSA)
+	clientSecretName := firstNonEmpty(gitrepo.Spec.ClientSecretName, defaultClientSecret)
+
+	if _, err := isAllowed(gitrepo.Spec.TargetNamespace, "", allowedTargetNS); err != nil {
 		return fmt.Errorf("disallowed targetNamespace %s: %w", gitrepo.Spec.TargetNamespace, err)
 	}
 
-	serviceAccount, err := isAllowed(gitrepo.Spec.ServiceAccount, defaultSA, allowedSAs)
-	if err != nil {
-		return fmt.Errorf("disallowed serviceAccount %s: %w", gitrepo.Spec.ServiceAccount, err)
+	if _, err := isAllowed(serviceAccount, "", allowedSAs); err != nil {
+		return fmt.Errorf("disallowed serviceAccount %s: %w", serviceAccount, err)
 	}
 
-	repo, err := isAllowedByRepo(gitrepo.Spec.Repo, grr.AllowedRepoPatterns, pol.GitAllowedRepoPatterns)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/cmd/controller/gitops/reconciler/restrictions.go
```diff
@@ -60,24 +60,26 @@ func AuthorizeAndAssignDefaults(ctx context.Context, c client.Client, gitrepo *f
...
-	serviceAccount, err := isAllowed(gitrepo.Spec.ServiceAccount, defaultSA, allowedSAs)
-	if err != nil {
-		return fmt.Errorf("disallowed serviceAccount %s: %w", gitrepo.Spec.ServiceAccount, err)
+	if err := isAllowed(serviceAccount, allowedSAs); err != nil {
+		return fmt.Errorf("disallowed serviceAccount %s: %w", serviceAccount, err)
 	}
 
-	repo, err := isAllowedByRepo(gitrepo.Spec.Repo, grr.AllowedRepoPatterns, pol.GitAllowedRepoPatterns)
-	if err != nil {
+	if _, err := isAllowedByRepo(gitrepo.Spec.Repo, grr.AllowedRepoPatterns, pol.GitAllowedRepoPatterns); err != nil {
```
weyfonk: nit: `isAllowedByRepo` seems to only be used here, so we could refactor it to return a single value.

internal/cmd/controller/gitops/reconciler/restrictions.go
```diff
@@ -60,24 +60,26 @@ func AuthorizeAndAssignDefaults(ctx context.Context, c client.Client, gitrepo *f
 		return errors.New("empty targetNamespace denied, because allowedTargetNamespaceSelector restriction is present")
 	}
 
-	targetNamespace, err := isAllowed(gitrepo.Spec.TargetNamespace, "", allowedTargetNS)
-	if err != nil {
+	// Apply defaults before validation, so a default that is not in the allow-list
+	// is rejected rather than silently accepted. This keeps the GitRepo path
+	// consistent with the HelmOp and Bundle paths.
+	serviceAccount := firstNonEmpty(gitrepo.Spec.ServiceAccount, defaultSA)
```
weyfonk, line 66: nit: if the target namespace validation fails, we don't even need to compute these.

weyfonk (approved):

internal/cmd/controller/gitops/reconciler/restrictions_test.go
```diff
@@ -210,6 +210,57 @@ func TestAuthorizeAndAssignDefaults(t *testing.T) {
...
+				},
+			},
+			expectedGr: fleet.GitRepo{
+				Spec: fleet.GitRepoSpec{
+					Repo: "http://foo.bar/baz",
+				},
+			},
+			expectedErr: "disallowed clientSecretName.*",
+		},
 		{
```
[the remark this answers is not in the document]
p-se: Those test are exercising the same path as GRR is to be removed eventually, it does make sense to have tests that cover Policy to make sure those aren't lost when GRR is removed.

Issue #5252, opened 2026-06-08
HelmOp: Secret cloned to downstream namespace cannot be updated failing on immutable type field


d3flex:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Deploying a HelmOp resource that pulls an OCI chart from a private registry (e.g., dp.apps.rancher.io) and requires secrets on both the upstream (chart pull) and downstream (image pull) clusters causes the bundle controller to enter an infinite reconcile loop.

The fleet-controller logs the following error every ~5 seconds indefinitely:

```bash
INFO  bundle  requeue event: failed to clone secret fleet-default/rancher-registry-auth to downstream cluster namespace: failed to create or update source secret cluster-fleet-default-downstream1-32d08b529583/rancher-registry-auth: Secret "rancher-registry-auth" is invalid: type: Invalid value: "fleet.cattle.io/bundle-helmops-access/v1alpha1": field is immutable
```

From investigation, the secret is successfully created in the downstream cluster namespace on the first attempt. However, on subsequent reconcile cycles Fleet attempts to update the same value of the type field (fleet.cattle.io/bundle-helmops-access/v1alpha1). Kubernetes rejects this because a Secret's type field is immutable once set. The controller treats this as a retryable error and loops indefinitely.

Symptoms observed:
- BUNDLEDEPLOYMENTS-READY 0/0 on the HelmOp resource
- kubectl describe helmop kafka-app shows no status conditions (generation=1, observedGeneration=0)
- The secret copy exists in the downstream cluster namespace but the deployment never proceeds
- No error is surfaced via kubectl describe helmop, only visible in fleet-controller pod logs

### Expected Behavior

Not sure about that. Fleet should successfully reconcile the HelmOp and deploy the Helm chart to the downstream cluster. Likely the secret cloned to the downstream cluster namespace should not have its type field modified after initial creation.

### Steps To Reproduce

1. ```bash
  kubectl create secret docker-registry rancher-registry-auth \
    --docker-server=dp.apps.rancher.io \
    --docker-username=<user> \
    --docker-password=<token> \
    -n fleet-default
```
 2. ```bash
  kubectl create secret docker-registry application-collection \
    --docker-server=dp.apps.rancher.io \
    --docker-username=<user> \
    --docker-password=<token> \
    -n default
```
3. `kubectl apply -f helmop_kafka.yaml`
```bash
apiVersion: fleet.cattle.io/v1alpha1
  kind: HelmOp
[...]

PR #5294, opened 2026-06-16
fix(bundle): do not update secret type when cloning downstream secrets

d3flex:
That prevents `cloneSecret` to try update the Secret.type. When the same secret was referenced both as a HelmOp access secret and as a downstreamResources entry, cloneSecret was called twice for the same target secret.

Because this error is treated as retryable, the controller requeued the bundle and rerun the same failing reconcile indefinitely, so the BundleDeployment could not reach the expected status.

It is unclear what should happen if a source secret changes its type between reconciles but this seems likely an unwanted behavior and a new Secret should be created instead.

Refers to https://github.com/rancher/fleet/issues/5252

## Additional Information

No behavior change

### Checklist


weyfonk (approved):
Nice one, LGTM.

PR #5296, opened 2026-06-17
benchmarks: avoid unreliable kubectl attach when fetching metrics

p-se:
Fixes the benchmarks in conjunction with https://github.com/rancher/fleet/pull/5267

### Checklist

### DIFF
--- a/benchmarks/record/record.go
+++ b/benchmarks/record/record.go
@@ -240,14 +240,16 @@ func getMetrics(res map[string]float64, url string, controllers ...string) {
 	)
 	Eventually(func() error {
 		pod := addRandomSuffix("curl")
+		defer func() {
+			_, _ = k.Run("delete", "pod", "--namespace", "cattle-fleet-system", pod, "--ignore-not-found")
+		}()
+
 		GinkgoWriter.Print("Fetching metrics from " + url + "\n")
 
 		// Create the pod without --attach to avoid kubectl's unreliable
 		// attach mechanism, which can duplicate output on fallback to logs.
 		_, _, err := k.RunStdout("run", "--restart=Never", pod, "--image=curlimages/curl", "--namespace", "cattle-fleet-system", "--command", "--", "curl", "-sf", url)
 		if err != nil {
-			// Pod may have been created despite the error; clean up.
-			_, _ = k.Run("delete", "pod", "--namespace", "cattle-fleet-system", pod, "--ignore-not-found")
 			return fmt.Errorf("kubectl run: %w", err)
 		}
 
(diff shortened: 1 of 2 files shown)


benchmarks/record/record.go
```diff
@@ -233,33 +233,42 @@ func Metrics(experiment *gm.Experiment, suffix string) {
...
+		pod := addRandomSuffix("curl")
 		GinkgoWriter.Print("Fetching metrics from " + url + "\n")
-		out, err := k.Run("run", "--rm", "--attach", "--quiet", "--restart=Never", pod, "--image=curlimages/curl", "--namespace", "cattle-fleet-system", "--command", "--", "curl", "-s", url)
+
+		// Create the pod without --attach to avoid kubectl's unreliable
+		// attach mechanism, which can duplicate output on fallback to logs.
+		_, _, err := k.RunStdout("run", "--restart=Never", pod, "--image=curlimages/curl", "--namespace", "cattle-fleet-system", "--command", "--", "curl", "-sf", url)
+		if err != nil {
+			// Pod may have been created despite the error; clean up.
+			_, _ = k.Run("delete", "pod", "--namespace", "cattle-fleet-system", pod, "--ignore-not-found")
```
weyfonk: Since this always needs to be run, couldn't it simply be the object of a `defer` call?

benchmarks/record/record.go
```diff
@@ -233,33 +233,42 @@ func Metrics(experiment *gm.Experiment, suffix string) {
...
+			return fmt.Errorf("waiting for pod: %w", err)
+		}
+
+		phase, _, _ := k.RunStdout("get", "pod", pod, "--namespace", "cattle-fleet-system", "-o", "jsonpath={.status.phase}")
+		if strings.TrimSpace(phase) != "Succeeded" {
+			_, _ = k.Run("delete", "pod", "--namespace", "cattle-fleet-system", pod, "--ignore-not-found")
+			return fmt.Errorf("curl pod %s finished with phase %s", pod, phase)
+		}
+
+		out, _, err := k.RunStdout("logs", "--namespace", "cattle-fleet-system", pod)
```
weyfonk, line 269: nit: in case we run into more troubles getting these logs, we could replace creating a new pod with a port forward for fetching metrics, as [done in the dump CLI](https://github.com/rancher/fleet/blob/main/internal/cmd/cli/dump/dump.go#L832).

e2e/testenv/kubectl/kubectl.go
```diff
@@ -85,30 +85,50 @@ func (c Command) Run(args ...string) (string, error) {
...
+func (c Command) RunStdout(args ...string) (stdout, stderr string, err error) {
+	if c.cnt != "" {
+		args = append([]string{"--context", c.cnt}, args...)
+	}
+
+	if c.ns != "" {
+		args = append([]string{"-n", c.ns}, args...)
+	}
+
+	GinkgoWriter.Printf("kubectl %s\n", strings.Join(args, " "))
```
weyfonk, line 106: Do we need this line, or might it be a leftover from testing?

e2e/testenv/kubectl/kubectl.go
```diff
@@ -85,30 +85,50 @@ func (c Command) Run(args ...string) (string, error) {
 	}
 
 	GinkgoWriter.Printf("kubectl %s\n", strings.Join(args, " "))
-	result, err := c.exec("kubectl", args...)
+	stdout, stderr, err := c.exec("kubectl", args...)
+	result := stdout + stderr
```
weyfonk: When simply concatenating the two, would we have any way of distinguishing what comes from stdout vs stderr? I wonder if the resulting output might be confusing.

p-se: Hmm, yeah. Let's preserve the previous behavior and add a new one instead.

e2e/testenv/kubectl/kubectl.go
p-se: It's intentional, it mirrors the same pattern in Run(), just showing stdout and stderr separately since RunStdout has them split.

benchmarks/record/record.go
p-se: Hope it won't be necessary, but yeah, we may have to do that.

weyfonk (approved):

PR #5301, opened 2026-06-18
Harden GitHub Actions workflows

thardeck:
Move GitHub Actions step outputs into environment variables, add least-privilege `permissions`, and restrict checkout/secret usage to address `zizmor` findings. No functional changes - outputs are set in `env:` and referenced safely in `run:` blocks.

- **Template-injection**: Move `steps.*.outputs.*` into env vars (e.g. `CACHE_HIT`) and reference `$CACHE_HIT` in `run` blocks.
- **Permissions & checkout**: Add explicit `permissions` and set `persist-credentials: false` for non-release `actions/checkout` steps.
- **Config**: Update [.zizmor.yml](.zizmor.yml) suppress non-actionable `cache-poisoning` because using caches should be fine outside of the release process to speed up CI and ignores warning about stored git credentials (`artipacked`) for two actions which need them during their runtime

### DIFF
--- a/.github/workflows/e2e-test-fleet-in-rancher.yml
+++ b/.github/workflows/e2e-test-fleet-in-rancher.yml
@@ -22,6 +22,9 @@ on:
     paths-ignore:
       - '*.md'
 
+permissions:
+  contents: read
+
 env:
   GOARCH: amd64
   CGO_ENABLED: 0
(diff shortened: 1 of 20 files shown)


kkaempf (approved):
lgtm

.github/workflows/e2e-test-fleet-in-rancher.yml
```diff
@@ -190,11 +196,14 @@ jobs:
 
       -
         name: Set up latest Rancher
+        env:
+          CHARTS_REPO: ${{ needs.release-against-test-charts.outputs.charts_repo }}
+          CHARTS_BRANCH: ${{ needs.release-against-test-charts.outputs.target_branch }}
```
d3flex, line 201: nit but I think it makes sense: s/CHARTS_BRANCH/TARGET_BRANCH/

Issue #5261, opened 2026-06-11
Garbage collector incorrectly uninstalls valid helm releases when BundleDeployment name exceeds 53 characters


nilsalex:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When a fleet.yaml sets an explicit name: field longer than 53 characters without also setting helm.releaseName, the agent's garbage collector periodically uninstalls the corresponding helm release, even though it is valid and actively managed. Fleet immediately redeploys it, and the GC uninstalls it again — indefinitely.
The agent logs the following repeatedly:
{"level":"info","logger":"bundledeployment.garbage-collect",
 "msg":"Deleting unknown bundle ID, helm uninstall",
 "bundleID":"<bundle-name>",
 "release":"<namespace>/<truncated-name>-<hash>",
 "expectedRelease":"<namespace>/<full-bundle-name>"}

### Expected Behavior

The GC recognises the release as belonging to the bundle deployment and leaves it untouched.

### Steps To Reproduce

See https://github.com/nilsalex/fleet-gc-bug-repro for a minimal reproduction with instructions.

### Environment

```markdown
- Architecture: amd64
- Fleet Version: v0.14.3
- Cluster: reproduced on k3d and EKS
```

### Logs

```markdown
{"level":"info","logger":"bundledeployment.garbage-collect",
 "msg":"Deleting unknown bundle ID, helm uninstall",
 "bundleID":"repro-bundle-name-that-exceeds-the-53-char-helm-limit-x",
 "release":"default/repro-bundle-name-that-exceeds-the-53-char-helm-98fdb",
 "expectedRelease":"default/repro-bundle-name-that-exceeds-the-53-char-helm-limit-x"}
```

### Root cause

The helm release name is derived from bd.Name via names.HelmReleaseName() in getOpts() (internal/helmdeployer/deployer.go), which truncates names longer than 53 characters and appends a 5-character hash. The GC's releaseKey() function (internal/cmd/agent/deployer/cleanup/cleanup.go) computes the expected release name using bd.Name directly, without applying the same truncation. The two values diverge whenever bd.Name exceeds 53 characters.
bd.Name exceeds 53 characters when a fleet.yaml sets an explicit name: field longer than 53 characters — the value is used raw as the Bundle and BundleDeployment name, bypassing the sanitization that auto-derived names go through. Setting helm.releaseName explicitly avoids the bug, since both getOpts() and releaseKey() use it directly without truncation.

[...]

PR #5302, opened 2026-06-18
fix: apply HelmReleaseName truncation in releaseKey for GC comparison

nilsalex:
Refers to #5261

When no explicit Helm.ReleaseName is configured, the deployer derives the helm release name via names.HelmReleaseName(bundleID), which truncates names longer than 53 characters and appends a 5-char hash. The releaseKey function in cleanup.go was using bd.Name directly, causing a mismatch against the actual truncated release name stored in helm secrets.

This caused the garbage collector to incorrectly flag valid releases as unknown and delete them, logging:

  "Deleting unknown bundle ID, helm uninstall" ...
  "release":"default/...-742e3","expectedRelease":"default/...-5nj4is"

Fix: apply names.HelmReleaseName() in releaseKey when no explicit release name is set, mirroring what deployer.getOpts() does.


0xavi0 (approved):
LGTM

PR #5305, opened 2026-06-18
Generate binaries for AMD64 Mac

weyfonk:
Users running Mac OS on amd64 architectures should be able to run the Fleet CLI and benchmark binaries.

### Checklist


kkaempf:
Hmm, the move to `aarch64` on MacOS [happened 6 years ago](https://en.wikipedia.org/wiki/Mac_transition_to_Apple_silicon).
Not sure supporting this has any market relevance.

PR #5318, opened 2026-06-19
fix: propagate Accepted=False to Ready condition when GitRepo job creation fails

### LINKED ISSUE #4865: GitRepo incorrectly marked as Ready upon failure

- [x] I have searched the existing issues

### Current Behavior

Rancher version: v2.14.0-alpha9
Fleet version: v0.15.0-beta.4

A GitRepo fails to be deployed correctly due to `cabundle` errors.  
Note this particular error seems to be happening when configuring a Private CA for Rancher. See [docs](https://ranchermanager.docs.rancher.com/getting-started/installation-and-upgrade/resources/update-rancher-certificate#updating-a-private-ca-certificate)

```yaml
kind: GitRepo
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"fleet.cattle.io/v1alpha1","kind":"GitRepo","metadata":{"annotations":{},"name":"gcp-cluster-class-kubeadm-344zh6","namespace":"fleet-local"},"spec":{"branch":"remove_cert_manager","forceSyncGeneration":1,"paths":["examples/clusterclasses/gcp/kubeadm"],"repo":"https://github.com/rancher/turtles","targetNamespace":"creategitops-gcp-kubeadm"}}
  creationTimestamp: 2026-03-19T08:45:49Z
  finalizers:
  - fleet.cattle.io/gitrepo-finalizer
  generation: 1
  name: gcp-cluster-class-kubeadm-344zh6
  namespace: fleet-local
  resourceVersion: '33557'
  uid: 78fd2c55-10a6-4c2b-9964-29550d54e022
spec:
  branch: remove_cert_manager
  forceSyncGeneration: 1
  paths:
  - examples/clusterclasses/gcp/kubeadm
  repo: https://github.com/rancher/turtles
  targetNamespace: creategitops-gcp-kubeadm
status:
  commit: 2d71fdbec9841cc696121c843796c4ec70fdf7a6
  conditions:
  - lastUpdateTime: 2026-03-19T08:45:50Z
    status: 'True'
    type: Ready
  - lastUpdateTime: 2026-03-19T08:45:50Z
    status: 'True'
    type: GitPolling
  - lastUpdateTime: 2026-03-19T08:45:52Z
    status: 'True'
    type: Accepted
  - lastUpdateTime: 2026-03-19T08:45:52Z
    status: 'False'
    type: Reconciling
  - lastUpdateTime: 2026-03-19T08:45:52Z
    status: 'False'
    type: Stalled
  desiredReadyClusters: 0
  display:
    readyBundleDeployments: 0/0
  gitJobStatus: Current
  lastPollingTriggered: 2026-03-19T09:15:57Z
  observedGeneration: 1
  pollingCommit: 2d71fdbec9841cc696121c843796c4ec70fdf7a6
  readyClusters: 0
  resourceCounts:
    desiredReady: 0
    missing: 0
    modified: 0
    notReady: 0
    orphaned: 0
    ready: 0
    unknown: 0
    waitApplied: 0
  summary:
    desiredReady: 0
    ready: 0
  updateGeneration: 1
```

There are a few problems with the status reporting:
1. `Ready` status is `True`, however no resource was applied in the target namespace and the only log instance for this GitRepo is an error:

```
{"level":"error","ts":"2026-03-19T08:45:52Z","msg":"Reconciler error","controller":"gitrepo","controllerGroup":"fleet.cattle.io","controllerKind":"GitRepo","GitRepo":{"name":"gcp-cluster-class-kubeadm-344zh6","namespace":"fleet-local"},"namespace":"fleet-local","name":"gcp-cluster-class-kubeadm-344zh6","reconcileID":"a87211a5-74e2-4b00-99f0-b1384f00555e","error":"error creating git job: Secret \"gcp-cluster-class-kubeadm-344zh6-rancher-cabundle\" not found","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.23.1/pkg/internal/controller/controller.go:495\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.23.1/pkg/internal/controller/controller.go:438\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func1.1\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.23.1/pkg/internal/controller/controller.go:313"}
```

2. The `Ready` condition timestamp does not make sense. It's too early compared to the log I posted above. It does feel like this condition is always defaulted to True.

3. The GitRepo is no longer reconciled. It does not seem to ever recover from this state. Deletion and recreation seem to be necessary.

### Expected Behavior

As user I expect the `GitRepo` resource to be Ready only when resources are correctly applied on all targeted clusters.

As a workaround, a user can currently exploit the `status.readyClusters` count, if the expected value is known beforehand.

### Steps To Reproduce

_No response_

### Environment

```markdown
- Architecture:
- Fleet Version:
- Cluster:
  - Provider:
  - Options:
  - Kubernetes Version:
```

### Logs

```markdown

```

### Anything else?

_No response_

mvanhorn:
Refers to #4865

## Summary

When a GitRepo fails before any bundles are created (e.g. a missing or invalid secret referenced by `HelmSecretNameForPaths` or `ClientSecretName`), the `gitjob_controller` sets `Accepted=False` via `updateErrorStatus`. However the `StatusReconciler` computes `Ready` independently from the bundle summary. With no bundles present the summary is 0/0, which satisfies `IsReady()`, so the GitRepo is reported as `Ready=True` even though nothing is deploying.

This PR adds `propagateAcceptedFailureToReady` in `status_controller.go`. After `setReadyStatusFromBundle` runs, the function checks whether the `Accepted` condition is `False`. If so it forces `Ready=False`, copying the `Accepted` message and reason onto the `Ready` condition. The check is additive: bundle-driven errors are still handled as before, and the propagation only fires when `Accepted` is explicitly `False`.

## Why this matters

Users who hit pre-job errors (missing secret, cabundle failure, restriction violation) see a healthy green `GitRepo` while their workload is silently not deploying. This change makes the error visible exactly where operators look first.

## Testing

Four unit tests added to `gitjob_test.go` (same package):

- `Test_PropagateAcceptedFailureToReady_SetsReadyFalse` - overwrites existing `Ready=True` when `Accepted=False`
- `Test_PropagateAcceptedFailureToReady_AddsReadyWhenMissing` - adds `Ready=False` when no Ready condition exists yet (the 0/0 empty-bundle case)
- `Test_PropagateAcceptedFailureToReady_NoOpWhenAcceptedTrue` - verifies no change when `Accepted=True`
- `Test_PropagateAcceptedFailureToReady_NoOpWhenNoAccepted` - verifies no change when no Accepted condition present

All 115 tests in the reconciler package pass (`go test ./internal/cmd/controller/gitops/reconciler/...`).

Fixes #4865

### Checklist

- [ ] I have updated the documentation via a pull request in the [fleet-product-docs](https://github.com/rancher/fleet-product-docs) repository.

### DIFF
--- a/internal/cmd/controller/gitops/reconciler/status_controller.go
+++ b/internal/cmd/controller/gitops/reconciler/status_controller.go
@@ -145,6 +145,12 @@ func (r *StatusReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		return ctrl.Result{}, err
 	}
 
+	// If the Accepted condition is False (e.g. missing secret, cabundle failure,
+	// restriction violation), propagate that failure to the Ready condition.
+	// Without this, a GitRepo with no bundles (because the job never ran) would
+	// report Ready=True because the empty bundle summary satisfies 0/0 == ready.
+	propagateAcceptedFailureToReady(gitrepo)
+
 	if err := r.updateStatus(ctx, orig, gitrepo); err != nil {
 		logger.Error(err, "Reconcile failed update to git repo status", "status", gitrepo.Status)
 		return ctrl.Result{RequeueAfter: durations.GitRepoStatusDelay}, nil
@@ -245,6 +251,49 @@ bundles:
 	gitrepo.Status.Conditions = newConditions
 
 	return nil
+}
+
+// propagateAcceptedFailureToReady ensures the Ready condition reflects an
+// Accepted=False error. When a pre-job error (missing secret, cabundle failure,
+// restriction violation) prevents any bundle from being created, the bundle
+// summary is 0/0, which would normally resolve to Ready=True. By copying the
+// Accepted condition's message onto Ready we surface the real error.
+func propagateAcceptedFailureToReady(gitrepo *fleet.GitRepo) {
+	var acceptedMsg string
+	var acceptedReason string
+	acceptedFalse := false
+	for _, c := range gitrepo.Status.Conditions {
+		if c.Type == fleet.GitRepoAcceptedCondition && c.Status == v1.ConditionFalse {
+			acceptedFalse = true
+			acceptedMsg = c.Message
+			acceptedReason = c.Reason
+			break
+		}
+	}
+	if !acceptedFalse {
+		return
+	}
+
+	found := false
+	newConditions := make([]genericcondition.GenericCondition, 0, len(gitrepo.Status.Conditions))
+	for _, c := range gitrepo.Status.Conditions {
+		if c.Type == string(fleet.Ready) {
+			c.Status = v1.ConditionFalse
+			c.Message = acceptedMsg
+			c.Reason = acceptedReason
+			found = true
+		}
+		newConditions = append(newConditions, c)
+	}
+	if !found {
+		newConditions = append(newConditions, genericcondition.GenericCondition{
+			Type:    string(fleet.Ready),
+			Status:  v1.ConditionFalse,
+			Message: acceptedMsg,
+			Reason:  acceptedReason,
+		})
+	}
+	gitrepo.Status.Conditions = newConditions
 }
 
 type forcedDelayingSource[R comparable] struct {
--- a/internal/cmd/controller/gitops/reconciler/gitjob_test.go
+++ b/internal/cmd/controller/gitops/reconciler/gitjob_test.go
@@ -3495,3 +3495,102 @@ func TestCreateJob_AlreadyExistsIsIgnored(t *testing.T) {
 		t.Fatal("createJob should return false when job already exists")
 	}
 }
+
+// Test_PropagateAcceptedFailureToReady_SetsReadyFalse verifies that when the
+// Accepted condition is False, propagateAcceptedFailureToReady forces Ready to
+// False with the same message and reason. This is the core fix for
+// https://github.com/rancher/fleet/issues/4865.
+func Test_PropagateAcceptedFailureToReady_SetsReadyFalse(t *testing.T) {
+	gitrepo := &fleetv1.GitRepo{}
+	gitrepo.Status.Conditions = []genericcondition.GenericCondition{
+		{
+			Type:    fleetv1.GitRepoAcceptedCondition,
+			Status:  "False",
+			Message: "failed to look up HelmSecretNameForPaths, error: secret not found",
+			Reason:  "Error",
+		},
+		{
+			Type:   "Ready",
+			Status: "True",
+		},
+	}
+
+	propagateAcceptedFailureToReady(gitrepo)
+
+	readyCond, found := getCondition(gitrepo, "Ready")
+	if !found {
+		t.Fatal("expected Ready condition to be present")
+	}
+	if readyCond.Status != "False" {
+		t.Errorf("expected Ready=False, got Ready=%s", readyCond.Status)
+	}
+	if readyCond.Message != "failed to look up HelmSecretNameForPaths, error: secret not found" {
+		t.Errorf("unexpected Ready message: %s", readyCond.Message)
+	}
+}
+
+// Test_PropagateAcceptedFailureToReady_AddsReadyWhenMissing verifies that when
+// Ready is absent and Accepted=False, a Ready=False condition is added.
+func Test_PropagateAcceptedFailureToReady_AddsReadyWhenMissing(t *testing.T) {
+	gitrepo := &fleetv1.GitRepo{}
+	gitrepo.Status.Conditions = []genericcondition.GenericCondition{
+		{
+			Type:    fleetv1.GitRepoAcceptedCondition,
+			Status:  "False",
+			Message: "missing cabundle secret",
+			Reason:  "Error",
+		},
+	}
+
+	propagateAcceptedFailureToReady(gitrepo)
+
+	readyCond, found := getCondition(gitrepo, "Ready")
+	if !found {
+		t.Fatal("expected Ready condition to be added")
+	}
+	if readyCond.Status != "False" {
+		t.Errorf("expected Ready=False, got Ready=%s", readyCond.Status)
+	}
+	if readyCond.Message != "missing cabundle secret" {
+		t.Errorf("unexpected Ready message: %s", readyCond.Message)
+	}
+}
+
+// Test_PropagateAcceptedFailureToReady_NoOpWhenAcceptedTrue verifies that when
+// Accepted=True, the Ready condition is not changed by propagateAcceptedFailureToReady.
+func Test_PropagateAcceptedFailureToReady_NoOpWhenAcceptedTrue(t *testing.T) {
+	gitrepo := &fleetv1.GitRepo{}
+	gitrepo.Status.Conditions = []genericcondition.GenericCondition{
+		{
+			Type:   fleetv1.GitRepoAcceptedCondition,
+			Status: "True",
+		},
+		{
+			Type:   "Ready",
+			Status: "True",
+		},
+	}
+
+	propagateAcceptedFailureToReady(gitrepo)
+
+	readyCond, found := getCondition(gitrepo, "Ready")
+	if !found {
+		t.Fatal("expected Ready condition to be present")
+	}
+	if readyCond.Status != "True" {
+		t.Errorf("expected Ready=True (unchanged), got Ready=%s", readyCond.Status)
+	}
+}
+
+// Test_PropagateAcceptedFailureToReady_NoOpWhenNoAccepted verifies that when
+// no Accepted condition is present, propagateAcceptedFailureToReady is a no-op.
+func Test_PropagateAcceptedFailureToReady_NoOpWhenNoAccepted(t *testing.T) {
+	gitrepo := &fleetv1.GitRepo{}
+	// No Accepted condition, no Ready condition - simulates fresh GitRepo with 0/0 bundles.
+	propagateAcceptedFailureToReady(gitrepo)
+
+	_, found := getCondition(gitrepo, "Ready")
+	if found {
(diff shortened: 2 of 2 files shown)


internal/cmd/controller/gitops/reconciler/status_controller.go
```diff
@@ -247,6 +253,49 @@ bundles:
...
+	found := false
+	newConditions := make([]genericcondition.GenericCondition, 0, len(gitrepo.Status.Conditions))
+	for _, c := range gitrepo.Status.Conditions {
+		if c.Type == string(fleet.Ready) {
+			c.Status = v1.ConditionFalse
+			c.Message = acceptedMsg
+			c.Reason = acceptedReason
+			found = true
+		}
+		newConditions = append(newConditions, c)
```
0xavi0, line 286: I'm not sure this could introduce an undesired`LastUpdateTime` update.

Let's say the first iteration of the status reconciler computes the desired state and, because there are no `Bundles` it considers `Ready` should be `true`.
Later this new function propagates the message and reason from the `Accepted` condition and status is now `False`.
That's fine.

The problem I see (I would need to confirm testing it myself) is that for the next reconcile it re-checks again, there are no Bundles and the `Ready` condition is found as `False`. 
When that happens, wrangler updates the `LastUpdateTime` when it sets the condition back to `True`. 
Later the new function only preserves the message and reason, so the `LastUpdateTime` has been updated (when it shouldn't). 

I think the tests are not exercising this particular scenario. I think it should be also covered 


---- 

Also, what happens if the `Ready` condition was already `False`? 
We would be overwritting unconditionally the message and reason to the `Accepted` values, right?

0xavi0 (commented):
Thanks, for your contribution!
I've left a few comments.

internal/cmd/controller/gitops/reconciler/status_controller.go
```diff
@@ -247,6 +253,49 @@ bundles:
...
+	}
+	if !found {
+		newConditions = append(newConditions, genericcondition.GenericCondition{
+			Type:    string(fleet.Ready),
+			Status:  v1.ConditionFalse,
+			Message: acceptedMsg,
+			Reason:  acceptedReason,
+		})
+	}
+	gitrepo.Status.Conditions = newConditions
```
anmazzotti, line 296: Doesn't this invalidate/remove all other conditions that may be existing on this object? 
I also expect the Ready condition to be a summary of other conditions, so maybe a good approach for this state is to introduce a new one? 

Not that familiar with the API to judge, but I feel like it would be a more clean approach.

mvanhorn:
Thanks both, this is the right pushback and I'd rather settle the design here before reworking.

Current state, so we're on the same page: the failure is already recorded on the `Accepted` condition (set False in `gitjob_controller` when job creation fails). The bug is only that `Ready`, computed by `summary.SetReadyConditions` from the bundle summary, comes back True when there are no bundles (0/0), so the `Accepted` failure never surfaces on `Ready`. My `propagateAcceptedFailureToReady` patched that after the fact by overwriting `Ready`, which is the part you're both right to dislike.

@anmazzotti: agreed that `Ready` should stay a summary. One clarification: the current code does preserve the other conditions (it copies them through), but your deeper point stands that overwriting the summary is the wrong mechanism. The distinct condition you're suggesting effectively already exists as `Accepted`. So rather than add a third condition that duplicates `Accepted`, my proposal is to fold `Accepted=False` into the `Ready` summary computation itself, so `Ready` is genuinely summarized (`Accepted=False` makes `Ready=False` with the `Accepted` reason) instead of being rewritten. If you'd rather I introduce a brand new condition type instead of reusing `Accepted`, I'm happy to, just let me know the name you'd want.

@0xavi0: the flip-flop you described is real, caused by the overwrite running after `SetReadyConditions` recomputes `Ready=True` each pass. Folding `Accepted` into the summary removes the recompute-then-overwrite cycle. On top of that I'd guard the write the same way `setAcceptedConditionHelm` already does for the helm path: DeepCopy the status, set the condition, and only call `LastUpdated` when `equality.Semantic.DeepEqual` reports an actual change. That keeps `LastUpdateTime` stable across reconciles. For your second question (Ready already False from a real bundle failure): a genuine bundle `Ready=False` should take precedence over the `Accepted` message rather than be clobbered, so I'd only let `Accepted` drive `Ready` when `Ready` isn't already False for a more specific reason. I'll add a test that runs two reconciles and asserts `LastUpdateTime` doesn't move on the second pass, which is the scenario the current tests miss.

Does folding `Accepted` into the `Ready` summary (rather than a new condition) match your intent? If so I'll rework along these lines.

0xavi0:
@mvanhorn sorry for the late reply. Sounds good to me. Ping me when you commit those changes so I can review again. Thanks!

mvanhorn:
Pushed @0xavi0 - both points addressed.

@0xavi0 on the LastUpdateTime churn: you were right, and it was reachable exactly the way you described. The propagation now goes through wrangler's `condition.Cond` and checks `MatchesError` first, so when Ready already carries the same status/reason/message the write is skipped entirely and LastUpdateTime stays put across reconciles. I also stopped calling `summary.SetReadyConditions` while Accepted is False: with no bundles it sets Ready=True from the 0/0 summary, which is what made the next reconcile flip the condition back and forth in the first place.

@anmazzotti yes, it did - good catch. The old code rebuilt `Status.Conditions` from scratch, so anything not copied across was dropped. Ready is now updated in place and the rest of the conditions are left alone. On your second point, I agree Ready should stay a summary of other conditions rather than a mirror of Accepted, which is why this only forces Ready=False while Accepted is False and otherwise leaves the normal summary path to own it.

Tests cover repeated reconciles (LastUpdateTime stable on the second pass), preservation of unrelated conditions, and the Accepted=False path. go build, go vet and the package tests pass locally.

0xavi0 (commented):
Thanks!  Conditions and no longer overwritten in the second reconciliation. 

I think Ready is still overwritten even when it's already False, though.

I've used the following `GitRepo`:

```yaml
kind: GitRepo
metadata:
  name: status-ready
  namespace: fleet-local
spec:
  repo: https://github.com/rancher/fleet-test-data
  branch: master
  paths:
  - simple
```

1. Deploy it and wait until it is green. The simple path deploys one Service into simple-example:

```
$ kubectl -n fleet-local get gitrepo status-ready
NAME           REPO                                         COMMIT     BUNDLEDEPLOYMENTS-READY   STATUS
status-ready   https://github.com/rancher/fleet-test-data   2b37e0f…   1/1
```

2. Produce a real, bundle-level failure by deleting one of the resources Fleet deployed:

```
$ kubectl -n simple-example delete service simple-service
service "simple-service" deleted

$ kubectl -n fleet-local get gitrepo status-ready
NAME           REPO                                         COMMIT     BUNDLEDEPLOYMENTS-READY   STATUS
status-ready   https://github.com/rancher/fleet-test-data   2b37e0f…   0/1                       Modified(1) [Cluster fleet-local/local]; service.v1 simple-example/simple-service missing
```

Ready correctly names the missing resource. Both conditions:

```
$ kubectl -n fleet-local get gitrepo status-ready -o jsonpath='{range .status.conditions[?(@.type=="Accepted")]}Accepted: {.status}  {.message}{"\n"}{end}{range .status.conditions[?(@.type=="Ready")]}Ready:    {.status}  {.message}{"\n"}{end}'
Accepted: True
Ready:    False  Modified(1) [Cluster fleet-local/local]; service.v1 simple-example/simple-service missing
```

3. Now trigger a pre-job error on the same GitRepo — point helmSecretNameForPaths at a secret that does not exist:

```
$ kubectl -n fleet-local patch gitrepo status-ready --type=merge \
    -p '{"spec":{"helmSecretNameForPaths":"does-not-exist"
gitrepo.fleet.cattle.io/status-ready patched                                                                                         
$ kubectl -n fleet-local get gitrepo status-ready -o jsonpath='{range .status.conditions[?(@.type=="Accepted")]}Accepted: {.status}  {.message}{"\n"}{end}{range .status.conditions[?(@.type=="s}  {.message}{"\n"}{end}'
Accepted: False  error validating external secrets: failed to look up HelmSecretNameForPaths, error: Secret "does-not-exist" not found
Ready:    False  error validating external secrets: failedForPaths, error: Secret "does-not-exist" not found
```

mvanhorn:
Good catch from the bot, and it was right. Fixed in `f13cbbf`.

`Reconcile` returns early when `AuthorizeAndAssignDefaults` fails, above the propagation this PR adds, while `gitjob_controller` records that failure as `Accepted=False`. So a GitRepo that was already `Ready=True` kept reporting `Ready=True` for restriction violations, which is exactly the case the PR claims to cover.

That path now calls the same `propagateAcceptedFailureToReady` helper and persists the result, rather than a second copy of the logic. The helper still goes through `condition.Cond` and checks `MatchesError` first, so the `LastUpdateTime` behaviour @0xavi0 asked about is unchanged: an identical status/reason/message is not rewritten.

Added `Test_StatusReconcile_RestrictionViolation_PropagatesAcceptedToReady`, which sets up a GitRepo that is `Ready=True` with `Accepted=False` from a targetNamespace restriction and asserts `Ready` flips to `False` carrying the Accepted reason, that a second reconcile leaves `LastUpdateTime` on its sentinel, and that the second pass still returns on the restriction path rather than computing display status.

The reconciler package is green locally.

Issue #4210, opened 2025-10-09
Bug: namespaceLabels/namespaceAnnotations targetCustomizations off-target


EdKingscote:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

With a fleet.yaml as follows:
```
targetCustomizations:
  - name: myapp-local
    namespaceAnnotations:
      edtest: local
    namespaceLabels:
      edtest: local
    clusterName: local

  - name: myapp-preprod
    namespaceAnnotations:
      fredtest: preprod
    namespaceLabels:
      fredtest: preprod
    clusterName: preprod

```

The namespace in my preprod cluster has labels and annotations from both target customizations : 

<img width="816" height="553" alt="Image" src="https://github.com/user-attachments/assets/8a40ed76-211f-493d-ad2d-f987d0dc3ddc" />

If the key is the same, then my local cluster ends up with the preprod values.

This is preventing me from assigning namespaces to projects without having to have lots of small extra wrapper charts just to help manage the namespaces.

### Expected Behavior

The target customizations should be targeted as expected!

### Steps To Reproduce

Across two K8s clusters have namepspaceLabels/namepaceAnnotations targeted separately and observe behavior. 

### Environment

```markdown
- Architecture: x86_64
- Rancher Version: v2.12.2
- Fleet Version: v0.13.2
- Cluster:
  - Provider: RKE2 on Ubuntu (VMs)
  - Options: 3 nodes per cluster
  - Kubernetes Version:  
      - local: v1.32.5+rke2r1
      - preprod: v1.32.6+rke2r1
```

### Logs

```markdown

```

p-se:
# QA Template

https://gist.github.com/p-se/ced41c219254f274a316f543ec8ea0a7

PR #5319, opened 2026-06-21
fix: scope namespace metadata to matched target customizations

immanuwell:
Refers to #4210

Target customizations were leaking `namespaceLabels` and `namespaceAnnotations` across clusters.
The controller was copying metadata from every `bundle.spec.targets` entry into every BundleDeployment, so cluster specific namespace metadata got mixed up.

Fix is pretty small: merge namespace metadata only from the matched target entries, then stop reapplying all targets later.
Added unit tests for the merge path and for the leak itself.

Repro:
1. Put two `targetCustomizations` in `fleet.yaml`, one for cluster `local` and one for `preprod`.
2. Set different `namespaceLabels` or `namespaceAnnotations` on each.
3. Deploy to both clusters.
4. Look at the generated BundleDeployments or namespaces.
5. Before this fix, `preprod` picks up `local` metadata too, which is not what you want.

Tests:
- `go test -count=1 ./internal/cmd/controller/...`


p-se:
@immanuwell Thank you very much for your contribution!

Unfortunately the fix introduces a regression: the target/customization namespaceLabels and namespaceAnnotations are no longer applied to the bundle deployments at all. They're merged into targetOpts, but then discarded one line later at builder.go:141, because options.Merge doesn't handle those two fields — only root-level namespace metadata survives. The existing e2e cases in target_customization_test.go ("override empty root configuration" is the clearest) should surface this.

May I suggest adding handling for namespaceLabels/namespaceAnnotations directly in Merge (calculate.go) instead of a separate helper alongside options.Merge? Merge already resolves every other BundleDeploymentOptions field, so this would be consistent and would let mergeTargetNamespaceMetadata be removed entirely.

p-se:
Quick update on the approach: just handling this in options.Merge wasn't enough. Turns out BundleTarget had its own NamespaceLabels/NamespaceAnnotations on top of the ones it already inherits from the embedded BundleDeploymentOptions. Same JSON tags, so YAML fills the shadowing ones and Merge only ever sees the empty embedded ones - nothing to merge. That shadow is also why the old apply-all loop existed (and why it leaked).

So I dropped the duplicate fields from BundleTarget, and now everything flows through Merge like every other option - no helper, no loop.

Sounds scary since it's in pkg/apis, but it's harmless: CRD comes out byte-identical (same field names), only deepcopy regenerated, and target.NamespaceLabels still compiles via promotion. Checked rancher + webhook too — nothing touches those fields on a BundleTarget.

Tested on a 2-downstream devenv: each namespace gets only its own projectId now, no leak.

weyfonk (approved):

0xavi0:
Before merging I think I've found something, but need some time to confirm

0xavi0:
I'm not sure if we want to address this now, but after applying the fix in this PR the labels remain for BundleDeployments that were already deployed before the fix.

What I did:

1. Deploy an upstream cluster from the main branch (non fixed) and 2 downstream clusters. (with name `dev` and `prod` and labels `env=dev` and `env=prod`)
2. Apply the following `GitRepo`
```yaml
kind: GitRepo
metadata:
  name: nslabel-leak
  namespace: fleet-default
spec:
  repo: https://github.com/0xavi0/fleet-examples
  branch: nslabels
  paths:
    - nslabels
  targets:
    - name: all
      clusterSelector: {}
```
3. Check that it deploys 2 `BundleDeployments` and that they both have the namespace.
You can use the following bash script that prints useful stuff:
```bash
#!/usr/bin/env bash
set -euo pipefail

NS=${NS:-fleet-default}
BD=${BD:-nslabel-leak-nslabels}
DEMO_NS=${DEMO_NS:-nslabel-demo}
UPSTREAM_CTX=${UPSTREAM_CTX:-k3d-upstream}

clusters=("$@")
[ ${#clusters[@]} -eq 0 ] && clusters=(prod dev)

for c in "${clusters[@]}"; do
  ns=$(kubectl --context "$UPSTREAM_CTX" get cluster -n "$NS" "$c" -o jsonpath='{.status.namespace}')
  echo "== $c"

  kubectl --context "$UPSTREAM_CTX" get bundledeployment -n "$ns" "$BD" -o json | jq -r '
    "   deploymentID       (options hash) \(.spec.deploymentID       | split(":") | last)",
    "   stagedDeploymentID (options hash) \(.spec.stagedDeploymentID | split(":") | last)",
    "   spec.options.namespaceLabels        \(.spec.options.namespaceLabels       // "<nil>")",
    "   spec.stagedOptions.namespaceLabels  \(.spec.stagedOptions.namespaceLabels // "<nil>")"'

  kubectl --context "k3d-$c" get ns "$DEMO_NS" -o json \
    | jq -r '"   ns labels                           \(.metadata.labels)"'
done
```

It should print something like:
```bash
== dev

   deploymentID:       s-0d6783b960cf93eddca1a29541cc400e1a813b12f69c34938b4d30f9bf721:942db1ebe4315ab8741a54c2bfff5f94ef7846d6aa671f48543854de9eaa73e8
   stagedDeploymentID: s-0d6783b960cf93eddca1a29541cc400e1a813b12f69c34938b4d30f9bf721:942db1ebe4315ab8741a54c2bfff5f94ef7846d6aa671f48543854de9eaa73e8
   options:            {"tier":"critical"}
   stagedOptions:      {"tier":"critical"}
   ns labels: {"kubernetes.io/metadata.name":"nslabel-demo","tier":"critical"}
== prod

   deploymentID:       s-0d6783b960cf93eddca1a29541cc400e1a813b12f69c34938b4d30f9bf721:942db1ebe4315ab8741a54c2bfff5f94ef7846d6aa671f48543854de9eaa73e8
   stagedDeploymentID: s-0d6783b960cf93eddca1a29541cc400e1a813b12f69c34938b4d30f9bf721:942db1ebe4315ab8741a54c2bfff5f94ef7846d6aa671f48543854de9eaa73e8
   options:            {"tier":"critical"}
   stagedOptions:      {"tier":"critical"}
   ns labels: {"kubernetes.io/metadata.name":"nslabel-demo","tier":"critical"}
```

Which shows that the label is leaked. 

4. Upgrade the controller to the version of this PR.
```bash
git checkout --detach <this PR>
kubectl config use-context k3d-upstream
./dev/update-controller-k3d
kubectl -n cattle-fleet-system wait --for=condition=ready pod -l app=fleet-controller --timeout=180s
```

Wait 20s or so to let the controller reconcile and inspect again:
The leaked label is still there:
```
== prod
   deploymentID       (options hash) 683a5170424f839cfc420ad4d92d7955e4065ad52a595d648d4670f49b934295
   stagedDeploymentID (options hash) 683a5170424f839cfc420ad4d92d7955e4065ad52a595d648d4670f49b934295
   spec.options.namespaceLabels        {"tier":"critical"}
   spec.stagedOptions.namespaceLabels  {"tier":"critical"}
   ns labels                           {"kubernetes.io/metadata.name":"nslabel-demo","tier":"critical"}
== dev
   deploymentID       (options hash) 942db1ebe4315ab8741a54c2bfff5f94ef7846d6aa671f48543854de9eaa73e8
   stagedDeploymentID (options hash) 942db1ebe4315ab8741a54c2bfff5f94ef7846d6aa671f48543854de9eaa73e8
   spec.options.namespaceLabels        {"tier":"critical"}
   spec.stagedOptions.namespaceLabels  <nil>
   ns labels                           {"kubernetes.io/metadata.name":"nslabel-demo","tier":"critical"}
```

Note the `stagedOptions` is now `nil`

If I delete the `BundleDeployment` as a workaround it gets re-generated fine but the label is still present in the agent side, so I have to patch manually. 

A possible fix would be to add:
```go
    t.Deployment.Spec.Options.NamespaceLabels = t.Deployment.Spec.StagedOptions.NamespaceLabels
    t.Deployment.Spec.Options.NamespaceAnnotations = t.Deployment.Spec.StagedOptions.NamespaceAnnotations
```

In partition.go right after:
```go
    t.Deployment.Spec.DeploymentID == t.Deployment.Spec.StagedDeploymentID {
    t.Deployment.Spec.Options.Diff = t.Deployment.Spec.StagedOptions.Diff
```

And... ensure it's dropped from the hash, so it applies the changes:
In calculate.go:

Add:
```go
    sanitized.NamespaceLabels = nil
    sanitized.NamespaceAnnotations = nil
```

after:
```go
     sanitized.CorrectDrift = nil
```

That fixes the `BundleDeployment`. 
I think we have still to patch manually the labels in the agent side.

As I said... I'm not sure if we should fix this now or simply a different ticket for it.

0xavi0 (approved):

PR #5346, opened 2026-06-24
Switch container base images to bci-nano and remove tini

thardeck:
Switch the `fleet-controller` and `fleet-agent` container images from `bci-busybox` to `bci-nano` to reduce image size and CVE exposure.

Both images no longer ship `tini`. The controller and agent now run the Fleet binaries directly as PID 1, which keeps signal handling correct without an init wrapper. The original controller-side `tini` usage was only needed when `log.sh` was the entrypoint; that shell wrapper has been removed, and the remaining Fleet processes do not spawn child processes that require zombie reaping.

Because `bci-nano` has no shell, `log.sh` is removed. Gitjob containers now use `TerminationMessageFallbackToLogsOnError` to preserve failure diagnostics by surfacing container logs as the termination message on non-zero exit. Their commands now invoke `fleet` directly instead of going through `tini --`.

Also wait for the ChartMuseum health endpoint before uploading charts in the E2E test environment setup to make the process more reliable.

Refers #5407

### DIFF
--- a/package/Dockerfile.agent
+++ b/package/Dockerfile.agent
@@ -1,29 +1,10 @@
 ARG BUILD_ENV=docker
-ARG BUILDPLATFORM
 ARG TARGETPLATFORM
 ARG TARGETARCH
 ARG ARCH=${TARGETARCH}
 
-# renovate: datasource=github-release-attachments depName=krallin/tini
-ARG TINI_VERSION=v0.19.0
-# renovate: datasource=github-release-attachments depName=krallin/tini digestVersion=v0.19.0
-ARG TINI_SUM_amd64=c5b0666b4cb676901f90dfcb37106783c5fe2077b04590973b885950611b30ee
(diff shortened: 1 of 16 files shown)


aruiz14 (approved):

package/Dockerfile.agent
```diff
@@ -19,4 +39,5 @@ COPY ${TARGETPLATFORM}/fleetagent-linux-* /usr/bin/fleetagent
 
 FROM copy_${BUILD_ENV}
 USER 1000
+ENTRYPOINT ["tini", "--"]
```
aruiz14: I'm not completely sure we need this.

aruiz14 (dismissed):

weyfonk (approved):

weyfonk (approved):

thardeck:
I have created a ticket to track this: https://github.com/rancher/fleet/issues/5407

Issue #5306, opened 2026-06-18
Warn users about GitRepoRestriction resource usage


p-se:
Support for GitRepoRestriction resource will soon be deprecated (see https://github.com/rancher/fleet/issues/5303).
In the meantime, Fleet should warn users about such usage, for instance by issuing warning log messages when they are being processed, possibly pointing to the migration guide.

This will be complements by a documentation enhancement (https://fleet.rancher.io/next/how-tos-for-operators/tenant-setup#_migration_from_gitreporestriction) encouraging users to migrate to Policy.

d3flex:
backport to 0v.16 https://github.com/rancher/fleet/pull/5395

d3flex:
# QA Template
## Additional QA
### Issue: #5306
 
### Problem
Gitrepo resource still utilize GitRepoRestrictions, as opposed to helmops which use the `Policy`. In order to drop the underlined resources in the future
release we prepare for the deprecation and we need to info somehow the users.

### Solution
log emmision when the `GitRepoRestrictions` are being applied and at least one is present during gitops reconcilation.

### Testing

### Engineering Testing
#### Manual Testing
## Repro

```yaml
# 1. GitRepoRestriction in the target namespace
apiVersion: fleet.cattle.io/v1alpha1
kind: GitRepoRestriction
metadata:
  name: deprecation-repro
  namespace: fleet-local
allowedServiceAccounts:
  - my-sa
---
# 2. A GitRepo in the same namespace
apiVersion: fleet.cattle.io/v1alpha1
kind: GitRepo
metadata:
  name: repro-repo
  namespace: fleet-local
spec:
  repo: https://github.com/rancher/fleet-examples
  paths: [ simple ]
  serviceAccount: my-sa
```

```bash
kubectl apply -f repro.yaml
kubectl -n cattle-fleet-system logs deploy/gitjob | grep "GitRepoRestriction will be removed"
```

**Expected:** the deprecation line appears once from each of two independent reconcilers in the `gitjob` pod, in sequence:
1. `"controller": "gitrepo"` (GitJobReconciler) reconciles first and logs
2. `"controller": "GitRepoStatus"` (StatusReconciler) reconciles next and logs again


## Test cases

**1. Happy path**
- Setup: ≥1 `GitRepoRestriction` + GitRepo in same ns (repro above)
- Expected: Deprecation `Info` log emitted; GitRepo reconciles normally

**2. No restrictions, no policy**
[...]

PR #5347, opened 2026-06-24
feat(gitrepo): announce GitRepoRestriction deprecation

d3flex:
GitRepoRestriction is being deprecated in favor of Policy. Seems already the case for `helmops`. With this commit we start deprecation on the next version (v2.16.0) pointing to the migration guide and announce removal from v2.17.0.

The log is gated on at least one restriction being present, so users who already moved to Policy are not warned.

Refers to #5306

### DIFF
--- a/internal/cmd/controller/gitops/reconciler/restrictions.go
+++ b/internal/cmd/controller/gitops/reconciler/restrictions.go
@@ -15,6 +15,7 @@ import (
 	fleet "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 // AuthorizeAndAssignDefaults applies restrictions from both GitRepoRestriction and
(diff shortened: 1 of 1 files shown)


internal/cmd/controller/gitops/reconciler/restrictions.go
```diff
@@ -37,6 +38,12 @@ func AuthorizeAndAssignDefaults(ctx context.Context, c client.Client, gitrepo *f
 		return nil
 	}
 
+	if len(restrictions.Items) > 0 {
+		log.FromContext(ctx).Info(
+			"GitRepoRestriction will be removed in v2.17.0 release; All restrictions must migrate to Policy. " +
```
weyfonk: This may not mean anything to Fleet standalone users. How about something like:
```suggestion
			"GitRepoRestriction will be removed in the next minor release; All restrictions must migrate to Policy. " +
```
or even:
```suggestion
			"GitRepoRestriction will be removed in the next minor release (Fleet v0.18.0 / Rancher v2.17.0); All restrictions must migrate to Policy. " +
```

d3flex: Done

internal/cmd/controller/gitops/reconciler/restrictions.go
```diff
@@ -37,6 +38,12 @@ func AuthorizeAndAssignDefaults(ctx context.Context, c client.Client, gitrepo *f
 		return nil
 	}
 
+	if len(restrictions.Items) > 0 {
+		log.FromContext(ctx).Info(
+			"GitRepoRestriction will be removed in the next minor release(Fleet v0.18.0 / Rancher v2.17.0); All restrictions must migrate to Policy. " +
```
weyfonk: nit: missing whitespace?
```suggestion
			"GitRepoRestriction will be removed in the next minor release (Fleet v0.18.0 / Rancher v2.17.0); All restrictions must migrate to Policy. " +
```

weyfonk (approved):
Thanks for this :)

PR #5351, opened 2026-06-25
fix(policy): introduce CreateNamespace option

p-se:
to, by default, disable the need for helm to create namespaces if a serviceAccount is required by Policy. This ensures tenant ServiceAccounts are not forced to hold the cluster-scoped namespaces.create verb, which cannot be scoped to a single tenant boundary.

When requireServiceAccount is true on an aggregated Policy and allowNamespaceCreation is not explicitly set, the controller stamps CreateNamespace=false on BundleDeploymentOptions. This only disables Helm's namespace creation on install; it does not affect namespaceLabels/namespaceAnnotations patching, which is still attempted (when set) as the deployment's ServiceAccount and gated by downstream RBAC.

Operators who have secured namespace creation through other means (ValidatingAdmissionPolicy, Kyverno, etc.) can re-enable it by setting allowNamespaceCreation: true on the Policy.

Refers to #5374

### Checklist

- [x]  [I have updated the documentation via a pull request](https://github.com/rancher/fleet-product-docs/pull/321) in the [fleet-product-docs](https://github.com/rancher/fleet-product-docs/pull/321) repository.

### DIFF
--- a/internal/cmd/agent/controller/bundledeployment_controller.go
+++ b/internal/cmd/agent/controller/bundledeployment_controller.go
@@ -235,36 +235,15 @@ func (r *BundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 
 	// helm deploy the bundledeployment
 	if status, err := r.Deployer.DeployBundle(ctx, bd, forceDeploy); err != nil {
-		// A denied namespace patch leaves the deployment not-ready: the Helm
-		// release installed, but the requested namespaceLabels/
-		// namespaceAnnotations could not be applied. It is handled as a
-		// controlled requeue rather than a failed reconcile so it does not
-		// tight-loop: keep Ready/Installed=false from the returned status, mark
-		// Deployed (the release is installed), persist, and requeue so the patch
-		// converges once the missing namespace RBAC is granted (granting it does
-		// not otherwise trigger a reconcile).
-		var namespaceForbiddenError *deployer.NamespaceForbiddenError
-		if errors.As(err, &namespaceForbiddenError) {
-			bd.Status = setCondition(status, nil, monitor.Cond(fleetv1.BundleDeploymentConditionDeployed))
-			if err := r.updateStatus(ctx, orig, bd); err != nil {
-				return ctrl.Result{}, err
-			}
-			logger.V(1).Info("Namespace patch forbidden, requeuing...", "error", namespaceForbiddenError)
-			return ctrl.Result{RequeueAfter: durations.NamespacePermissionRequeueInterval}, nil
+		if handled, res, err := r.requeueIfNamespaceForbidden(ctx, orig, bd, status, err); handled {
+			return res, err
 		}
 
 		// do not use the returned status, instead set the condition and possibly a timestamp
 		bd.Status = setCondition(bd.Status, err, monitor.Cond(fleetv1.BundleDeploymentConditionDeployed))
 
-		// Not-ready dependencies should not be treated as an error.
-		// Instead, a controlled requeue should happen until the conditions are met.
-		var notReadyDependenciesError *deployer.NotReadyDependenciesError
-		if errors.As(err, &notReadyDependenciesError) {
-			if err := r.updateStatus(ctx, orig, bd); err != nil {
-				return ctrl.Result{}, err
-			}
-			logger.V(1).Info("Dependencies not ready, requeuing...", "pending", notReadyDependenciesError.Pending)
-			return ctrl.Result{RequeueAfter: durations.WaitForDependenciesReadyRequeueInterval}, nil
+		if handled, res, err := r.requeueIfDependenciesNotReady(ctx, orig, bd, err); handled {
+			return res, err
 		}
 
 		logger.V(1).Info("Failed to deploy bundle", "status", status, "error", err)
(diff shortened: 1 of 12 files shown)


internal/cmd/controller/target/builder.go
```diff
@@ -139,6 +146,9 @@ func (m *Manager) Targets(ctx context.Context, bundle *fleet.Bundle, manifestID
 			if namespaceSelector != nil {
 				opts.AllowedTargetNamespaceSelector = namespaceSelector
 			}
+			if createNamespace != nil {
+				opts.CreateNamespace = createNamespace
+			}
```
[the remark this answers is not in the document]
p-se: createNamespace is Policy-controlled, not tenant-controlled and it doesn't make sense to use it without a service account restriction, in which case fleet-agent falls back unrestricted access.

internal/cmd/controller/policyrestrictions/policyrestrictions.go
```diff
@@ -44,6 +45,9 @@ func Aggregate(policies []fleet.Policy) Merged {
 		if p.RequireServiceAccount {
 			m.RequireServiceAccount = true
 		}
+		if p.AllowNamespaceCreation {
+			m.AllowNamespaceCreation = true
+		}
```
weyfonk, line 55: Interesting, so if multiple Policies exist in the same namespace but only one of them allows namespace creation, then it will be allowed. Let's be sure we document this.

internal/cmd/controller/target/builder.go
```diff
@@ -466,3 +476,29 @@ func (m *Manager) getNamespaceSelectorForBundle(ctx context.Context, bundle *fle
...
+		return nil, nil
+	}
+
+	pol := policyrestrictions.Aggregate(policies.Items)
+	if pol.RequireServiceAccount && !pol.AllowNamespaceCreation {
+		v := false
+		return &v, nil
+	}
+
+	return nil, nil
```
weyfonk, line 503: Simple remark: it seems that `createNamespace` is never actually set to `true`, being either `nil` or a pointer to `false`.

internal/cmd/agent/deployer/deployer.go
```diff
@@ -208,6 +208,9 @@ func (d *Deployer) setNamespaceLabelsAndAnnotations(ctx context.Context, bd *fle
 	if bd.Spec.Options.NamespaceLabels == nil && bd.Spec.Options.NamespaceAnnotations == nil {
 		return nil
 	}
+	if bd.Spec.Options.CreateNamespace != nil && !*bd.Spec.Options.CreateNamespace {
```
weyfonk: If I understand this well, `CreateNamespace` evaluating to `false` also means that even on _existing_ namespaces, labels and annotations will not be propagated, as `updateNamespace` will be skipped.
Which makes me wonder:
- if this is intended behaviour
- if we may want to make this more flexible in the future
- if the name of the option (`CreateNamespace`) might be a little misleading. It could perhaps be something like `MutateNamespace` (because `CreateOrUpdateNamespace` feels a bit long 😛).

p-se: Fleet agent always used the cluster-wide access for patching namespaces, so I went with a different implementation where the service account in the bundle is used for patching the namespaces if a service account is provided in Policy or GitRepoRestriction. Makes this additional guard unnecessary and CreateNamespace is true to its name.

internal/cmd/agent/deployer/deployer_test.go
```diff
@@ -155,6 +157,129 @@ func TestSetNamespaceLabelsAndAnnotations(t *testing.T) {
 	}
 }
 
+// TestSetNamespaceLabelsAndAnnotations_CreateNamespaceFalse verifies that
+// disabling Helm namespace creation (CreateNamespace=false) does not prevent
+// Fleet from applying namespaceLabels/namespaceAnnotations to the (already
+// existing) namespace. CreateNamespace only governs creation; mutation is gated
+// by the deployment's service account RBAC, not by this flag.
+func TestSetNamespaceLabelsAndAnnotations_CreateNamespaceFalse(t *testing.T) {
```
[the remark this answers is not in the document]
p-se: Updated.

internal/cmd/agent/deployer/deployer.go
```diff
@@ -266,6 +331,14 @@ const podSecurityLabelPrefix = "pod-security.kubernetes.io/"
...
 // and any existing `pod-security.kubernetes.io/*` labels. Labels with the
 // `pod-security.kubernetes.io/` prefix in optLabels are ignored.
+//
+// This filtering is intentionally unconditional and independent of the
+// service-account impersonation used for the namespace patch: it must also hold
+// for deployments that run as the agent (no service account pinned), where
+// there is no downstream RBAC gating at all. It is the only safeguard against a
+// bundle escalating pod-security enforcement on its target namespace. To set
+// pod-security labels on a namespace, declare them on the Namespace resource in
+// the bundle instead.
```
weyfonk, line 344: Nice!

internal/cmd/agent/deployer/deployer.go
```diff
@@ -235,13 +273,36 @@ func (d *Deployer) setNamespaceLabelsAndAnnotations(ctx context.Context, bd *fle
...
+
+// namespaceClient returns the client to use for namespace label/annotation
+// mutations. When the deployment resolves to a service account (pinned, or the
+// "fleet-default" fallback), it returns a client impersonating that account, so
+// the mutation is authorized against the downstream RBAC of the tenant rather
+// than the agent's cluster-admin credentials. Otherwise it returns the agent
+// client, preserving the previous behaviour.
+func (d *Deployer) namespaceClient(ctx context.Context, bd *fleet.BundleDeployment) (client.Client, error) {
+	if d.helm != nil {
+		c, err := d.helm.ImpersonatedClient(ctx, bd.Spec.Options.ServiceAccount)
```
weyfonk, line 290: The _current_ chain of calls prevents `bd` from being nil (otherwise `bd.Status` would panic in `DeployBundle`), yet I think it would make sense to be defensive here and return an error if `bd` is nil. Our future selves may thank us :D

internal/helmdeployer/impersonate.go
```diff
@@ -11,8 +11,46 @@ import (
...
 
+// ImpersonatedClient returns a controller-runtime client that impersonates the
+// service account resolved for the given deployment, to be used for namespace
+// label/annotation mutations. Performing those mutations as the same identity
+// the deployment runs as ensures they are gated by the same downstream RBAC,
+// rather than by the agent's own (cluster-admin) credentials.
+//
+// Service account resolution is identical to the Helm install path
+// (getServiceAccount): an explicitly pinned account is used as-is, otherwise it
+// falls back to "fleet-default" when present. It returns a nil client only when
```
weyfonk, line 25: Do we have that service account documented anywhere? 🤔 (I mean our docs website, away from our [sources](https://github.com/rancher/fleet/blob/main/internal/helmdeployer/impersonate.go#L20))

internal/cmd/agent/controller/bundledeployment_controller.go
```diff
@@ -523,6 +520,47 @@ func (r *BundleDeploymentReconciler) updateStatus(ctx context.Context, orig *fle
...
+// installed), persist, and requeue so the patch converges once the missing
+// namespace RBAC is granted (granting it does not otherwise trigger a
+// reconcile). Returns handled=false when err is not a NamespaceForbiddenError.
+func (r *BundleDeploymentReconciler) requeueIfNamespaceForbidden(ctx context.Context, orig, bd *fleetv1.BundleDeployment, status fleetv1.BundleDeploymentStatus, err error) (bool, ctrl.Result, error) {
+	var namespaceForbiddenError *deployer.NamespaceForbiddenError
+	if !errors.As(err, &namespaceForbiddenError) {
+		return false, ctrl.Result{}, nil
+	}
+
+	bd.Status = setCondition(status, nil, monitor.Cond(fleetv1.BundleDeploymentConditionDeployed))
```
weyfonk, line 537: Why do we need this line? Is it a reset prior to calling `updateStatus`?

p-se: It sets the BundleDeployment condition `Deployed` to `true`. This doesn't make it ready and that's deliberate, because patching the namespace failed. But in this path it means it has actually been already deployed, so we make the status reflect that. It mirrors the success path in the reconcile in line 253. A forbidden namespace patch is treated like a successful deployment rather than like a failure, but it does requeue because the RBAC can only externally be resolved and the reconciler doesn't watch it.

internal/helmdeployer/impersonate.go
p-se: Underrepresented but existent. I'll make sure the docs update covers this as well.

internal/cmd/agent/controller/bundledeployment_controller.go
weyfonk: Oh, good to know, thanks for clarifying.

weyfonk (approved):

PR #5365, opened 2026-06-29
GC incorrectly uninstalls helm releases when BundleDeployment name exceeds 53 chars

0xavi0:
- `releaseKey()` in the garbage collector now applies `names.HelmReleaseName()` when no explicit `helm.releaseName` is set, mirroring what `helmdeployer.getOpts()` does at install time
- Without this, names >53 chars produced different strings in the two paths (raw vs truncated+hashed), causing the GC to treat a valid release as orphaned and uninstall it on every reconcile loop

Refers to: https://github.com/rancher/fleet/issues/5261

### Checklist


0xavi0:
closing this as there was already a PR open to fix the same :/

PR #5316, opened 2026-06-19
Warn when Helm credentials are set but helmRepoURLRegex is empty

thardeck:
When Helm credentials are configured in an `Auth` object but `helmRepoURLRegex` is empty, `addRemoteCharts` silently drops the credentials. Log a Warn-level message so operators know credentials are not forwarded and how to fix it by setting `spec.helmRepoURLRegex`.

PR #5367, opened 2026-06-29
Surface HelmRepoURLRegex hint in bundle errors

thardeck:
Track when Helm credentials are stripped because `helmRepoURLRegex` is empty and carry that marker on the chart directory metadata.

Append an explicit guidance hint to content-loading errors for those directories so the status message shown in Rancher UI explains how to fix the configuration.

Follow up to #5316

### DIFF
--- a/internal/bundlereader/loaddirectory_internal_test.go
+++ b/internal/bundlereader/loaddirectory_internal_test.go
@@ -14,25 +14,30 @@ import (
 func TestMaybeAddHelmRepoURLRegexHint(t *testing.T) {
 	t.Run("adds hint when credentials were stripped for empty regex", func(t *testing.T) {
 		err := errors.New("unauthorized")
-		got := maybeAddHelmRepoURLRegexHint(err, directory{strippedCredentialsForEmptyHelmRepoURLRegex: true})
+		got := maybeAddHelmRepoURLRegexHint(err, directory{strippedCreds: true})
 		require.Error(t, got)
 		require.ErrorIs(t, got, err)
 		assert.Contains(t, got.Error(), helmRepoURLRegexUIHint)
 	})
 
+	t.Run("returns nil for nil error", func(t *testing.T) {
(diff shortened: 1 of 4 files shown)


internal/bundlereader/resources.go
```diff
@@ -148,6 +148,8 @@ type directory struct {
 	version string
 	// auth is the auth to use for the chart URL
 	auth Auth
+	// indicates auth was stripped because helmRepoURLRegex was empty
+	strippedCredentialsForEmptyHelmRepoURLRegex bool
```
weyfonk: nit: this variable name is... long.
Should the reason for stripping the creds be featured in the variable name? Or would the comment be enough? In the latter case, we could shorten this as e.g. `strippedCreds`.

internal/bundlereader/loaddirectory_internal_test.go
```diff
@@ -1,14 +1,37 @@
...
+		assert.Contains(t, got.Error(), helmRepoURLRegexUIHint)
+	})
+
+	t.Run("keeps original error when marker is not set", func(t *testing.T) {
+		err := errors.New("unauthorized")
+		got := maybeAddHelmRepoURLRegexHint(err, directory{strippedCredentialsForEmptyHelmRepoURLRegex: false})
+		assert.Equal(t, err, got)
+	})
+
+	t.Run("keeps context cancellation errors unchanged", func(t *testing.T) {
```
weyfonk, line 34: We could cover the `nil` error case here as well.

weyfonk (approved):
LGTM, and I'm wondering: has this been tested with the Rancher UI?

thardeck:
> has this been tested with the Rancher UI

Not yet, that's why I have added Manuel in CC on Slack.

thardeck:
Verified in Rancher ui:
image

weyfonk (approved):

PR #5369, opened 2026-06-29
Fix false-positive diffs for controller-injected webhook caBundle

susesamu:
issue: https://github.com/rancher/fleet/issues/5401


internal/cmd/agent/deployer/desiredset/diff.go
```diff
@@ -234,16 +234,273 @@ func newNormalizers(live objectset.ObjectByGVK, bd *fleet.BundleDeployment) (dif
...
+	_, desiredHasCABundle := desiredCC["caBundle"]
+	return !desiredHasCABundle
+}
+
+func filterEmptyWebhooks(webhooksList []any) []any {
+	filteredWebhooks := make([]any, 0, len(webhooksList))
+	for _, webhook := range webhooksList {
+		webhookMap, ok := webhook.(map[string]any)
+		if !ok {
+			filteredWebhooks = append(filteredWebhooks, webhook)
```
weyfonk, line 439: Why add `webhook` to filtered webhooks in this case? This could perhaps benefit from a comment.

internal/cmd/agent/deployer/desiredset/diff.go
```diff
@@ -234,16 +234,273 @@ func newNormalizers(live objectset.ObjectByGVK, bd *fleet.BundleDeployment) (dif
...
+		return false, nil
+	}
+
+	// Check if this is a nested patch or wholesale replacement
+	if !hasNestedCABundleKey(webhooksList) {
+		return handleWholesaleReplacement(actual, desiredWebhooks, patchData, patch)
+	}
+
+	// Process nested caBundle keys
+	return processNestedCABundles(webhooksList, desiredWebhooks, patchData, patch)
```
weyfonk: nit: this would read more easily to me this way:
```suggestion
	// Check if this is a nested patch or wholesale replacement
	if hasNestedCABundleKey(webhooksList) {
		// Process nested caBundle keys
		return processNestedCABundles(webhooksList, desiredWebhooks, patchData, patch)
	}

	return handleWholesaleReplacement(actual, desiredWebhooks, patchData, patch)
```

internal/cmd/agent/deployer/desiredset/diff.go
```diff
@@ -234,16 +234,273 @@ func newNormalizers(live objectset.ObjectByGVK, bd *fleet.BundleDeployment) (dif
...
+	return false
+}
+
+func handleWholesaleReplacement(
+	actual *unstructured.Unstructured,
+	desiredWebhooks []any,
+	patchData map[string]any,
+	patch *[]byte,
+) (bool, error) {
+	actualWebhooks, _, _ := unstructured.NestedSlice(actual.Object, "webhooks")
```
weyfonk: We may want to catch any errors from this call, along with the boolean indicating whether webhooks are found (as already done in `getDesiredWebhooks`).

internal/cmd/agent/deployer/desiredset/diff_test.go
```diff
@@ -746,3 +746,397 @@ func hpa(
...
+							"namespace": ns,
+						},
+						// No caBundle
+					},
+				},
+			},
+		},
+	}
+
+	live := desired.DeepCopy()
```
weyfonk: nit:
```suggestion
    // Simulate CA bundle injection into a live MutatingWebhookConfiguration by a controller
	live := desired.DeepCopy()
```

internal/cmd/agent/deployer/desiredset/diff_test.go
```diff
@@ -746,3 +746,397 @@ func hpa(
 		},
 	}
 }
+
+func TestDiff_WebhookCABundle_ControllerInjected(t *testing.T) {
```
weyfonk: These new test cases pass even when commenting out the call to `normalizeWebhookCABundlePatch` 🤔
This makes me think that they may not be validating what we expect them to.

internal/cmd/agent/deployer/desiredset/diff.go
susesamu: it's just a defensive handler to avoid any data loss. I added a explanatory comment. ty!

internal/cmd/agent/deployer/desiredset/diff_test.go
susesamu: thanks a lot for catching that,the Argo diff normalizers (from the vendored code) run before the normalization and already handle webhook differences in those cases I tested. I replicated the customer case and changed the tests.

internal/cmd/agent/deployer/desiredset/webhook_test.go
```diff
@@ -0,0 +1,333 @@
...
+	emptied, err := normalizeWebhookCABundlePatch(desired, actual, &patch)
+	if err != nil {
+		t.Fatalf("normalizeWebhookCABundlePatch failed: %v", err)
+	}
+
+	// Patch should be emptied since only the controller-injected caBundle differs
+	if !emptied {
+		t.Errorf("Expected nested patch to be emptied, but got: %s", string(patch))
+	}
+}
```
weyfonk, line 333: We could add a test case for diffs not related to validating nor mutated webhooks, which should then be unaffected even if they differ by, say, a CA bundle.

weyfonk (approved):

0xavi0:
As this one targets `main` (0.17 now), we need to backport to `v2.15.1` and `v2.14.5`

PR #5394, opened 2026-07-03
feat(gitrepo): announce GitRepoRestriction deprecation

d3flex:
GitRepoRestriction is being deprecated in favor of Policy. Seems already the case for `helmops`. With this commit we start deprecation on the next version (v2.16.0) pointing to the migration guide and announce removal from v2.17.0.

The log is gated on at least one restriction being present, so users who already moved to Policy are not warned.

Refers to #5306

(cherry picked from commit 7ea7b79edb3b8518217e9fe24c4d90615d5b523c)


d3flex:
wrong branch again

PR #5396, opened 2026-07-03
Split pkg/helmvalues into its own Go module

0xavi0:
Promote pkg/helmvalues to a standalone module (like pkg/apis) so it can be released and consumed independently by rancher/rancher.

Also adds check-module-tag-sync.sh + workflow to verify pkg/apis and pkg/helmvalues are tagged in lockstep (same version, same commit)

### Checklist


pkg/helmvalues/go.mod
```diff
@@ -0,0 +1,31 @@
+module github.com/rancher/fleet/pkg/helmvalues
+
+go 1.26.0
+
+toolchain go1.26.4
+
+replace github.com/rancher/fleet/pkg/apis => ../apis
+
+require github.com/rancher/fleet/pkg/apis v0.15.2
```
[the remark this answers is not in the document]
0xavi0: Replace directives are ignored for non-main modules, so the released go.mod's require github.com/rancher/fleet/pkg/apis vX is what downstream consumers use — the ../apis replace only applies to local builds inside this repo.

.github/scripts/release-against-rancher.sh
```diff
@@ -68,15 +69,22 @@ git add build.yaml pkg/buildconfig/constants.go
...
     popd > /dev/null
 
     git add go.mod go.sum pkg/apis/go.mod pkg/apis/go.sum
 fi
 
+# Bump the Fleet helmvalues module when a pkg/helmvalues tag for this exact version
+# exists in the fleet repo.
+if git -C ../fleet tag -l "pkg/helmvalues/v${NEW_FLEET_VERSION}" | grep -q .; then
+    bump_fleet_module pkg/helmvalues
+    git add go.mod go.sum
```
weyfonk: Why not add `pkg/helmvalues/go.{mod,sum}` as well?

.github/workflows/release-against-rancher.yml
```diff
@@ -94,6 +96,21 @@ jobs:
             fi
           fi
 
+          # Same guard for the pkg/helmvalues module. Rancher may reference it from a different
+          # go.mod than pkg/apis, so scan all of rancher's module files for a require line.
+          # -o restricts the match to "<module> <version>" so a trailing " // indirect" is
+          # excluded; otherwise awk would read "indirect" as the version and drop the match.
+          if ! git -C fleet tag -l "pkg/helmvalues/v${NEW_FLEET}" | grep -q .; then
+            RANCHER_FLEET_HELMVALUES_VERSION=$(grep -rhsoE 'github.com/rancher/fleet/pkg/helmvalues[[:space:]]+v[0-9][^[:space:]]*' rancher --include=go.mod | awk '{print $NF}' | grep -- - | head -n1 || true)
```
weyfonk: This will only work if `pkg/helmvalues` is referenced from a _single_ `go.mod` file (`head -n1`). Not sure if that is intended.

.github/scripts/check-module-tag-sync.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/bash
+#
+# Verify that the pkg/apis and pkg/helmvalues modules are released in lockstep.
+#
+# pkg/helmvalues depends on pkg/apis and both are tagged with the same version
```
weyfonk, line 5: Why could `pkg/helmvalues` not be updated even without `pkg/apis` being bumped?
Is this a convention for the sake of simplicity (e.g. not referencing multiple Fleet tags from Rancher)?

.github/scripts/release-against-rancher.sh
0xavi0: Yup, I've also changed `bump_fleet_module` so it discovers every go.mod in rancher that requires the module and, for each, runs go get/go mod tidy and git add "${moddir}/go.mod" "${moddir}/go.sum". So helmvalues' (and any nested module's) go.{mod,sum} are now staged.

.github/scripts/check-module-tag-sync.sh
0xavi0: I thought it was easier to maintain.
The new module is pretty simple and I thought that it was simpler to have the main Fleet and this one to be tagged in sync. Later when importing it in Rancher is also easier to maintain. And for releasing it's also easier.
Happy to talk about it :)

.github/workflows/release-against-rancher.yml
0xavi0: Changed now, as described earlier

.github/scripts/check-module-tag-sync.sh
weyfonk: Right, so it's for the sake of simplicity. No issue against that, I also think it makes sense :)

.github/workflows/release-against-rancher.yml
```diff
@@ -94,6 +96,24 @@ jobs:
...
           fi
 
+          # Same guard for the pkg/helmvalues module. Rancher may reference it from a different
+          # go.mod than pkg/apis, and potentially from more than one, so scan all of rancher's
+          # module files and consider every match (not just the first) to avoid missing a
+          # pre-release reference in a later file.
+          # -o restricts the match to "<module> <version>" so a trailing " // indirect" is
+          # excluded; otherwise awk would read "indirect" as the version and drop the match.
+          if ! git -C fleet tag -l "pkg/helmvalues/v${NEW_FLEET}" | grep -q .; then
+            RANCHER_FLEET_HELMVALUES_PRERELEASES=$(grep -rhsoE 'github.com/rancher/fleet/pkg/helmvalues[[:space:]]+v[0-9][^[:space:]]*' rancher --include=go.mod | awk '{print $NF}' | grep -- - | sort -u || true)
```
weyfonk, line 106: I don't yet understand why any version found here would be a pre-release version 🤔

weyfonk (approved):

.github/workflows/release-against-rancher.yml
weyfonk: Nice, thanks!

.github/workflows/release-against-rancher.yml
0xavi0: Because during development rancher references it via a Go pseudo-version (for example, v0.15.4-0.20260115…-abcdef), which is a semver pre-release, so grep -- - matches it — same reasoning as the pkg/apis guard just above, but in this case I changed the variable name to have the PRERELEASES suffix.

weyfonk: Ack, thanks for clarifying :bow:

Issue #4813, opened 2026-03-13
(Visual) feedback on upgrade status when upgrading between Rancher minor versions


kkaempf:
### Is your feature request related to a problem?

https://github.com/rancher/fleet/issues/4131#issuecomment-4052790360 is just the latest of several reports, where Rancher minor upgrades have downstream fleet-agent stuck in an old version.

While the [upgrade checklist](https://support.scc.suse.com/s/kb/360051866152?language=en_US) explicitly states that adding a grace period is recommended, we are lacking to give users (visual) feedback about version inconsistencies between upstream and downstream.

### Solution you'd like

Ultimately we need to find the root cause why upgrades sometime get stuck downstream.
Until then, we should improve the feedback we give to users about the state of an ongoing upgrade, like
* a cli command to see "where's my cluster on its upgrade journey"
* a UI status / event to visualize upgrade steps
* a UI status message to indicate when downstream clusters (resp. their fleet-agent) is stuck at an old version
* ...

kkaempf:
Another interesting insight: https://github.com/rancher/fleet/issues/4131#issuecomment-4061652942

p-se:
# QA Template – Agent version reporting and `AgentVersionUpToDate` condition

Issue [#4813](https://github.com/rancher/fleet/issues/4813), PR [#5404](https://github.com/rancher/fleet/pull/5404), first shipped in **fleet 0.17.0-alpha.1**. The previous release without the feature is **0.16.1**.

The fleet-agent now reports its own binary version in `Cluster.status.agent.version`. The controller compares it against the **controller's own binary version** (not the configured agent image tag) and sets an `AgentVersionUpToDate` condition on the `Cluster`. The condition is **informational only** and must never affect cluster readiness.

Upstream plus at least one downstream cluster.

Helper used throughout:

```bash
cond() { kubectl -n "$1" get cluster "$2" -o jsonpath\
='{range .status.conditions[?(@.type=="AgentVersionUpToDate")]}{.status} [{.reason}] {.message}{"\n"}{end}'; }
```

## T1 – Agents report their version

```bash
kubectl get cluster -A -o custom-columns=NS:.metadata.namespace,NAME:.metadata.name,AGENT:.status.agent.version
```

**Pass:** every cluster whose agent has checked in shows `v0.17.0-alpha.1`. Note the leading `v`: the chart version has none, the reported binary version does. This includes `fleet-local/local`.

## T2 – Matching versions give `True`

```bash
[...]

PR #5404, opened 2026-07-06
feat(agent): report agent version and surface version mismatch

p-se:
## report agent version and surface version mismatch as a cluster condition

The agent reports its binary version in AgentStatus on every check-in. The controller compares it against its own binary version and sets an AgentVersionUpToDate condition on the Cluster resource, making version drift visible without having to inspect pods directly. The controller and agent are built and released together, so the controller's own version is the version agents are expected to run, independent of the configured agent image (which may be a moving tag, a custom image, or otherwise not reflect the actual released version).

Versions that cannot be compared as semver (unstamped "dev" builds) result in Unknown. An agent that has checked in but reports no version is marked outdated (False): every version-reporting agent writes its version alongside LastSeen, so an empty version can only mean the agent predates version reporting. An agent that has never checked in stays Unknown. Detecting an agent that checked in before but has since gone silent is the job of the WaitCheckIn state in the cluster reconciler, not this write-triggered handler.

The condition is informational only. It is legitimately False for a short window during a rolling upgrade while downstream agents catch up to the controller, so it must not feed cluster readiness (Ready is derived solely from the bundle summary), and its Reason is deliberately left unset to keep Rancher's condition summarizer from surfacing routine version skew as a cluster error.

Refers to #4813

### Checklist

### DIFF
--- a/internal/cmd/controller/agentmanagement/controllers/manageagent/manageagent.go
+++ b/internal/cmd/controller/agentmanagement/controllers/manageagent/manageagent.go
@@ -14,6 +14,7 @@ import (
 	"fmt"
 	"sort"
 
+	"github.com/Masterminds/semver/v3"
 	"github.com/sirupsen/logrus"
 
 	"github.com/rancher/fleet/internal/cmd"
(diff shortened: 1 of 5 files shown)


internal/cmd/controller/agentmanagement/controllers/manageagent/manageagent.go
```diff
@@ -143,9 +146,60 @@ func (h *handler) onClusterStatusChange(cluster *fleet.Cluster, status fleet.Clu
 		h.namespaces.Enqueue(cluster.Namespace)
 	}
 
+	setAgentVersionCondition(&status, version.Version)
```
d3flex, line 148: This will run only when the `SkipCluster(cluster) || cluster.Labels[fleet.LocalAgentDisabledLabel] == "true"`. is there a race condition or inconsistency on `status` if it returns early? Have you tested this scenario?

p-se: It runs only if the cluster is _not_ skipped. If it is skipped, the status then isn't updated. That is consistent with how updating the status for the cluster is also skipped as well as the agent environment variables.

d3flex (approved):

PR #5405, opened 2026-07-06
agentmanagement: config and cluster testing coverage.

0xavi0:
This PR adds testing coverage for the config and cluster controllers in the agentmanagement.

### DIFF
--- a/integrationtests/agentmanagement/cluster_test.go
+++ b/integrationtests/agentmanagement/cluster_test.go
@@ -112,16 +112,28 @@ var _ = Describe("cluster namespace lifecycle", func() {
 	})
 
 	Describe("bundledeployment changes", func() {
-		It("does not error or create any namespace when a BundleDeployment changes in a namespace without cluster annotations", func() {
+		It("does not re-create a deleted cluster namespace when a BundleDeployment changes in a namespace without cluster annotations", func() {
+			clusterName := "bd-no-requeue-cluster"
+			expectedNS := clusterNamespaceName(regNamespace, clusterName)
+
+			// Stand up a cluster and let its generated namespace appear, so
+			// that a spurious enqueue of the cluster would have an observable
+			// effect: the namespace coming back.
+			cluster := newCluster(regNamespace, clusterName)
+			Expect(k8sClient.Create(ctx, cluster)).To(Succeed())
+			namespaceExists(expectedNS).Should(Succeed())
+
+			// Drop the generated namespace, so that only a further reconcile of
+			// the cluster can bring it back. Deleting it does not enqueue the
+			// cluster.
+			deleteNamespace(expectedNS)
+			namespaceIsGone(expectedNS).Should(Succeed())
+
+			// The BundleDeployment lives in a namespace without cluster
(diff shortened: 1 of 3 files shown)


integrationtests/agentmanagement/config_test.go
```diff
@@ -0,0 +1,105 @@
...
+	It("ignores a ConfigMap with the right name in the wrong namespace", func() {
+		cfg := config.DefaultConfig()
+		cfg.AgentImage = "rancher/fleet-agent:should-be-ignored-namespace"
+
+		Expect(k8sClient.Create(ctx, newConfigMap("default", config.ManagerConfigName, cfg))).To(Succeed())
+
+		Consistently(func(g Gomega) {
+			g.Expect(config.Get().AgentImage).To(Equal("rancher/fleet-agent:config-test-updated"))
+		}).Should(Succeed())
+	})
```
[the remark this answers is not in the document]
0xavi0: The test does not fail.

integrationtests/agentmanagement/cluster_test.go
```diff
@@ -0,0 +1,164 @@
...
+			g.Expect(apierrors.IsNotFound(err)).To(BeTrue())
+		}).Should(Succeed())
+
+		Consistently(func(g Gomega) {
+			g.Expect(k8sClient.Get(ctx, types.NamespacedName{Name: expectedNS}, &corev1.Namespace{})).To(Succeed())
+		}).Should(Succeed())
+	})
+
+	Describe("bundledeployment changes", func() {
+		It("does not error or create any namespace when a BundleDeployment changes in a namespace without cluster annotations", func() {
```
weyfonk: Could we check that the opposite also works? i.e. a namespace _may_ be created when a bundle deployment changes in a namespace with cluster annotations.

0xavi0: Good catch!. Added a test for the positive branch — and it exposed that my existing annotated-namespace spec couldn't fail anyway: the cluster namespace already exists when the BundleDeployment is created, so the watch firing has no observable effect.

weyfonk: Thanks for adding this new test case :)

>  and it exposed that my existing annotated-namespace spec couldn't fail anyway: the cluster namespace already exists when the BundleDeployment is created, so the watch firing has no observable effect.

Does this mean that the negative test case should delete the cluster namespace, as the positive test case does, to ensure that the watch is fired?

0xavi0: I've changed the negative test case to mirror the positive case: create the cluster, wait for its namespace, delete it, then create the BundleDeployment in an unannotated namespace and assert the namespace stays gone. Now if findClusters wrongly enqueued the cluster, it'd reappear and fail.

weyfonk (approved):

integrationtests/agentmanagement/cluster_test.go
weyfonk: Nice, that's clearer, thanks :)

PR #5411, opened 2026-07-07
ci: pin k3s to v1.35.6 for Rancher E2E test

d3flex:
Downgraded the k3s reference for Test Fleet in Rancher as the latest kubectl version is incompatible with Kubernetes v1.36.0+k3s1. This sort of reverts part of the d5566e86 and is a workaround for now to make the test work.


thardeck (approved):
I would change the title to E2E test but besides that lgtm.

thardeck (approved):

PR #5412, opened 2026-07-07
Use port forwarding for fetching metrics from benchmark tests

weyfonk:
Spawning an additional pod to run `curl` and extracting its logs is brittle. Instead, benchmark tests now extract metrics through port forwarding, as already done by the `fleet dump` command.
This should fix workflow failures such as [this one](https://github.com/rancher/fleet/actions/runs/28846455661/job/85552154941).

Follow-up to #5296.

### Checklist

### DIFF
--- a/benchmarks/record/record.go
+++ b/benchmarks/record/record.go
@@ -7,8 +7,11 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"io"
 	"maps"
 	"math/rand/v2"
+	"net/http"
+	"os/exec"
 	"runtime"
 	"slices"
 	"strings"
@@ -208,9 +211,9 @@ func Header(s string) string {
 func Metrics(experiment *gm.Experiment, suffix string) {
 	res := map[string]float64{}
 
-	getMetrics(res, "monitoring-fleet-controller.cattle-fleet-system.svc.cluster.local:8080/metrics", "bundle", "bundledeployment", "cluster", "clustergroup", "imagescan")
+	getMetrics(res, "monitoring-fleet-controller", "cattle-fleet-system", 8080, "bundle", "bundledeployment", "cluster", "clustergroup", "imagescan")
 
-	getMetrics(res, "monitoring-gitjob.cattle-fleet-system.svc.cluster.local:8081/metrics", "GitRepoStatus", "gitrepo")
+	getMetrics(res, "monitoring-gitjob", "cattle-fleet-system", 8081, "GitRepoStatus", "gitrepo")
 
 	for k, v := range res {
 		n := k + suffix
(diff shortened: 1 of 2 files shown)


benchmarks/record/record.go
```diff
@@ -251,45 +254,52 @@ var requiredMetricFamilies = []string{
...
-		// attach mechanism, which can duplicate output on fallback to logs.
-		_, _, err := k.RunStdout("run", "--restart=Never", pod, "--image=curlimages/curl", "--namespace", "cattle-fleet-system", "--command", "--", "curl", "-sf", url)
-		if err != nil {
-			return fmt.Errorf("kubectl run: %w", err)
-		}
+	ctx, cancel := context.WithCancel(context.Background())
+	Eventually(func(g Gomega) {
+		// Note on the `nolint: gosec` comment below: We are looking for an available port number; this can afford to be
+		// fairly predictable.
+		hostPort = port + rand.IntN(57535) // Highest possible port: 65534
```
d3flex: This doesnt really ensure the claim in the comments. Taking a port 8080 + 57535 will return 65615. I would like to suggest two approaches:

`port + rand.IntN(65535-port+1)` or using modulo (this is my favorite but maybe not the simplest) 

nit: three are a few magic numbers; would you like to use constants for those? (ex 57535)

weyfonk: Oh, thanks for raising this; I'd copied this logic over from `fleet dump`, and as you'd noticed the magic numbers did not quite apply to this case.
Replaced with this, which I think is a bit clearer:
```
hostPort = port + rand.IntN(65535-port) // TCP port range: 0-65535
```

d3flex (approved):

benchmarks/record/record.go
```diff
@@ -251,45 +250,56 @@ var requiredMetricFamilies = []string{
...
+		defer resp.Body.Close()
 
-		out, _, err := k.RunStdout("logs", "--namespace", "cattle-fleet-system", pod)
+		body, err := io.ReadAll(resp.Body)
 		if err != nil {
 			return err
 		}
 
-		mfs, err = parser.TextToMetricFamilies(bytes.NewBufferString(out))
+		defer cancel()
```
p-se: Do we want the port forwarding to close after we have "successfully" retrieved (resp.StatusCode != http.StatusOK) the metrics once? It won't work for anymore for the checks below, in case we were still waiting for metric families to appear that haven't been found.

benchmarks/record/record.go
```diff
@@ -251,45 +250,56 @@ var requiredMetricFamilies = []string{
...
+
+	ctx, cancel := context.WithCancel(context.Background())
+	Eventually(func(g Gomega) {
+		// Note on the `nolint: gosec` comment below: We are looking for an available port number; this can afford to be
+		// fairly predictable.
+		hostPort = port + rand.IntN(65535-port) // TCP port range: 0-65535
+
+		cmd := exec.CommandContext(ctx, "kubectl", "-n", namespace, "port-forward", "service/"+name, fmt.Sprintf("%d:%d", hostPort, port))
+
+		err := cmd.Start()
```
p-se: cmd.Start() doesn't wait for the command to complete. So if it fails to bind the port, it still returns err=nil.

```go
package main

import (
	"fmt"
	"os/exec"
)

func main() {
	// A command that starts fine but fails at runtime (exit code 3).
	// Stand-in for "kubectl port-forward that can't bind the port".
	cmd := exec.Command("sh", "-c", "exit 3")

	fmt.Printf("Start() error: %v\n", cmd.Start()) // <nil> — looks like success
	fmt.Printf("Wait()  error: %v\n", cmd.Wait())  // exit status 3 — the real result
}

/*
Start() error: <nil>
Wait()  error: exit status 3
*/
```

benchmarks/record/record.go
```diff
@@ -251,45 +250,56 @@ var requiredMetricFamilies = []string{
...
 		if err != nil {
-			return fmt.Errorf("kubectl run: %w", err)
+			return err
 		}
 
-		// Wait for the pod to terminate. Use condition=Ready=false to
-		// detect both Succeeded and Failed without blocking for the full
-		// timeout on a failed curl.
-		_, _, err = k.RunStdout("wait", "--for=condition=Ready=false", "--namespace", "cattle-fleet-system", "pod/"+pod, "--timeout=30s")
+		cli := &http.Client{}
```
p-se: Not having a timeout on http.Client could potentially result in a hanging test.

benchmarks/record/record.go
```diff
@@ -34,15 +35,13 @@ import (
 )
 
 var (
-	k         kubectl.Command
 	k8sClient client.Client
 	workspace string
 )
 
 func Setup(w string, k8s client.Client, kcmd kubectl.Command) {
```
p-se: kcmd is unused.

benchmarks/record/record.go
weyfonk: Indeed; the challenge here is that port forwarding does not return once ports have been forwarded.
One way to work around this would be to duplicate port forwarding code from `fleet dump` instead of calling `kubectl`, to have better control over the port forwarding process.

benchmarks/record/record.go
weyfonk: Good point; added a timeout, thanks.

benchmarks/record/record.go
weyfonk: Replaced with a blank identifier, leaving the function signature unaltered for compatibility.

benchmarks/record/record.go
weyfonk: Thanks for catching this; moved the `defer cancel()` outside of the `Eventually` block, which should ensure cancellation only happens once either that block has timed out, or the full family of metrics has been extracted.

benchmarks/record/record.go
weyfonk: I have implemented an alternative to this, which uses a `net.Listener` to check if a given random port is open. A port will only be used for port forwarding once we know that it can be successfully listened on.
This should reduce the risk of errors going undetected.
I have tested this by temporarily using a hard-coded forwarding port number (instead of computing it randomly) and listening on that port from another program, which revealed failures to listen on the (busy) port, confirming that this mechanism works.

Happy to discuss ;)

p-se (approved):

Issue #5414, opened 2026-07-08
[backport v2.15] Add a `CreateNamespace` option to `Policy`


rancherbot:
This is a backport issue for #5374, automatically created via [GitHub Actions workflow](https://github.com/rancher/fleet/actions/runs/28950005917) initiated by @p-se

Original issue body:

## Summary

Add a `CreateNamespace` option to `BundleDeploymentOptions` so that Policy can control whether Helm creates the target namespace during installs.

## Motivation

When a Policy requires a ServiceAccount for deployments, namespace creation should be opt-in rather than automatic. Today there is no way to disable Helm's namespace creation behavior on a per-deployment basis.

weyfonk:
## QA Template

https://gist.github.com/p-se/d11b23fd8f3139e7fa23d5e2d89bf9f3

sbulage:
## QA Verification — PASS

Verified against the QA template ([gist d11b23fd…](https://gist.github.com/p-se/d11b23fd8f3139e7fa23d5e2d89bf9f3)).

**Result: 8/8 scenarios pass, 39/39 assertions, 0 failures.**

| | |
| --- | --- |
| Rancher | `v2.15-22606a1ec939f9b2db95fba5b3894e3ee52abd96-head` |
| Fleet | `110.0.1+up0.16.1-beta.2` → `rancher/fleet:v0.16.1-beta.2`, `rancher/fleet-agent:v0.16.1-beta.2` |
| Setup | single-node k3s upstream; everything runs in `fleet-local` against the `local` cluster — agent ns `cattle-fleet-local-system`, target ns `project1-workloads` |
| ServiceAccounts | `tenant-1-deployer` (pinned via `spec.serviceAccount`, scenarios B–G) and `fleet-default` (the fallback, scenario H) — **both in the agent ns `cattle-fleet-local-system`** |

---

### Scenarios

#### What each scenario performs

- **A** — nothing breaks for existing users; the namespace is still created
- **B** — a tenant can deploy successfully while never holding `namespaces.create`
- **C** — Fleet genuinely refuses; the namespace does not appear
- **D** — the opt-in flag restores creation
- **E** — a missing grant is actionable and recovers on its own
- **F** — a tenant cannot escalate pod security through `namespaceLabels`
- **G** — Policy aggregation is OR; one permissive Policy wins for the namespace
- **H** — the namespace patch runs as `fleet-default`, not as cluster-admin

| # | Scenario | Result | Evidence |
[...]

PR #5415, opened 2026-07-08
[v0.16.1] fix(policy): introduce CreateNamespace option (#5351)

p-se:
* fix(policy): introduce CreateNamespace option

to, by default, disable the need for helm to create namespaces if a serviceAccount is required by Policy. This ensures tenant ServiceAccounts are not forced to hold the cluster-scoped namespaces.create verb, which cannot be scoped to a single tenant boundary.

When requireServiceAccount is true on an aggregated Policy and allowNamespaceCreation is not explicitly set, the controller stamps CreateNamespace=false on BundleDeploymentOptions. Namespace label/annotation patching is also skipped in this case.

Operators who have secured namespace creation through other means (ValidatingAdmissionPolicy, Kyverno, etc.) can re-enable it by setting allowNamespaceCreation: true on the Policy.

* fix(policy): document CreateNamespace side-effects, add test coverage

Clarify that CreateNamespace also controls namespace label/annotation patching and add a unit test asserting no namespace GET/UPDATE when CreateNamespace is false.

* fix(policy): patch target namespace as the deployment's service account

Namespace label/annotation mutation previously ran on the agent's own (cluster-admin) client, while only the Helm workload ran under the impersonated service account. This left a tenant-isolation gap: a bundle could make the privileged agent add/remove arbitrary namespace labels and annotations on its target namespace.

Patch the namespace through a client that impersonates the deployment's service account (pinned, or the fleet-default fallback), so the get/update is gated by the same downstream RBAC as the deployment itself. When no service account resolves, fall back to the agent client, preserving the previous behaviour.

Drop the CreateNamespace=false short-circuit so CreateNamespace governs only Helm's namespace creation, not label/annotation mutation. Surface permission failures as Ready/Installed status conditions via forbiddenToStatus instead of requeuing forever, with an actionable, IsForbidden-detectable error. The unconditional pod-security label filter is kept as the sole safeguard on the no-service-account path.

Document that Policy AllowNamespaceCreation aggregates with least-restrictive OR semantics across Policies in a namespace.

* fix(policy): requeue on namespace permission errors so they converge

A denied namespace label/annotation patch was recorded as a status condition and the reconcile then stopped. Because the missing downstream RBAC is not a watched resource and the agent's BundleDeployment controller filters out the periodic cache resync, granting the permission would never re-trigger a reconcile: the deployment stayed Ready=false indefinitely even after the operator fixed the RBAC.

Return a typed NamespaceForbiddenError from DeployBundle and handle it in the controller like NotReadyDependenciesError: keep the Ready/Installed=false conditions (the Helm release installed, so Deployed stays true), persist the status, and requeue every NamespacePermissionRequeueInterval (2m). The patch then converges within ~2m of the RBAC grant, without tight-looping or emitting reconcile-error metrics. The error unwraps to the underlying Forbidden error so apierrors.IsForbidden still reports true.

* fix(policy): correct forbiddenToStatus comment to match requeue flow

The comment claimed namespace permission errors are surfaced "instead of looping," but the caller wraps them in NamespaceForbiddenError so the controller does a controlled requeue that converges once the missing RBAC is granted. Describe the actual behaviour.

* fix(policy): satisfy golangci-lint on namespace permission changes

Extract the namespace-forbidden and not-ready-dependencies handling from Reconcile into requeueIfNamespaceForbidden and requeueIfDependenciesNotReady helpers, keeping the function under the funlen statement limit. Also use errors.New instead of fmt.Errorf for the static error string in the tests (perfsprint). No behaviour change.

* fix(policy): Fix typo

* nil guard

(cherry picked from commit a8aaf14768d5e6cbdcba7d69da93647d9e1e4c5c)

Refers to #5414

### Checklist

### DIFF
--- a/charts/fleet-crd/templates/crds.yaml
+++ b/charts/fleet-crd/templates/crds.yaml
@@ -318,6 +318,27 @@ spec:
                             in the helm history.
                           type: boolean
                       type: object
+                    createNamespace:
+                      description: 'CreateNamespace controls whether Fleet creates
+                        the target namespace on
+
+                        downstream clusters during Helm installs. When nil, the default
+                        behavior
+
+                        is to create the namespace (backward-compatible). Set to false
+                        by the
+
+                        controller when Policy requires a ServiceAccount and does
+                        not explicitly
+
+                        allow namespace creation. This does not affect namespaceLabels/
+
+                        namespaceAnnotations patching, which is always attempted (when
+                        set) as
+
+                        the deployment''s ServiceAccount and gated by downstream RBAC.'
+                      nullable: true
+                      type: boolean
                     defaultNamespace:
                       description: 'DefaultNamespace is the namespace to use for resources
                         that do not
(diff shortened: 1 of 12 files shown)


d3flex (approved):
approved by revewing:
- naming and layering 
- logic correctness
- commit description
- for obvious bugs

PR #5435, opened 2026-07-13
Simplify GoReleaser dockers_v2 config

thardeck:
Reduce repeated `dockers_v2` fields in `.goreleaser.yaml` with YAML anchors for shared IDs, tags, labels, build args, platforms, and flag sets while preserving image names, tags, dockerfiles, and signing IDs.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -112,25 +112,25 @@ dockers_v2:
   -
     id: fleet-public
     disable: "{{ .Env.IS_HOTFIX }}"
-    ids:
+    ids: &fleet_controller_ids
     - fleet-controller
     - fleet-cli
     images:
     - "docker.io/rancher/fleet"
-    tags:
+    tags: &docker_tags
     - "{{ .Tag }}"
     dockerfile: package/Dockerfile
-    labels:
+    labels: &docker_labels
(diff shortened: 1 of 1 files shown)


.goreleaser.yaml
```diff
@@ -112,107 +112,72 @@ dockers_v2:
...
-      "org.opencontainers.image.title": "{{.ProjectName}}"
-      "org.opencontainers.image.revision": "{{.FullCommit}}"
-      "org.opencontainers.image.version": "{{.Version}}"
-      "org.opencontainers.image.source": "https://github.com/rancher/fleet"
-    build_args:
-      BUILD_ENV: "goreleaser"
-    flags:
+    labels: *docker_labels
+    build_args: *docker_build_args
+    flags: &private_docker_flags
```
d3flex, line 162: I believe you forgot to remove the flags here

thardeck: There are public and private docker flags, and here are the `private_docker_flags` defined.

d3flex (approved):

.goreleaser.yaml
d3flex: my bad. In second look was not what I thought i saw yesterday

PR #5436, opened 2026-07-13
Fix Schedule race condition and flaky e2e tests

0xavi0:
This PR addresses some Schedule e2e tests that are flaky and a race condition in the Scheduler controller.

### DIFF
--- a/e2e/single-cluster/agent_scheduling_customization_test.go
+++ b/e2e/single-cluster/agent_scheduling_customization_test.go
@@ -18,20 +18,6 @@ func ptr[T any](v T) *T {
 }
 
 var _ = Describe("Agent Scheduling Customization", func() {
-	var (
-		cluster *fleet.Cluster
-	)
-
-	BeforeEach(func() {
-		var err error
-		cluster = &fleet.Cluster{}
-		err = clientUpstream.Get(context.TODO(), client.ObjectKey{
-			Namespace: env.Namespace,
-			Name:      "local",
-		}, cluster)
-		Expect(err).ToNot(HaveOccurred())
-	})
-
 	When("agentSchedulingCustomization.PriorityClass is configured on the cluster resource", func() {
 		BeforeEach(func() {
 			// Update the cluster with agentSchedulingCustomization
@@ -68,6 +54,13 @@ var _ = Describe("Agent Scheduling Customization", func() {
 				g.Expect(err).ToNot(HaveOccurred())
 
 				latestCluster.Spec.AgentSchedulingCustomization = nil
+				// Bump RedeployAgentGeneration to force a prompt agent re-import.
+				// Removing the customization clears the status hash quickly, but the
+				// re-import that prunes fleet-agent-priority-class is otherwise only
+				// triggered on the controller's periodic resync (~5m). That cadence
+				// races this AfterEach's PriorityClass-deletion wait and makes the
+				// spec flaky, so we force the redeploy here to prune it immediately.
+				latestCluster.Spec.RedeployAgentGeneration++
 				err = clientUpstream.Update(context.TODO(), latestCluster)
 				g.Expect(err).ToNot(HaveOccurred())
 			}).Should(Succeed(), "Should be able to update cluster to remove agentSchedulingCustomization")
@@ -79,7 +72,7 @@ var _ = Describe("Agent Scheduling Customization", func() {
 					Name:      "local",
 				}, latestCluster)
 				g.Expect(err).ToNot(HaveOccurred())
-				g.Expect(cluster.Status.AgentSchedulingCustomizationHash).To(Equal(""))
+				g.Expect(latestCluster.Status.AgentSchedulingCustomizationHash).To(Equal(""))
 			}).Should(Succeed(), "Should have cleared the AgentSchedulingCustomizationHash")
 
 			// Wait for PriorityClass to be deleted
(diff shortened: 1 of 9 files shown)


e2e/single-cluster/agent_scheduling_customization_test.go
```diff
@@ -239,9 +235,32 @@ var _ = Describe("Agent Scheduling Customization", func() {
...
+				g.Expect(latestCluster.Status.AgentSchedulingCustomizationHash).To(Equal(""))
+			}).Should(Succeed(), "Should have cleared the AgentSchedulingCustomizationHash")
+
+			Eventually(func() bool {
+				pc := &schedulingv1.PriorityClass{}
+				err := clientUpstream.Get(context.TODO(), client.ObjectKey{
+					Name: "fleet-agent-priority-class",
+				}, pc)
+				return errors.IsNotFound(err)
+			}).Should(BeTrue(), "PriorityClass should be deleted")
```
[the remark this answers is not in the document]
0xavi0: I've added a new function also because the TearDown code was being repeated at least 3 times

0xavi0:
A race condition has been found in the controller code.

internal/cmd/controller/reconciler/job_registry.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/reugn/go-quartz/quartz"
+)
+
+// jobRegistry tracks the CronDurationJobs owned by a ScheduleReconciler, keyed by quartz job key.
+//
+// It exists because the quartz scheduler cannot answer the question "does this Schedule have a
+// job?": quartz pops a job off its queue before running it, and a CronDurationJob only puts itself
+// back once its start or stop action is done. A running job is therefore reported as missing by
+// quartz.GetScheduledJob, and re-creating a job from that answer would cancel the active window the
+// running job has just opened.
```
weyfonk, line 16: Crystal clear, thank you 🤩

internal/cmd/controller/reconciler/job_registry.go
```diff
@@ -0,0 +1,88 @@
...
+type jobRegistry struct {
+	mu   sync.RWMutex
+	jobs map[string]*CronDurationJob
+}
+
+func (r *jobRegistry) get(key *quartz.JobKey) (*CronDurationJob, bool) {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+
+	job, found := r.jobs[key.String()]
```
weyfonk, line 32: Could `key` be `nil`?
Same question for `store` and `delete`.

internal/cmd/controller/reconciler/job_registry.go
```diff
@@ -0,0 +1,88 @@
...
+			jobs = append(jobs, job)
+		}
+	}
+
+	return jobs
+}
+
+// isClusterScheduled returns true if the given cluster is targeted by any registered job.
+func (r *jobRegistry) isClusterScheduled(cluster, namespace string) bool {
+	return len(r.jobsForCluster(cluster, namespace)) > 0
```
weyfonk: This method and `clustersNotScheduled` both use `jobsForCluster` (itself calling `targets()` on each job, and incurring mutex operations...), which computes the full list of jobs targeting the cluster, while only being interested in whether _at least one_ job does target the cluster.

We may have room for optimisation there (e.g. using a separate method such as `clusterHasJob`, with a similar body to `jobsForCluster`, but returning `true`, instead of appending jobs to a slice, as soon as a matching job is found).

internal/cmd/controller/reconciler/schedule_controller_test.go
```diff
@@ -192,6 +192,72 @@ var _ = Describe("ScheduleReconciler", func() {
...
+			Expect(newJob).NotTo(BeIdenticalTo(oldJob))
+
+			// Quartz may already have dispatched a start for the old job. Running it would flag the
+			// cluster as active and re-arm the old job over the one which replaced it.
+			Expect(oldJob.Execute(ctx)).To(Succeed())
+
+			updatedCluster := &fleet.Cluster{}
+			err = k8sclient.Get(ctx, client.ObjectKeyFromObject(cluster), updatedCluster)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(updatedCluster.Status.ActiveSchedule).To(BeFalse())
```
weyfonk, line 254: Since we've reconciled the _schedule_ and are validating expectations against a _cluster_ here, I think it would make sense to wrap this into an `Eventually` statement to reduce risks of flakiness. WDYT?

internal/cmd/controller/reconciler/job_registry.go
0xavi0: In practice no: every caller passes the result of scheduleKey(), which always returns quartz.NewJobKey(...), that's never nil. But given this is living in a separate file I think your point is totally valid and I've added the nil checks. 

Note: passing a nil key would mean a panic anyway in the controller code (but that should never happen anyway)

internal/cmd/controller/reconciler/job_registry.go
0xavi0: I've implemented the `clusterHasJob` as suggested, thanks!!

internal/cmd/controller/reconciler/schedule_controller_test.go
0xavi0: The assertion at line 254 follows a synchronous `oldJob.Execute(ctx)` against a fake client, (and ActiveSchedule stays false) — Eventually would pass immediately anyway

internal/cmd/controller/reconciler/job_registry_test.go
```diff
@@ -0,0 +1,94 @@
...
+		It("returns only clusters which no job targets", func() {
+			notScheduled := registry.clustersNotScheduled(
+				[]string{"cluster1", "cluster2", "cluster3", "cluster4"},
+				"default",
+			)
+
+			Expect(notScheduled).To(Equal([]string{"cluster3", "cluster4"}))
+		})
+
+		It("returns no cluster when the registry is empty", func() {
```
weyfonk: This does return a cluster though 🤔

0xavi0: oops! fixed now. Thanks!

weyfonk (approved):

internal/cmd/controller/reconciler/job_registry_test.go
weyfonk: This makes more sense :) Thanks for the fix!

Issue #3720, opened 2025-05-26
Webhooks not working in SSH mode when SSH hostname is different from web UI hostname


sestegra:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

After receiving a webhook, `gitjob` doesn't evaluate it, because it can't find related `GitRepo` object.
As the find query is only based on Web hostname, no `GitRepo` is found by this query.

See `parsePayload` function
https://github.com/rancher/fleet/blob/main/pkg/webhook/webhook.go#L272

### Expected Behavior

Received webhooks must be evaluated using the SSH URLs as well.

### Steps To Reproduce

1. Install Rancher 2.11.2 with URL equals to `rancher.my.labs`
2. Deploy Ingress for Fleet Webhook
3. Install Gitea/Forgejo/Gogs with HTTP URL equals to `gitea.my.labs` and SSH URL equals to `gitea-ssh.my.labs`(different hostname)
4. Add GitRepo using `repo: git@gitea-ssh.my.labs:fruit-salad/orange.git`

### Environment

```markdown
- Architecture: amd64
- Fleet Version: 0.12.3
- Cluster:
  - Provider: RKE2
  - Options:
  - Kubernetes Version: v1.32.4+rke2r1
```

### Logs

```markdown
`gitjob` logs, only webhook payload is displayed, no further logs.

[...]

PR #5439, opened 2026-07-13
fix(webhook): evaluate SSH URLs from webhook payloads

d3flex:
- Use `parsePayload` to include giturl and allow the outer loop to process git along with https url
- Apply fix to all the providers for consistency
- dedup payload URLs and skip empty SSH paths
- update and refactor webhook_test.go to cover each edge case from `sshURLToParsable` and prevent regression

Refers to #3720

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -116,6 +116,9 @@ func (w *Webhook) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
 		}
 		seen[key] = struct{}{}
 
+		if u.EscapedPath() == "" {
+			continue
+		}
 		path := strings.Replace(regexp.QuoteMeta(u.EscapedPath()[1:]), `/_git/`, `(/_git)?/`, 1)
 		regexpStr := `(?i)(http://|https://|\w+@|ssh://(\w+@)?|git@(ssh\.)?)` + regexp.QuoteMeta(u.Hostname()) +
 			"(:[0-9]+|)[:/](v\\d/)?" + path + "(\\.git)?$"
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -99,22 +99,17 @@ func (w *Webhook) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// A payload can yield several URLs (e.g. a web URL and an SSH URL) that
-	// resolve to the same host and path, most commonly when the SSH host equals
-	// the web host (github.com). Track those with `seen` so a matching GitRepo is
-	// processed once per request instead of letting it be processed repeatedly.
-	seen := make(map[string]struct{}, len(repoURLs))
+	// A payload can yield several URLs (e.g. a web URL and an SSH URL), and more
+	// than one can match the same GitRepo (most commonly when the SSH host equals
+	// the web host, github.com). Track the GitRepos already updated in `seen` so
+	// each is processed once per request, no matter how many URLs resolve to it.
+	seen := make(map[types.NamespacedName]struct{})
 	for _, repo := range repoURLs {
 		u, err := url.Parse(repo)
 		if err != nil {
 			w.logAndReturn(rw, err)
 			return
 		}
-		key := u.Hostname() + "\x00" + u.EscapedPath()
-		if _, ok := seen[key]; ok {
-			continue
-		}
-		seen[key] = struct{}{}
 
 		if u.EscapedPath() == "" {
 			continue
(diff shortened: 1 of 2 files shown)


kkaempf:
This is missing tests 🤔

d3flex:
> This is missing tests 🤔

I thought that covering some input with the `sshURLToParsable` and add one case which has git repo in `TestGitRepoURLMatch` will do. what test do you have in mind?

d3flex:
> This is missing tests 🤔

I rafactor the pkg/webhook/webhook_test.go to rectify the missing test cases. Those are now should cover:

- github.PushPayload
- gitlab.PushEventPayload
- gitlab.TagEventPayload
- gogsclient.PushPayload

pkg/webhook/webhook.go
```diff
@@ -99,12 +99,23 @@ func (w *Webhook) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
...
 			w.logAndReturn(rw, err)
 			return
 		}
+		key := u.Hostname() + "\x00" + u.EscapedPath()
+		if _, ok := seen[key]; ok {
+			continue
+		}
+		seen[key] = struct{}{}
+
 		path := strings.Replace(regexp.QuoteMeta(u.EscapedPath()[1:]), `/_git/`, `(/_git)?/`, 1)
```
p-se, line 117: If there is no path provided, `u.EscapedPath()[1:]` will panic. Maybe we can fix this is with this PR as well?

pkg/webhook/webhook.go
```diff
@@ -99,12 +99,23 @@ func (w *Webhook) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// A payload can yield several URLs (e.g. a web URL and an SSH URL) that
+	// resolve to the same host and path, most commonly when the SSH host equals
+	// the web host (github.com). Track those with `seen` so a matching GitRepo is
+	// processed once per request instead of letting it be processed repeatedly.
+	seen := make(map[string]struct{}, len(repoURLs))
```
p-se: Tracking and deduplicating by GitRepo resource we've already updated could be cleaner.

p-se (commented):
It looks like for scp-form URLs the `.git` suffix is stripped which makes it optional, but for scheme-form payloads (gitlab on a non-22 SSH port, e.g. `ssh://git@gitlab.com:2222/owner/repo.git`) they are required and don't match spec.repo URLs without it (but scp-form URLs would).

pkg/webhook/webhook.go
```diff
@@ -281,14 +321,23 @@ func parsePayload(payload any) (revision, branch, tag string, repoURLs []string)
 		branch, tag = getBranchTagFromRef(t.Ref)
 		revision = t.After
 		repoURLs = append(repoURLs, t.Repository.HTMLURL)
+		if sshURL := sshURLToParsable(t.Repository.SSHURL); sshURL != "" {
+			repoURLs = append(repoURLs, sshURL)
+		}
```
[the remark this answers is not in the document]
d3flex: no change

pkg/webhook/webhook.go
d3flex: done - a small check preventing a panic letting the loop continue applied

pkg/webhook/webhook.go
d3flex: after our call I removed my approach and I adapted to make use of the `gitrepo.NamespacedName` type. definetely an improvement. thanks for review and assist

pkg/webhook/webhook.go
```diff
@@ -272,6 +286,35 @@ func getBranchTagFromRef(ref string) (string, string) {
...
 }
 
+// sshURLToParsable converts an scp-style git remote such as
+// "git@github-ssh.example.com:owner/repo.git" into an "ssh://" URL that
+// net/url can parse, so the webhook matching loop can extract its hostname and
+// path.
+func sshURLToParsable(raw string) string {
+	if raw == "" {
+		return ""
+	}
```
0xavi0, line 296:
```suggestion
    }
    // Drop a trailing ".git" so the "(\.git)?$" in the match regexp stays
    // optional and matches GitRepos configured with or without the suffix.
    // This must happen before the "://" check so it applies to both scp-style
    // remotes and ready-made "ssh://" URLs.
    raw = strings.TrimSuffix(raw, ".git")
```

pkg/webhook/webhook.go
```diff
@@ -272,6 +286,35 @@ func getBranchTagFromRef(ref string) (string, string) {
...
+	if !ok {
+		return ""
+	}
+	host, path, ok := strings.Cut(rest, ":")
+	if !ok {
+		return ""
+	}
+	// Drop a trailing ".git" so the "(\.git)?$" in the match regexp stays
+	// optional and matches GitRepos configured with or without the suffix.
+	path = strings.TrimSuffix(path, ".git")
```
0xavi0:
```suggestion
```

I believe trimming the suffix before makes the following cases work fine:

### Payload
git_ssh_url = ssh://git@gitlab.example.com:2222/group/repo.git

### GitReposURL
ssh://git@gitlab.example.com:2222/group/repo
git@gitlab.example.com:group/repo

pkg/webhook/webhook.go
0xavi0: See next comment

pkg/webhook/webhook.go
d3flex: done

0xavi0 (approved):

p-se (approved):

Issue #5444, opened 2026-07-15
Performance: LabelSelectorAsSelector compiled fresh on every reconcile in clusterGroupsForCluster — O(clusters × ClusterGroups) allocation storm at scale


himabindugit:
### Is there an existing issue for this?

- [x] I have searched the existing issues

## What happened?

In `pkg/controllers/bundle/target.go` (func `clusterGroupsForCluster`), `metav1.LabelSelectorAsSelector(cg.Spec.Selector)` is called fresh on every reconcile for every ClusterGroup × every cluster. The compiled selector is never cached — it is created, used once, and GC'd.

## Root cause (code location)

pkg/controllers/bundle/target.go:113-135

```go
for _, cg := range cgs {
    sel, _ := metav1.LabelSelectorAsSelector(cg.Spec.Selector) // compiled fresh EVERY call
    if sel.Matches(labels.Set(cluster.Labels)) { ... }
}
CPU profile evidence (production pprof)
LabelSelectorAsSelector : 999s — 61% of all CPU
NewRequirement : 644s — 39%
mallocgc : 901s — 55% (GC overwhelmed by short-lived objects)
sweepLocked.sweep : 413s — 25%
Scale impact
At our fleet size:

~3,900 Cluster objects × 10 ClusterGroups = 39,130 fresh compilations per OnBundleChange event. Every OnClusterChange does N bundles × 10 ClusterGroups compilations. The compiled selector is never cached — it invalidates on every call regardless of whether the ClusterGroup spec changed.


// pseudocode
cacheKey := cg.Namespace + "/" + cg.Name + "@" + cg.ResourceVersion
if cached, ok := selectorCache.Load(cacheKey); ok {
    sel = cached.(labels.Selector)
} else {
    sel, _ = metav1.LabelSelectorAsSelector(cg.Spec.Selector)
    selectorCache.Store(cacheKey, sel)
}
Fleet version
v0.9.x (issue likely present in all versions — not seen fixed in changelogs up to v0.15)

### Expected Behavior

`LabelSelectorAsSelector(cg.Spec.Selector)` should be compiled **once**  per ClusterGroup and cached — recompiled only when the ClusterGroup's `Spec.Selector` actually changes (detectable via `ResourceVersion`).

At 3,900 clusters × 10 ClusterGroups, each `OnBundleChange` event should trigger **10 selector lookups** (cache hits), not 39,130 
fresh compilations.

[...]

PR #5445, opened 2026-07-15
target: cache compiled LabelSelector per ClusterGroup to reduce CPU

himabindugit:
https://github.com/rancher/fleet/issues/5444
Refers to https://github.com/rancher/fleet/issues/5444

## Problem
`clusterGroupsForCluster` calls `metav1.LabelSelectorAsSelector` on every reconcile loop for every ClusterGroup, with no caching. At scale this causes an allocation storm.

With ~3900 clusters × 10 ClusterGroups: 39,130 `LabelSelectorAsSelector` compilations per `OnBundleChange` event. pprof showed `LabelSelectorAsSelector` consuming 61% CPU on the fleet-controller, leading to OOM and lease-lock failures.

Reported in: #5444

## Fix

Cache the compiled `labels.Selector` in a `sync.Map` on the `Manager`, keyed by `namespace/name@resourceVersion`. The `ResourceVersion` ensures the selector is recompiled when a ClusterGroup is updated.

ClusterGroups change infrequently and their count is small, so stale entries from old ResourceVersions are acceptable and not evicted.

Added unit tests covering:
- Matching and non-matching selectors
- Nil selector skipped
- Invalid selector not cached
- Cache populated after first call
- New ResourceVersion gets a new cache entry

## AI-Assisted Code Review

This change was analyzed using **Claude Code** (Anthropic, claude-sonnet-4-6) as part of a broader investigation into fleet-controller CPU instability (production incident IM-890045).

**What the analysis found:**
- `LabelSelectorAsSelector` is called inside nested per-cluster × per-bundle loops in `target.go`, recompiling selectors on every reconcile cycle
- At production scale (~3,900 clusters, ~4,000 bundles, ~50 ClusterGroups), this results in ~600M recompilations per cycle — the root cause of the sustained CPU pin
- The memoization fix (cache keyed by ClusterGroup UID + resourceVersion)
  reduces this to O(ClusterGroups) ≈ 50 unique compilations

**Evidence:**
- Production pprof CPU trace (30-second capture, port-forwarded to the live pod) confirmed `LabelSelectorAsSelector` as the dominant hot path
- Fleet-controller logs from the July 7 incident showed a 3,990-line reconciliation burst within one minute of lease acquisition after restart

**Correctness conclusion:**
The AI analysis concluded that correctness is maintained — cache entries are invalidated on resourceVersion change, ensuring stale selectors are never used.
This was independently verified by Stephen O'Neal ([sonealv](https://github.com/sonealv)), and GitHub Copilot.

### Checklist


himabindugit:
Hi @aruiz14 — following up on this PR. 
We've been running this in production with ~3,900 clusters and consistently seeing 90%+ CPU on the fleet-controller node. The fix addresses the `LabelSelectorAsSelector` recompile storm we identified via pprof, bringing operations from ~716M per cycle down dramatically.

Note:
1. This PR change was analyzed using Claude Code, PR description was updated saying this.
2. We have a load test running at production scale to collect metrics and will share results here once available.

Could you let us know if there's any additional information needed to move this forward, or an expected timeline for review?
Thanks

himabindugit:
Load test + production incident evidence

Environment: Load test (2026-07-28):
Sandbox cluster — 3,500+ downstream clusters, fleet-controller at 97% CPU on a 32-core node.

30-second CPU profile (before fix):


Duration: 30.20s, Total samples = 401.12s (1328.14%)

      flat  flat%   cum    cum%
     5.10s  1.27%  222.79s 55.54%  k8s.io/apimachinery/pkg/apis/meta/v1.LabelSelectorAsSelector
LabelSelectorAsSelector accounts for 55.54% of all CPU (222.79s out of 401.12s sampled) — consistent with the original production profile where it was 61.49%.

Production incident (2026-07-27):

A post-incident investigation on our production cluster (3,932 clusters / 4,153 bundles / 25,865 bundledeployments) found fleet-controller listing all bundles every ~1.5–1.9 seconds with no resourceVersion and no limit — ~40 full-namespace etcd scans per minute. This drove the etcd node's EBS volume to its 3,000 IOPS ceiling (9.6 GB read from disk in one minute from a 1 GB database), freezing etcd and crashing fleet-controller.

The excessive LIST frequency is a direct consequence of the O(N³) reconciliation loop. The caching fix in this PR reduces reconciliation frequency dramatically, which in turn reduces LIST pressure on etcd.

d3flex (changes requested):
Can you please squash some commits in the history into a couple of logical commits? Right now the history has a lot of fixup/WIP ones. It would make easier the review and keep the history tidy.  And while fixing that, two other things missing: commit body and signatures. add a few words on the commits . Mostly are empty (especially the caching change)

himabindugit:
> * clusterGroupsForCluster

Thanks for the feedback! Addressed all three points:

- Squashed commits — consolidated into 2 logical commits: one for the LabelSelector caching fix and one for the loop hoist.
- Commit bodies — added descriptions explaining the problem, the approach, and the performance impact to both commits.
- Signed-off-by — added signatures to both commits.

himabindugit:
> ## Pull request overview
> 
> This PR optimizes target selection by caching compiled Kubernetes `LabelSelector` objects per `ClusterGroup` (keyed by `namespace/name@resourceVersion`) to reduce repeated `metav1.LabelSelectorAsSelector` compilations during reconcile loops.
> 
> **Changes:**
> 
>     * Implement per-`Manager` selector compilation caching via `sync.Map` in `clusterGroupsForCluster`.
> 
>     * Add unit tests for selector matching behavior and caching behavior.
> 
>     * Extend the `target.Manager` struct to hold the selector cache.
> 
> 
> ### Reviewed changes
> 
> Copilot reviewed 3 out of 3 changed files in this pull request and generated 2 comments.
> File 	Description
> internal/cmd/controller/target/query.go 	Adds selector caching in `clusterGroupsForCluster` to reduce CPU/allocations during target queries.
> internal/cmd/controller/target/query_test.go 	Adds unit tests for selector matching/caching behavior.
> internal/cmd/controller/target/builder.go 	Adds a `sync.Map` field to `Manager` to store the selector cache.
> 
> 💡 [Add Copilot custom instructions](/rancher/fleet/new/main?filename=.github/instructions/*.instructions.md) for smarter, more guided reviews. [Learn how to get started](https://docs.github.com/en/copilot/customizing-copilot/adding-repository-custom-instructions-for-github-copilot).

- Fixed — using cgCopy := cg before appending &cgCopy to avoid the range-variable alias bug.

- Added in TestClusterGroupsForCluster_NewResourceVersionCreatesNewCacheEntry — swaps to a second fake client returning rv=2 to verify a new cache entry is created.

internal/cmd/controller/target/query.go
```diff
@@ -87,7 +87,38 @@ func (m *Manager) getBundlesInScopeForCluster(ctx context.Context, cluster *flee
...
+		if cached, ok := m.selectorCache.Load(cacheKey); ok {
+			sel = cached.(labels.Selector)
+		} else {
+			sel, err = metav1.LabelSelectorAsSelector(cg.Spec.Selector)
+			if err != nil {
+				logger.Error(err, "invalid selector on clusterGroup", "namespace", cg.Namespace, "name", cg.Name,
+					"selector", cg.Spec.Selector)
+				continue
+			}
+			m.selectorCache.Store(cacheKey, sel)
```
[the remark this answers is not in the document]
himabindugit: Good observation — matchingClusters and getClusterSchedules do bypass the cache via the exported ClusterGroupsForCluster. 
However, these are schedule-reconciler paths (cron-triggered, low frequency) and not the hot path that caused the original CPU spike. 
To keep this PR focused and minimal, I'd prefer to address those callers in a follow-up if profiling shows they contribute meaningfully.

internal/cmd/controller/target/query.go
```diff
@@ -27,11 +32,6 @@ func (m *Manager) BundlesForCluster(ctx context.Context, cluster *fleet.Cluster)
 			continue
 		}
 
-		cgs, err := m.clusterGroupsForCluster(ctx, cluster)
-		if err != nil {
-			return nil, nil, err
-		}
-
 		match := bm.Match(cluster.Name, ClusterGroupsToLabelMap(cgs), cluster.Labels)
```
weyfonk: nit: this `ClusterGroupsToLabelMaps(cgs)` call could also be moved out of the loop.

internal/cmd/controller/target/query.go
```diff
@@ -87,7 +87,38 @@ func (m *Manager) getBundlesInScopeForCluster(ctx context.Context, cluster *flee
 }
 
 func (m *Manager) clusterGroupsForCluster(ctx context.Context, cluster *fleet.Cluster) (result []*fleet.ClusterGroup, _ error) {
-	return ClusterGroupsForCluster(ctx, m.client, cluster)
+	cgs := &fleet.ClusterGroupList{}
```
weyfonk, line 91: This would leave us with 2 almost identical implementations of `{C,c}lusterGroupsForCluster`: one with cache, one without.
Instead, the changes could be made to `ClusterGroupsForCluster`, which could take a (nullable) `sync.Map`, leading to caching when that map is non-nil, as in the case of calls from `clusterGroupsForCluster` (where the map would be `m.selectorCache`).

Then, `ClusterGroupsForCluster` would load and store entries from/to the cache when such a map is available, skipping those operations otherwise.

internal/cmd/controller/target/query.go
weyfonk, line 115: Note: we don't seem to ever delete entries (i.e. cluster group keys with obsolete versions, or keys for cluster groups which have been deleted) from the cache.

weyfonk (commented):
Thank you for tackling this issue; I would feel more comfortable merging it with a couple of improvements, suggested in comments. Happy to discuss.

internal/cmd/controller/target/query.go
weyfonk: Oh, just noticed [this earlier comment](https://github.com/rancher/fleet/pull/5445#discussion_r4026680259); a follow-up could work as well.

internal/cmd/controller/target/query.go
himabindugit: Good point — hoisted it out of the loop; `cgs` is invariant across bundle iterations so there's no need to recompute the map on every bundle.

internal/cmd/controller/target/query.go
himabindugit: right, to keep this PR focused and minimal to cache compiled LabelSelctor changes. I'd prefer not to make further structural changes beyond what's already in this PR to avoid introducing risk ahead of review.

internal/cmd/controller/target/query.go
himabindugit: Good catch — correctness is preserved since the cache key includes ResourceVersion (stale selectors are never served), but you're right that old entries accumulate. Added a `TODO to track eviction` to handle this later.

weyfonk: Indeed, I am not concerned about correctness here, but rather about memory leaks.
They can be avoided pretty simply; [here](https://github.com/weyfonk/fleet/commit/8292af3346fbe638c331ec44f87a6e5a99b2ed94) is an example of a naive implementation for purging obsolete cache entries.
Other options include:
- having the cache be a map of maps (namespace/name → resource version → selector), enabling keys not pertaining to the current resource version to be purged when a new resource version is found
- using 2 different maps:
    * cg.Namespace/cg.Name → cg.ResourceVersion
    * cacheKey → selector (the current one)
    * → before updating the resource version in the first map, delete the obsolete selector from second map based on the previous resource version

Issue #4883, opened 2026-03-23
[SURE-10854] fleet apply to aggregate bundles errors and fail later


kkaempf:
### Business case
Cluster admins might not have access/rights over gitrepos, so when a bundle fail they will need to report back to their dev team. However, if we aggregate the bundle errors and fail later, they will have an opportunity to deliver other bundles from the repo. 

### Request description:
`fleet apply` to aggregate bundle errors and fail later

### Actual behavior:
once hits an error, `fleet apply` fails and report back

### Expected behavior:
if there are more bundles, and one fail, go to the next before reporting the errors

### Workaround:
Is a workaround available and implemented? no
What is the workaround:

### Additional notes:
one might argue that failing early is the best approach, so there is no half deployments.

See SURE-10854 for further details

thardeck:
## QA Template

- Run `fleet apply` with valid bundles before and after one malformed bundle.
- Confirm valid bundles are created despite the malformed bundle.
- Confirm command returns an error identifying malformed bundle path.
- Confirm existing bundles are not pruned after incomplete discovery.

mmartin24:
FTR:

I tested this in `0.17.0-alpha.1` and seemed to work but I was expecting this not to work in `v0.16.0` without the fix but it did so not sure if my testing is wrong or if this was actually fixed before Fleet `0.17.0-alpha.1`

This is how I tested it:



Install Fleet v0.16.0:

Use this example: [fleet-error-test.zip](https://github.com/user-attachments/files/31795739/fleet-error-test.zip)

Check the test repo structure:

```console
➜  fleet-error-test cd fleet-error-test

# List all bundles in alphabetical/processing order
find . -name "fleet.yaml" -exec dirname {} \; | sed 's|^\./||' | sort

# Count bundles before bundle2z-mongodb-bad (should be 4)
find . -name "fleet.yaml" -exec dirname {} \; | sed 's|^\./||' | sort | grep -B999 '^bundle2z-mongodb-bad$' | grep -v '^bundle2z-mongodb-bad$' | wc -l

# Count bundles after bundle2z-mongodb-bad (should be 0)
find . -name "fleet.yaml" -exec dirname {} \; | sed 's|^\./||' | sort | grep -A999 '^bundle2z-mongodb-bad$' | grep -v '^bundle2z-mongodb-bad$' | wc -l
cd: no such file or directory: fleet-error-test
bundle1
bundle2
bundle2z-mongodb-bad
bundle3
bundle4
2
2
```

[...]

PR #5446, opened 2026-07-15
Apply all bundles before returning errors

thardeck:
Ensure `fleet apply` continues reading and writing every bundle path after individual failures.

Aggregate read, prune, and write errors after processing completes so valid bundles can still be applied while failures remain visible. Skip pruning when bundle discovery is incomplete to preserve existing bundles from a partial scan.

Refers #4883


internal/cmd/cli/apply/apply.go
```diff
@@ -141,25 +162,29 @@ func CreateBundles(pctx context.Context, client client.Client, r record.EventRec
 	// 3. We use another goroutine to wait for all goroutines to finish, then close `bundlesChan`, finally unblocking the main function.
 
 	bundlesChan := make(chan *bundleWithOpts)
-	eg, ctx := errgroup.WithContext(pctx)
+	var eg errgroup.Group
```
aruiz14: Does it make sense to keep an "errgroup" here when we don't need the context cancellation and error propagation it provides? Back when I introduced it, I was preserving the original behavior of aborting the process when any error was found.
You could consider using a [WaitGroup](https://pkg.go.dev/sync#WaitGroup) with regular goroutines instead, and a [semaphore](https://pkg.go.dev/golang.org/x/sync@v0.22.0/semaphore) for controlling max concurrency.

internal/cmd/cli/apply/apply.go
aruiz14: Please rename it to `ctx` instead. Previous code needed to be able to still access the parent

internal/cmd/cli/apply/apply.go
```diff
@@ -141,25 +162,29 @@ func CreateBundles(pctx context.Context, client client.Client, r record.EventRec
...
+	var readErrors errorCollector
```
aruiz14: You could also use an error channel + goroutine aggregating the errors into an slice.

aruiz14 (changes requested):

internal/cmd/cli/apply/apply.go
thardeck: Replaced `errgroup` with `sync.WaitGroup` and a bounded semaphore channel. This should preserve bounded concurrent bundle processing without error-driven cancellation.

internal/cmd/cli/apply/apply.go
thardeck: Replaced the mutex-backed error collector with an error channel and a dedicated aggregation goroutine. Read and write errors are joined after their respective workers finish.

internal/cmd/cli/apply/apply.go
```diff
@@ -128,76 +142,79 @@ func getEffectiveMaxConcurrency(configured int) int {
...
-				return fmt.Errorf("invalid path glob %s: %w", baseDir, err)
+				readErrorChan <- fmt.Errorf("invalid path glob %s: %w", baseDir, err)
+				continue
 			}
 			for _, baseDir := range matches {
 				if err := filepath.WalkDir(baseDir, func(path string, entry fs.DirEntry, err error) error {
 					if err != nil {
-						return fmt.Errorf("failed walking path %q: %w", path, err)
+						readErrorChan <- fmt.Errorf("failed walking path %q: %w", path, err)
+						return nil
```
aruiz14: This one can't be `nil`, otherwise it will continue walking. WalkDir continues to run the WalkDirFunc propagating the error from previous executions. I know it's a bit annoying

internal/cmd/cli/apply/apply.go
```diff
@@ -128,76 +142,79 @@ func getEffectiveMaxConcurrency(configured int) int {
...
-	// 2. The main function will read from `bundlesChan`, hence unblocking the goroutines. This will continue to read from `bundlesChan` until it is closed.
-	// 3. We use another goroutine to wait for all goroutines to finish, then close `bundlesChan`, finally unblocking the main function.
-
 	bundlesChan := make(chan *bundleWithOpts)
-	eg, ctx := errgroup.WithContext(pctx)
-	eg.SetLimit(maxConcurrency + 1) // extra goroutine for WalkDir loop
-	eg.Go(func() error {
+	readErrorChan := make(chan error)
+	readErrorsDone := collectErrors(readErrorChan)
+	semaphore := make(chan struct{}, maxConcurrency)
```
aruiz14: Sorry I should have been more explicit. I was referring to https://pkg.go.dev/golang.org/x/sync@v0.22.0/semaphore, which has context support.

internal/cmd/cli/apply/apply.go
aruiz14, line 334: Hmm I just thinking now, do we need to wait before all reading is done to start writing bundles? Can we now just write straightaway and simplify our operations? (we'd still need to record `gitRepoBundlesMap` for the prune step though)

internal/cmd/cli/apply/apply.go
```diff
@@ -128,110 +143,116 @@ func getEffectiveMaxConcurrency(configured int) int {
...
 							}
-							return err
+							readErrorChan <- err
+							return
 						}
 						select {
 						case <-ctx.Done():
-							return ctx.Err()
+							readErrorChan <- ctx.Err()
 						case bundlesChan <- &bundleWithOpts{bundle: bundle, scans: scans, opts: &opts}:
```
aruiz14, line 207: Sorry, I meant to do the write here, so we don't need 2 separate waitgroups.

internal/cmd/cli/apply/apply.go
thardeck: Updated the callback to propagate the incoming walk error, which stops that affected directory walk while allowing the remaining base paths to continue.

internal/cmd/cli/apply/apply.go
thardeck: Replaced the channel semaphores with context-aware `semaphore.Weighted` instances for both read and write work.

internal/cmd/cli/apply/apply.go
thardeck: Updated `fleet apply` to begin writing each successfully read bundle immediately. The full bundle map is still collected before pruning is considered, and pruning remains disabled when discovery reports errors.

internal/cmd/cli/apply/apply.go
thardeck: I tried moving the write into the read worker, but that writes bundles before `pruneBundlesNotFoundInRepo` can populate `Spec.Overwrites` from bundles being removed.

That breaks the existing overwrite behavior, so the separate read and write phases need to remain. The bundle handoff also ensures the complete `gitRepoBundlesMap` is available before pruning and updates.

aruiz14: I'm not sure I understand... `pruneBundlesNotFoundInRepo` is always called after all reads have already finished, and while some of the writes may have already started. There shouldn't be a dependency on `pruneBundlesNotFoundInRepo` for writers.

> The bundle handoff also ensures the complete gitRepoBundlesMap is available before pruning and updates.

`gitRepoBundlesMap` is only needed for the bundle names, so you we could change the `map[string]*fleet.Bundle` to a `Set[string]`, and you'd still need a range over bundlesChan:
```
	for bundleName := range bundlesChan {
		gitRepoBundlesMap.Insert(bundleName)
	}
```

aruiz14: Okay, I'm just seeing this:
https://github.com/rancher/fleet/blob/6ae77ffe87d3a472648a1c6eeb4750d238d1cac7/internal/cmd/cli/apply/apply.go#L432

I honestly do not think `pruneBundlesNotFoundInRepo` should be mutating any of the data inside a cleanup function...

aruiz14: @weyfonk I see this code is relatively new, any thoughts?

internal/cmd/cli/apply/apply_test.go
```diff
@@ -117,3 +143,131 @@ func Test_newOCISecret_usesInsecureSkipTLSKey(t *testing.T) {
...
+		name          string
+		createBundles func(context.Context, client.Client, record.EventRecorder, string, []string, Options) error
+	}{
+		{name: "recursive scan", createBundles: CreateBundles},
+		{name: "driven scan", createBundles: CreateBundlesDriven},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			output := &bytes.Buffer{}
+			err := test.createBundles(context.Background(), nil, nil, "repo", []string{"first", "invalid", "last"}, Options{
```
weyfonk, line 185: Isn't there a risk of a false negative here when using a `nil` client? In other words, couldn't the aggregated error message contain the names of all bundles, but for different reasons (the invalid one would fail to read, whereas the other two would fail to write)?

internal/cmd/cli/apply/apply.go
weyfonk: #3770 has more context as to why this new logic with `Overwrites` is needed; but yes, we would make this cleaner by decoupling the cleanup (`pruneBundlesNotFoundInRepo`) from writing new bundles.
In turn, this could allow the mutation (populating `Overwrites`) to be decoupled from the cleanup.

internal/cmd/cli/apply/apply_test.go
thardeck: `Output` is set in this test, so `writeBundle` serializes the valid bundles and does not use the nil client. The invalid bundle should be the only possible source of the asserted read error.

internal/cmd/cli/apply/apply.go
thardeck: I have separated bundle lookup, `Overwrites` population, and deletion. The map still needs full bundle values because overwrite population compares resources, not only bundle names.

I have also restored the required ordering: discovery and overwrite population complete before writers persist new bundles.

internal/cmd/cli/apply/apply.go
0xavi0, line 170: A WalkDir error stops the walk completely. Is that what we want?

0xavi0 (commented):
Are the Bundles failing identified? 
I mean, if 3 Bundles fail...we'll get the errors aggregated, but... how do we know what's the Bundle for each error?

internal/cmd/cli/apply/apply.go
thardeck: A `WalkDir` error on one entry no longer stops the whole scan - it's reported via `readErrorChan` and the walk continues, skipping only the affected directory (or entry) instead of aborting.

thardeck:
> @0xavi0 Are the Bundles failing identified?
> I mean, if 3 Bundles fail...we'll get the errors aggregated, but... how do we know what's the Bundle for each error?

Read and write errors are now wrapped with the bundle path/name (e.g. `path: %w`), so aggregated errors can be attributed to a specific bundle.

thardeck:
I had to rebase against main and thus to force push. I did not changed the commits or code though, besides [switching from `logrus` to `log.Log`](https://github.com/rancher/fleet/pull/5540).

internal/cmd/cli/apply/apply.go
```diff
@@ -206,32 +229,43 @@ func CreateBundles(pctx context.Context, client client.Client, r record.EventRec
...
+			}
+			writeErrorChan <- nil
+		})
 	}
 
 	if len(gitRepoBundlesMap) == 0 {
+		close(writeErrorChan)
+		if readErr != nil {
+			return readErr
+		}
```
weyfonk, line 257: Would we want to join read and possible write errors in this case, as done on L268?

internal/cmd/cli/apply/apply.go
```diff
@@ -300,32 +340,43 @@ func CreateBundlesDriven(pctx context.Context, client client.Client, r record.Ev
...
+				writeErrorChan <- fmt.Errorf("%s: %w", b.bundle.Name, err)
+				return
+			}
+			writeErrorChan <- nil
+		})
 	}
 
 	if len(gitRepoBundlesMap) == 0 {
+		close(writeErrorChan)
+		if readErr != nil {
```
weyfonk: Same remark as above, on joining read and write errors.

internal/cmd/cli/apply/apply.go
```diff
@@ -206,32 +229,43 @@ func CreateBundles(pctx context.Context, client client.Client, r record.EventRec
...
-	for _, b := range bundlesToWrite {
-		egWrite.Go(func() error {
-			return writeBundle(ctx, client, r, b.bundle, b.scans, *b.opts)
-		})
-	}
-
-	return egWrite.Wait()
+	writeWg.Wait()
+	close(writeErrorChan)
+	return errors.Join(readErr, pruneErr, <-writeErrorsDone)
```
weyfonk, line 264: nit: wouldn't the linter complain about a missing newline before this `return`?
(same remark about L379)
```suggestion

	return errors.Join(readErr, pruneErr, <-writeErrorsDone)
```

weyfonk (approved):
LGTM, with a few nitpicks :)

internal/cmd/cli/apply/apply.go
```diff
@@ -102,6 +103,20 @@ type bundleWithOpts struct {
 	opts   *Options
 }
 
+func collectErrors(errorChan <-chan error) <-chan error {
+	done := make(chan error, 1)
+	go func() {
```
aruiz14, line 108:
```suggestion
	go func() {
		defer close(done)
```

internal/cmd/cli/apply/apply.go
```diff
@@ -128,76 +143,84 @@ func getEffectiveMaxConcurrency(configured int) int {
...
 			if err != nil {
-				return fmt.Errorf("invalid path glob %s: %w", baseDir, err)
+				readErrorChan <- fmt.Errorf("invalid path glob %s: %w", baseDir, err)
+				continue
 			}
 			for _, baseDir := range matches {
 				if err := filepath.WalkDir(baseDir, func(path string, entry fs.DirEntry, err error) error {
 					if err != nil {
-						return fmt.Errorf("failed walking path %q: %w", path, err)
+						readErrorChan <- fmt.Errorf("failed walking path %q: %w", path, err)
```
aruiz14: I think we should just `return err` here, as this is the "shortcut" in case of failures. The error may have already been written to the error channel, so this would duplicate it. Also, I don't really understand the condition below.

If we want to continue walking when there is an error, we should make that explicit by returning nil there (we are already doing it).

internal/cmd/cli/apply/apply.go
```diff
@@ -206,32 +229,43 @@ func CreateBundles(pctx context.Context, client client.Client, r record.EventRec
...
+			writeErrorChan <- fmt.Errorf("%s: %w", b.bundle.Name, err)
+			continue
 		}
+		writeWg.Go(func() {
+			defer writeSemaphore.Release(1)
+			if err := writeBundle(ctx, client, r, b.bundle, b.scans, *b.opts); err != nil {
+				writeErrorChan <- fmt.Errorf("%s: %w", b.bundle.Name, err)
+				return
+			}
+			writeErrorChan <- nil
```
aruiz14: Can this be removed?

internal/cmd/cli/apply/apply.go
thardeck: Done, both early returns now do `errors.Join(readErr, <-writeErrorsDone)` before falling back to the "no resource found" error.

internal/cmd/cli/apply/apply.go
thardeck: Returning `err` directly stops `filepath.WalkDir` for the whole `baseDir`, but  isn't that the whole point of this PR (apply all bundles before returning errors)?
Sending to `readErrorChan` and returning `nil`/`SkipDir` is what keeps that guarantee.

aruiz14: But I don't think we should continue processing at all costs. There could be some errors, maybe internal from WalkDir, that we actually want to use to stop the loop. That's why I'd keep the error check at the top of the WalkDirFunc, but explicit "hide" errors in every place where it could produce them but are acceptable for continuing the processing. Does that make sense?

thardeck: @aruiz14 Agreed, changed the code so that return `WalkDir` errors to abort only the affected base path while continuing subsequent configured paths.
Is it that what you had in mind?

internal/cmd/cli/apply/apply.go
```diff
@@ -243,55 +273,61 @@ func CreateBundles(pctx context.Context, client client.Client, r record.EventRec
...
+	var readWg sync.WaitGroup
+	readWg.Go(func() {
 		for _, baseDir := range baseDirs {
 			opts := opts
-			eg.Go(func() error {
-				var err error
-				baseDir, opts.BundleFile, err = getPathAndFleetYaml(baseDir, opts.DrivenScanSeparator)
+
+			if err := readSemaphore.Acquire(ctx, 1); err != nil {
+				readErrorChan <- fmt.Errorf("%s: %w", baseDir, err)
```
aruiz14, line 293: nit: this will probably only fail if the context is canceled, in which case it does not make much sense to `continue` and duplicate error messages.

aruiz14 (approved):
LGTM! Thanks for the changes @thardeck!

0xavi0 (approved):

internal/cmd/cli/apply/apply.go
thardeck: Done, thank you.

PR #5452, opened 2026-07-16
feat(fleet-controller): skip config re-apply and callbacks for unmodified reloads

d3flex:
The version guard added in the previous commit kept a running controller's
config in memory when a mismatched-version ConfigMap was detected, but it
returned no signal to the caller. Without that signal the live reload paths
kept doing their work even for a rejected ConfigMap: reloadConfig re-fired
every OnChange callback via SetAndTrigger, and Reconcile re-applied the
retained config, when nothing was actually adopted.
    
ReadConfig and Lookup now report whether the config was adopted, and both
live reload paths skip re-applying and re-triggering callbacks when it was
not.
    
Depends-on: https://github.com/rancher/fleet/pull/5397
Refers to https://github.com/rancher/fleet/issues/5809


0xavi0 (commented):
I think the Copilot comments are valid suggestions. 

Aside from that.. looks good to me

d3flex:
> I think the Copilot comments are valid suggestions.
> 
> Aside from that.. looks good to me

both accepted and done

0xavi0 (approved):

PR #5456, opened 2026-07-17
Fix GitHub workflow lint errors

thardeck:
Quote shell expansions, avoid unsafe filename listing, and align reusable workflow inputs with their defaults.

### DIFF
--- a/.github/workflows/port-issue.yml
+++ b/.github/workflows/port-issue.yml
@@ -18,12 +18,12 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.ADD_TO_PROJECT_PAT }}
         run: |
-          if gh api orgs/${GITHUB_REPOSITORY_OWNER}/members --paginate | jq -e --arg GITHUB_ACTOR "$GITHUB_ACTOR" '.[] | select(.login == $GITHUB_ACTOR)' > /dev/null; then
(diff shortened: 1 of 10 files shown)


.github/workflows/port-issue.yml
```diff
@@ -18,12 +18,12 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.ADD_TO_PROJECT_PAT }}
         run: |
-          if gh api orgs/${GITHUB_REPOSITORY_OWNER}/members --paginate | jq -e --arg GITHUB_ACTOR "$GITHUB_ACTOR" '.[] | select(.login == $GITHUB_ACTOR)' > /dev/null; then
-              echo "${GITHUB_ACTOR} is a member"
-              echo "is_member=true" >> $GITHUB_ENV
+          if gh api "orgs/${GITHUB_REPOSITORY_OWNER}/members" --paginate | jq -e --arg ASSIGNEE "$ASSIGNEE" '.[] | select(.login == $ASSIGNEE)' > /dev/null; then
+            echo "${GITHUB_ACTOR} is a member"
```
d3flex: 1. where the `$ASSIGNEE` is defined? 
2. was `${GITHUB_ACTOR}` meant to be `$ASSIGNEE` (in echo and on the rest statements)?

thardeck: Good point, `$ASSIGNEE` is defined only in the later `for ASSIGNEE in $ASSIGNEES` loop.

I have added a fix for that part.

d3flex (approved):

d3flex (approved):

Issue #4564, opened 2026-01-23
Cannot move Namespace into a Rancher Project if it were deployed via Fleet


jamescooke-xyz:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

For any Namespaces created as part of a Fleet deployment, after being moved to a Rancher Project (manually or otherwise) the next time Fleet reconciles, it will revert this move.

In theory we could set the project label/annotation in the fleet deployment, but in practice this is not viable. We will be deploying to hundreds of downstream clusters which will have different ids for the 'target' project.

### Expected Behavior

Either:

- labels and annotations not specified in fleet should not be deleted by fleet when it reconciles, or
- add a mechanism to ignore certain labels and annotations for fleet deployments to allow us control this, or
- provide a mechanism to move a namespace into a project using a displayName or predetermined label

Note:

- unlike spec changes, labels and annotation changes don't trigger drift detection so don't seem to be considered 'desired state' per say (but they are reverted when the next reconciliation occurs)
- certain labels and annotations remain, but these are hardcoded (e.g. [Keep the `kubernetes.io/metadata.name`](https://github.com/rancher/fleet/blob/1708279c067339164814524a3f7f9bab8a7d8037/internal/cmd/agent/deployer/deployer.go#L240))
- this issue stretched beyond just Rancher Projects, as labels and annotations are metadata and may be used for other reasons that cannot always be predetermined

### Steps To Reproduce

1. Deploy a namespace via fleet
2. Move the namespace into a Rancher Project
3. Force Reconcile of the GitRepo

### Environment

```markdown
- Architecture: amd64
- Fleet Version: 0.11.4 (default version with Rancher 2.10.3)
- Cluster:
  - Provider: EKS
  - Options: 2 nodes, basic configuration
  - Kubernetes Version: 1.30.14
```

### Logs

```markdown

```

kkaempf:
This is not (yet) expected to work. Added to Rancher 2.15 backlog.

EdKingscote:
@jamescooke-xyz 

I have adopted the pattern of creating projects with known IDs whereby they have the clustername embedded, which then allows targetCustomizations to set the matching annotations easily. (e.g. `p-cluster1-myproject` `p-cluster2-myproject`) - the displayName for the project is the same across all clusters (but it doesn't have to be)

There are a few issues I've worked around :
 
[...]

PR #5460, opened 2026-07-21
feat(agent): Server-Side Apply for namespace labels and annotations

p-se:
## SSA for namespaces

Manage the release namespace's labels and annotations with server-side apply instead of a read-modify-write update.

The previous approach fetched the namespace, merged in the configured labels/annotations, and then deleted every key not present in the options, preserving only a hardcoded allowlist (kubernetes.io/metadata.name). That clobbered metadata set by other actors -- most notably field.cattle.io/projectId, which Rancher adds when a namespace is moved into a Project -- and the allowlist could never be made complete. See issue #4564.

Fleet now applies with a dedicated field manager ("fleet-agent-namespace-metadata"), so it owns exactly the keys it declares: keys owned by other managers are left untouched, and a key Fleet stops declaring is pruned by the apiserver. ForceOwnership adopts keys previously written by the old update, so the first apply after upgrade does not conflict.

Labels and annotations are applied independently. A bundle that sets only namespaceLabels never asserts ownership of annotations (and so never prunes them), and vice versa.

Security-sensitive labels such as pod-security.kubernetes.io/* are not filtered out of the options; #5687 removed that filter. A bundle that declares one applies it like any other label, gated by the downstream RBAC of the deployment's service account, since the namespace is patched as that account. Deployments that resolve to no service account still run as the agent, so restricting them requires pinning a service account, either in the bundle or through a Policy. A pod-security label already on the namespace that the bundle does not declare is now preserved implicitly: Fleet never declares it, so the apply does not touch it.

The deployment's service account now needs 'patch' on the namespace instead of 'update'; the RBAC error messages are updated accordingly. Namespace creation remains Helm's responsibility -- fetchExistingNamespace only verifies the namespace is present and surfaces a clear RBAC error before the apply.

## Migrate legacy namespace managed fields to enable SSA pruning

Namespaces that predate the switch to server-side apply still carry a "fleetagent" Update field-manager entry that owns their labels and annotations. ForceOwnership on the first post-upgrade apply makes Fleet's SSA manager co-own those keys but does not strip the stale Update entry, and a field is only pruned once no manager owns it. As a result, a key dropped from a bundle stays behind forever on any such namespace.

Add a self-healing, idempotent migration that moves ownership of the metadata.labels/metadata.annotations keys the bundle currently declares from the legacy Update entry into the SSA apply entry, leaving every other field the general-purpose "fleetagent" manager owns untouched, including finalizers and any label or annotation the bundle does not declare. This is scoped deliberately rather than absorbing the whole entry via csaupgrade. It runs before the namespace metadata apply and is a no-op once migrated.

Scoping the migration to the declared keys is what keeps it from taking metadata Fleet has no claim to. The legacy manager cannot be identified any more precisely than by name, and that name is shared: client-go derives it from the binary, and so does Helm. The entry Helm leaves behind when it creates the release namespace, carrying its "name: <namespace>" label, is therefore indistinguishable from the entry the old read-modify-write update left. Absorbing it wholesale would make Fleet own Helm's label and prune it on the next apply. A key Fleet declares, by contrast, is one it is about to own anyway. With Helm 4 the collision is limited to namespaces that predate the SSA switch and to bundles using takeOwnership or force, because Helm 4 otherwise creates the namespace with server-side apply and records the label under an Apply entry of its own.

Known trade-off: a key dropped from the bundle in the same change that upgraded Fleet is never absorbed and stays on the namespace. It self-heals if the key is re-added and then dropped again.

## Tests

Unit tests for the managed-fields migration in internal/cmd/agent/deployer, integration tests in ./integrationtests/agent that drive real BundleDeployment reconciles against envtest, and e2e specs in ./e2e/single-cluster.

Refers to #4564

### Checklist

### DIFF
--- a/internal/cmd/agent/deployer/deployer.go
+++ b/internal/cmd/agent/deployer/deployer.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"maps"
 	"regexp"
 	"slices"
 	"strings"
@@ -23,6 +22,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	corev1ac "k8s.io/client-go/applyconfigurations/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
@@ -231,7 +231,20 @@ func (d *Deployer) helmdeploy(ctx context.Context, logger logr.Logger, bd *fleet
 	return resourceID, nil
 }
 
-// setNamespaceLabelsAndAnnotations updates the namespace for the release, applying all labels and annotations to that namespace as configured in the bundle spec.
+// setNamespaceLabelsAndAnnotations applies the labels and annotations configured
+// in the bundle spec to the release namespace.
+//
+// It server-side applies with a dedicated Fleet field manager, so Fleet owns
+// exactly the keys it declares. Labels and annotations set by other actors (for
+// example Rancher's field.cattle.io/projectId, added when a namespace is moved
+// into a Project) are left untouched, and a key that Fleet stops declaring is
+// pruned. This replaced an earlier read-modify-write that deleted every key not
+// present in the options and could only preserve a hardcoded allowlist (see
+// issue #4564).
+//
+// Labels and annotations are applied independently: a bundle that sets only
+// namespaceLabels never asserts ownership of (and therefore never prunes)
+// annotations, and vice versa.
 func (d *Deployer) setNamespaceLabelsAndAnnotations(ctx context.Context, bd *fleet.BundleDeployment, releaseID string) error {
 	if bd.Spec.Options.NamespaceLabels == nil && bd.Spec.Options.NamespaceAnnotations == nil {
 		return nil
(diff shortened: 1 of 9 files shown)


d3flex (approved):

internal/cmd/agent/deployer/namespace_managed_fields.go
```diff
@@ -0,0 +1,201 @@
...
+// namespaceFieldOwner's Apply entry, leaving any other fields the legacy
+// manager owns untouched. It returns a nil patch (and nil error) if the legacy
+// manager holds no such Update entry, or if that entry does not own any
+// labels/annotations fields.
+func buildManagedFieldsMigrationPatch(ns *corev1.Namespace) ([]byte, error) {
+	entries := ns.ManagedFields
+
+	legacyIdx := -1
+	for i, e := range entries {
+		if e.Manager == legacyNamespaceFieldManager && e.Operation == metav1.ManagedFieldsOperationUpdate && e.Subresource == "" {
```
0xavi0, line 78: What happens with fields owned by Helm?

If my understanding is correct, two things write as manager `fleetagent`, operation Update:

- the old c.Update(ctx, ns) in setNamespaceLabelsAndAnnotations — the intended target
- Helm, when it creates the release namespace with labels: {name: <ns>}, because getManagedFieldsManager() returns filepath.Base(os.Args[0]) and Fleet never overrides it

I think that name: <ns> owned by Helm is gone, right?

internal/cmd/agent/deployer/deployer_ssa_envtest_test.go
```diff
@@ -0,0 +1,213 @@
...
+// TestSetNamespaceLabelsAndAnnotations_ServerSideApply exercises the namespace
+// metadata sync against a real API server (envtest), which — unlike the
+// controller-runtime fake client — tracks server-side-apply field ownership.
+// This is the authoritative check for issue #4564: foreign metadata is
+// preserved, and a key that Fleet stops declaring is pruned.
+//
+// It is skipped unless KUBEBUILDER_ASSETS points at envtest binaries (e.g. via
+// `setup-envtest use -p env`).
+func TestSetNamespaceLabelsAndAnnotations_ServerSideApply(t *testing.T) {
+	if os.Getenv("KUBEBUILDER_ASSETS") == "" {
```
0xavi0, line 28: Is this value going to be valid here? (if it's not the test self skips)

Shouldn't this whole file be moved to `integrationtests` instead?

internal/cmd/agent/deployer/deployer.go
```diff
@@ -333,71 +353,53 @@ func (d *Deployer) namespaceClient(ctx context.Context, bd *fleet.BundleDeployme
...
-	maps.Copy(nsLabels, optLabels)
-
-	// Delete labels not defined in the options.
-	// Keep the `kubernetes.io/metadata.name` label as it is added by kubernetes when creating the namespace.
-	for k := range nsLabels {
-		if _, ok := optLabels[k]; k != corev1.LabelMetadataName && !ok {
-			delete(nsLabels, k)
-		}
-	}
-}
```
0xavi0, line 389: Is this logic from PR #5687 missing in the new code? (I can't find it)

internal/cmd/agent/deployer/namespace_managed_fields.go
p-se: Good catch! The migration now only takes ownership of the keys the bundle's namespaceLabels/-Annotations currently declare, it never touches keys Fleet didn't declare. Like `name`.

internal/cmd/agent/deployer/deployer_ssa_envtest_test.go
p-se: Agreed, it was silently skipped in CI because the unit-test job doesn't set KUBEBUILDER_ASSETS. Moved to integrationtests/agent/bundle_deployment_namespace_metadata_test.go.

internal/cmd/agent/deployer/deployer.go
p-se: Nothing's missing. #5687 removed that logic, and this branch is rebased on top of it. Only the comment explaining why got lost, and I've restored it.

PR #5462, opened 2026-07-22
Skip unsplit Fleet submodules in release preflight

thardeck:
The workflow required a tag for every Fleet submodule on every new final minor version, even when that submodule has not been split out on the release branch yet (e.g. `pkg/helmvalues` before it is backported to a `release/vX.Y` branch). Skip the requirement when `<mod>/go.mod` is absent from the checked-out branch, since rancher/rancher cannot consume it either way.

Also verify that any tag that is required actually points at a commit containing `<mod>/go.mod`, and add the same guard to `bump_fleet_module()` in `release-against-rancher.sh`. Without it, a tag cut before the module split silently resolves to the root github.com/rancher/fleet module instead of failing loudly.

This was the error during the [release run](https://github.com/rancher/fleet/actions/runs/29909263651/job/88888003649):
```
2026-07-22T09:49:11.4332737Z go: downloading github.com/xiang90/probing v0.0.0-20221125231312-a49e3df8f510
2026-07-22T09:49:11.6092101Z ERROR: no go.mod in rancher/rancher requires github.com/rancher/fleet/pkg/helmvalues
2026-07-22T09:49:11.6100320Z ##[error]Process completed with exit code 1.
```

And this is the original commit 0b65aed1b37492c13963a5b12ded0982758e99b5 introducing this additional check, but these changes were not part of `release/v0.16`, so `pkg/helmvalues` was missing a go.mod and can not be imported from this branch.

Successful [release run](https://github.com/rancher/fleet/actions/runs/29938694901) with this branch.

### DIFF
--- a/.github/scripts/release-against-rancher.sh
+++ b/.github/scripts/release-against-rancher.sh
@@ -15,6 +15,17 @@ bump_fleet_module() {
     # module and its nested pkg/apis module), so update every go.mod that
     # requires the module, not just the root one.
     local module="github.com/rancher/fleet/$1"
+    local tag="$1/v${NEW_FLEET_VERSION}"
+
+    # Guard against a tag that predates the module being split out of the main
+    # Fleet module: without its own go.mod, `go get` silently falls back to
+    # resolving the root `github.com/rancher/fleet` module instead.
+    if ! git -C ../fleet cat-file -e "${tag}:$1/go.mod" 2>/dev/null; then
+        printf 'ERROR: %s does not contain %s/go.mod\n' "${tag}" "$1" >&2
+        printf 'The tag does not point at a standalone Go module; refusing to bump.\n' >&2
+        exit 1
+    fi
+
     local modfiles
     modfiles=$(grep -rlE "${module}[[:space:]]" --include=go.mod . || true)
     if [ -z "${modfiles}" ]; then
--- a/.github/scripts/release-against-rancher.sh
+++ b/.github/scripts/release-against-rancher.sh
@@ -15,6 +15,7 @@ bump_fleet_module() {
     # module and its nested pkg/apis module), so update every go.mod that
     # requires the module, not just the root one.
     local module="github.com/rancher/fleet/$1"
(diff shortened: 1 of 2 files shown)


d3flex (approved):

.github/scripts/release-against-rancher.sh
```diff
@@ -15,8 +15,20 @@ bump_fleet_module() {
...
+    # resolving the root `github.com/rancher/fleet` module instead.
+    if ! git -C ../fleet cat-file -e "${tag}:$1/go.mod" 2>/dev/null; then
+        printf 'ERROR: %s does not contain %s/go.mod\n' "${tag}" "$1" >&2
+        printf 'The tag does not point at a standalone Go module; refusing to bump.\n' >&2
+        exit 1
+    fi
+
     local modfiles
-    modfiles=$(grep -rlE "${module}[[:space:]]" --include=go.mod . || true)
+    modfiles=$(grep -rlE "^[[:space:]]*require[[:space:]]+${escaped_module}[[:space:]]|^[[:space:]]+${escaped_module}[[:space:]]" --include=go.mod . || true)
```
weyfonk, line 31: I've tested that this works, but would be interested to better understand how 😅

weyfonk (approved):

.github/scripts/release-against-rancher.sh
thardeck: The regex handles both `go.mod` forms:

- `require github.com/rancher/fleet/pkg/helmvalues v...`
- A dependency line inside a grouped `require (...)` block.

The module path is escaped first because `.` otherwise means “any character” in regex. Anchoring the match to either `require` or whitespace before the module also prevents matching `replace` directives or similarly named modules.

`grep -rl` returns every matching `go.mod`, and the script updates each one. The `|| true` prevents `set -e` from exiting before the script can emit its own clear error when no Rancher module requires the Fleet submodule.

weyfonk: Got it now; thanks for clarifying!

PR #5466, opened 2026-07-23
fix(agent): add jitter to cluster status ticker to prevent thundering…

himabindugit:
## Problem

When agents start together — for example, after a fleet-controller restart or a batch of aircraft powering on in the same window — those agents begin their check-in tickers in sync. Without jitter, every agent in that cohort PATCHes its cluster status at the same instant every 15 minutes.

In steady state, clusters added at different times are naturally spread out. But for clusters that started together (e.g., 50–100 agents from the same restart or power-up window), the synchronised PATCHes arrive as a burst, producing a concentrated spike of watch events on the fleet-controller. This worsens the cache recompilation storm tracked in #5444, since the burst fires the hot path simultaneously across all affected clusters.

Note: The FleetController logs confirm the burst pattern but cannot distinguish cold-start events from synchronized agent PATCHes — Kubernetes audit logs are needed to isolate the agent PATCH timing specifically — investigation ongoing.

## Fix

Add a random sleep of `[0, checkinInterval)` before the periodic ticker loop begins (`ticker.go`). The sleep uses a `select` so the goroutine exits cleanly if the context is cancelled during the jitter window rather than blocking on `time.After`.

Because the offset is random per-agent, the spread is permanent — not just on the first tick:
- Agent A: jitter=3min → fires at 3, 18, 33 ... min
- Agent B: jitter=7min → fires at 7, 22, 37 ... min

## Tests

Two new test cases added to `ticker_test.go`:

- **Context cancellation during jitter** — verifies the goroutine exits cleanly when the context is cancelled before the jitter window expires, so no goroutine leak occurs.
- **Thundering herd spread** — starts 5 concurrent agents simultaneously and asserts their first periodic patches are spread across time rather than bunched at t=0.

Existing test (`should patch the cluster status after checkinInterval`) continues to pass — jitter does not prevent eventual check-in.

Refers to https://github.com/rancher/fleet/issues/5444

## Additional Information

This fix was developed in collaboration with [Claude Code](https://claude.ai/code) (Anthropic) as part of a broader
investigation into fleet-controller CPU spikes at production scale.

### Checklist

- [ ] I have updated the documentation via a pull request in the [fleet-product-docs](https://github.com/rancher/fleet-product-docs) repository.

### DIFF
--- a/internal/cmd/agent/clusterstatus/ticker.go
+++ b/internal/cmd/agent/clusterstatus/ticker.go
@@ -44,14 +44,18 @@ func Ticker(ctx context.Context, client client.Client, agentNamespace string, cl
 		}
 	}()
 	go func() {
-		if checkinInterval == 0 {
+		if checkinInterval <= 0 {
 			checkinInterval = durations.DefaultClusterCheckInterval
 		}
 		// Spread agent check-ins across the interval window to prevent a thundering
 		// herd on the fleet-controller when many agents start at the same time.
+		timer := time.NewTimer(rand.N(checkinInterval))
 		select {
-		case <-time.After(rand.N(checkinInterval)):
+		case <-timer.C:
 		case <-ctx.Done():
+			if !timer.Stop() {
+				<-timer.C
+			}
 			return
 		}
 		for range ticker.Context(ctx, checkinInterval) {
--- a/internal/cmd/agent/clusterstatus/ticker.go
+++ b/internal/cmd/agent/clusterstatus/ticker.go
@@ -3,6 +3,7 @@ package clusterstatus
 
 import (
 	"context"
+	"math/rand/v2"
 	"time"
 
 	fleet "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
@@ -45,6 +46,13 @@ func Ticker(ctx context.Context, client client.Client, agentNamespace string, cl
 	go func() {
 		if checkinInterval == 0 {
 			checkinInterval = durations.DefaultClusterCheckInterval
+		}
+		// Spread agent check-ins across the interval window to prevent a thundering
+		// herd on the fleet-controller when many agents start at the same time.
+		select {
+		case <-time.After(rand.N(checkinInterval)):
+		case <-ctx.Done():
+			return
 		}
 		for range ticker.Context(ctx, checkinInterval) {
 			logger.V(1).Info("Reporting cluster status")
--- a/internal/cmd/agent/clusterstatus/ticker_test.go
+++ b/internal/cmd/agent/clusterstatus/ticker_test.go
@@ -102,8 +102,8 @@ var _ = Describe("ClusterStatus Ticker", func() {
 	})
 
 	It("should stop periodic patches when context is cancelled", func() {
-		// Verifies the jitter select respects ctx.Done so the goroutine
-		// exits cleanly instead of blocking forever on time.After.
+		// Ensures the periodic ticker loop stops when ctx is cancelled,
+		// so the agent doesn't keep patching after shutdown.
 		checkinInterval = time.Millisecond * 50
 
 		var patchCount atomic.Int32
@@ -145,7 +145,7 @@ var _ = Describe("ClusterStatus Ticker", func() {
 		checkinInterval = time.Millisecond * 200
 
 		var mu sync.Mutex
-		firstPatchTimes := make([]time.Time, 0, agentCount)
+		firstPatchTimes := make(map[string]time.Time, agentCount)
 		allDone := make(chan struct{})
 
 		buildClient := func(name string) client.Client {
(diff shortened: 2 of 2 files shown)


d3flex (changes requested):
The new `rand.N(checkinInterval)` thing only staggers the periodic ticks that come after the startup. If the production incident was a startup burst rather than middle lifecycle drift, this doesn't cover it. is that intentionally out of scope?

himabindugit:
> The new `rand.N(checkinInterval)` thing only staggers the periodic ticks that come after the startup. If the production incident was a startup burst rather than middle lifecycle drift, this doesn't cover it. is that intentionally out of scope?

Good catch — you're right, the startup burst was not covered.

More details:
Fixed: the startup check-in goroutine now sleeps ClusterRegisterDelay + rand.N(checkinInterval) instead of a fixed ClusterRegisterDelay. This spreads initial check-ins across the same interval window as periodic ticks, covering the scenario where many agents restart simultaneously (e.g. after a fleet-controller recovery). Context cancellation is also handled properly during the extended delay.

d3flex:
Same as https://github.com/rancher/fleet/pull/5445#pullrequestreview-4971447075. commit history needs some work, if you do not mind

himabindugit:
> Same as [#5445 (review)](https://github.com/rancher/fleet/pull/5445#pullrequestreview-4971447075). commit history needs some work, if you do not mind

Done — consolidated into 2 logical commits with bodies and signatures:

- fix(agent): add jitter to cluster status ticker to prevent thundering herd — periodic ticker jitter + test fixes
- fix(agent): add startup jitter to spread initial check-ins on agent restart — startup goroutine jitter covering the mass-restart scenario

internal/cmd/agent/clusterstatus/ticker.go
```diff
@@ -39,16 +40,37 @@ func Ticker(ctx context.Context, client client.Client, agentNamespace string, cl
 	}
 
 	go func() {
-		time.Sleep(durations.ClusterRegisterDelay)
+		// Fixed registration delay plus jitter to spread startup check-ins when
+		// many agents restart simultaneously (e.g. after a fleet-controller recovery).
+		timer := time.NewTimer(durations.ClusterRegisterDelay + rand.N(checkinInterval))
```
d3flex, line 50: it will be wise to add a check for the `checkinInterval`. in the case of `<= 0` it will panic. see https://pkg.go.dev/math/rand/v2#N

d3flex (commented):
Thanks for the changes but I would like to ask you to rebase the 6 commits and keep only yours. let me know if you need any help. I will continue review in the meantime.

himabindugit:
> Thanks for the changes but I would like to ask you to rebase the 6 commits and keep only yours. let me know if you need any help. I will continue review in the meantime.

Done — rebased onto upstream/main and force-pushed. Branch now contains only my 2 commits.

internal/cmd/agent/clusterstatus/ticker.go
himabindugit: Good catch! Moved the <= 0 guard to before both goroutines so rand.N is protected in the startup jitter path as well. Also removed the duplicate guard from inside the second goroutine since it's now handled once at the top.

internal/cmd/agent/clusterstatus/ticker_test.go
```diff
@@ -98,4 +100,107 @@ var _ = Describe("ClusterStatus Ticker", func() {
...
+		for _, t := range firstPatchTimes {
+			if earliest.IsZero() || t.Before(earliest) {
+				earliest = t
+			}
+			if t.After(latest) {
+				latest = t
+			}
+		}
+		spread := latest.Sub(earliest)
+		Expect(spread).To(BeNumerically(">", checkinInterval/10),
```
weyfonk: Where does this `10` come from? How can we be sure that it will not make this test case flaky?

internal/cmd/agent/clusterstatus/ticker_test.go
```diff
@@ -98,4 +100,107 @@ var _ = Describe("ClusterStatus Ticker", func() {
...
+			"jitter should spread %d agents' first check-in across time, not bunch them at t=0", agentCount)
```
weyfonk: Where does this `10` come from? How can we be sure that it will not make this test case flaky, since the implementation does not mandate anything about the _distribution_ of those timestamps?
Running this repeatedly on my machine made it fail after a few hundred attempts.

Perhaps simply checking that each check-in timestamp is different to all others could be enough, e.g.:
```suggestion
        var foundTimestamps []time.Time
        for _, t := range firstPatchTimes {
            Expect(t.IsZero()).To(BeFalse(), "no check-in timestamp should be 0")
            Expect(foundTimestamps).ToNot(ContainElement(t), "no check-in timestamp should be duplicated between agents")
            foundTimestamps = append(foundTimestamps, t)
        }
```

d3flex (changes requested):
Apply a commit to address commits, remove duplication and introduce a jitter window on initial propose

d3flex (approved):

Issue #4511, opened 2026-01-09
Store fields relative to helm values secrets in the cli monitor


0xavi0:
It would be interesting to store the `Bundle.Spec.ValuesHash` and the `BundleDeployment.Spec.ValuesHash` so we can verify that those are propagated correctly upon a new commit that changes the values.

We could also store the hash of the secret, so we can compare if the supposed hash that is stored in _Bundles_ and _BundleDeployments_ matches the secret. 

We're facing some issues with certain customers that tend to modify helm values only in their commits. 
When that happens _Content_ resources are not updated (resources of the _Bundle_ are the same) and  only the helm values secrets and its related fields are updated.

d3flex:
#5467 reveals the ValuesHash on the monitor's output. 
```bash
❯ fleetcli monitor -n fleet-default | jq -r '.bundles.[].valuesHash'
null
null
9eb5dab08e428001d159f9c2a0c7bc15effcdb039732aa8e5d808d14d2cccecb
null
```

However i do not know if that satisfy the aim of this ticket. Which in my understanding is to help when some values are mismatch.
Is it likely that other values are expected to assist debugging the situation described in the description? 
would it make sense to add a detectSmth* for the `analyze` command?

d3flex:
## Additional QA
### Issue

Refers to #4511

### Problem

When a commit changes only the Helm values, with no manifest or template change,
the bundle's resources stay identical, so nothing people normally look at moves. If
the new values fail to reach a cluster, there was no way to see that and no way to
tell how far they got.

### Solution

Exposed the hash of the values from Bundle and on each BundleDeployment to the `fleetcli`
monitor and have the analyze subcommand to detect and report inconsistenies.

Because both the Bundle and the BundleDeployment are checked, a mismatch tells you
*where* the values stopped: at the Bundle, or on the way to a particular cluster.

The check stays quiet in situations that only look wrong: a secret being deleted,
one whose owner is gone or was recreated, and a BundleDeployment flagged as waiting
for its values, where the secret is missing or not yet propagated and the
disagreement is expected.

### Testing

Deploy a Bundle with Helm values (GitRepo or HelmOp), then change only those
values. Compare the hashes in `fleetcli monitor` before and after.

### Engineering Testing
#### Manual Testing

Tested on a single-cluster k3d setup (`fleet-local`, sharded controller), using a
HelmOp deploying `oci://ghcr.io/stefanprodan/charts/podinfo` with inline
[...]

PR #5467, opened 2026-07-24
feat(cli): surface helm values hash in monitor snapshot

d3flex:
Provide a way to tell whether a resource reconciled with the helm values it was expected to.

Refs #4511

### DIFF
--- a/pkg/troubleshooting/analyzer.go
+++ b/pkg/troubleshooting/analyzer.go
@@ -294,6 +294,16 @@ func OutputIssues(w io.Writer, snapshots []*Snapshot) error {
 		}
 	}
 
+	if len(diag.SecretsWithValuesHashMismatch) > 0 {
+		hasIssues = true
+		fmt.Fprintln(w)
+		printWarning(w, fmt.Sprintf("Secrets with ValuesHash Mismatch (%d)", len(diag.SecretsWithValuesHashMismatch)))
+		for _, s := range diag.SecretsWithValuesHashMismatch {
+			fmt.Fprintf(w, "  • %s %s/%s (secret: %s, spec: %s)\n",
+				s.OwnerKind, s.Namespace, s.Name, s.SecretHash, s.SpecHash)
+		}
+	}
+
 	if !hasIssues {
 		printSuccess(w, "No issues detected!")
 	}
@@ -489,6 +499,10 @@ func printDiagnosticsSummary(w io.Writer, diag *Diagnostics) {
 	// Commit Mismatches
 	commitIssues := len(diag.GitReposWithCommitMismatch)
 	fmt.Fprintf(tw, "  Commit Mismatches:\t%d\t%s\n", commitIssues, statusIcon(commitIssues))
+
+	// ValuesHash Mismatches
+	valuesHashIssues := len(diag.SecretsWithValuesHashMismatch)
+	fmt.Fprintf(tw, "  ValuesHash Mismatches:\t%d\t%s\n", valuesHashIssues, statusIcon(valuesHashIssues))
 
 	// GitRepos last polled too long ago
 	pollingIssues := len(diag.GitReposUnpolled)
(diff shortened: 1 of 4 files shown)


pkg/troubleshooting/analyzer.go
```diff
@@ -294,6 +294,16 @@ func OutputIssues(w io.Writer, snapshots []*Snapshot) error {
 		}
 	}
 
+	if len(diag.SecretsWithValuesHashMismatch) > 0 {
+		hasIssues = true
+		fmt.Fprintln(w)
+		printWarning(w, fmt.Sprintf("Secrets with ValuesHashes Mismatch (%d)", len(diag.SecretsWithValuesHashMismatch)))
+		for _, s := range diag.SecretsWithValuesHashMismatch {
```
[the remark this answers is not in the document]
d3flex: leave it as is because I follow the convention from the others `fmt.Sprintf`

pkg/troubleshooting/collector.go
```diff
@@ -828,6 +846,64 @@ func (col *Collector) filterSecretsWithInvalidOwners(orphanedSecrets []corev1.Se
...
+	bundleHashes := make(map[string]string)
+	bundleUIDs := make(map[string]string)
+	for _, b := range bundles {
+		key := b.Namespace + "/" + b.Name
+		bundleHashes[key] = b.Spec.ValuesHash
+		bundleUIDs[key] = string(b.UID)
+	}
+
+	bundleDeploymentHashes := make(map[string]string)
+	bundleDeploymentUIDs := make(map[string]string)
```
0xavi0, line 863: Fleet also tracks `BundleDeployments` that could be healing its `BundleDeployment` secret. 
In that case `bd.Spec.WaitingForValues == true`
When that happens the Hash is empty, and we we can skip it.


I would add a new map here:


```suggestion
	bundleDeploymentUIDs := make(map[string]string)
	bundleDeploymentsWaiting := make(map[string]bool)
```

pkg/troubleshooting/collector.go
```diff
@@ -828,6 +846,64 @@ func (col *Collector) filterSecretsWithInvalidOwners(orphanedSecrets []corev1.Se
...
+	for _, bd := range bundleDeployments {
+		key := bd.Namespace + "/" + bd.Name
+		bundleDeploymentHashes[key] = bd.Spec.ValuesHash
+		bundleDeploymentUIDs[key] = string(bd.UID)
```
0xavi0, line 868: Continues from previous message:
```suggestion
		bundleDeploymentUIDs[key] = string(bd.UID)
		bundleDeploymentsWaiting[key] = bd.Spec.WaitingForValues
```

pkg/troubleshooting/collector.go
```diff
@@ -828,6 +846,64 @@ func (col *Collector) filterSecretsWithInvalidOwners(orphanedSecrets []corev1.Se
...
+		}
+
+		owner := secret.OwnerReferences[0]
+		key := secret.Namespace + "/" + owner.Name
+		var specHash, expectedUID string
+		switch owner.Kind {
+		case "Bundle":
+			specHash = bundleHashes[key]
+			expectedUID = bundleUIDs[key]
+		case "BundleDeployment":
```
0xavi0, line 885: Continues from the previous message:

```suggestion
		case "BundleDeployment":
		if bundleDeploymentsWaiting[key] {
              continue
      }
```

pkg/troubleshooting/collector.go
```diff
@@ -828,6 +846,64 @@ func (col *Collector) filterSecretsWithInvalidOwners(orphanedSecrets []corev1.Se
 	return invalidOwners
 }
 
+// detectSecretsWithValuesHashMismatch detects bundle lifecycle secrets whose
+// content hash does not match their owner's Spec.ValuesHash.
+func (col *Collector) detectSecretsWithValuesHashMismatch(bundleSecrets []corev1.Secret, bundles []fleet.Bundle, bundleDeployments []fleet.BundleDeployment) []ValuesHashMismatch {
```
0xavi0, line 852: Can you add tests for this function?

pkg/troubleshooting/collector.go
```diff
@@ -354,11 +357,24 @@ func (col *Collector) convertSecrets(secrets []corev1.Secret) []SecretInfo {
 			info.OwnerUID = string(owner.UID)
 		}
 
+		info.ValuesHash = secretValuesHash(s)
 		result = append(result, info)
 	}
 	return result
 }
 
+func secretValuesHash(s corev1.Secret) string {
```
0xavi0, line 367: Please add tests for this function

0xavi0 (approved):

PR #5469, opened 2026-07-27
Run Ginkgo suites through go test

thardeck:
Use Go test for Ginkgo suites to remove redundant CLI setup. Disable test caching so CI preserves one-run suite behavior.

`gingko unfocus` was accidentally introduced by 4f84c1598e8fa0627b71dcfbbb84fdaad7caa787.

### DIFF
--- a/.github/scripts/check-for-auto-generated-changes.sh
+++ b/.github/scripts/check-for-auto-generated-changes.sh
@@ -2,7 +2,6 @@
 set -euo pipefail
 
 go generate
-ginkgo unfocus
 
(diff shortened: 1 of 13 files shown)


0xavi0 (commented):
I wonder if we'd need the `-v` flag to keep the `GinkgoWriter` working as before this PR.
Some tests, for example, the ones using ` --flake-attempts=2` will be silent and we'll don't know if it retried or not.

0xavi0 (approved):

PR #5509, opened 2026-08-04
ci(release): checkout Fleet twice to fix ref skew in release workflow

d3flex:
The "Release Fleet against rancher/rancher" workflow checked out the Fleet repo once, at inputs.fleet_branch, into fleet/, and ran every script from that checkout. A workflow_dispatch always executes the workflow YAML itself from the dispatched ref (main), so this paired main's calling convention with whatever copy of the scripts happened to sit on the release branch.

That skew broke fleet_branch=release/v0.15: PR #4796 dropped the should_bump_api input and rewrote release-against-rancher.sh to take 2 args, deriving the API bump from a pkg/apis tag instead. main's workflow now passes 2 args, but release/v0.15 still has the pre-#4796 script, whose line 9 reads $3 under set -ue and fails with "unbound variable". PR #5508 backported compute-rancher-versions.sh to that branch and only moved the failure one step further down the chain, because it fixed the same skew for one script without removing the underlying mismatch.

Design choice: check Fleet out twice instead of backporting scripts. fleet/ has no ref, so it lands on the workflow's own ref, and every script now runs from there. fleet-release/ stays pinned to inputs.fleet_branch with fetch-depth: 0, and is read only as a data source for git tags and which submodules are split out on that branch. The release checkout is passed to the scripts as FLEET_REPO_DIR, absolute for the release-against-rancher.sh step because that script pushd's into the rancher checkout before reading it. In the script, three hardcoded `git -C ../fleet` reads become `git -C "${FLEET_REPO_DIR}"`, defaulting to the script's own checkout so manual invocation is unaffected.

Behavior change: none for the tag reads themselves. fetch-depth: 0 makes git tag/cat-file lookups branch-independent, so pointing FLEET_REPO_DIR at either checkout returns the same answer today. The change is about sourcing scripts and tooling from a single consistent ref, not about changing what is read.

Intentionally left out: no backport to release branches. Each release branch already carries a self-consistent workflow+script pair (its own manual-input workflow and matching 3-arg script). The old failure only happened because main's YAML was mixed with a release branch's scripts; running scripts from the workflow's own ref permanently removes that mix, so release branches never need this backport again.

### DIFF
--- a/.github/scripts/release-against-rancher.sh
+++ b/.github/scripts/release-against-rancher.sh
@@ -21,7 +21,7 @@ bump_fleet_module() {
     # Guard against a tag that predates the module being split out of the main
     # Fleet module: without its own go.mod, `go get` silently falls back to
     # resolving the root `github.com/rancher/fleet` module instead.
-    if ! git -C ../fleet cat-file -e "${tag}:$1/go.mod" 2>/dev/null; then
+    if ! git -C "${FLEET_REPO_DIR}" cat-file -e "${tag}:$1/go.mod" 2>/dev/null; then
         printf 'ERROR: %s does not contain %s/go.mod\n' "${tag}" "$1" >&2
         printf 'The tag does not point at a standalone Go module; refusing to bump.\n' >&2
         exit 1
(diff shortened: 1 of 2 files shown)


d3flex:
closing in favor of https://github.com/rancher/fleet/pull/5511

Issue #4455, opened 2025-12-17
Event generation around deployment states


weyfonk:
Fleet already [generates events](https://fleet.rancher.io/observability) related to git operations, such as received commits.
However, once git operations have succeeded, nothing is reported on the state of deployments.

Since the unit of deployment on a given target cluster is a bundle deployment, event generation could be added to the bundle deployment controller (part of upstream Fleet controllers), reporting at least failed deployments.

While this could lead to the Kubernetes API server being flooded with calls on setups involving large numbers of bundle deployments, client-go supports [event aggregation](https://github.com/kubernetes/client-go/blob/a296bd743c2dd954f56664f6ee7bc8558f7d8e32/tools/record/events_cache.go#L161) with a configurable time window. That aggregation works against events with different messages, which would fit our use case.

## Acceptance criteria

- [ ] Fleet generates events for failed bundle deployments
- [ ] A large number (e.g. 100 or 1000) of bundle deployments failing within a given window (e.g. 1 minute) lead to a _single_ event being generated
    - [ ] Can that event contain information about every single failing bundle deployment?

This could be extended to successful bundle deployments too.

PR #5515, opened 2026-08-05
Report bundle deployment state as Kubernetes events

0xavi0:
Fleet exposes deployment failures only in bundle status, which is easy to miss. This adds events on the bundle when its deployments fail, and when they all become ready again, so `kubectl describe bundle` and any event- based tooling can see them.

Events are created directly rather than through client-go's recorder: the recorder keeps the note of the first event in a series, which would drop notes describing different failures of the same bundle. Deduplication happens on a fingerprint of the failure causes and how many deployments are affected instead, so a burst collapses into one accurate event, and comparing against the last persisted status keeps that true across controller restarts.

Configurable via the `deploymentEvents` block in the fleet chart: debounce, minimum interval per object, recovery reporting, how many causes an event names, and optional per-bundle-deployment events (off by default). Reporting is on by default and can be turned off entirely.

Refers to: https://github.com/rancher/fleet/issues/4455

### Checklist

### DIFF
--- a/charts/fleet/templates/rbac.yaml
+++ b/charts/fleet/templates/rbac.yaml
@@ -34,6 +34,7 @@ rules:
   - '*'
 - apiGroups:
     - ""
+    - "events.k8s.io"
   resources:
     - 'events'
   verbs:
--- a/charts/fleet/values.yaml
+++ b/charts/fleet/values.yaml
@@ -30,6 +30,34 @@ garbageCollectionInterval: "15m"
 # Whether you want to allow cluster upon registration to specify their labels.
 ignoreClusterRegistrationLabels: false
 
+# Events reporting the deployment state of bundles, on the bundle they are about.
+deploymentEvents:
+  # Turns off events about the deployment state of bundles. Unsetting the whole
+  # block keeps the defaults below, which report.
+  disabled: false
+  # How long to wait for a burst of failures to settle before reporting it, so
+  # that the event describes the whole burst instead of only its first failure.
+  # A bundle failing on 500 clusters over 3s reports once, as "500/500 bundle
+  # deployments failing", instead of first reporting "1/500".
+  debounce: 5s
+  # The minimum time between two events about the same object. It bounds how
+  # often a flapping deployment reports: one failing and recovering every 10s
+  # reports about once a minute, instead of on every transition.
+  minInterval: 1m
+  # Report bundles whose deployments all became ready again after a failure.
+  # Only bundles which were failing report, so healthy ones stay silent.
+  reportRecovery: true
+  # Additionally report each failing bundle deployment, in the namespace of its
+  # cluster. This is more detailed, but produces one event per failing
+  # deployment, of which there is one per cluster a bundle is deployed to: a
+  # bundle failing on 500 clusters reports 501 events instead of 1.
+  perDeployment: false
+  # How many distinct failure causes a single event describes. The bundle's
+  # status.summary.nonReadyResources lists more of them, up to 10, which is
+  # therefore the highest value with an effect. Causes beyond it are reported as
+  # "(+117 more, see status.summary.nonReadyResources)".
+  maxCauses: 3
+
 # Counts from gitrepo are out of sync with bundleDeployment state.
 # Just retry in a number of seconds as there is no great way to trigger an event that doesn't cause a loop.
 # If not set default is 15 seconds.
--- a/internal/cmd/controller/bundleevents/note.go
+++ b/internal/cmd/controller/bundleevents/note.go
@@ -1,0 +1,188 @@
+package bundleevents
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+
+	fleet "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
+)
+
+const (
+	// noteMaxLength stays below the 1kB limit the API enforces on an event's note.
+	noteMaxLength = 1000
+
+	// messageMaxLength bounds a single cause, so that one long Helm error
+	// cannot take up the whole note.
+	messageMaxLength = 200
+
+	// statusHint points at the bundle status, which lists more causes than
+	// fit into a note.
+	statusHint = "see status.summary.nonReadyResources"
+)
+
+// failureCounts returns the number of bundle deployments which failed to deploy
+// or are deployed but not ready, together with the state to report for them.
+// Transient states, like Pending or WaitApplied, are not failures: they are
+// reported once they either settle or turn into a failure.
+func failureCounts(s fleet.BundleSummary) (int, fleet.BundleState) {
+	switch {
+	case s.ErrApplied > 0:
+		// ErrApplied outranks NotReady, see fleet.StateRank.
+		return s.ErrApplied + s.NotReady, fleet.ErrApplied
+	case s.NotReady > 0:
+		return s.NotReady, fleet.NotReady
+	}
+
+	return 0, fleet.Ready
+}
+
+// failureReason is the reason to report for a failure state.
+func failureReason(state fleet.BundleState) string {
+	if state == fleet.NotReady {
+		return ReasonNotReady
+	}
+
+	return ReasonDeployFailed
+}
+
+// failureFingerprint describes the failures of a summary: how many deployments
+// are affected, the reason to report them under, and a fingerprint which
+// changes whenever the failures are worth reporting again.
+func failureFingerprint(s fleet.BundleSummary) (int, string, fingerprint) {
+	failing, state := failureCounts(s)
+	if failing == 0 {
+		return 0, "", fingerprint{}
+	}
+
+	reason := failureReason(state)
+
+	return failing, reason, fingerprintOf("failing", reason, magnitude(failing), causeKeys(s))
+}
+
+// failureCauses returns the failing entries of the summary's non-ready
+// resources. The summary stores at most 10 of them.
+func failureCauses(s fleet.BundleSummary) []fleet.NonReadyResource {
+	causes := make([]fleet.NonReadyResource, 0, len(s.NonReadyResources))
+	for _, r := range s.NonReadyResources {
+		if r.State == fleet.ErrApplied || r.State == fleet.NotReady {
+			causes = append(causes, r)
+		}
+	}
+
+	return causes
+}
+
+// causeKeys returns the distinct state and message pairs of the failing
+// resources, sorted, so that the same set of causes always produces the same
+// keys. Cluster names are deliberately left out: the same failure on more
+// clusters is not a new cause and must not produce another event.
+func causeKeys(s fleet.BundleSummary) []string {
+	seen := make(map[string]struct{})
+	keys := make([]string, 0, len(s.NonReadyResources))
+
+	for _, c := range failureCauses(s) {
+		key := string(c.State) + "|" + c.Message
+		if _, ok := seen[key]; ok {
+			continue
+		}
+		seen[key] = struct{}{}
+		keys = append(keys, key)
+	}
+	sort.Strings(keys)
+
+	return keys
+}
+
+// magnitude buckets a number of failures, so that a deployment failing on one
+// more cluster does not warrant another event, while an order of magnitude
+// more, or fewer, failures does.
+func magnitude(n int) string {
+	switch {
+	case n <= 1:
+		return "1"
+	case n <= 5:
+		return "2-5"
+	case n <= 20:
+		return "6-20"
+	case n <= 100:
+		return "21-100"
+	case n <= 1000:
+		return "101-1000"
+	}
+
+	return "1000+"
+}
+
+// failureNote describes how many bundle deployments are failing and why, for up
+// to maxCauses distinct clusters. Causes which do not fit are represented by a
+// count and a pointer to the bundle status, which holds more of them.
+func failureNote(s fleet.BundleSummary, maxCauses int) string {
+	failing, _ := failureCounts(s)
+
+	var b strings.Builder
+	fmt.Fprintf(&b, "%d/%d bundle deployments failing", failing, s.DesiredReady)
+	if s.ErrApplied > 0 && s.NotReady > 0 {
+		fmt.Fprintf(&b, " (errApplied %d, notReady %d)", s.ErrApplied, s.NotReady)
+	}
+	b.WriteString(".")
+
+	// Leave room for the trailing hint, which is more useful than one more cause.
+	budget := noteMaxLength - len(statusHint) - 32
+
+	causes := failureCauses(s)
+	written := 0
+	for _, c := range causes {
+		if written >= maxCauses {
+			break
+		}
+
+		cause := fmt.Sprintf(" %s: %s: %s;", c.Name, c.State, truncate(c.Message, messageMaxLength))
+		if b.Len()+len(cause) > budget {
+			break
+		}
+
+		b.WriteString(cause)
+		written++
+	}
+
+	// The hint counts the causes left in the status, not the failing
+	// deployments: the summary records at most 10 causes for any number of
+	// them, so counting deployments would point at causes which are not
+	// there. How many deployments are failing is already in the first
+	// sentence.
+	if remaining := len(causes) - written; remaining > 0 {
+		fmt.Fprintf(&b, " (+%d more, %s)", remaining, statusHint)
+	}
+
+	return truncate(b.String(), noteMaxLength)
+}
+
+// readyNote describes a bundle whose deployments are all ready again.
+func readyNote(s fleet.BundleSummary) string {
+	return fmt.Sprintf("%d/%d bundle deployments ready", s.Ready, s.DesiredReady)
+}
+
+// deploymentNote describes the state of a single bundle deployment.
+func deploymentNote(bundle string, state fleet.BundleState, message string) string {
+	if bundle == "" {
+		bundle = "unknown"
+	}
+	if message == "" {
+		return fmt.Sprintf("bundle %s: %s", bundle, state)
+	}
+
+	return truncate(fmt.Sprintf("bundle %s: %s: %s", bundle, state, message), noteMaxLength)
+}
+
+// truncate shortens s to max bytes, without cutting a rune in half.
+func truncate(s string, max int) string {
+	if len(s) <= max {
+		return s
+	}
+	if max <= 3 {
+		return strings.ToValidUTF8(s[:max], "")
+	}
+
+	return strings.ToValidUTF8(s[:max-3], "") + "..."
+}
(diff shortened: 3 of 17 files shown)


internal/cmd/controller/bundleevents/notifier.go
```diff
@@ -0,0 +1,638 @@
...
+	namespace := snap.regarding.Namespace
+	if namespace == "" {
+		namespace = metav1.NamespaceDefault
+	}
+
+	return e.client.Create(ctx, &eventsv1.Event{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s.%x", snap.regarding.Name, now.UnixNano()),
+			Namespace: namespace,
+		},
```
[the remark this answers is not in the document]
0xavi0: Valid comment, changed as suggested

internal/cmd/controller/bundleevents/notifier.go
```diff
@@ -0,0 +1,638 @@
...
+
+		if err := e.create(ctx, p.snap); err != nil {
+			e.createFailed(ctx, p.uid, p.snap, now, err)
+
+			continue
+		}
+
+		e.recordEmit(p.uid, p.snap.fp, now, opts)
+	}
+}
```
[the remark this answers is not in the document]
0xavi0: Changed as suggested

p-se:
I've tested it and I'm a bit confused about the events it reports:

```
  Summary:
    Desired Ready:  1
    Ready:          1
  Unavailable:      0
Events:
  Type     Reason                  Age   From               Message
  ----     ------                  ----  ----               -------
  Warning  BundleNotReady          85s   fleet-bundle-ctrl  1/1 bundle deployments failing. fleet-default/bds-events-multi-ds1: NotReady: ;
  Normal   BundleDeploymentsReady  24s   fleet-bundle-ctrl  0/1 bundle deployments ready
```

## Reporting ready when it is not

`Normal   BundleDeploymentsReady  38m   fleet-bundle-ctrl  0/1 bundle deployments ready`

I'm confused that it reports ready while it isn't (`0/1`). Could we resolve this by adding a generic "InProgress" state? Or is that an issue with our states?

## No Reason

It looks like there should be a reason for why it isn't ready, but it's missing in all cases where `BundleNotReady` is emitted.

## How I reproduced it

Two clusters, no special timing, the agent stays running and the whole thing takes about a minute.

A bundle that fails, because the image does not exist:

```yaml
kind: Bundle
metadata:
  name: repro-5515
  namespace: fleet-default
spec:
  resources:
    - name: deploy.yaml
      content: |
        apiVersion: apps/v1
        kind: Deployment
        metadata:
          name: repro-5515
          namespace: default
        spec:
          replicas: 1
          selector:
            matchLabels:
              app: repro-5515
          template:
            metadata:
              labels:
                app: repro-5515
            spec:
              containers:
                - name: app
                  image: registry.invalid/nope:v1
  targets:
    - clusterSelector: {}
```

A fix that is ready the moment it is applied:

```yaml
kind: Bundle
metadata:
  name: repro-5515
  namespace: fleet-default
spec:
  resources:
    - name: cm.yaml
      content: |
        apiVersion: v1
        kind: ConfigMap
        metadata:
          name: repro-5515
          namespace: default
        data:
          fixed: "true"
  targets:
    - clusterSelector: {}
```

```sh
kubectl apply -f broken.yaml   # goes NotReady
sleep 15                       # let the failure event be written
kubectl apply -f fixed.yaml    # ready ~2s later
sleep 60
kubectl -n fleet-default describe bundle repro-5515
```

0xavi0:
Thanks for catching this, @p-se! It is indeed a bug when aggregating the BDs. 
I'm working on fixing this and also noticed the event reason could be better.

0xavi0:
I've added changes to report correctly the number of BDs that failed as well as changes for reporting the events reasons.

I can see the events as below now:
```
  Type     Reason          Age   From               Message
  ----     ------          ----  ----               -------
  Warning  BundleNotReady  69s   fleet-bundle-ctrl  2/2 bundle deployments failing. fleet-default/dev: NotReady: ; fleet-default/prod: NotReady: ;
  Normal   BundleReady     9s    fleet-bundle-ctrl  2/2 bundle deployments ready
```

internal/cmd/controller/bundleevents/notifier.go
```diff
@@ -0,0 +1,813 @@
...
+	// ReasonBundleDeploymentNotReady reports a single bundle deployment which applied its resources, but those are not ready.
+	ReasonBundleDeploymentNotReady = "BundleDeploymentNotReady"
+
+	// ReasonBundleDeploymentReady reports a single bundle deployment which is ready again.
+	ReasonBundleDeploymentReady = "BundleDeploymentReady"
+
+	// actionDeploy is the operation all events reported here relate to.
+	actionDeploy = "Deploy"
+
+	// tickInterval is how often pending events are checked for being due.
```
weyfonk, line 65: A pending event is an event which exists in this controller's memory, but which has not yet been created from the K8s API's point of view (possibly waiting for a state to settle before generation, for instance); correct?

internal/cmd/controller/bundleevents/notifier.go
```diff
@@ -0,0 +1,813 @@
...
+// Notifier observes objects after their status has been updated, and reports
+// state changes worth an event.
+type Notifier interface {
+	// ObserveBundle reports failures and recoveries of a bundle's
+	// deployments. Old is the bundle as previously persisted, which makes
+	// the state change survive a restart of the controller.
+	ObserveBundle(ctx context.Context, old, cur *fleet.Bundle)
+
+	// ObserveBundleDeployment reports a single bundle deployment, if
+	// per-deployment reporting is enabled.
```
weyfonk, line 96: nit: the comment on what `old` represents, present in `ObserveBundle` documentation, could be added here too to help our future selves and their IDEs :)

internal/cmd/controller/bundleevents/notifier.go
```diff
@@ -0,0 +1,813 @@
...
+	reason    string
+	note      string
+	fp        fingerprint
+}
+
+// entry is the emitter's memory of one object.
+type entry struct {
+	// emitted is the fingerprint of the last event created for the object.
+	emitted fingerprint
+	hasSent bool
```
weyfonk, line 120:
```suggestion

    // hasSent is true if an event has already been sent for this object, independently of possible pending events.
	hasSent bool
```

internal/cmd/controller/bundleevents/options.go
```diff
@@ -0,0 +1,106 @@
...
+func OptionsFromConfig(cfg *config.Config) Options {
+	opts := DefaultOptions()
+	if cfg == nil {
+		return opts
+	}
+
+	events := cfg.DeploymentEvents
+
+	if events.Disabled {
+		opts.Enabled = false
```
weyfonk, line 80: nit: does it even make sense to configure the rest if event creation is disabled?

internal/cmd/controller/bundleevents/options.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	if events.MaxCauses > 0 {
+		opts.MaxCauses = events.MaxCauses
+	}
+
+	return opts
+}
+
+// OptionsFromGlobalConfig reads the options from the fleet controller's config.
+// It is read on every use, so that changes to the configmap take effect without
```
weyfonk, line 102: The controller's config is read each time this function is called, yet in practice, with this function only being called when instantiating the controller (through its `operator.go`), my understanding is that the config is only read at controller start-up time. Subsequent updates to the config map would be missed by this logic, unless the controller is restarted.
What am I missing? 🤔

charts/fleet/values.yaml
```diff
@@ -30,6 +30,34 @@ garbageCollectionInterval: "15m"
 # Whether you want to allow cluster upon registration to specify their labels.
 ignoreClusterRegistrationLabels: false
 
+# Events reporting the deployment state of bundles, on the bundle they are about.
+deploymentEvents:
+  # Turns off events about the deployment state of bundles. Unsetting the whole
+  # block keeps the defaults below, which report.
+  disabled: false
```
weyfonk, line 37: Why name the toggle `disabled` and have it default to `false`, instead of `enabled: true`? The latter would perhaps incur less cognitive load on users. Besides, it would be consistent with the way other parts of Fleet can be toggled (e.g. gitOps, helmOps, metrics...).

charts/fleet/values.yaml
```diff
@@ -30,6 +30,34 @@ garbageCollectionInterval: "15m"
...
+  reportRecovery: true
+  # Additionally report each failing bundle deployment, in the namespace of its
+  # cluster. This is more detailed, but produces one event per failing
+  # deployment, of which there is one per cluster a bundle is deployed to: a
+  # bundle failing on 500 clusters reports 501 events instead of 1.
+  perDeployment: false
+  # How many distinct failure causes a single event describes. The bundle's
+  # status.summary.nonReadyResources lists more of them, up to 10, which is
+  # therefore the highest value with an effect. Causes beyond it are reported as
+  # "(+117 more, see status.summary.nonReadyResources)".
```
weyfonk, line 58: nit:
```suggestion
  # e.g. "(+117 more, see status.summary.nonReadyResources)".
```

charts/fleet/templates/rbac.yaml
weyfonk, line 41: nit: perhaps we could be more restrictive about this, ensuring that events can only be created. Does Fleet need to do anything else about them? Reading already sent events about a given object (upon controller restart)?

internal/config/config.go
```diff
@@ -166,6 +167,40 @@ type Config struct {
...
+	// reports.
+	MinInterval metav1.Duration `json:"minInterval,omitzero"`
+
+	// ReportRecovery reports bundles whose deployments all became ready
+	// again after a failure, defaults to true.
+	ReportRecovery *bool `json:"reportRecovery,omitempty"`
+
+	// PerDeployment additionally reports each failing bundle deployment, in
+	// the namespace of its cluster. This is more detailed, but produces one
+	// event per failing deployment.
```
weyfonk, line 198: nit:
```suggestion
	// event per failing deployment. Defaults to false.
```

internal/cmd/controller/bundleevents/notifier.go
```diff
@@ -0,0 +1,813 @@
...
+	ent.attempts = 0
+	if ent.pending != nil {
+		// Something was observed while this event was created.
+		ent.dueAt = e.dueTime(now, ent.lastEmit, opts)
+	}
+}
+
+// createFailed retries the event, unless it has been tried too often or the
+// error says that trying again can only fail the same way. Giving up drops the
+// event, which lets the entry be forgotten again: what the event describes is
```
weyfonk, line 613: Dropping the event would be the responsibility of the caller though, as that is not done here.

internal/cmd/controller/bundleevents/note.go
```diff
@@ -0,0 +1,210 @@
...
+		}
+	}
+
+	return causes
+}
+
+// causeKeys returns the distinct state and message pairs of the failing
+// resources, sorted, so that the same set of causes always produces the same
+// keys. Cluster names are deliberately left out: the same failure on more
+// clusters is not a new cause and must not produce another event.
```
weyfonk, line 101: Interesting stance :)
I'd consider it debatable, weren't it for the sake of limiting event bursts on large setups.

internal/cmd/controller/bundleevents/notifier.go
```diff
@@ -0,0 +1,813 @@
...
+	// bundle, so that a failure recurring later is reported again.
+	e.clearEmitted(cur.UID)
+
+	if !opts.ReportRecovery {
+		e.stopAwaitingRecovery(cur.UID)
+		return
+	}
+
+	// wasFailing covers the failure and the recovery landing in one status
+	// update, which is the only form a restart can still detect.
```
weyfonk, line 286: Hm, not sure I understand this. How could a failure and a recovery be part of the same status update?

internal/cmd/controller/bundleevents/notifier.go
```diff
@@ -0,0 +1,813 @@
...
+		ent = &entry{touched: now}
+		e.entries[uid] = ent
+		e.evict(opts.MaxTracked, uid)
+	}
+	ent.touched = now
+	ent.awaitingRecovery = true
+}
+
+// hasPending reports whether an event for the object is waiting to be created.
+func (e *Emitter) hasPending(uid types.UID) bool {
```
weyfonk, line 390: nit: this method could be moved up, so that all recovery-related methods would live in a common block.

internal/cmd/controller/bundleevents/notifier_test.go
```diff
@@ -0,0 +1,908 @@
...
+
+	// createErr decides how creating an event fails. Nil lets it succeed.
+	createErr func() error
+}
+
+// failingClient lets a test fail event creation the way the API server would.
+type failingClient struct {
+	client.Client
+
+	fixture *fixture
```
weyfonk, line 60: Why do we need _two_ clients inside a `failingClient` (including the one living inside the `fixture`)?

Edit: oh, one has to be a failing client to simulate failing event creation, whereas the other one may be used for other purposes, such as listing events.

internal/cmd/controller/bundleevents/notifier_test.go
```diff
@@ -0,0 +1,908 @@
...
+	if byKind["Bundle"] != ReasonBundleDeployFailed {
+		t.Errorf("unexpected reason on the bundle: %q", byKind["Bundle"])
+	}
+	if byKind["BundleDeployment"] != ReasonBundleDeploymentNotReady {
+		t.Errorf("unexpected reason on the deployment: %q", byKind["BundleDeployment"])
+	}
+}
+
+// bundleUID is the UID of the bundle the fixture reports on.
+const bundleUID = types.UID("7b3d")
```
weyfonk, line 678: nit: wouldn't this definition be better located at the top of the file, with other constants?

internal/cmd/controller/bundleevents/notifier_test.go
```diff
@@ -0,0 +1,908 @@
...
+	f.flush(t)
+
+	if got := f.events(t); len(got) != 0 {
+		t.Fatalf("expected the event to be rate limited, got %d", len(got))
+	}
+
+	f.emitter.mu.Lock()
+	defer f.emitter.mu.Unlock()
+
+	ent, ok := f.emitter.entries[types.UID("7b3d")]
```
weyfonk, line 494: This would be a bit clearer:
```suggestion
	ent, ok := f.emitter.entries[bundleUID]
```

internal/cmd/controller/bundleevents/notifier_test.go
```diff
@@ -0,0 +1,908 @@
...
+
+		return serverBusy()
+	}
+
+	f.emitter.ObserveBundle(context.TODO(), bundle(ready()), bundle(failing(500, renderError)))
+	f.flush(t)
+
+	if got := f.events(t); len(got) != 0 {
+		t.Fatalf("expected the first attempt to fail, got %d events", len(got))
+	}
```
weyfonk, line 710: nit: we could run a similar check here to the one on L717, with `ent.attempts == 1`.

internal/cmd/controller/bundleevents/notifier_test.go
```diff
@@ -0,0 +1,908 @@
...
+	f.emitter.ObserveBundle(context.TODO(), bundle(ready()), bundle(failing(500, renderError)))
+	f.flush(t)
+	f.retry(2 * maxCreateAttempts)
+
+	if got := f.events(t); len(got) != 0 {
+		t.Fatalf("expected no event while creation fails, got %d", len(got))
+	}
+
+	// Whatever made creation fail is fixed, and the bundle fails for a new
+	// reason: nothing was lost permanently.
```
weyfonk, line 793: Arguably the event reporting the rendering error was; we can probably live with that though :)

internal/cmd/controller/bundleevents/notifier_test.go
```diff
@@ -0,0 +1,908 @@
...
+	f.emitter.ObserveBundle(
+		context.TODO(),
+		otherBundle(newUID, ready()),
+		otherBundle(newUID, failing(1, nsForbidden)),
+	)
+	f.flush(t)
+
+	if got := f.eventsFor(t, newUID); got != 1 {
+		t.Errorf("expected the newly tracked bundle to be reported once, got %d events", got)
+	}
```
weyfonk, line 907: A check similar to the one on L884, on the size of `f.emitter.entries`, could make sense here, validating that the `MaxTracked` size is indeed exceeded this time.

internal/cmd/controller/bundleevents/note.go
```diff
@@ -0,0 +1,210 @@
...
+
+	var b strings.Builder
+	fmt.Fprintf(&b, "%d/%d bundle deployments failing", failing, s.DesiredReady)
+	if s.ErrApplied > 0 && s.NotReady > 0 {
+		fmt.Fprintf(&b, " (errApplied %d, notReady %d)", s.ErrApplied, s.NotReady)
+	}
+	b.WriteString(".")
+
+	// Leave room for the trailing hint, which is more useful than one more cause.
+	budget := noteMaxLength - len(statusHint) - 32
```
weyfonk, line 153: Where does this `32` come from?

internal/cmd/controller/bundleevents/note.go
```diff
@@ -0,0 +1,210 @@
...
+
+	return truncate(fmt.Sprintf("bundle %s: %s: %s", bundle, state, message), noteMaxLength)
+}
+
+// truncate shortens s to max bytes, without cutting a rune in half.
+func truncate(s string, max int) string {
+	if len(s) <= max {
+		return s
+	}
+	if max <= 3 {
```
weyfonk, line 205: `3` being the length of `...` ...

internal/cmd/controller/bundleevents/note_test.go
```diff
@@ -0,0 +1,176 @@
...
+	truncated := truncate(strings.Repeat("ü", 20), 10)
+
+	if len(truncated) > 10 {
+		t.Errorf("expected at most 10 bytes, got %d", len(truncated))
+	}
+	if !strings.HasSuffix(truncated, "...") {
+		t.Errorf("expected an ellipsis, got %q", truncated)
+	}
+	for _, r := range truncated {
+		if r == '�' {
```
weyfonk, line 172: I would like to better understand this line :D

weyfonk (commented):
Nice one!
My review brings mostly nitpicks, with a couple of questions around configuration.

Besides, I think it could make sense to add an end-to-end test for this, even a simple-ish happy case, just to be sure that assumptions made by unit tests about how a bundle's _first_ deployment-related event would look do make sense.

PR #5538, opened 2026-08-10
agentmanagement tests - cluster registration token

0xavi0:
This PR adds testing coverage for the cluster registration token controller.

### DIFF
--- a/integrationtests/agentmanagement/clusterregistrationtoken_test.go
+++ b/integrationtests/agentmanagement/clusterregistrationtoken_test.go
@@ -227,18 +227,39 @@ var _ = Describe("ClusterRegistrationToken", func() {
 			objectGone(token).Should(Succeed())
 		})
 
-		It("does not delete a token before its TTL has elapsed", func() {
-			// Populate the SA token Secret rather than creating the token
-			// bare: the reconcile that observes the ServiceAccount otherwise
-			// parks forever in the handler's unbounded wait for that Secret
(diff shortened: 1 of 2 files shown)


integrationtests/agentmanagement/clusterregistrationtoken_test.go
```diff
@@ -0,0 +1,271 @@
...
+		It("does not delete a token before its TTL has elapsed", func() {
+			// Populate the SA token Secret rather than creating the token
+			// bare: the reconcile that observes the ServiceAccount otherwise
+			// parks forever in the handler's unbounded wait for that Secret
+			// (envtest populates no service account tokens), holding a worker
+			// for the rest of the suite. This also settles the token, so the
+			// assertion below observes a fully reconciled object.
+			token, _ := createTokenWithPopulatedSecret("long-ttl", &metav1.Duration{Duration: time.Hour}, "long-ttl-value")
+
+			Consistently(func(g Gomega) {
```
d3flex, line 256: the only thing that confuses me is this part. from the spec title it should verify that the token is present for *1h*. but that cant be validated in the timespan of the test. not sure what `consistently` uses but there should be a timeout or something, which the token must verified within those limits, right?

0xavi0: Good catch! the test wasn't too useful. Now it should be checking for a more realistic time

p-se (approved):

PR #5540, opened 2026-08-11
Remove logrus usage

thardeck:
Replace Fleet logging calls with controller-runtime logr and standard library logging while retaining transitive module dependencies required by upstream packages.

Refers #3554

### DIFF
--- a/internal/cmd/agent/deployer/normalizers/jsonpatch.go
+++ b/internal/cmd/agent/deployer/normalizers/jsonpatch.go
@@ -1,8 +1,6 @@
 package normalizers
 
 import (
-	"fmt"
-
 	jsonpatch "github.com/evanphx/json-patch"
 	"github.com/rancher/fleet/internal/cmd/agent/deployer/objectset"
 	"sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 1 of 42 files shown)


internal/cmd/agent/deployer/normalizers/jsonpatch.go
```diff
@@ -36,7 +38,7 @@ func (j JSONPatchNormalizer) Normalize(un *unstructured.Unstructured) error {
 	gvk := un.GroupVersionKind()
 	metaObj, err := meta.Accessor(un)
 	if err != nil {
-		logrus.Errorf("Failed to normalize obj with json patch, error: %v", err)
+		log.Log.Error(nil, fmt.Sprintf("Failed to normalize obj with json patch, error: %v", err))
```
0xavi0: Question: Isn't this dropping the original error error? 


```suggestion
		log.Log.Error(err, fmt.Sprintf("Failed to normalize obj with json patch"))
```

`logr.Log.Error` is called several times this way, and other passing the original error. 

The difference would be:

Passing nil:
```json
{"level":"error","msg":"Failed to register agent: dial tcp: connection refused"}
```

Passing the error:
```json
{"level":"error","msg":"Failed to register agent","error":"dial tcp: connection refused"}
```

internal/cmd/cli/apply/apply.go
```diff
@@ -611,7 +610,7 @@ func save(ctx context.Context, c client.Client, bundle *fleet.Bundle) (*fleet.Bu
 			if err := deleteOCIManifest(ctx, c, bundle, ocistorage.OCIOpts{}); err != nil {
 				// return the error, since the OCI registry is an external entity to the cluster
 				// and we may encounter various types of transient errors (such as connection or access issues).
-				logrus.Warnf("deleting OCI artifact: %v", err)
+				log.Log.Info(fmt.Sprintf("deleting OCI artifact: %v", err))
```
0xavi0:
```suggestion
				log.Log.Info(err, "deleting OCI artifact: %v")
```
I think this option looks more the way `logr` is designed to be. (And the way we use it in our controllers)

It is used in other places (just mentioning this one)

0xavi0 (approved):

Issue #4797, opened 2026-03-11
Bundle dependencies: use specific state in dependent bundle if dependency not in accepted state


weyfonk:
Work on #2552 has resulted in users being able to specify a set of accepted states for a dependency bundle, allowing dependent bundles to move forward based on those states (docs [acceptedStates](https://fleet.rancher.io/reference/ref-fleet-yaml) and [dependsOn](https://fleet.rancher.io/reference/ref-crds)).

However, when a bundle is _not_ in one of its dependent bundle's _accepted states_, the dependent bundle still bears state `ErrApplied`.
We should change this, using a different state to clarify that this is its own type of error, and that it may well be temporary.
Possible names for this new state: `WaitingForDependency`, `DependencyInUnsupportedState` (perhaps a bit too long)...

Relates to https://github.com/rancher/dashboard/issues/13121 for UI work.

kkaempf:
UI is waiting for this to be fixed.

mmartin24:
Verified along with 2.15.3 [here](https://github.com/rancher/fleet/issues/5690#issuecomment-5679426426).
Test Fleet-207 on dependsOn [passing now](https://github.com/rancher/fleet-e2e/actions/runs/34957208396/job/104342201081#step:10:418) after adaption on `2.16-136b11a0062247c1b28f107d82283bdb29bc07ae-head` with `fleet-111.0.0+up0.17.0-alpha.3`

PR #5549, opened 2026-08-12
Add WaitingForDependency bundle state

0xavi0:
Introduces a new WaitingForDependency bundle/bundledeployment state so bundles blocked on an unaccepted dependency are reported distinctly from actual deployment errors, instead of surfacing as ErrApplied. Includes CRD, summary, and metrics updates plus e2e coverage.

Refers to: https://github.com/rancher/fleet/issues/4797

### DIFF
--- a/e2e/multi-cluster/depends_on_test.go
+++ b/e2e/multi-cluster/depends_on_test.go
@@ -226,6 +226,35 @@ var _ = Describe("Bundle Depends On", func() {
 				}
 				return out
 			}, duration, interval).Should(Equal(string(fleet.WaitingForDependency)))
+
+			By("restoring the ConfigMap, which makes the dependency acceptable again")
+			_, err = k.Namespace(namespace).Run(
+				"patch", "configmap", "root-will-be-modified",
+				"--type=merge", "-p", `{"data":{"value":"original"}}`,
+			)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("waiting for the required bundle to be ready again")
+			Eventually(func() string {
+				out, err := k.Namespace(env.Namespace).Get("bundle", required, "-o=jsonpath={.status.display}")
+				if err != nil {
+					return ""
+				}
+				var d fleet.BundleDisplay
+				_ = json.Unmarshal([]byte(out), &d)
+				return d.ReadyClusters
+			}, 2*time.Minute, interval).Should(Equal("1/1"))
+
+			By("verifying the dependent bundle leaves WaitingForDependency and deploys")
+			Eventually(func() string {
+				out, err := k.Namespace(env.Namespace).Get("bundle", dependsOn, "-o=jsonpath={.status.display}")
+				if err != nil {
+					return ""
+				}
+				var d fleet.BundleDisplay
+				_ = json.Unmarshal([]byte(out), &d)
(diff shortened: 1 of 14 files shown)


d3flex (approved):

internal/cmd/controller/summary/summary_test.go
```diff
@@ -64,6 +65,142 @@ func TestGetSummaryState(t *testing.T) {
...
+			Name:  "test",
+			State: fleet.WaitingForDependency,
+		},
+		{
+			Name:  "test",
+			State: fleet.ErrApplied,
+		},
+	}
+	bundleState = summary.GetSummaryState(s)
+	if bundleState != fleet.ErrApplied {
```
weyfonk, line 107: The only thing that concerns me here is that we (probably accidentally) always expect the actual summary state to reflect the _last_ element of `s.NonReadyResources`, across all 4 test cases within this fixture.
Meaning that replacing the current, `StateRank`-based implementation of `GetSummaryState` with this dumb one would still make tests pass:
```
        var state fleet.BundleState

        if len(summary.NonReadyResources) == 0 {
                return state
        }

        return summary.NonReadyResources[len(summary.NonReadyResources)-1].State
```

Maybe flip `s.NonReadyResources` elements in at least one test case, or add a dummy element?

internal/bundlereader/validate_fleetyaml.go
```diff
@@ -39,8 +39,21 @@ func validateBundleRef(index int, dep fleet.BundleRef) error {
 	return nil
 }
 
-// isValidBundleState checks if a BundleState is valid by checking against StateRank
+// statesRejectedAsAccepted lists states which are ranked in StateRank, but which
```
weyfonk, line 42: Great explanation, thanks!

e2e/multi-cluster/depends_on_test.go
```diff
@@ -162,4 +162,70 @@ var _ = Describe("Bundle Depends On", func() {
...
+
+			By("keeping that state instead of falling back to ErrApplied")
+			Consistently(func() string {
+				out, err := k.Namespace(env.Namespace).Get("bundle", dependsOn, "-o=jsonpath={.status.summary.nonReadyResources[0].bundleState}")
+				if err != nil {
+					return ""
+				}
+				return out
+			}, duration, interval).Should(Equal(string(fleet.WaitingForDependency)))
+		})
```
weyfonk, line 258: nit: we _could_ check that undoing the change on the dependency (config map) allows the dependent bundle to move on to `Ready` (and I say "could", because that's kind of already covered in the first test case contained in this file).

weyfonk (commented):
Nice one! Leaving a few comments, only one of them not labeled as a nitpick, although arguably it could be considered one 🙃

e2e/multi-cluster/depends_on_test.go
0xavi0: Suggestion implemented.

internal/cmd/controller/summary/summary_test.go
0xavi0: yup, nice suggestion. Implemented

weyfonk (approved):

Issue #5406, opened 2026-07-06
Fleet will delete all resources on agent timeout


kimdane:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

I am using fleet shipped with rancher 2.14.3 for deploying production services and applications using helm. After a while the clusters in rancher/fleet has a short timeout on the agent and becomes unavailable. After that the cluster comes back to rancher but is left unavailable to all repos and bundles in Fleet. Worse yet is that once the connection is severed something (likely fleet agent) starts removing everything Managed by fleet in the cluster. I might have been stupid to use Fleet for this, but given I have used Fleet to deploy most of the CRDs, and postgresql databases using the cnpg operator, this has now deleted several CNPG production databases with catastrophic loss. 

I have now started to mark all the GitRepos with "Always keep resources", but not even that seems to hold correctly. The cascading effect of managing to remove in use CRDs from a production cluster is honestly impressive, but also completely horrifying! 

### Expected Behavior

Any glitch in connection between Fleet and cluster would result in fleet agent on cluster leaving all resources it manages in place.

### Steps To Reproduce

Install Fleet as part of Rancher. Import a Talos 13.5 cluster using Cilium 1.19.5 CNI. Try installing anything using CRDs and other resources, make connection glitch for the fleet agent. etcd re-election etc are good candidates to get a good glitch for the agent.

### Environment

```markdown
- Architecture: Talos 13.5 cluster using Cilium 1.19.5 CNI
- Fleet Version: fleet:109.0.4+up0.15.4
- Cluster: Talos 13.5
  - Provider:
  - Options:
  - Kubernetes Version: 1.36.1
```

### Logs

```markdown

```

weyfonk:
Hi, we'd like to better understand the context around this issue.
A few questions:
* did it happen in the middle of an upgrade?
* had all bundle deployments been deployed successfully before this happened?

0xavi0:
I managed to recreate the problem with an integration test. (I also tried with a full cluster but this seems to be a race condition and it's not so simple to recreate).

The problem seems to be a query to an out of sync controller's cache.
It returns `NotFound` even when there's a connection problem making the Cache to be empty, and Fleet uninstalls the release because it considers the `BundleDeployment` is gone.

I could not recreate the problem (unistalling releases) when using `keepResources`, though.

PR #5585, opened 2026-08-18
Confirm deletion upstream before uninstalling releases

0xavi0:
The agent reads BundleDeployments through the manager's cache. A CacheReader returns a plain NotFound whenever the object is missing from the local informer store, with no way to tell "deleted" apart from "our store is empty or behind". Both the reconciler and the garbage collector took that NotFound as a deletion and uninstalled the helm release, wiping live resources on a stale cache.

Both paths now confirm with a live read (mgr.GetAPIReader()) before uninstalling anything, and treat a read they could not complete as "do not delete" rather than as a deletion. While the cache is behind, the reconciler requeues with an exponential backoff instead of cleaning up.

Refers to: https://github.com/rancher/fleet/issues/5406

### Checklist

~ [ ]  I have updated the documentation via a pull request in the [fleet-product-docs](https://github.com/rancher/fleet-product-docs) repository.~

### DIFF
--- a/internal/cmd/agent/controller/bundledeployment_controller_test.go
+++ b/internal/cmd/agent/controller/bundledeployment_controller_test.go
@@ -29,8 +29,8 @@ func TestStaleCacheRequeueAfter(t *testing.T) {
 }
 
 // TestStaleCacheHits checks that the counter for a BundleDeployment starts at
-// one, grows while its cache entry is missing, and is forgotten once the agent
-// sees the object again.
+// one, grows by one every time a stale cache hit is recorded, and is reset to 0
+// when stale cache hits are forgotten.
 func TestStaleCacheHits(t *testing.T) {
 	r := &BundleDeploymentReconciler{}
 	key := "cluster-ns/bd"
--- a/internal/cmd/agent/controller/bundledeployment_controller_test.go
+++ b/internal/cmd/agent/controller/bundledeployment_controller_test.go
@@ -1,0 +1,54 @@
+package controller
+
+import (
+	"testing"
+	"time"
+
+	"github.com/rancher/fleet/pkg/durations"
+)
+
+// TestStaleCacheRequeueAfter checks that repeated stale cache observations back
+// off, and that the wait stops growing at the cap: a cache which never catches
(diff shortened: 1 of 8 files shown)


internal/cmd/agent/controller/bundledeployment_controller_test.go
```diff
@@ -0,0 +1,54 @@
...
+	for _, c := range cases {
+		if got := staleCacheRequeueAfter(c.hits); got != c.expected {
+			t.Errorf("staleCacheRequeueAfter(%d) = %s, expected %s", c.hits, got, c.expected)
+		}
+	}
+}
+
+// TestStaleCacheHits checks that the counter for a BundleDeployment starts at
+// one, grows while its cache entry is missing, and is forgotten once the agent
+// sees the object again.
```
weyfonk: nit: this is not exactly what's under test here (no reconciler is involved in this test case):
```suggestion
// TestStaleCacheHits checks that the counter for a BundleDeployment starts at
// one, grows by one every time a stale cache hit is recorded, and is reset to 0
// when stale cache hits are forgotten.
```

pkg/durations/durations.go
```diff
@@ -46,6 +46,15 @@ const (
...
 	// converge once the permission is added.
 	NamespacePermissionRequeueInterval = time.Minute * 2
+	// StaleCacheRequeue is how long the Fleet agent waits before looking again
+	// at a BundleDeployment which is missing from its cache but still present
+	// on the API server. It doubles on every further attempt, up to
+	// StaleCacheRequeueMax.
+	StaleCacheRequeue = time.Second * 30
+	// StaleCacheRequeueMax caps the wait between those attempts, so that a
+	// cache which never catches up stops polling the API server every few
+	// seconds.
```
weyfonk: To be quite nitpicky about this: the comment could read as "a cache which never catches up will poll the API server, and every few seconds it will stop doing so". How about something like this instead?
```suggestion
	// cache which never catches up polls the API server with this periodicity.
```

weyfonk (approved):
LGTM :)

Issue #5426, opened 2026-07-10
"fleet apply --output" fails on empty kubeconfig


mearsten:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Take an valid empty kubeconfig, then try to render a bundle locally. Fleet CLI is supposed to work with no k8s cluster accessible:

```text
% cat kubeconfig-empty.yaml
apiVersion: v1
kind: Config
preferences: {}
clusters: []
users: []
contexts: []

% env KUBECONFIG=kubeconfig-empty.yaml ./fleet-linux-amd64-0.15.4 apply -o bundle.yaml some ./some || echo "rc=$?"
rc=1

% env KUBECONFIG=kubeconfig-empty.yaml ./fleet-linux-amd64-0.16.0-rc.4 apply -o bundle.yaml some ./some || echo "rc=$?"
rc=1

% ls -l bundle.yaml
"bundle.yaml": No such file or directory (os error 2)

% ./fleet-linux-amd64-0.15.4 --version; ./fleet-linux-amd64-0.16.0-rc.4 --version
fleet version v0.15.4 (7b97db764df7de385e0572eab6da959e0bd22569)
fleet version v0.16.0-rc.4 (4b6976899c8cfc16317e7e5d610c74caa56ddd34)
```

It is working fine with a dummy cluster added:

```text
% cat kubeconfig-dummy.yaml
apiVersion: v1
kind: Config
preferences: {}
clusters:
  - {name: dummy, cluster: {server: whatever}}
contexts:
  - {name: dummy, context: {cluster: dummy}}
current-context: dummy

% env KUBECONFIG=kubeconfig-dummy.yaml ./fleet-linux-amd64-0.15.4 apply -o bundle.yaml some ./some || echo "rc=$?"

% ls -l bundle.yaml
.rw------- 86k user 10 Jul 13:30 bundle.yaml
```

### Expected Behavior

`fleet apply -o bundle.yaml …` should produce output with valid empty kubeconfig.

### Steps To Reproduce

All info is in the "Current behaviour" section.

### Environment

```markdown
- Architecture: amd64
- Fleet Version: 0.15.4, 0.16.0-rc.4
- Cluster: N/A
```

### Logs

```markdown

```

### Anything else?

It seems the CLI tries to detect whether it is run locally or in a pod and fails:

```text
[...]

PR #5621, opened 2026-08-20
Don't require a kubeconfig for fleet apply --output

0xavi0:
This PR fixes an issue when calling `fleet apply --ouput` without a valid kubeconfig.

fleet apply --output renders bundles locally and never talks to a cluster, but it built a Kubernetes client before deciding that — so it failed on any machine without a loadable kubeconfig, and failed silently (exit 1, no message), because ctrl.GetConfigOrDie logs through controller-runtime's logger that fleet apply never initializes and then calls os.Exit.

The client and event recorder are now built only when --output is unset, so the offline path never needs a kubeconfig; and every CLI command uses ctrl.GetConfig and returns the error instead of dying, so a genuinely missing kubeconfig now prints a valid message.

Refers to: https://github.com/rancher/fleet/issues/5426

### Checklist

### DIFF
--- a/internal/cmd/cli/apply_offline_test.go
+++ b/internal/cmd/cli/apply_offline_test.go
@@ -1,0 +1,144 @@
+package cli
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/spf13/cobra"
+
+	// Registers the fleet GVKs with the wrangler scheme used to render
+	// bundles, the same way cmd/fleetcli does.
+	_ "github.com/rancher/fleet/pkg/generated/controllers/fleet.cattle.io"
+)
+
+// setupOfflineApply prepares a working directory holding a single resource in
+// ./some, and makes any attempt to reach a cluster fail by pointing the
+// kubeconfig at a valid but empty one.
+func setupOfflineApply(t *testing.T) {
+	t.Helper()
+
+	t.Chdir(t.TempDir())
+
+	if err := os.MkdirAll("some", 0o700); err != nil {
+		t.Fatalf("failed to create bundle dir: %v", err)
+	}
+	configMap := `apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: test
+data:
+  key: value
+`
+	if err := os.WriteFile(filepath.Join("some", "configmap.yaml"), []byte(configMap), 0o600); err != nil {
+		t.Fatalf("failed to write configmap: %v", err)
+	}
+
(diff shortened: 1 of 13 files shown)


internal/cmd/cli/apply_offline_test.go
```diff
@@ -0,0 +1,144 @@
...
+		"file":   "bundle.yaml",
+		"stdout": "-",
+	} {
+		t.Run(name, func(t *testing.T) {
+			setupOfflineApply(t)
+
+			a := newApplyForTest()
+			a.Output = output
+
+			if err := a.Run(testCommand(), []string{"repo", "some"}); err != nil {
```
weyfonk: What does "repo" stand for here? It isn't an `apply` flag 🤔

internal/cmd/cli/kubeconfig.go
```diff
@@ -0,0 +1,60 @@
...
+
+// addGoFlags converts fs, together with the controller-runtime flags, to
+// pflags and adds them to cmd. Every subcommand which talks to a cluster has
+// to go through here, so that --kubeconfig reaches the loader used by
+// getKubeconfig.
+//
+// ctrl.RegisterFlags binds --kubeconfig via the stdlib flag package, which has
+// no notion of shorthands. pflag only records a shorthand while adding a flag
+// to a set, in AddFlag, so "k" has to be set on the way in: assigning
+// Shorthand afterwards updates the usage output but leaves -k unparsable.
```
weyfonk: Does this explain why we're creating a `pfs` variable here, instead of reusing `fs` and doing something like `fs.AddGoFlagSet(fs)`?

weyfonk (commented):
Looks mostly good to me, pending clarification on a couple of doubts :)

internal/cmd/cli/apply_offline_test.go
0xavi0: "repo" and "some" are random names that would be the parameters passed to `fleet apply`.
I've changed "repo" to "test-bundle" to avoid confusions

internal/cmd/cli/kubeconfig.go
0xavi0: Yes, that's what it's meant to explain, but it stops one step short — reworded to name pfs explicitly.

weyfonk (approved):

internal/cmd/cli/kubeconfig.go
weyfonk: This is clearer, thanks :bow:

Issue #4663, opened 2026-02-17
Feature Request:  Add early validation for incorrect regexp malformation when deploying jobs


mmartin24:
### Is your feature request related to a problem?

Follow up on https://github.com/rancher/fleet/issues/748
Related comment: https://github.com/rancher/fleet/issues/748#issuecomment-3914911946

---

Issue: 

When passing an incorrect Regexp (ie: `name: ttl-test-["·Ewdfsdfsad0?-job`) an error is trigered at `agent` level pod:

```
k logs -n cattle-fleet-local-system pods/fleet-agent-7f944bbd75-bvj4d | tail
```

Output:

```json
[...]

PR #5669, opened 2026-08-27
Validate diff.comparePatches fields early

0xavi0:
Bad values in `diff.comparePatches` are currently accepted by `fleet apply` and then discarded on the agent, where the failure is either silent or only visible at `-v=1`. This validates four of them up front, so a typo fails at apply time with the offending path and value.

Validated fields:

- `diff.comparePatches[].name` — must be a valid Go regular expression
- `diff.comparePatches[].operations[].op` — one of `add`, `ignore`, `remove`, `replace`, `test` (`copy`/`move` are unsupported: `Operation` has no `from` field to encode them)
- `diff.comparePatches[].operations[].path` — non-empty JSON pointer starting with `/`; skipped for `ignore`, which never reads the path
- `diff.comparePatches[].jsonPointers` — same rule

Refers to: https://github.com/rancher/fleet/issues/4663

### Checklist

### DIFF
--- a/charts/fleet-crd/templates/crds.yaml
+++ b/charts/fleet-crd/templates/crds.yaml
@@ -388,7 +388,7 @@ spec:
                                   applies, which is
 
                                   laxer than RFC 6901: "~0" and "~1" are read as "~"
-                                  and "/", and any other
+                                  and "/", respectively, and any other
 
                                   "~" sequence is left as written.'
                                 items:
@@ -928,7 +928,7 @@ spec:
                                   applies, which is
 
                                   laxer than RFC 6901: "~0" and "~1" are read as "~"
-                                  and "/", and any other
+                                  and "/", respectively, and any other
 
                                   "~" sequence is left as written.'
                                 items:
@@ -2043,7 +2043,7 @@ spec:
                               applies, which is
 
                               laxer than RFC 6901: "~0" and "~1" are read as "~" and
-                              "/", and any other
+                              "/", respectively, and any other
 
                               "~" sequence is left as written.'
                             items:
@@ -3149,7 +3149,7 @@ spec:
                                     applies, which is
 
                                     laxer than RFC 6901: "~0" and "~1" are read as
-                                    "~" and "/", and any other
+                                    "~" and "/", respectively, and any other
 
                                     "~" sequence is left as written.'
                                   items:
@@ -8495,7 +8495,7 @@ spec:
                               applies, which is
 
                               laxer than RFC 6901: "~0" and "~1" are read as "~" and
-                              "/", and any other
+                              "/", respectively, and any other
 
                               "~" sequence is left as written.'
                             items:
@@ -9633,7 +9633,7 @@ spec:
                                     applies, which is
 
                                     laxer than RFC 6901: "~0" and "~1" are read as
-                                    "~" and "/", and any other
+                                    "~" and "/", respectively, and any other
 
                                     "~" sequence is left as written.'
                                   items:
--- a/charts/fleet-crd/templates/crds.yaml
+++ b/charts/fleet-crd/templates/crds.yaml
@@ -378,13 +378,19 @@ spec:
                                 description: 'JSONPointers ignore diffs at the given
                                   JSON pointers, e.g. /spec/replicas.
 
-                                  Each entry must be a non-empty RFC 6901 pointer
-                                  starting with a slash; a
+                                  Each entry must be a non-empty pointer starting
+                                  with a slash; a
 
                                   Kubernetes-style field path such as spec.replicas
                                   addresses nothing and is
 
-                                  rejected.'
+                                  rejected. Escaping is the one the JSON patch library
+                                  applies, which is
+
+                                  laxer than RFC 6901: "~0" and "~1" are read as "~"
+                                  and "/", and any other
+
+                                  "~" sequence is left as written.'
                                 items:
                                   type: string
                                 nullable: true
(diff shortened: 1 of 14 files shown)


charts/fleet-crd/templates/crds.yaml
```diff
@@ -406,12 +421,41 @@ spec:
...
+
+                                        supported, because an Operation has no "from"
+                                        field to encode them with.
+
+                                        Any other value, including an empty one, makes
+                                        the whole patch fail to
+
+                                        apply.'
                                       nullable: true
                                       type: string
```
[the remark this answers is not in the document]
0xavi0: This would mean an API breaking change.

pkg/apis/fleet.cattle.io/v1alpha1/bundledeployment_types.go
```diff
@@ -169,13 +169,20 @@ type ComparePatch struct {
...
 	Name string `json:"name,omitempty"`
 	// Namespace is the namespace of the resource to match.
 	// +nullable
 	Namespace string `json:"namespace,omitempty"`
-	// JSONPointers ignore diffs at a certain JSON path.
+	// JSONPointers ignore diffs at the given JSON pointers, e.g. /spec/replicas.
+	// Each entry must be a non-empty pointer starting with a slash; a
+	// Kubernetes-style field path such as spec.replicas addresses nothing and is
+	// rejected. Escaping is the one the JSON patch library applies, which is
+	// laxer than RFC 6901: "~0" and "~1" are read as "~" and "/", and any other
```
weyfonk: nit:
```suggestion
	// laxer than RFC 6901: "~0" and "~1" are read as "~" and "/", respectively, and any other
```

schemas/fleet.yaml.json
```diff
@@ -165,7 +165,7 @@
             "type": "string"
           },
           "type": "array",
-          "description": "JSONPointers ignore diffs at a certain JSON path."
+          "description": "JSONPointers ignore diffs at the given JSON pointers, e.g. /spec/replicas.\nEach entry must be a non-empty pointer starting with a slash; a\nKubernetes-style field path such as spec.replicas addresses nothing and is\nrejected. Escaping is the one the JSON patch library applies, which is\nlaxer than RFC 6901: \"~0\" and \"~1\" are read as \"~\" and \"/\", and any other\n\"~\" sequence is left as written."
```
weyfonk: nit:
```suggestion
          "description": "JSONPointers ignore diffs at the given JSON pointers, e.g. /spec/replicas.\nEach entry must be a non-empty pointer starting with a slash; a\nKubernetes-style field path such as spec.replicas addresses nothing and is\nrejected. Escaping is the one the JSON patch library applies, which is\nlaxer than RFC 6901: \"~0\" and \"~1\" are read as \"~\" and \"/\", respectively, and any other\n\"~\" sequence is left as written."
```

internal/validation/comparepatches.go
```diff
@@ -0,0 +1,247 @@
...
+	ValidateComparePatchNames,
+	ValidateComparePatchOperations,
+	ValidateComparePatchOperationPaths,
+	ValidateComparePatchJSONPointers,
+}
+
+// ValidateComparePatchNames checks that every diff.comparePatches[].name in opts compiles
+// as a regular expression. The agent matches a patch by exact name first and falls
+// back to matching the name as a regex, so a name which does not compile is dead
+// config: it is dropped by the agent instead of being reported to the user.
```
weyfonk: "dropped by the agent" suggests that it would _reach_ the Fleet agent; isn't the whole point of this PR to prevent this? 🤔 
Edit: it seems to me that this phrases in an indicative mood instead of using a _conditional_ tense to denote what _would_ happen on the agent side if this controller-side validation were not in place.

internal/validation/comparepatches_test.go
weyfonk, line 1: This does not cover patch names nor operations; is that deliberate?

internal/cmd/controller/helmops/reconciler/helmop_controller_test.go
```diff
@@ -1354,3 +1485,198 @@ func expectCABundleLookup(client *mocks.MockK8sClient) {
 		return k8serrors.NewNotFound(schema.GroupResource{}, "tls-ca-additional")
 	})
 }
+
+// TestValidateBundleSpec covers the rules a HelmOp shares with fleet.yaml through
+// its embedded BundleSpec. A HelmOp never passes through
+// bundlereader.validateFleetYAML, so without these the values would only fail on
+// the agent, where the failure is logged and swallowed.
```
weyfonk: nit: Glp! The following may be clearer:
```suggestion
// the agent, where the failure is logged and not surfaced to the user.
```

internal/bundlereader/validate_fleetyaml_test.go
```diff
@@ -198,3 +198,828 @@ func TestValidBundleStatesList_SortedByRank(t *testing.T) {
...
+		},
+		{
+			name:          "wrong case",
+			diff:          diffWithPatchOps("Remove"),
+			expectedError: `diff.comparePatches[0].operations[0].op: unsupported operation "Remove"`,
+		},
+		{
+			// An operation with a path but no op reads as a no-op and is not one:
+			// it makes the whole patch fail. Rejecting it is a deliberate
+			// tightening over what fleet apply accepted before.
```
weyfonk, line 504: Great explanation, I think we could surface this in user-facing docs.

internal/cmd/agent/deployer/desiredset/diff.go
```diff
@@ -58,9 +59,11 @@ func Diff(logger logr.Logger, plan Plan, bd *fleet.BundleDeployment, ns string,
...
 				re, err := regexp.Compile(key.Name)
 				if err != nil {
-					// XXX: enable detection of such issues earlier, for instance through CLI validating
-					// fleet.yaml syntax; see fleet#4533.
-					logger.V(1).Error(
+					// fleet apply rejects a name which does not compile (see
+					// validation.ValidateComparePatchNames), but BundleDeployments
+					// created by an older Fleet version, or from a HelmOp, still reach
+					// this point, hence the check and the error log.
+					logger.Error(
```
weyfonk, line 66: Nice, thanks for making these errors more visible!

weyfonk (approved):
LGTM, with a few nitpicks :)

internal/validation/comparepatches_test.go
0xavi0: Covered now, thanks!

internal/validation/comparepatches.go
0xavi0: changed

Issue #5690, opened 2026-09-01
[backport v2.15] Bundle dependencies: use specific state in dependent bundle if dependency not in accepted state


rancherbot:
This is a backport issue for #4797, automatically created via [GitHub Actions workflow](https://github.com/rancher/fleet/actions/runs/33488310097) initiated by @0xavi0

Original issue body:

Work on #2552 has resulted in users being able to specify a set of accepted states for a dependency bundle, allowing dependent bundles to move forward based on those states (docs [acceptedStates](https://fleet.rancher.io/reference/ref-fleet-yaml) and [dependsOn](https://fleet.rancher.io/reference/ref-crds)).

However, when a bundle is _not_ in one of its dependent bundle's _accepted states_, the dependent bundle still bears state `ErrApplied`.
We should change this, using a different state to clarify that this is its own type of error, and that it may well be temporary.
Possible names for this new state: `WaitingForDependency`, `DependencyInUnsupportedState` (perhaps a bit too long)...

Relates to https://github.com/rancher/dashboard/issues/13121 for UI work.

mmartin24:
Our CI already spotted the change:

<img width="1537" height="923" alt="Image" src="https://github.com/user-attachments/assets/02d3fbf2-e357-4b4d-b20d-0c9362138a13" />

Adaptation implemented and verified through automation in `v2.15-735ca34da520e2f4aeb1f58c589dc4e5d8c14e50-head`  | `fleet-110.0.2+up0.16.2-rc.1`  in test fleet-207  [here](https://github.com/rancher/fleet-e2e/actions/runs/34957150329/job/104342010877#step:10:419).

<img width="2476" height="1314" alt="Image" src="https://github.com/user-attachments/assets/d17715b5-7f23-4da2-b61b-25aee83fc022" />

PR #5691, opened 2026-09-01
[v0.16] - Add WaitingForDependency bundle state (#5549)

0xavi0:
Backports: https://github.com/rancher/fleet/pull/5549
Refers to: https://github.com/rancher/fleet/issues/5690


schemas/fleet.yaml.json
```diff
@@ -0,0 +1,809 @@
+{
```
weyfonk: Isn't this file the result of a [feature](https://github.com/rancher/fleet/issues/1962) implemented for v0.17? In other words, I don't think we need to backport changes made to it.

0xavi0: Good catch! deleted now

weyfonk (approved):

PR #5717, opened 2026-09-03
Fix GitRepo/HelmOp status races that drop concurrent conditions

0xavi0:
Fleet has multiple writers per GitRepo/HelmOp status (main reconciler, status reconciler, polling job), each patching from a stale snapshot — a merge patch replaces the whole conditions array, so one writer could silently overwrite conditions another had just set (e.g. Accepted reverted by a concurrent poll).

- Add status.MergeConditions to merge a reconciler's desired conditions into live instead of replacing wholesale, and apply it in the GitRepo/HelmOp reconcilers.
- Add an optimistic lock to every status patch/update path, retrying on conflict; new StatusConflictRequeue (1s) requeues fast when that happens.
- Stop reverting PollingCommit/LastPollingTime from the stale snapshot in the GitRepo reconciler, which was delaying new-commit pickup by a full polling interval.
- New integration tests exercise concurrent writers via a stale-read test client.

### Checklist

### DIFF
--- a/internal/cmd/controller/helmops/reconciler/helmop_controller.go
+++ b/internal/cmd/controller/helmops/reconciler/helmop_controller.go
@@ -27,13 +27,13 @@ import (
 	"github.com/Masterminds/semver/v3"
 	"github.com/go-logr/logr"
 	"github.com/rancher/wrangler/v3/pkg/condition"
-	"github.com/rancher/wrangler/v3/pkg/genericcondition"
 	"github.com/reugn/go-quartz/quartz"
 
 	"github.com/rancher/fleet/internal/bundlereader"
 	fleetutil "github.com/rancher/fleet/internal/cmd/controller/errorutil"
 	"github.com/rancher/fleet/internal/cmd/controller/finalize"
 	ctrlquartz "github.com/rancher/fleet/internal/cmd/controller/quartz"
+	"github.com/rancher/fleet/internal/cmd/controller/status"
 	"github.com/rancher/fleet/internal/metrics"
 	fleet "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
 	"github.com/rancher/fleet/pkg/cert"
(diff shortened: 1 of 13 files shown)


internal/cmd/controller/helmops/reconciler/helmop_status.go
```diff
@@ -98,13 +99,31 @@ func (r *HelmOpStatusReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, err
 	}
 
-	statusPatch := client.MergeFrom(orig)
-	if patchData, err := statusPatch.Data(helmop); err == nil && string(patchData) != "{}" {
-		// skip update if patch is empty
-		if err := r.Status().Patch(ctx, helmop, statusPatch); err != nil {
-			logger.Error(err, "Reconcile failed update to HelmOp status", "status", helmop.Status)
-			return ctrl.Result{RequeueAfter: durations.HelmOpStatusDelay}, nil
+	if equality.Semantic.DeepEqual(orig.Status, helmop.Status) {
```
d3flex, line 102: this part looks very identical to the `updateStatus` from the internal/cmd/controller/gitops/reconciler/status_controller.go. The diferrence is that one is wrapped in a function, the other is inline. So for consistency it would be nice to export the function in some helper module and replace the lines here. wdyt?

internal/cmd/controller/helmops/reconciler/helmop_controller.go
```diff
@@ -442,58 +446,71 @@ func usesPolling(helmop fleet.HelmOp) bool {
...
-			t.Status.Version = objToPatchFrom.Status.Version
-			// (#3883)
-			// If orig.Status.Version is, for example, equal to 1.0.0, when
-			// assigning the Version to t.Status.Version both will be 1.0.0.
-			// When calculating the Patch data, Status.Version will be ignored because
-			// both objects have the same value.
-			// The following cleanup prevents that so Status.Version is taken into
-			// account when calculating the patch data.
-			objToPatchFrom.Status.Version = ""
+		if desired.Status.Version != "" {
```
d3flex, line 483: `if desired.Status.Version != "" && desired.Status.Version != base.Version {`

This took me some time to figure it out. And I am still not 100% sure. Why do you remove the versions incompatibility check?

PR #5775, opened 2026-09-11
Constrain Helm values file paths

thardeck:
Resolve each `valuesFiles` entry relative to its bundle or checkout root before reading. Constrain resolved paths to that root.

### DIFF
--- a/internal/bundlereader/resources_test.go
+++ b/internal/bundlereader/resources_test.go
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"os"
+	"path/filepath"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


internal/bundlereader/resources_test.go
```diff
@@ -113,6 +115,82 @@ func TestValueMerge(t *testing.T) {
...
+}
+
+// A relative base (the common case: gitjob invokes fleet apply with base ".")
+// combined with a values file that is an absolute symlink resolving inside
+// that base must still be accepted.
+func TestGenerateValuesAllowsAbsoluteSymlinkWithinRelativeBase(t *testing.T) {
+	base := t.TempDir()
+	target := filepath.Join(base, "actual-values.yaml")
+	require.NoError(t, os.WriteFile(target, []byte("foo: bar"), 0644))
+	require.NoError(t, os.Symlink(target, filepath.Join(base, "values.yaml")))
```
weyfonk, line 179: Is this an absolute symlink? In other words, wouldn't something like this better reflect the use case we want to cover?
```suggestion
    absTarget, err := filepath.Abs(target)
    require.NoError(t, err)
	require.NoError(t, os.Symlink(absTarget, filepath.Join(base, "values.yaml")))
```

thardeck: > Is this an absolute symlink? 

Yes, `t.TempDir()` always returns an absolute path. But it is not documented, that is weird.

weyfonk (approved):

Issue #5793, opened 2026-09-15
helmRepoURLRegex set but non-matching silently strips helmSecretName credentials (misleading 401)


aisensiy:
### Current Behavior

When `spec.helmRepoURLRegex` is set on a GitRepo but does not match a chart's repository URL, Fleet silently strips the credentials referenced by `helmSecretName` and fetches the chart anonymously. Private registries then answer `401 authentication required`, which presents as an authentication failure (wrong password, missing permissions, registry incompatibility) rather than a regex misconfiguration.

There is a log line for the **empty**-regex case (`"helmRepoURLRegex is empty: Helm credentials will not be forwarded to any repository; ..."`, in `internal/bundlereader/resources.go`, `addRemoteCharts`), but the **set-but-not-matching** case logs nothing. From the job log alone there is no way to tell that credentials were dropped client-side.

### Expected Behavior

A log line explaining that credentials were not forwarded because `helmRepoURLRegex` does not match this chart's URL, mirroring the existing empty-regex warning. (No behavior change — stripping stays, it is a deliberate security property. Only the silence hurts.)

### Steps To Reproduce

Fleet 0.16.1 standalone, GitRepo in `fleet-local` with:

```yaml
spec:
  repo: https://github.com/org/deploy.git
  branch: main
  paths: [k8s/env]
  helmSecretName: my-registry-creds   # Secret with username/password, pull works
  targets:
  - clusterName: local
```

`k8s/env/fleet.yaml` in that repo:

```yaml
helm:
  repo: oci://registry.example.com/team/mychart
  version: 1.2.3
```

1. Set `helmRepoURLRegex` with a regex that does not match. In our real incident this was an escaping accident while patching: JSON needs `\\.` to produce a regex `\.` (literal dot), and one extra level produced `\\.` in the stored value, which matches a **literal backslash** — never present in any URL:

   ```bash
   # correct — stored value: ^oci://registry\.example\.com/  (matches)
   kubectl patch gitrepo demo --type merge \
     -p '{"spec":{"helmRepoURLRegex":"^oci://registry\\.example\\.com/"}}'

   # what we actually ran — stored value: ^oci://registry\\.example\\.com/  (never matches)
   kubectl patch gitrepo demo --type merge \
     -p '{"spec":{"helmRepoURLRegex":"^oci://registry\\\\.example\\\\.com/"}}'
   ```

   Both commands succeed silently; the difference in `kubectl get gitrepo -o yaml` is easy to miss.

[...]

PR #5794, opened 2026-09-15
Log warning when helmRepoURLRegex strips credentials from a non-matching chart URL

aisensiy:
Refers #5793

## Description

When `spec.helmRepoURLRegex` is set on a GitRepo but does not match a chart's repository URL, the credentials from `helmSecretName` are silently stripped. The chart is then fetched anonymously and private registries respond with `401 authentication required`, which looks like an authentication failure instead of a regex misconfiguration (see the issue for a real-world timeline this cost us).

The empty-regex case already logs a warning. This adds the missing counterpart for the set-but-non-matching case, mirroring the existing pattern in `addRemoteCharts`:

```
helmRepoURLRegex "^oci://registry\\.example\\.com/" does not match chart repo=oci://registry.example.com/team/mychart chart= version=1.2.3: Helm credentials will not be forwarded to this repository
```

No behavior change: credentials are still stripped, transport settings are still preserved.

## Testing

Extended the table in `TestAddRemoteChartsWarnsMissingRegex` (`internal/bundlereader/resources_test.go`):

- `SSH key set, regex non-matching — warn` (new)
- `no credentials, regex non-matching — no warn` (new)
- `credentials set, regex matching — no warn` (replaces the previous `regex provided` case, whose chart URL never actually matched the regex — with the new warning that case now logs)
- `credentials set, regex non-matching — warn` (new)

`go test ./internal/bundlereader/ -count=1` passes; `gofmt` clean.

## AI Disclosure

This change was developed with AI assistance (Claude, in an agentic workflow driven by the signer). The author reviewed the change, ran the tests, and can explain every line, per `AI_POLICY.md`. The commit carries an `Assisted-by:` trailer.

### Checklist

- [x]  Documentation not affected: log line only, no user-facing configuration or API change.


internal/bundlereader/resources.go
```diff
@@ -245,8 +245,17 @@ func addRemoteCharts(ctx context.Context, directories []directory, base string,
...
-				if helmRepoURLRegex == "" && (auth.Username != "" || auth.Password != "" || len(auth.SSHPrivateKey) > 0) {
+				hasCredentials := auth.Username != "" || auth.Password != "" || len(auth.SSHPrivateKey) > 0
+				if helmRepoURLRegex == "" && hasCredentials {
 					strippedCredentialsForEmptyRegex = true
+				} else if helmRepoURLRegex != "" && hasCredentials {
+					// A non-empty regex that does not match this chart's URL also
+					// strips credentials, but silently: the resulting download
+					// failure is a registry-side 401 that looks like an
+					// authentication problem. Say why credentials were dropped.
+					log.Log.Info(fmt.Sprintf(
```
thardeck, line 256: The `warnedOnce` check is not used for the new log message.

thardeck (changes requested):
Thank you for your pr.

internal/bundlereader/resources.go
thardeck: Never mind, it is debatable if the log message makes sense for all charts in the list, but it is not really wrong since each chart would be affected by the issue.

thardeck (approved):
Looks good to me.

Issue #5438, opened 2026-07-13
Removing a cluster-specific `imagePullSecrets` doesn't restore the global  default one


mmartin24:
### Is there an existing issue for this?

- [x] I have searched the existing issues

## Summary

If you set a custom image pull secret for one downstream cluster, then later remove it, Fleet doesn't switch back to using the global/default secret. It keeps using the old (removed) one, and the agent never gets updated. Found while testing #5254.

## Environment

- Fleet Standalone (no Rancher), built from `main`, tested 2026-07-08
- Downstream cluster `downstream1` registered via manager-initiated registration
- Setup guide: https://github.com/rancher/fleet/issues/5254#issuecomment-4925245959 (see "Standalone Installation Guide" section)
- Full test steps/output: https://github.com/rancher/fleet/issues/5254#issuecomment-4925245959 (see "Test 2" under "Testing Guide (Standalone Fleet)")

## Steps to Reproduce

1. Set up Fleet with a global image pull secret (e.g. `zot-creds`) — confirm the downstream cluster's agent uses it.
2. Set a cluster-specific override — e.g. patch the Cluster with `agentPullSecrets: [{name: cluster-zot-creds}]`.
3. Confirm the agent switches to using `cluster-zot-creds` (this part works fine).
4. Remove the override (delete `agentPullSecrets` from the Cluster).
5. Check what secret the agent is using now.

## Actual behavior

The agent still uses `cluster-zot-creds`, even though it was removed from the Cluster config. It never goes back to the global `zot-creds`. Nothing happens in the controller logs when the override is removed — Fleet doesn't notice the change at all.

## Expected behavior

Removing the override should make Fleet pick the global secret back up automatically.

Confirmed by @weyfonk

> The agent management controller does not trigger a new cluster import when cluster-level secrets are removed. This is something we should change.

https://github.com/rancher/fleet/issues/5254#issuecomment-4936186845

### Environment

```markdown
- Architecture:
- Fleet Version:
- Cluster:
  - Provider:
  - Options:
  - Kubernetes Version:
```

### Logs

```markdown

```

PR #5795, opened 2026-09-15
fix(manageagent) reconcile agent when agentPullSecret is modified

d3flex:
`GetAgentPullSecrets` propagates the agent's secret only when the field is `nil`. Similar every other agentic field is reconciled via the `updateClusterStatus` which signals a change.

Use the same logic to teach `hashChanged` about corev1.LocalObjectReference (the AgentPullSecrets type) and reset the field to nil will make the process propagate the global value agin after the old is removed.

Refers to https://github.com/rancher/fleet/issues/5438


0xavi0 (approved):

0xavi0:
@d3flex, Please create the PR (and the issue) for `main` also.

d3flex:
> @d3flex, Please create the PR (and the issue) for `main` also.

done - https://github.com/rancher/fleet/pull/5816

Issue #5410, opened 2026-07-07
Fleet HelmOp cannot resolve OCI charts whose tags use the `+up` / `_up` convention


e-minguez:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

A Fleet `HelmOp` pointing at an OCI Helm chart whose **only** published tags use
the Helm build-metadata convention (semver `X.Y.Z+upA.B.C`, stored in the
registry as the OCI-safe tag `X.Y.Z_upA.B.C`) never becomes `Accepted`. Fleet's
OCI version resolver treats `spec.helm.version` as a semver **constraint** and
fails to resolve a chart for **every** value of `version` — a specific
build-metadata version, the underscore tag, the plain `X.Y.Z`, and empty.

The same chart installs fine with the Helm CLI, which requests the tag directly
instead of listing + semver-matching.

### Expected Behavior

Fleet should resolve and install the chart. In particular, a constraint of
`109.0.1` (or the exact `109.0.1+up1.0.2`) should match the published chart
version `109.0.1+up1.0.2`, since semver comparison ignores build metadata.

## Contrast: the Helm CLI works

Helm requests the tag directly (converting `+` -> `_`) instead of
listing + semver-matching, so it succeeds:

```bash
helm show chart oci://registry.suse.com/rancher/charts/rancher-ai-agent --version 109.0.1+up1.0.2
# Pulled: registry.suse.com/rancher/charts/rancher-ai-agent:109.0.1_up1.0.2
# ... version: 109.0.1+up1.0.2

helm pull oci://registry.suse.com/rancher/charts/rancher-ai-agent --version 109.0.1+up1.0.2
# downloads rancher-ai-agent-109.0.1_up1.0.2.tgz

### Steps To Reproduce

```bash
kubectl apply -f helmop.yaml

# Read the Accepted condition:
kubectl get helmop liz-agent-repro -n fleet-local \
  -o jsonpath='{.status.conditions[?(@.type=="Accepted")].status} {.status.conditions[?(@.type=="Accepted")].message}{"\n"}'
```

To test the other `version` values, edit `spec.helm.version` in `helmop.yaml`,
re-apply, and re-read the condition.

Cleanup:

```bash
kubectl delete -f helmop.yaml
```

The helmop file:
```
# Minimal Fleet HelmOp that reproduces the OCI "+up"/"_up" tag resolution
# failure. No Rancher AI Factory operator involved — this is pure Fleet.
#
[...]

PR #5808, opened 2026-09-18
Support OCI build metadata in Helm chart versions

0xavi0:
OCI registries don't allow + in tags, so Helm publishes 109.0.1+up1.0.2 under the tag 109.0.1_up1.0.2. Fleet didn't account for that: version 109.0.1 never reached the chart, and the tag spelling copied from a registry was rejected.

- New internal/helmversion package: converts between both spellings, parses versions strictly, and orders them with build metadata taken into account (semver ignores it), comparing metadata identifier by identifier so up1.0.10 ranks above up1.0.2.
- GetOCITag resolves an exact version to the highest build of it; spelling out metadata pins one build. Resolution no longer depends on the order a registry lists tags in, and a missing match returns a typed NoMatchingTagError.
- Charts in git repos now resolve their OCI version at bundle-build time, matching HelmOp's reach. Best-effort: on failure the version goes to Helm unchanged, as before.
- HelmOps and fleet.yaml accept the tag spelling and store the semver one; usesPolling no longer polls a version written that way.

Refers to: https://github.com/rancher/fleet/issues/5410

### Checklist


internal/helmversion/helmversion.go
```diff
@@ -0,0 +1,166 @@
...
+	// them takes precedence.
+	return cmp.Compare(len(aIDs), len(bIDs))
+}
+
+// compareIdentifier orders two dot separated identifiers, following the
+// precedence rules the semver specification gives for pre-release identifiers:
+// numeric identifiers compare numerically, and rank below alphanumeric ones.
+func compareIdentifier(a, b string) int {
+	aNum, aErr := strconv.ParseUint(a, 10, 64)
+	bNum, bErr := strconv.ParseUint(b, 10, 64)
```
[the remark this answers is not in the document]
0xavi0: Not too realistic, but still a valid remark.
I've changed it so compareIdentifier no longer parses identifiers into a uint64. It classifies them with isNumeric (digits alone, per the semver definition) and orders numeric ones with compareNumeric, which strips leading zeroes and compares by length, then lexically — correct at any width. The strconv import is gone.

internal/bundlereader/charturl.go
```diff
@@ -46,21 +48,21 @@ var (
 // ChartVersion returns the version of the helm chart from a helm repo server, by
 // inspecting the repo's index.yaml
 func ChartVersion(ctx context.Context, location fleet.HelmOptions, a Auth) (string, error) {
+	// Users may spell build metadata the way a registry tag does, with an
+	// underscore instead of a plus sign.
+	location.Version = helmversion.Normalize(location.Version)
```
d3flex, line 53: If I understand correctly, the /normalization/ is not needed at the OCI registry. like the versions should be as they are supposed to be. So does the `Normalize` is unnecessary for not OCI path??

d3flex (approved):
Approved with one small question for my curiosity. In any case the few test cases I run as described in the ticket, all looked good.

internal/bundlereader/charturl.go
0xavi0: Good question — it's actually the other way round: the OCI path would survive without it,
and it's the plain Helm repo path that needs it here.

`Normalize` only converts the *user supplied* version — what someone wrote in fleet.yaml or
in a HelmOp — from the tag spelling (109.0.1_up1.0.2) to the semver one (109.0.1+up1.0.2).
It never touches what the registry reports; that direction is `FromTag`, inside `GetOCITag`.

PR #5825, opened 2026-09-22
Fix controller runtime logger setup

thardeck:
fleetcli `apply`, and the agentmanagement and cleanup fleet-controller subcommands, never called `ctrl.SetLogger`. Since `logrus` was replaced with the controller-runtime `log.Log` delegating logger, calls to it in these entrypoints are silently dropped after a 30s timeout, hiding messages such as the `helmRepoURLRegex` credential-stripping warning and any logging from the agentmanagement and cleanup controllers.

Wire up `zap.Options` and `ctrl.SetLogger` in these entrypoints the same way gitops and helmops already do, and drop the now-unused `registerKubeconfigFlags` helper.

Refers #3554


internal/cmd/controller/agentmanagement/root_test.go
```diff
@@ -0,0 +1,50 @@
+package agentmanagement
```
weyfonk, line 1: nit: this regression is now in place for a few locations, but if say, the `bundlediff` CLI were modified not to call `ctrl.SetLogger` anymore, it would not be covered.

weyfonk (approved):

d3flex (approved):

weyfonk (approved):

